PR #1, opened 2024-08-20
Add Pulp deployment

archanaserver:


roles/pulp/templates/pulp.container.j2
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Pulp Container
+After=network.target
+
+[Service]
```
ehelms, line 5: This needs to be a `[Container]` section for a quadlet, and all of the `Exec` commands you have below are unneeded. You can translate the options in the run command to top-level options. For example:

Instead of `-v /var/lib/pulp/settings:/etc/pulp:Z` you can do:

```
Volume=/var/lib/pulp/settings:/etc/pulp:Z
```

Take a look at https://github.com/ehelms/foreman-quadlet/blob/master/roles/candlepin/templates/candlepin.container.j2#L5-L26

ehelms: Oh yea, Isee you have it here -- https://github.com/ehelms/foreman-quadlet/pull/1/files#diff-b96d3b06048d3fbc532b73220a5bd877d7549e44a11ab70ae001f82a8b38b9b5

Move that file here instead inside the role and deploy it.

ekohl (commented):
My recommendation is to try out https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html to create the container.

playbooks/setup.yaml
```diff
@@ -32,3 +32,21 @@
...
+      - /var/lib/pulp/settings
+      - /var/lib/pulp/pulp_storage
+      - /var/lib/pulp/pgsql
+      - /var/lib/pulp/containers
+
+  - name: Create settings.py for Pulp
+    ansible.builtin.lineinfile:
+      path: /var/lib/pulp/settings/settings.py
+      line: "CONTENT_ORIGIN='http://{{ ansible_hostname }}:8080'"
+      create: yes
```
ekohl, line 51: I wonder why https://pulpproject.org/pulp-oci-images/docs/admin/tutorials/quickstart/#create-the-directories-and-settings suggests writing a `settings.py` instead of using environment variables. I recall you could also set `PULP_CONTENT_ORIGIN` as an environment variable instead to configure it but I'll check that with the Pulp people.

ekohl: Quoting @mdellweg from Matrix:

> I think that's a matter of taste. Both work. Some complex settings are easier coded in settings.py. OTOH, having executable settings always makes me shiver.

So my recommendation would be to define it in the `.container` file as an `Environment` key to simplify things. Then you can avoid the entire `settings.py` file

roles/pulp/templates/pulp.container.j2
```diff
@@ -0,0 +1,16 @@
...
+After=network.target
+
+[Container]
+Image=pulp/pulp:stable
+ContainerName=pulp
+PublishPort=8080:80
+Volume=/var/lib/pulp/settings:/etc/pulp:Z
+Volume=/var/lib/pulp/pulp_storage:/var/lib/pulp:Z
+Volume=/var/lib/pulp/pgsql:/var/lib/pgsql:Z
+Volume=/var/lib/pulp/containers:/var/lib/containers:Z
```
ekohl, line 12: If you don't care about where the volume is on the host system you can also use relative paths:
```suggestion
Volume=pulp_settings:/etc/pulp
Volume=pulp_storage:/var/lib/pulp
Volume=pulp_pgsql:/var/lib/pgsql
Volume=pulp_containers:/var/lib/containers
```

Benefit is that you don't need to ensure the directory (simplifying your Ansible file). Downside is that you can't just take a backup of `/var/lib/pulp`. On the other hand, you may want to approach backups in a different way anyway.

roles/pulp/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+pulp_image: pulp/pulp:stable
+pulp_ports:
+  - "8080:80"
+pulp_volumes:
```
ehelms, line 5: Is this variable unused now that they are defined in the `container` file?

roles/pulp/templates/pulp.container.j2
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Pulp 
+After=network.target
+
+[Container]
+Image=pulp/pulp:stable
+ContainerName=pulp
+PublishPort=8080:80
+Volume=pulp_settings:/etc/pulp
```
ehelms, line 9: How does this declaration work? Is that a `pulp_settings` directory relative to the current directory this gets run from?

ekohl: When I tested it, it created a volume that you could also see with `podman volume ls`. You can also define specific volumes with (for example) `pulp_settings.volume`.

roles/pulp/defaults/main.yml
archanaserver: Yes, this seem redundant here, but what is the best practice to define it where

roles/pulp/templates/pulp.container.j2
archanaserver: Yes, This `pulp_settings:/etc/pulp` is a Podman named volume here. We can see it with `podman volume ls`. I can make it more explicit as @ekohl suggested using `.volume` syntax

ehelms: Where on the filesystem does this volume get created? I tried to read up on it but I didn't feel like I found a clear answer.

ekohl: I think podman creates its volumes in `/var/lib/containers`

ekohl (changes requested):
I'll reiterate: this PR is reproducing [containers.podman.podman_container](https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html) and a I think we should use that instead of reinventing the wheel. I'll note I haven't used it myself, so perhaps there are practical limitations, but it's maintained in the same GitHub org as podman itself.

ehelms:
> I'll reiterate: this PR is reproducing [containers.podman.podman_container](https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html) and a I think we should use that instead of reinventing the wheel. I'll note I haven't used it myself, so perhaps there are practical limitations, but it's maintained in the same GitHub org as podman itself.

I am simply avoiding it right now in order to better understand and see what is happening with the quadlet definition file. While I love a good abstraction, I sometimes hate fighting with them early on, or having to apply them in order to see and understand.

I started with the just define it myself over in https://github.com/ehelms/foreman-quadlet/blob/master/roles/candlepin/templates/candlepin.container.j2 for that reason. I am also using Ansible as "light weight" as possible right now, because I didn't want to commit to it just yet.

One item of note, the `podman_container` module does not appear to have "native" support for the `[Kube]` method and would rely upon use of supplying an inline template to `quadlet_options`.

ekohl:
That's fair and thanks for writing it out. I think that's important to share.

ehelms:
This fails to load the service due to issues with the specification:

```
Sep 16 14:22:02 quadlet.example.com systemd[1]: Reloading.
Sep 16 14:22:02 quadlet.example.com quadlet-generator[47766]: Warning: pulp.container specifies the image "pulp/pulp:stable" which not a fully qualified image name. This is not ideal for performance and security reasons. See the podman-pull manpage discussion of short-name-aliases.conf for details.
Sep 16 14:22:02 quadlet.example.com quadlet-generator[47766]: converting "pulp.container": unsupported key 'Device' in group 'Container' in /etc/containers/systemd/pulp.container
Sep 16 14:22:02 quadlet.example.com systemd-rc-local-generator[47777]: /etc/rc.d/rc.local is not marked executable, skipping.
Sep 16 14:22:02 quadlet.example.com systemd[47763]: /usr/lib/systemd/system-generators/podman-system-generator failed with exit status 1.
```

archanaserver:
> converting "pulp.container"



> This fails to load the service due to issues with the specification:
> 
> ```
> Sep 16 14:22:02 quadlet.example.com systemd[1]: Reloading.
> Sep 16 14:22:02 quadlet.example.com quadlet-generator[47766]: Warning: pulp.container specifies the image "pulp/pulp:stable" which not a fully qualified image name. This is not ideal for performance and security reasons. See the podman-pull manpage discussion of short-name-aliases.conf for details.
> Sep 16 14:22:02 quadlet.example.com quadlet-generator[47766]: converting "pulp.container": unsupported key 'Device' in group 'Container' in /etc/containers/systemd/pulp.container
> Sep 16 14:22:02 quadlet.example.com systemd-rc-local-generator[47777]: /etc/rc.d/rc.local is not marked executable, skipping.
> Sep 16 14:22:02 quadlet.example.com systemd[47763]: /usr/lib/systemd/system-generators/podman-system-generator failed with exit status 1.
> ```

Thanks, I was ignoring this log earlier, but I found out that, Quadlet has a limited set of supported options in the [Container] section, and Device is not one of them. So I did some changes and used Podman `--device` flag to mount `/dev/fuse` directly in the [Service] section using ExecStart like this. Do you think this is a good approach?

roles/pulp/defaults/main.yml
archanaserver: Not anyone.

roles/pulp/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+  - pulp_settings.volume:/etc/pulp:Z
```
ekohl, line 6: Since you're providing the setting through an environment variable and don't actually write out a settings file, I think you can drop this volume. Not entirely sure if Pulp actually creates files at runtime though.

roles/pulp/templates/pulp.container.j2
```diff
@@ -0,0 +1,18 @@
...
+After=network.target
+
+[Container]
+Image={{ pulp_image }}
+ContainerName={{ pulp_container_name }}
+PublishPort={{ pulp_ports }}
+{% for volume in pulp_volumes %}
+Volume={{ volume }}
+{% endfor %}
+Environment=PULP_CONTENT_ORIGIN=http://{{ ansible_hostname }}:8080
```
ekohl, line 12: I think this port should be the same as what's used as the external port in your `pulp_ports` variable.

That's now structured as `8080` (the port on the host), which you also use here. Then the port `80` is the one inside the container and that's fixed (because it's part of the container image itself).

My recommendation would be to change the `pulp_ports` variable to something that's more like `pulp_port=8080`. Then you can use `PublishPort={{ pulp_port }}:80` and here:
```suggestion
Environment=PULP_CONTENT_ORIGIN=http://{{ ansible_hostname }}:{{ pulp_port }}
```

tests/pulp_test.py
```diff
@@ -0,0 +1,23 @@
...
+    assert pulp_process.exists
+
+# def test_pulp_container(host):
+#     container_info = host.run("podman inspect pulp").stdout
+#     assert "pulp/pulp:stable" in container_info
+#     assert "Running" in container_info
+
+def test_systemd_reload(host):
+    result = host.run("systemctl daemon-reload")
+    assert result.succeeded
```
ekohl, line 23: This tests you can run `daemon-reload`, but that doesn't assert any behavior we want. What you want to assert is that a deamon-reload isn't actually needed (for example, because it already ran). https://www.baeldung.com/linux/systemctl-daemon-reload tells me `systemctl show pulp --property=NeedDaemonReload` should return `NeedDaemonReload=no`, but I didn't test this with containers. You also want to try out what it returns when the `pulp.container` doesn't exist, though that's probably already covered by your `test_pulp_service` test. And https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.service.Service tells me you can even extend it with `assert pulp.systemd_properties['NeedDaemonReload'] == 'no'`, which makes it really clean.

tests/pulp_test.py
```diff
@@ -0,0 +1,23 @@
+import testinfra
+
+def test_pulp_service(host):
+    pulp = host.service("pulp")
+    assert pulp.is_running
+    assert pulp.is_enabled
```
ekohl, line 6: A good rule when writing tests is that you want a single assertion per test. That way you get all the needed output reported at once so I'd recommend:
```suggestion
    assert pulp.is_running and pulp.is_enabled
```

But I'd go a bit further and try:
```suggestion
    assert pulp.exists and pulp.is_running and pulp.is_enabled
```

And I'd also play with it and see how it reports a (partial) failure. For example, force it to fail with 
```suggestion
    assert not pulp.exists and pulp.is_running and pulp.is_enabled
```

I think pytest will report all variables, but that's useful to know.

roles/pulp/templates/pulp.container.j2
```diff
@@ -0,0 +1,18 @@
...
+
+[Service]
+ExecStart=/usr/bin/podman run --name %n --rm --device /dev/fuse {{ pulp_image }}
```
ehelms, line 15: This section has to be dropped, quadlet's generate a `ExecStart` and this causes a duplicate which breaks the service.

roles/pulp/templates/pulp.container.j2
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Pulp 
+After=network.target
+
+[Container]
+Image={{ pulp_image }}
+ContainerName={{ pulp_container_name }}
+PublishPort={{ pulp_ports }}
```
ehelms, line 8: This also throws errors... I think to start with we need to go the route I did with Candlepin, and use `Network=host`.

roles/pulp/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+  - pulp_storage.volume:/var/lib/pulp:Z
+  - pulp_pgsql.volume:/var/lib/pgsql:Z
+  - pulp_containers.volume:/var/lib/containers:Z
```
ehelms, line 9: I learned something today. Adding the `.volume` causes quadlets to treat this as a service providing the volume and not just a simple filesystem volume. This then results in those services being required and failing, here is the output from the generated service file:

```
[Unit]
Wants=network-online.target
After=network-online.target
Description=Pulp
After=network.target
SourcePath=/etc/containers/systemd/pulp.container
RequiresMountsFor=%t/containers
Requires=pulp_settings-volume.service
After=pulp_settings-volume.service
Requires=pulp_storage-volume.service
After=pulp_storage-volume.service
Requires=pulp_pgsql-volume.service
After=pulp_pgsql-volume.service
Requires=pulp_containers-volume.service
After=pulp_containers-volume.service
```

TLDR; Drop the `.volume`

roles/pulp/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+pulp_image: pulp/pulp:stable
```
ehelms, line 2:
```suggestion
pulp_image: quay.io/pulp/pulp:stable
```

roles/pulp/templates/pulp.container.j2
ehelms: Ignore that comment, `Network=host` breaks the all-in-one Pulp container because it has a postgres within the container and it fails looking for the postgresql running on the host.

roles/pulp/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+pulp_ports:
+  - "8080:80"
```
ehelms, line 4:
```suggestion
pulp_port: "8080:80"
```

`PublishPort` does not accept an array in the Quadlet file. Which if you were trying to look at https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html and map it, the Ansible module abstraction does allow you to define an array of ports, however, in that case it's looping over the array to create multiple `PublishPort` entries. Similar to what you did with the volumes.

tests/pulp_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_pulp_port(host):
+    pulp = host.addr("localhost")
+    assert pulp.port("8080").is_reachable
+
+def test_pulp_process_running(host):
+    pulp_process = host.process.filter(comm="pulp")
+    assert pulp_process.exists
```
ehelms, line 14: This is handled by your first test of the service and can be deleted.

roles/pulp/defaults/main.yml
archanaserver: That's interesting, thanks for sharing. Updated it.

ehelms:
> > I'll reiterate: this PR is reproducing [containers.podman.podman_container](https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html) and a I think we should use that instead of reinventing the wheel. I'll note I haven't used it myself, so perhaps there are practical limitations, but it's maintained in the same GitHub org as podman itself.
> 
> I am simply avoiding it right now in order to better understand and see what is happening with the quadlet definition file. While I love a good abstraction, I sometimes hate fighting with them early on, or having to apply them in order to see and understand.
> 
> I started with the just define it myself over in https://github.com/ehelms/foreman-quadlet/blob/master/roles/candlepin/templates/candlepin.container.j2 for that reason. I am also using Ansible as "light weight" as possible right now, because I didn't want to commit to it just yet.
> 
> One item of note, the `podman_container` module does not appear to have "native" support for the `[Kube]` method and would rely upon use of supplying an inline template to `quadlet_options`.

Today I learned, that you handle the Kube method in a similar abstraction but using the `podman_play` module: https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_play_module.html#ansible-collections-containers-podman-podman-play-module

ehelms:
Closing in favor of https://github.com/ehelms/foreman-quadlet/pull/2

PR #2, opened 2024-09-09
Add Pulp deployment using containers.podman.podman_container module

archanaserver:


roles/pulp/templates/pulp.container.j2
```diff
@@ -0,0 +1,17 @@
+# [Unit]
```
ehelms, line 1: Using this method, is this file needed?

archanaserver: This one is definitely not needed here. I intend to drop this.

roles/pulp/defaults/main.yaml
```diff
@@ -0,0 +1,10 @@
+---
+pulp_image: pulp/pulp:stable
+pulp_ports:
+  - "8080:80"
+pulp_volumes:
+  - /var/lib/pulp/settings:/etc/pulp:Z
+  - /var/lib/pulp/pulp_storage:/var/lib/pulp:Z
+  - /var/lib/pulp/pgsql:/var/lib/apgsql:Z
```
evgeni, line 8:
```suggestion
  - /var/lib/pulp/pgsql:/var/lib/pgsql:Z
```

;)

roles/pulp/defaults/main.yaml
```diff
@@ -0,0 +1,10 @@
+---
+pulp_image: pulp/pulp:stable
```
evgeni, line 2:
```suggestion
pulp_image: quay.io/pulp/pulp:stable
```

podman wants fully qualified container names

roles/pulp/tasks/main.yaml
```diff
@@ -0,0 +1,24 @@
...
+- name: Deploy Pulp Container
+  containers.podman.podman_container:
+    name: "{{ pulp_container_name }}"
+    image: "{{ pulp_image }}"
+    state: started
+    restart_policy: always
+    ports:  
+      - "{{ pulp_ports }}"
+    volumes:
+      - "{{ pulp_volumes }}"
```
evgeni, line 16:
```suggestion
    ports: "{{ pulp_ports }}"
    volumes: "{{ pulp_volumes }}"
```
both, `pulp_ports` and `pulp_volumes` are already lists, so you don't need to make them into lists here, as otherwise you're passing `ports: [ ['8080:80'] ]` here and that doesn't work.

roles/pulp/defaults/main.yaml
```diff
@@ -0,0 +1,10 @@
...
+pulp_ports:
+  - "8080:80"
+pulp_volumes:
```
evgeni, line 5: nothing creates those paths in `/var/lib/pulp`, so the deployment fails to mount them

evgeni:
I've opened https://github.com/archanaserver/foreman-quadlet/pull/1 against your fork with my changes

evgeni:
Question: this currently uses the Pulp "all in one" container, which brings own PostgreSQL etc. Should we rather use their "minimal" container and spawn it multiple times for API, Workers and Content? And connect to the main PostgreSQL we have?

ehelms:
The idea was this:

- start with the simple AIO Pulp container, get it working and learn from it
- add a basic set of tests that verify Pulp as a whole is working
- start breaking it up piece by piece
- use the existing tests to verify
- add more tests

roles/pulp/defaults/main.yaml
archanaserver: Thanks for addressing this by adding a task.

roles/pulp/defaults/main.yaml
```diff
@@ -0,0 +1,10 @@
+---
+pulp_image: quay.io/pulp/pulp:stable
+pulp_ports:
+  - "8080:80"
+pulp_volumes:
+  - /var/lib/pulp/settings:/etc/pulp:Z
+  - /var/lib/pulp/pulp_storage:/var/lib/pulp:Z
+  - /var/lib/pulp/pgsql:/var/lib/pgsql:Z
+  - /var/lib/pulp/containers:/var/lib/containers:Z
```
ehelms, line 9: `/var/lib/containers` is a path that is owned and used by already by container related tooling, I don't think we should mess with that and instead keep this scoped to `/var/lib/pulp/containers`

evgeni: That's a path *inside* the Pulp container, and that's following their docs:
https://pulpproject.org/pulp-oci-images/docs/admin/reference/available-images/multi-process-images/

roles/pulp/defaults/main.yaml
ehelms: This does not look resolved

archanaserver: Ah my bad, let me update this.

ehelms (approved):

PR #5, opened 2024-09-27
Foreman deployment

evgeni:


roles/foreman/defaults/main.yaml
```diff
@@ -0,0 +1,4 @@
+---
+foreman_image: quay.io/evgeni/foreman-rpm:3.12
```
evgeni, line 2: This is built using https://github.com/evgeni/foreman-container/blob/devel/Containerfile

roles/foreman/templates/katello.yaml.j2
```diff
@@ -0,0 +1,15 @@
...
+    :url: https://{{ ansible_fqdn }}:23443/candlepin
+    :oauth_key: "katello"
+    :oauth_secret: "{{ candlepin_oauth_secret }}"
+    :ca_cert_file: /etc/foreman/katello-default-ca.crt
+
+  :candlepin_events:
+    :broker_host: {{ ansible_fqdn }}
+    :ssl_cert_file: /etc/foreman/client_cert.pem
+    :ssl_key_file: /etc/foreman/client_key.pem
+    :ssl_ca_file: /etc/foreman/katello-default-ca.crt
```
evgeni, line 15: This seems not to work :(

```
2024-09-27T10:20:16 [E|app|6da49aa8] Error occurred while starting Katello::CandlepinEventListener
2024-09-27T10:20:16 [E|app|6da49aa8] SSL_read: ssl/tls alert certificate unknown
2024-09-27T10:20:16 [E|app|6da49aa8] /usr/share/ruby/openssl/buffering.rb:80:in `sysread'
 6da49aa8 | /usr/share/ruby/openssl/buffering.rb:80:in `fill_rbuff'
 6da49aa8 | /usr/share/ruby/openssl/buffering.rb:230:in `gets'
 6da49aa8 | /usr/share/gems/gems/stomp-1.4.10/lib/connection/netio.rb:18:in `_interruptible_gets'
 6da49aa8 | /usr/share/gems/gems/stomp-1.4.10/lib/connection/netio.rb:587:in `_init_line_read'
 6da49aa8 | /usr/share/gems/gems/stomp-1.4.10/lib/connection/netio.rb:42:in `block (2 levels) in _receive'
```

evgeni: in the candlepin container, I see
```
2024-09-27 10:34:02,927 [thread=Thread-2 (activemq-netty-threads)] [=, org=, csid=] WARN  org.apache.activemq.artemis.core.server - AMQ222208: SSL handshake failed for client from /127.0.0.1:51586: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target.
```

evgeni: when I disable auth in artemis, I get
```
2024-09-27 10:43:34,297 [thread=Thread-1 (activemq-netty-threads)] [=, org=, csid=] WARN  org.apache.activemq.artemis.core.protocol.stomp - AMQ332069: Sent ERROR frame to STOMP client 127.0.0.1:55048: AMQ339016 Error creating subscription eeaeb84eb3d3b805af4c093375c363150f85e758
```
Wha?

evgeni: got it

roles/candlepin/tasks/artemis.yml
```diff
@@ -5,7 +5,7 @@
 
 - name: Set candlepin_artemis_client_dn
   ansible.builtin.set_fact:
-    candlepin_artemis_client_dn: "{{ openssl_response.stdout | replace('subject=', '') }}"
+    candlepin_artemis_client_dn: "{{ openssl_response.stdout | replace('subject=', '') | replace(' ', '') | replace(',', ', ') }}"
```
evgeni, line 8: Otherwise you get

```
2024-09-30 07:31:55,292 [thread=Thread-1 (activemq-netty-threads)] [=, org=, csid=] WARN  org.apache.activemq.artemis.core.server - AMQ222216: Security problem while authenticating: AMQ229031: Unable to validate user from 127.0.0.1:43804. Username: null; SSL certificate subject DN: CN=quadlet.example.com, O=Foreman, ST=NC, C=US
2024-09-30 07:31:55,304 [thread=Thread-1 (activemq-netty-threads)] [=, org=, csid=] WARN  org.apache.activemq.artemis.core.protocol.stomp - AMQ332069: Sent ERROR frame to STOMP client 127.0.0.1:43804: Security Error occurred: User name [null] or password is invalid
```

roles/candlepin/tasks/certs.yml
```diff
@@ -10,6 +10,17 @@
...
+- name: Generate truststore
+  community.crypto.openssl_pkcs12:
+    action: export
+    passphrase: "{{ candlepin_keystore_password }}"
+    path: "/root/candlepin.truststore"
+    friendly_name: 'artemis-client'
+    privatekey_path: "{{ candlepin_client_key }}"
+    certificate_path: "{{ candlepin_client_certificate }}"
+    other_certificates: "{{ candlepin_ca_certificate }}"
+    state: present
```
evgeni, line 22: This is not really a truststore, more a keystore.
But when I remove `privatekey_path`, Ansible generated an empty keystore :thinking: 

But I need a store that has `artemis-client` in it.

roles/foreman/tasks/main.yaml
```diff
@@ -0,0 +1,95 @@
...
+    state: started
+
+- name: Start the Dynflow Services
+  ansible.builtin.systemd:
+    name: "dynflow-sidekiq@{{ item }}"
+    enabled: true
+    state: started
+  loop:
+    - orchestrator
+    - worker
```
evgeni, line 89: this is missing the `hosts_queue` worker, because it's not in the default config and I was too lazy to create a config yet

needed for Katello

roles/foreman-proxy/tasks/main.yaml
```diff
@@ -0,0 +1,45 @@
...
+    network: host
+    hostname: "{{ ansible_fqdn }}"
+    secrets:
+      - 'foreman-proxy-settings-yaml,type=mount,target=/etc/foreman-proxy/settings.yml'
+      - 'candlepin-ca-cert,type=mount,target=/etc/foreman-proxy/foreman_ssl_ca.pem'
+      - 'candlepin-client-cert,type=mount,target=/etc/foreman-proxy/foreman_ssl_cert.pem'
+      - 'candlepin-client-key,type=mount,target=/etc/foreman-proxy/foreman_ssl_key.pem'
+      - 'candlepin-ca-cert,type=mount,target=/etc/foreman-proxy/ssl_ca.pem'
+      - 'candlepin-client-cert,type=mount,target=/etc/foreman-proxy/ssl_cert.pem'
+      - 'candlepin-client-key,type=mount,target=/etc/foreman-proxy/ssl_key.pem'
```
evgeni, line 27: meh,
```
Unable to communicate with the proxy: ERF12-2530 [ProxyAPI::ProxyException]: Unable to detect features ([RestClient::SSLCertificateNotVerified]: SSL_connect returned=1 errno=0 state=error: certificate verify failed (unsuitable certificate purpose)) for proxy https://quadlet.tanso.example.com:9090/v2/features
Please check the proxy is configured and running on the host.
```

roles/candlepin/tasks/artemis.yml
ehelms: `replace(' ', '')` I am surprised at this one as it compresses names, e.g. "O=Red Hat" would turn into "O=RedHat" which would be incorrect I'd think? If I look at our normal deployments, I think subject should be the only problem.

```
CN=hostname, OU=PUPPET, O=FOREMAN, ST=North Carolina, C=US
```

evgeni: So the real problem I was solving is the `, ` vs `,` -- java really wants those spaces between the DN parts.

You are right the "kill all spaces" first is bad.

roles/candlepin/tasks/certs.yml
evgeni: Yeah will do.

roles/foreman-proxy/defaults/main.yaml
```diff
@@ -0,0 +1,2 @@
+---
+foreman_proxy_image: quay.io/evgeni/foreman-proxy-rpm:3.12
```
ehelms, line 2: I recall we were saying we would avoid putting smart-proxy in a container to start with. I assume you added this because you need it for Pulp to work?

roles/foreman/tasks/main.yaml
```diff
@@ -0,0 +1,114 @@
...
+  containers.podman.podman_secret:
+    state: present
+    name: candlepin-tomcat-cert
+    path: "{{ candlepin_tomcat_certificate }}"
+
+- name: Create the podman secret for Candlepin tomcat key
+  containers.podman.podman_secret:
+    state: present
+    name: candlepin-tomcat-key
+    path: "{{ candlepin_tomcat_key }}"
```
ehelms, line 47: Why are the tomcat cert/key secrets in this file?

roles/foreman/tasks/main.yaml
```diff
@@ -0,0 +1,114 @@
...
+    hostname: "{{ ansible_fqdn }}"
+    secrets:
+      - 'foreman-database-url,type=env,target=DATABASE_URL'
+      - 'foreman-settings-yaml,type=mount,target=/etc/foreman/settings.yaml'
+      - 'foreman-katello-yaml,type=mount,target=/etc/foreman/plugins/katello.yaml'
+      - 'candlepin-ca-cert,type=mount,target=/etc/foreman/katello-default-ca.crt'
+      - 'candlepin-client-cert,type=mount,target=/etc/foreman/client_cert.pem'
+      - 'candlepin-client-key,type=mount,target=/etc/foreman/client_key.pem'
+    etc_hosts:
+      "quadlet.example.com": "127.0.0.1"
```
ehelms, line 65: Rather environment specific - short term solution?

roles/foreman-proxy/defaults/main.yaml
evgeni: Correct, it is essentially a stub that provides the config to katello. Probably should redesign this as part of this effort?

roles/foreman/tasks/main.yaml
evgeni: Yes, it's a hack. Capital H.

tests/foreman_test.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+FOREMAN_HOST = 'localhost'
+FOREMAN_PORT = 3000
+
+
+def test_foreman_service(host):
+    foreman = host.service("foreman")
+    assert foreman.is_running
+    assert foreman.is_enabled
```
ehelms, line 11: I saw @ekohl commenting that we should do an assert per test and thus combine them, e.g. `assert foreman.running and foreman.is_enabled`. Which style are you in favor of?

roles/redis/defaults/main.yaml
```diff
@@ -0,0 +1,2 @@
+---
+redis_image: docker.io/library/redis:6
```
ehelms, line 2: We can figure this out later (maybe we file issues for things we find?). Is this image running a version of Redis licensed the way we'd want?

evgeni: No idea. We can replace it with valkey to be sure. I just needed something for a start.

roles/foreman/defaults/main.yaml
ehelms: Is that included in this PR?

roles/foreman-proxy/defaults/main.yaml
ehelms: How should we track that? An issue here? Or did you mean re-design more broadly?

roles/foreman/defaults/main.yaml
evgeni: Nope

roles/foreman-proxy/defaults/main.yaml
evgeni: More broadly.
Today Katello talks directly to pulp (using the Ruby bindings), also on proxies.
But it does assume `smart_proxy_pulp` is present to derive URLs etc from it.
I find having a proxy plugin for a few strings a bit much :)

ehelms: I'll let you discuss that one with @ekohl :)

ekohl: When I talked with @ianballou he was thinking about always talking through `smart_proxy_pulp`.

tests/foreman_test.py
evgeni: I guess "it depends" :D

Pytest's `assert` is smart when it comes to displaying details of the assertion. Like when you do `assert a == b` it will actually display the diff if there is any (and you don't need `assert_eql` like in other frameworks…)
No idea if it's still smart if we start chaining things with `a and b`, but we can try that.

OTOH, I don't think it brings a lot of value in this particular example and in other cases I'd rather write multiple tests that use a shared fixture:
https://github.com/evgeni/foreman-quadlet/commit/ba8b616f659d5131a2f0f2e58eebb32934ee13ad

roles/redis/defaults/main.yaml
evgeni: I checked, Redis changed the license in 7.4 -- so earlier versions are fine.

roles/candlepin/tasks/certs.yml
evgeni: https://github.com/ehelms/foreman-quadlet/pull/8

roles/foreman/defaults/main.yaml
ehelms: Do you want to move that Containerfile into this repo into the container-images directory? Or do you have thoughts on how we should manage the container images we build? (dedicated repo for just them, individual git repos for each image, etc.)

tests/foreman_test.py
ehelms: > OTOH, I don't think it brings a lot of value in this particular example and in other cases I'd rather write multiple tests that use a shared fixture:
> [evgeni@ba8b616](https://github.com/evgeni/foreman-quadlet/commit/ba8b616f659d5131a2f0f2e58eebb32934ee13ad)

This looks nice and clean.

evgeni: I've used this approach here and in #11, so enjoy? :)

roles/foreman/templates/settings.yaml.j2
```diff
@@ -0,0 +1,12 @@
+---
```
evgeni, line 1: As with proxy, I have opened #17 to track "basic config, probably needs more"

roles/foreman/tasks/main.yaml
evgeni: #18

roles/foreman/tasks/main.yaml
evgeni: I've left this one as the one "extra" commit in this PR.

Do you want to tackle the "Candlepin should use `localhost` cert if it only listens on localhost" issue here, or in a separate PR?

evgeni:
This is now as clean as it gets and ready for review!

roles/foreman/tasks/main.yaml
evgeni: Let's see: https://github.com/ehelms/foreman-quadlet/pull/19

roles/foreman/defaults/main.yaml
```diff
@@ -0,0 +1,3 @@
+---
+foreman_container_image: "quay.io/evgeni/foreman-rpm"
```
ehelms, line 2: Don't forget to add this to the repo

roles/foreman/tasks/main.yaml
```diff
@@ -0,0 +1,115 @@
+---
+- name: Pull the Foreman container image
```
ehelms, line 2: We haven't done this yet with the other services -- but I tend to agree I like the explicit pull instead of the quadlet doing it.

ehelms:
@evgeni Are you ready for this to go in as is or doing work on it still?

evgeni:
> @evgeni Are you ready for this to go in as is or doing work on it still?

This is ready!

roles/foreman/defaults/main.yaml
evgeni: https://github.com/ehelms/foreman-quadlet/pull/20

roles/foreman/defaults/main.yaml
evgeni: https://github.com/ehelms/foreman-quadlet/pull/20

ehelms (approved):

PR #6, opened 2024-09-30
add GHA based CI

evgeni:


evgeni:
green run is green: https://github.com/evgeni/foreman-quadlet/actions/runs/11109335129/job/30864437261

ad molecule: yeah, idk… molecule is only really useful when we want verification of *ansible*, while so far to me the ansible in this repo was more of a side thing?

ehelms (approved):

PR #8, opened 2024-10-01
Artemis auth

evgeni:


roles/candlepin/tasks/artemis.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Extract the DN from the client certificate
-  ansible.builtin.command: "openssl x509 -in {{ candlepin_client_certificate }} -noout -subject -nameopt rfc2253"
+  ansible.builtin.command: "openssl x509 -in {{ candlepin_client_certificate }} -noout -subject -nameopt rfc2253,sep_comma_plus_space"
```
evgeni, line 3: The tests pass without this, as they only test an SSL connection can be made, not "we could *really* auth to Artemis.

For the later, we'd need a stomp client. And I am not sure I want to dig *that* deep today.

ehelms: That's fine, our puppet module does not have that level of testing either.

ehelms (approved):

PR #12, opened 2024-10-02
wait for candlepin

evgeni:


evgeni:
this was discovered in #11

ehelms (approved):

ehelms:
I'm fine with this approach for now. We do not do this in the puppet modules right? So what's the driver to do it here?

evgeni:
> I'm fine with this approach for now. We do not do this in the puppet modules right? So what's the driver to do it here?

Right now: the tests are scheduled when tomcat is still booting (while the systemd unit returned "ready").
I am not sure this is a difference between systemd native vs container, or something else tho.

PR #16, opened 2024-10-03
Update Candlepin to use podman secrets

ehelms:


ehelms:
This is ready for review now

roles/candlepin/tasks/artemis.yml
```diff
@@ -7,42 +7,32 @@
...
-  ansible.builtin.template:
-    src: broker.xml.j2
-    dest: /etc/candlepin/broker.xml
-    owner: root
-    group: tomcat
-    mode: '0640'
+- name: Create Candlepin broker.xml
+  containers.podman.podman_secret:
+    state: present
+    name: broker.xml
```
evgeni, line 13: should this be `candlepin-broker.xml`? or, more generally, `${role}-${file}`?

gives us a bit of a namespace (imagine two services having a settings.yaml) and ease the "huh, where the f does this secret come from" when you look at `podman secrets ls`

ehelms: Yea - that's fair. Take a look at the example here: https://github.com/ehelms/foreman-quadlet/pull/16/files#diff-6e185dcb895442d3f858c32b6dbd8fbe1854d41eeabe02378b89cc7fa0d0cca3R22-R26

I was thinking context will be helpful, and that we can use Labels to provide context. These are not visible through `podman secret ls` but they are visible when using `podman secret inspect`. I captured my thoughts of conventions to start:

https://github.com/ehelms/foreman-quadlet/pull/16/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R47-R62

roles/candlepin/tasks/artemis.yml
```diff
@@ -7,42 +7,47 @@
...
-    group: tomcat
-    mode: '0640'
+- name: Create Candlepin broker.xml
+  containers.podman.podman_secret:
+    state: present
+    name: candlepin-artemis-broker-xml
+    data: "{{ lookup('ansible.builtin.template', 'broker.xml.j2') }}"
+    labels:
+      filename: broker.xml
+      app: artemis
```
evgeni, line 17: is "artemis" an app? or is it "candlepin-artemis" and "candlepin-tomcat"?

OTOH artemis runs inside tomcat too (even if not explicitly as an web-app…)

complicated.

roles/candlepin/tasks/artemis.yml
evgeni: I like that! Commented some bikeshedding, but that doesn't need an absolute resolution just now.

roles/candlepin/tasks/artemis.yml
ehelms: Yea, it's complicated... I wanted to express why it exists, what consumes it. So while it runs inside Tomcat, it's because of the internal Artemis.

roles/candlepin/tasks/main.yml
```diff
@@ -1,4 +1,54 @@
...
+  - /var/log/candlepin
+  - /var/log/tomcat
+
+- ansible.builtin.include_tasks:
+    file: certs.yml
+
+- name: Create Candlepin configuration
+  containers.podman.podman_secret:
+    state: present
+    name: candlepin-candlepin-conf
```
evgeni, line 19: Wouldn't this be candlepin(role)-candlepin(app)-candlepin(file)-conf?

(It probably shouldn't, for my sanitys sake, but that's how I read your naming convention)

I wonder if the role prefix is necessary for all the cases where rolename==appname?

evgeni (approved):

roles/candlepin/tasks/main.yml
ehelms: > Wouldn't this be candlepin(role)-candlepin(app)-candlepin(file)-conf?
> 
> (It probably shouldn't, for my sanitys sake, but that's how I read your naming convention)
> 
> I wonder if the role prefix is necessary for all the cases where rolename==appname?

This ^ and I updated the README to indicate so

PR #22, opened 2024-10-04
Exclude container-images path from current CI tests

ehelms:


ehelms:
We should add tests in this area, but for now, let's just make things more efficient (https://github.com/ehelms/foreman-quadlet/issues/23)

evgeni:
Shouldn't this PR have triggered tests still?

ehelms:
Should it?

evgeni:
The PR changes files outside of container images, so yes

.github/workflows/test.yml
```diff
@@ -5,7 +5,11 @@ on:
   push:
     branches:
       - master
+    paths:
+      - '!container-images/**'
```
evgeni, line 9:
```suggestion
    paths-ignore:
      - 'container-images/**'
```
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

the bang-syntax is only to remove previously included paths, not to ignore whole structures

.github/workflows/test.yml
```diff
@@ -5,7 +5,11 @@ on:
   push:
     branches:
       - master
+    paths-ignore:
+      - 'container-images/**'
   pull_request:
+    paths:
+      - '!container-images/**'
```
evgeni, line 12:
```suggestion
    paths-ignore:
      - 'container-images/**'
```

evgeni (approved):

PR #24, opened 2024-10-04
Reduce Container sizes

Odilhao:
Before adding `--nodocs` and `dnf clean all`

```
REPOSITORY              TAG         IMAGE ID      CREATED         SIZE
quay.io/ehelms/foreman  nightly     a20e228905bb  10 seconds ago  884 MB
quay.io/ehelms/foreman-proxy  nightly     2d3a73f1ded9  29 seconds ago  458 MB


```

After `--nodocs`` and `dnf clean all`

```
REPOSITORY              TAG             IMAGE ID      CREATED        SIZE
quay.io/ehelms/foreman  nightly-nodocs  ab9ff2fb57a0  5 seconds ago  774 MB
quay.io/ehelms/foreman-proxy  nightly-nodocs  dfe53a6f8f3c  12 minutes ago  351 MB
```

### DIFF
--- a/container-images/foreman-proxy/Containerfile
+++ b/container-images/foreman-proxy/Containerfile
@@ -4,9 +4,10 @@ ARG FOREMAN_VERSION=nightly
 ARG KATELLO_VERSION=nightly
 ARG FOREMAN_PROXY_PLUGINS="pulp"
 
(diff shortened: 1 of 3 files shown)


Odilhao:
I was thinking about a way to have each plugin installed here https://github.com/ehelms/foreman-quadlet/blob/b19ae38891af8971da78ec80781a99b554d81ccd/container-images/foreman/Containerfile#L9 in it's own layer, this would make pull faster in the future when we add more plugins, I'm still not sure on how to do that.

evgeni (approved):

PR #25, opened 2024-10-07
configure pulp proxy integration

evgeni:


roles/pulp/tasks/main.yaml
```diff
@@ -37,3 +37,9 @@
     host: "{{ ansible_hostname }}"
     port: 8080
     timeout: 300
+
+# Only needed until we have cert auth configured
```
evgeni, line 41: Split out to #26

roles/foreman/tasks/main.yaml
```diff
@@ -113,3 +113,11 @@
   register: foreman_tasks_status
   when:
     - "'katello' in foreman_status.json['results']"
+
+- name: Configure Foreman Proxy
+  theforeman.foreman.smart_proxy:
```
ehelms, line 118: That's fun!

ehelms (approved):
After rebase

roles/foreman/tasks/main.yaml
evgeni: You mean dogfooding?

PR #27, opened 2024-10-07
add httpd

evgeni:


roles/httpd/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
...
+    path: /etc/httpd/conf.d/welcome.conf
+    state: absent
+
+- name: Create cert directories
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: directory
+  loop:
+    - /etc/pki/katello/certs
+    - /etc/pki/katello/private
```
ehelms, line 26: I'm OK with this for parity, but I would love to change this location.
I suppose if we were to run Apache as a container these would be secrets and it'd be a moot point.

roles/httpd/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
+---
+- name: Install Apache httpd
```
ehelms, line 2: We could use `ubi9/httpd-24` but based on the list here we'd have to do some minor customization of it?

roles/httpd/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
...
+      dest: "private/katello-apache.key"
+
+- name: Configure foreman-ssl vhost
+  ansible.builtin.template:
+    src: foreman-ssl-vhost.conf.j2
+    dest: /etc/httpd/conf.d/foreman-ssl.conf
+  vars:
+    pulp_api_backend: http://localhost:8080/
+    pulp_content_backend: http://localhost:8080/
+    foreman_backend: http://localhost:3000/
```
ehelms, line 50: Do you think we need to worry about making these variables or stick with a "static / opinionated" implementation?

roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -0,0 +1,104 @@
+<VirtualHost *:443>
+ServerName {{ ansible_fqdn }}
```
ehelms, line 2:
```suggestion
  ServerName {{ ansible_fqdn }}
```

roles/httpd/tasks/main.yml
evgeni: Tell me a good location and I'll move :)

roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -0,0 +1,104 @@
...
+    RequestHeader unset REMOTE_USER
+    RequestHeader set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN_CN
+    ProxyPass {{ pulp_api_backend }}/v2/
+    ProxyPassReverse {{ pulp_api_backend }}/v2/
+  </Location>
+
+  ProxyPass /pulp/container/ {{ pulp_content_backend }}/pulp/container/
+  ProxyPassReverse /pulp/container/ {{ pulp_content_backend }}/pulp/container/
+
+  <Location "/pulp/isos">
```
ehelms, line 40: Reminds me that this doesn't hurt anything, but we should track removing it.

roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -0,0 +1,104 @@
...
+    RequestHeader unset X-CLIENT-CERT
+    RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT
+    ProxyPass {{ pulp_content_backend }}/pulp/content disablereuse=on timeout=600
+    ProxyPassReverse {{ pulp_content_backend }}/pulp/content
+  </Location>
+
+  <Location "/pulp/repos">
```
ehelms, line 47: Reminds me that this doesn't hurt anything, but we should track removing it.

roles/httpd/tasks/main.yml
evgeni: I think I'd make them variables, yes.

roles/httpd/templates/foreman-ssl-vhost.conf.j2
evgeni: yeah, I literally copied over the 05-ssl.conf from an existing machine as a template :see_no_evil:

roles/httpd/tasks/main.yml
ehelms: Normalization of services, configurations mostly.

roles/httpd/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+httpd_ssl_dir: /etc/pki/katello
```
ehelms, line 1: Ideas:

- /etc/pki/httpd
- /etc/pki/tls

roles/httpd/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+httpd_ssl_dir: /etc/pki/httpd
+httpd_pulp_api_backend: http://localhost:8080
+httpd_pulp_content_backend: http://localhost:8080
+httpd_foreman_backend: http://localhost:3000
```
ehelms, line 4: I read a little bit, but seemed that Unix sockets were unfriendly in his container environment with podman and definitely do not extend to Openshift. Is that your read on it as well?

evgeni: That's also my thinking here, yes.

ehelms (approved):

roles/httpd/defaults/main.yml
ekohl: We have proxy issues (HTTP 502 IIRC) when we used HTTP that went away when we switched to UNIX sockets. We also saw those issues with proxies from content proxies that went away with HTTP/2. Does Puma support HTTP/2 or how can we avoid introducing those issues again?

Also, you can actually use socket activation with containers where systemd opens the socket on the host and passes the file descriptor to a container. I use that with gunicorn (on something that isn't Pulp) and works well for me. Should we investigate that?

tests/httpd_test.py
```diff
@@ -0,0 +1,37 @@
...
+    assert httpd.port(HTTP_PORT).is_reachable
+
+
+def test_https_port(host):
+    httpd = host.addr(HTTP_HOST)
+    assert httpd.port(HTTPS_PORT).is_reachable
+
+
+def test_https_foreman_ping(host):
+    cmd = host.run('curl --cacert /root/certificates/certs/ca.crt --silent --output /dev/null --write-out \'%{http_code}\' https://quadlet.example.com/api/v2/ping')
```
ekohl, line 23: I'd like to see more fixtures, like the CA file and hostname.

roles/httpd/defaults/main.yml
ehelms: Moved to https://github.com/ehelms/foreman-quadlet/issues/31 to allow continued discussion

PR #29, opened 2024-10-07
configure cert auth for pulp

evgeni:


evgeni:
contains #27 and #29 for (obvious?) reasons :)

evgeni:
huh, you can't tell `smart_proxy_pulp` which CA to use, so if the CA is not imported into the system store, things fail.
fun.

evgeni:
:green_apple: 

keeping it draft as I think we shouldn't be using `nightly` images -- unless you're OK with that?

ehelms:
> 🍏
> 
> keeping it draft as I think we shouldn't be using `nightly` images -- unless you're OK with that?

I'm OK with that.

ehelms (approved):

PR #34, opened 2024-10-15
add pulp services

archanaserver:
implementation of pulp services(api, content and worker)

### DIFF
--- a/roles/pulp/defaults/main.yaml
+++ b/roles/pulp/defaults/main.yaml
@@ -1,10 +1,20 @@
 ---
-pulp_image: quay.io/pulp/pulp:stable
-pulp_ports:
-  - "8080:80"
+pulp_api_image: quay.io/pulp/pulp-minimal:stable
+pulp_content_image: quay.io/pulp/pulp-minimal:stable
+pulp_worker_image: quay.io/pulp/pulp-minimal:stable
+
+pulp_api_ports:
+  - "24817:80"
+pulp_content_ports:
+  - "24816:80"
+pulp_worker_count: 2
+
 pulp_volumes:
   - /var/lib/pulp/settings:/etc/pulp:Z
   - /var/lib/pulp/pulp_storage:/var/lib/pulp:Z
   - /var/lib/pulp/pgsql:/var/lib/pgsql:Z
   - /var/lib/pulp/containers:/var/lib/containers:Z
-pulp_container_name: pulp
+
+pulp_api_container_name: pulp-api
+pulp_content_container_name: pulp-content
+pulp_worker_container_name: pulp-worker
--- a/roles/pulp/defaults/main.yaml
+++ b/roles/pulp/defaults/main.yaml
@@ -11,7 +11,7 @@ pulp_content_ports:
 pulp_worker_count: 2
 
 pulp_volumes:
-  - /var/lib/pulp/pulp_storage:/var/lib/pulp
+  - /var/lib/pulp:/var/lib/pulp
 
 pulp_api_container_name: pulp-api
 pulp_content_container_name: pulp-content
--- a/roles/pulp/defaults/main.yaml
+++ b/roles/pulp/defaults/main.yaml
@@ -4,10 +4,6 @@ pulp_api_image: "{{ pulp_image }}"
 pulp_content_image: "{{ pulp_image }}"
 pulp_worker_image: "{{ pulp_image }}"
 
-pulp_api_ports:
-  - "24817:80"
-pulp_content_ports:
-  - "24816:80"
 pulp_worker_count: 2
 
 pulp_volumes:
--- a/roles/pulp/defaults/main.yaml
+++ b/roles/pulp/defaults/main.yaml
@@ -10,10 +10,8 @@ pulp_content_ports:
 pulp_worker_count: 2
 
 pulp_volumes:
-  - /var/lib/pulp/settings:/etc/pulp:Z
-  - /var/lib/pulp/pulp_storage:/var/lib/pulp:Z
-  - /var/lib/pulp/pgsql:/var/lib/pgsql:Z
-  - /var/lib/pulp/containers:/var/lib/containers:Z
+  - /var/lib/pulp/pulp_storage:/var/lib/pulp
+  - /var/lib/pulp/containers:/var/lib/containers
 
 pulp_api_container_name: pulp-api
 pulp_content_container_name: pulp-content
--- a/roles/pulp/defaults/main.yaml
+++ b/roles/pulp/defaults/main.yaml
@@ -1,7 +1,8 @@
 ---
-pulp_api_image: quay.io/pulp/pulp-minimal:stable
-pulp_content_image: quay.io/pulp/pulp-minimal:stable
-pulp_worker_image: quay.io/pulp/pulp-minimal:stable
+pulp_image: quay.io/pulp/pulp-minimal:stable
+pulp_api_image: "{{ pulp_image }}"
+pulp_content_image: "{{ pulp_image }}"
+pulp_worker_image: "{{ pulp_image }}"
 
 pulp_api_ports:
   - "24817:80"
@@ -11,7 +12,6 @@ pulp_worker_count: 2
 
 pulp_volumes:
   - /var/lib/pulp/pulp_storage:/var/lib/pulp
-  - /var/lib/pulp/containers:/var/lib/containers
 
 pulp_api_container_name: pulp-api
 pulp_content_container_name: pulp-content
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -40,11 +40,16 @@
   args:
     creates: /var/lib/pulp/database_fields.symmetric.key
 
+- name: Load database symmetric key
+  ansible.builtin.slurp:
+    src: /var/lib/pulp/database_fields.symmetric.key
+  register: pulp_key
+
 - name: Create database symmetric key secret
   containers.podman.podman_secret:
     state: present
     name: pulp-symmetric-key
-    data: "{{ lookup('file', '/var/lib/pulp/database_fields.symmetric.key') }}"
+    data: "{{ pulp_key['content'] | b64decode }}"
 
 - name: Wait for PostgreSQL to be ready
   ansible.builtin.wait_for:
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -51,6 +51,8 @@
     command: pulp-api
     ports: "{{ pulp_api_ports }}"
     volumes: "{{ pulp_volumes }}"
+    security_opt:
+      - "label=disable"
     secrets:
       - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
       - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
@@ -67,6 +69,8 @@
     command: pulp-content
     ports: "{{ pulp_content_ports }}"
     volumes: "{{ pulp_volumes }}"
+    security_opt:
+      - "label=disable"
     secrets:
       - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
       - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
@@ -88,6 +92,8 @@
     state: quadlet
     command: pulp-worker
     volumes: "{{ pulp_volumes }}"
+    security_opt:
+      - "label=disable"
     secrets:
       - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
       - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -127,16 +127,21 @@
   ansible.builtin.systemd:
     daemon_reload: true
 
+- name: Migrate the Pulp database
+  containers.podman.podman_container:
+    name: pulpcore-manager-migrate
+    image: "{{ pulp_api_image }}"
+    command: pulpcore-manager migrate --noinput
+    network: host
+    secrets:
+      - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
+      - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
+
 - name: Start the Pulp API services
   ansible.builtin.systemd:
     name: pulp-api
     enabled: true
     state: started
-
-- name: Migrate the Pulp database
-  containers.podman.podman_container_exec:
-    name: "{{ pulp_api_container_name }}"
-    command: pulpcore-manager migrate --noinput
 
 - name: Wait for Pulp API service to be accessible
   ansible.builtin.wait_for:
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -127,6 +127,11 @@
     enabled: true
     state: started
 
+- name: Migrate the Pylp database
+  containers.podman.podman_container_exec:
+    name: "{{ pulp_api_container_name }}"
+    command: pulpcore-manager migrate --noinput
+
 - name: Start the Pulp Content services
   ansible.builtin.systemd:
     name: pulp-content
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -63,7 +63,7 @@
     image: "{{ pulp_api_image }}"
     state: quadlet
     command: pulp-api
-    ports: "{{ pulp_api_ports }}"
+    network: host
     volumes: "{{ pulp_volumes }}"
     security_opt:
       - "label=disable"
@@ -81,7 +81,7 @@
     image: "{{ pulp_content_image }}"
     state: quadlet
     command: pulp-content
-    ports: "{{ pulp_content_ports }}"
+    network: host
     volumes: "{{ pulp_volumes }}"
     security_opt:
       - "label=disable"
@@ -105,6 +105,7 @@
     image: "{{ pulp_worker_image }}"
     state: quadlet
     command: pulp-worker
+    network: host
     volumes: "{{ pulp_volumes }}"
     security_opt:
       - "label=disable"
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -20,6 +20,15 @@
     mode: "0755"
   loop: "{{ pulp_volumes }}"
 
+- name: Create Pulp storage subdirs
+  ansible.builtin.file:
+    path: "/var/lib/pulp/pulp_storage/{{ item }}"
+    state: directory
+    mode: "0755"
+  loop:
+    - tmp
+    - assets
+
 - name: Create settings config secret
   containers.podman.podman_secret:
     state: present
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -1,6 +1,16 @@
-- name: Pull the Pulp container image
+- name: Pull the Pulp API container image
   containers.podman.podman_image:
-    name: "{{ pulp_image }}"
+    name: "{{ pulp_api_image }}"
+    state: present
+
+- name: Pull the Pulp Content container image
+  containers.podman.podman_image:
+    name: "{{ pulp_content_image }}"
+    state: present
+
+- name: Pull the Pulp Worker container image
+  containers.podman.podman_image:
+    name: "{{ pulp_worker_image }}"
     state: present
 
 - name: Create Pulp storage
@@ -16,15 +26,71 @@
     name: pulp-settings-py
     data: "{{ lookup('ansible.builtin.template', 'settings.py.j2') }}"
 
-- name: Deploy Pulp Container
+- name: Generate database symmetric key
+  ansible.builtin.command: "bash -c 'openssl rand -base64 32 | tr \"+/\" \"-_\" > /var/lib/pulp/database_fields.symmetric.key'"
+  args:
+    creates: /var/lib/pulp/database_fields.symmetric.key
+
+- name: Create database symmetric key secret
+  containers.podman.podman_secret:
+    state: present
+    name: pulp-symmetric-key
+    data: "{{ lookup('file', '/var/lib/pulp/database_fields.symmetric.key') }}"
+
+- name: Wait for PostgreSQL to be ready
+  ansible.builtin.wait_for:
+    host: "localhost"
+    port: 5432
+    timeout: 300
+
+- name: Deploy Pulp API Container
   containers.podman.podman_container:
-    name: "{{ pulp_container_name }}"
-    image: "{{ pulp_image }}"
+    name: "{{ pulp_api_container_name }}"
+    image: "{{ pulp_api_image }}"
     state: quadlet
-    ports: "{{ pulp_ports }}"
+    command: pulp-api
+    ports: "{{ pulp_api_ports }}"
     volumes: "{{ pulp_volumes }}"
     secrets:
       - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
+      - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
+    quadlet_options:
+      - |
+        [Install]
+        WantedBy=default.target
+
+- name: Deploy Pulp Content Container
+  containers.podman.podman_container:
+    name: "{{ pulp_content_container_name }}"
+    image: "{{ pulp_content_image }}"
+    state: quadlet
+    command: pulp-content
+    ports: "{{ pulp_content_ports }}"
+    volumes: "{{ pulp_volumes }}"
+    secrets:
+      - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
+      - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
+    quadlet_options:
+      - |
+        [Install]
+        WantedBy=default.target
+
+- name: Wait for Pulp API service to be accessible
+  ansible.builtin.wait_for:
+    host: "{{ ansible_hostname }}"
+    port: 24817
+    timeout: 300
+
+- name: Deploy Pulp Worker Container
+  containers.podman.podman_container:
+    name: "{{ pulp_worker_container_name }}"
+    image: "{{ pulp_worker_image }}"
+    state: quadlet
+    command: pulp-worker
+    volumes: "{{ pulp_volumes }}"
+    secrets:
+      - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
+      - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
     quadlet_options:
       - |
         [Install]
@@ -34,17 +100,29 @@
   ansible.builtin.systemd:
     daemon_reload: true
 
-- name: Start the Pulp Service
+- name: Start the Pulp API services
   ansible.builtin.systemd:
-    name: pulp
+    name: pulp-api
     enabled: true
-    state: restarted
+    state: started
 
-- name: Wait for Pulp service to be accessible
+- name: Start the Pulp Content services
+  ansible.builtin.systemd:
+    name: pulp-content
+    enabled: true
+    state: started
+
+- name: Wait for Pulp Content service to be accessible
   ansible.builtin.wait_for:
     host: "{{ ansible_hostname }}"
-    port: 8080
-    timeout: 300
+    port: 24816
+    timeout: 600
+
+- name: Start the Pulp Worker service
+  ansible.builtin.systemd:
+    name: pulp-worker
+    enabled: true
+    state: started
 
 # Only needed until we have cert auth configured
 - name: Set Pulp admin password
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -75,6 +75,9 @@
         [Install]
         WantedBy=default.target
         Wants=postgresql.service
+        [Service]
+        Restart=always
+        RestartSec=3
 
 - name: Deploy Pulp Content Container
   containers.podman.podman_container:
@@ -94,6 +97,9 @@
         [Install]
         WantedBy=default.target
         Wants=postgresql.service
+        [Service]
+        Restart=always
+        RestartSec=3
 
 - name: Deploy Pulp Worker Container
   containers.podman.podman_container:
@@ -113,6 +119,9 @@
         [Install]
         WantedBy=default.target
         Wants=postgresql.service
+        [Service]
+        Restart=always
+        RestartSec=3
 
 - name: Run daemon reload to make Quadlet create the service files
   ansible.builtin.systemd:
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -74,6 +74,7 @@
       - |
         [Install]
         WantedBy=default.target
+        Wants=postgresql.service
 
 - name: Deploy Pulp Content Container
   containers.podman.podman_container:
@@ -92,6 +93,7 @@
       - |
         [Install]
         WantedBy=default.target
+        Wants=postgresql.service
 
 - name: Deploy Pulp Worker Container
   containers.podman.podman_container:
@@ -110,6 +112,7 @@
       - |
         [Install]
         WantedBy=default.target
+        Wants=postgresql.service
 
 - name: Run daemon reload to make Quadlet create the service files
   ansible.builtin.systemd:
@@ -121,7 +124,7 @@
     enabled: true
     state: started
 
-- name: Migrate the Pylp database
+- name: Migrate the Pulp database
   containers.podman.podman_container_exec:
     name: "{{ pulp_api_container_name }}"
     command: pulpcore-manager migrate --noinput
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -153,5 +153,5 @@
 # Only needed until we have cert auth configured
 - name: Set Pulp admin password
   containers.podman.podman_container_exec:
-    name: "{{ pulp_container_name }}"
+    name: "{{ pulp_api_container_name }}"
     command: pulpcore-manager reset-admin-password --password CHANGEME
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -22,7 +22,7 @@
 
 - name: Create Pulp storage subdirs
   ansible.builtin.file:
-    path: "/var/lib/pulp/pulp_storage/{{ item }}"
+    path: "/var/lib/pulp/{{ item }}"
     state: directory
     mode: "0755"
   loop:
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -93,12 +93,6 @@
         [Install]
         WantedBy=default.target
 
-- name: Wait for Pulp API service to be accessible
-  ansible.builtin.wait_for:
-    host: "{{ ansible_hostname }}"
-    port: 24817
-    timeout: 300
-
 - name: Deploy Pulp Worker Container
   containers.podman.podman_container:
     name: "{{ pulp_worker_container_name }}"
@@ -131,6 +125,12 @@
   containers.podman.podman_container_exec:
     name: "{{ pulp_api_container_name }}"
     command: pulpcore-manager migrate --noinput
+
+- name: Wait for Pulp API service to be accessible
+  ansible.builtin.wait_for:
+    host: "{{ ansible_hostname }}"
+    port: 24817
+    timeout: 300
 
 - name: Start the Pulp Content services
   ansible.builtin.systemd:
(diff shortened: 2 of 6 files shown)


archanaserver:
There are few things which I don't understand clearly yet, first how can we ensure dependencies here like we have `depends_on` in compose way to do it https://github.com/pulp/pulp-oci-images/blob/latest/images/compose/compose.yml#L93. I didn't find the much info. Another things is with multiple services that have dependencies on each other, how should secrets be managed? What is the recommended way to configure secrets to ensure secure communication between these services?
Also how when testing the deployment manually like podman run way, it works as expected. However, when using my Ansible configuration, I encounter a timeout error while waiting for the pulp-api service to be accessible.

ehelms:
* Dependency handling

This could be handled via systemd `Wants/Requires` and the `After` option to provide some ordering. I do feel we should also make the container services smart enough to check for connections before performing initialization operations like migrating. You can see examples of this last part (https://github.com/pulp/pulp-oci-images/blob/latest/images/assets/pulp-api#L3-L4) (https://github.com/pulp/pulp-oci-images/blob/latest/images/assets/wait_on_postgres.py)

* Secrets handling

We are using podman secrets right now, and I recently added to the repository a naming scheme for how to define the secrets (https://github.com/ehelms/foreman-quadlet?tab=readme-ov-file#naming-convention). For services that need the same secret, we can define the secret once with podman secrets and then it can be declared within each quadlet file. See https://github.com/ehelms/foreman-quadlet/blob/master/roles/candlepin/tasks/main.yml#L59-L71

* Timeout issue

I am not sure about this. If you ssh into the VM and try to start it manually what happens?

evgeni:
* Dependency / Secrets -- I think Eric answered it well already.
* Timeout - could it be that the DB doesn't get migrated (or not migrated in time) and thus services don't start? The compose file you linked has an explicit `migration_service` with runs `pulpcore-manager migrate --noinput` and the other services depend on.

roles/pulp/tasks/main.yaml
```diff
@@ -41,6 +74,72 @@
...
+  containers.podman.podman_container:
+    name: "{{ pulp_worker_container_name }}"
+    image: "{{ pulp_worker_image }}"
+    state: quadlet
+    volumes: "{{ pulp_volumes }}"
+    secrets:
+      - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
+    quadlet_options:
+      - |
+        After=network.target pulp-api.service pulp-content.service
```
evgeni, line 110:
```suggestion
        [Unit]
        After=network.target pulp-api.service pulp-content.service
```

roles/pulp/tasks/main.yaml
```diff
@@ -24,6 +39,24 @@
...
+    quadlet_options:
+      - |
+        [Install]
+        WantedBy=default.target
+  
+- name: Deploy Pulp API Container
+  containers.podman.podman_container:
+    name: "{{ pulp_api_container_name }}"
+    image: "{{ pulp_api_image }}"
+    state: quadlet
```
evgeni, line 51: as we just found out, those containers don't have a default command, so they don't start.
if you add `command: pulp-api` here, it should generate a container file with `Exec=pulp-api` and that should be able to start then (same for the worker and content containers)

the startup then will still fail on DB things, but at least it tries to start something now.

archanaserver: yes, i'm on doing db setup

ehelms:
We'll want to drop the all-in-one Pulp container /service https://github.com/theforeman/foreman-quadlet/pull/34/files#diff-6d531cd8425f805f75bfb941463407fb86c42731bd3036def1633d683def9cf0R34

roles/pulp/tasks/main.yaml
```diff
@@ -46,6 +76,75 @@
     port: 8080
     timeout: 300
 
+- name: Wait for Pulp API service to be accessible
```
ehelms, line 79: This code is failing to find the pulp-api port because the service isn't started until further down the file (https://github.com/theforeman/foreman-quadlet/pull/34/files#diff-6d531cd8425f805f75bfb941463407fb86c42731bd3036def1633d683def9cf0R124).

Line 67-77 are starting the old all-in-one and not the `pulp-api` service so this will never pass in it's current spot.

ehelms:
This fixed some issues for me:

```
diff --git a/roles/pulp/tasks/main.yaml b/roles/pulp/tasks/main.yaml
index c7b019c..87c5e23 100644
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -31,20 +31,6 @@
     name: pulp-settings-py
     data: "{{ lookup('ansible.builtin.template', 'settings.py.j2') }}"
 
-- name: Deploy Pulp Container
diff --git a/roles/pulp/tasks/main.yaml b/roles/pulp/tasks/main.yaml
index c7b019c..87c5e23 100644
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -31,20 +31,6 @@
     name: pulp-settings-py
     data: "{{ lookup('ansible.builtin.template', 'settings.py.j2') }}"
 
-- name: Deploy Pulp Container
-  containers.podman.podman_container:
-    name: "{{ pulp_container_name }}"
-    image: "{{ pulp_image }}"
-    state: quadlet
-    ports: "{{ pulp_ports }}"
-    volumes: "{{ pulp_volumes }}"
-    secrets:
-      - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
-    quadlet_options:
-      - |
-        [Install]
-        WantedBy=default.target
-  
 - name: Deploy Pulp API Container
@@ -60,28 +46,6 @@
         [Install]
         WantedBy=default.target
 
-- name: Run daemon reload to make Quadlet create the service files
-  ansible.builtin.systemd:
-    daemon_reload: true
-
-- name: Start the Pulp Service
-  ansible.builtin.systemd:
-    name: pulp
-    enabled: true
-    state: restarted
-
-- name: Wait for Pulp service to be accessible
-  ansible.builtin.wait_for:
-    host: "{{ ansible_hostname }}"
-    port: 8080
-    timeout: 300
-
-- name: Wait for Pulp API service to be accessible
-  ansible.builtin.wait_for:
-    host: "{{ ansible_hostname }}"
-    port: 24817 
-    timeout: 600
-
 - name: Deploy Pulp Content Container
     name: "{{ pulp_content_container_name }}"
@@ -133,6 +97,12 @@
     enabled: true
     state: started
 
+- name: Wait for Pulp API service to be accessible
+  ansible.builtin.wait_for:
+    host: "{{ ansible_hostname }}"
+    port: 24817
+    timeout: 600
+
 - name: Wait for Pulp Content service to be accessible
   ansible.builtin.wait_for:
     host: "{{ ansible_hostname }}"
```

The `pulp-api` service still fails to be ready but with this error that can be fixed:

```
Oct 28 13:54:51 quadlet.example.com pulp-api[74976]: django.core.exceptions.ImproperlyConfigured: Could not load DB_ENCRYPTION_KEY file '/etc/pulp/certs/database_fields.symmetric.key': [Errno 13] Permission denied: '/etc/pulp/certs/database_fields.symmetric.key'
```

archanaserver:
> The `pulp-api` service still fails to be ready but with this error that can be fixed:
> 
> ```
> Oct 28 13:54:51 quadlet.example.com pulp-api[74976]: django.core.exceptions.ImproperlyConfigured: Could not load DB_ENCRYPTION_KEY file '/etc/pulp/certs/database_fields.symmetric.key': [Errno 13] Permission denied: '/etc/pulp/certs/database_fields.symmetric.key'
> ```

I did some config setup to resolve it, but I still see the pulp-api.service up and then failing due to the same issue, I'm not sure why this is happening, Can anyone point me why this is still persist?

roles/pulp/tasks/main.yaml
```diff
@@ -46,6 +76,75 @@
...
+    command: pulp-content
+    ports: "{{ pulp_content_ports }}"
+    volumes: "{{ pulp_volumes }}"
+    secrets:
+      - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
+    quadlet_options:
+      - |
+        [Unit]
+        After=network.target pulp-api.service
+        Requires=pulp-api.service
```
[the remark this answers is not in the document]
archanaserver: I did read this somewhere, so added the dependency part but I can't find it for reference at this moment so dropping it.

roles/pulp/templates/settings.py.j2
```diff
@@ -1,7 +1,21 @@
...
 REDIS_HOST="localhost"
 REDIS_PORT=6379
+
+DATABASES = {
+	'default': {
+		'ENGINE': 'django.db.backends.postgresql',
+		'NAME': 'pulp',
+		'USER': 'pulp',
+		'PASSWORD': '{{ pulp_db_password }}',
+		'HOST': 'localhost',
```
evgeni, line 14: Right now, the pulp containers are deployed with a dedicated (tunneled/bridged) network, which means "localhost" is the container, not the whole VM we're running in, so it won't be able to reach the database.

For Foreman (and Candlepin), we configure the containers to use "host networking":
https://github.com/theforeman/foreman-quadlet/blob/98f6a4110998d2f36257b9b18415b76571a2f6c4/roles/foreman/tasks/main.yaml#L54
This allows the containers to reach the whole VM and all services running on it. This has the "downside" that you don't need to expose any ports as the container can just open ports on the host.
I am not sure this will trivially work with the Pulp containers (no idea how they expose the various HTTP services they offer), but might make the DB setup easier.

evgeni:
> > The `pulp-api` service still fails to be ready but with this error that can be fixed:
> > ```
> > Oct 28 13:54:51 quadlet.example.com pulp-api[74976]: django.core.exceptions.ImproperlyConfigured: Could not load DB_ENCRYPTION_KEY file '/etc/pulp/certs/database_fields.symmetric.key': [Errno 13] Permission denied: '/etc/pulp/certs/database_fields.symmetric.key'
> > ```
> 
> I did some config setup to resolve it, but I still see the pulp-api.service up and then failing due to the same issue, I'm not sure why this is happening, Can anyone point me why this is still persist?

I think I understand the issue now.

When you look at the whole output (via `journalctl`), the error is:
```
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: Waiting on postgresql to start...
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: Traceback (most recent call last):
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/asgiref/local.py", line 89, in _lock_storage
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     asyncio.get_running_loop()
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: RuntimeError: no running event loop
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: 
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: During handling of the above exception, another exception occurred:
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: 
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: Traceback (most recent call last):
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/utils/connection.py", line 58, in __getitem__
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     return getattr(self._connections, alias)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/asgiref/local.py", line 118, in __getattr__
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     return getattr(storage, key)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: AttributeError: '_thread._local' object has no attribute 'default'
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: 
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: During handling of the above exception, another exception occurred:
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: 
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: Traceback (most recent call last):
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/pulpcore/app/settings.py", line 476, in <module>
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     with open(DB_ENCRYPTION_KEY, "rb") as key_file:
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: PermissionError: [Errno 13] Permission denied: '/etc/pulp/certs/database_fields.symmetric.key'
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: 
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: During handling of the above exception, another exception occurred:
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: 
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: Traceback (most recent call last):
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/bin/wait_on_postgres.py", line 12, in <module>
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     connection.ensure_connection()
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/utils/connection.py", line 15, in __getattr__
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     return getattr(self._connections[self._alias], item)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/utils/connection.py", line 60, in __getitem__
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     if alias not in self.settings:
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/utils/functional.py", line 57, in __get__
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     res = instance.__dict__[self.name] = self.func(instance)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/utils/connection.py", line 45, in settings
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     self._settings = self.configure_settings(self._settings)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/db/utils.py", line 148, in configure_settings
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     databases = super().configure_settings(databases)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/utils/connection.py", line 50, in configure_settings
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     settings = getattr(django_settings, self.settings_name)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/conf/__init__.py", line 102, in __getattr__
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     self._setup(name)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/conf/__init__.py", line 89, in _setup
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     self._wrapped = Settings(settings_module)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/django/conf/__init__.py", line 217, in __init__
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     mod = importlib.import_module(self.SETTINGS_MODULE)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/lib64/python3.9/importlib/__init__.py", line 127, in import_module
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     return _bootstrap._gcd_import(name[level:], package, level)
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "<frozen importlib._bootstrap_external>", line 850, in exec_module
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:   File "/usr/local/lib/python3.9/site-packages/pulpcore/app/settings.py", line 479, in <module>
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]:     raise ImproperlyConfigured(
Nov 04 11:59:55 quadlet.example.com pulp-api[60981]: django.core.exceptions.ImproperlyConfigured: Could not load DB_ENCRYPTION_KEY file '/etc/pulp/certs/database_fields.symmetric.key': [Errno 13] Permission denied: '/etc/pulp/certs/database_fields.symmetric.key'
```

So it tries to reach the DB ("Waiting on postgresql to start..."), and while doing so it fails and also fails to load the secret we're mounting. Especially, the "permission denied" is confusing, right?

After a bit of poking, I realized that the permission error is because we both try to have a secret, but also actually mount things on-top of `/etc/pulp:
https://github.com/theforeman/foreman-quadlet/blob/98f6a4110998d2f36257b9b18415b76571a2f6c4/roles/pulp/defaults/main.yaml#L6
Once I drop this line, and add `network: host` to the container definitions it can actually reach the database and the pulp services start.
The DB is still not getting migrated and that's what breaks pulp from actually working, but at least things start)

(and the `lookup` thing I mention inline)

roles/pulp/tasks/main.yaml
```diff
@@ -16,15 +26,71 @@
...
+- name: Generate database symmetric key
+  ansible.builtin.command: "bash -c 'openssl rand -base64 32 | tr \"+/\" \"-_\" > /var/lib/pulp/database_fields.symmetric.key'"
+  args:
+    creates: /var/lib/pulp/database_fields.symmetric.key
+
+- name: Create database symmetric key secret
+  containers.podman.podman_secret:
+    state: present
+    name: pulp-symmetric-key
+    data: "{{ lookup('file', '/var/lib/pulp/database_fields.symmetric.key') }}"
```
evgeni, line 38: this looks up the file on the node Ansible running on (so "laptop"), but it's created inside the `quadlet` VM.

I changed this
```patch
+- name: Load database symmetric key
+  ansible.builtin.slurp:
+    src: /var/lib/pulp/database_fields.symmetric.key
+  register: pulp_key
+
 - name: Create database symmetric key secret
     state: present
     name: pulp-symmetric-key
-    data: "{{ lookup('file', '/var/lib/pulp/database_fields.symmetric.key') }}"
+    data: "{{ pulp_key['content'] | b64decode }}"
```

and that made it available properly

roles/pulp/defaults/main.yaml
```diff
@@ -1,10 +1,18 @@
...
-pulp_ports:
-  - "8080:80"
+pulp_api_image: quay.io/pulp/pulp-minimal:stable
+pulp_content_image: quay.io/pulp/pulp-minimal:stable
+pulp_worker_image: quay.io/pulp/pulp-minimal:stable
+
+pulp_api_ports:
+  - "24817:80"
+pulp_content_ports:
+  - "24816:80"
```
ehelms, line 9: I don't think we want both of these exposing their ports on "80" but rather just exposing their respective ports, i.e. `24816:24816".

evgeni: We switched to host networking now, so this won't be used anymore.
Makes me wonder tho, how will we scale the API: can't start more than one container with the same port (no idea if that's needed tho).

ekohl: Today we only have a single service that exposes `pulpcore-api`'s `--workers` parameter. Isn't that the same with a container?

roles/pulp/defaults/main.yaml
```diff
@@ -1,10 +1,18 @@
...
+pulp_worker_count: 2
+
 pulp_volumes:
-  - /var/lib/pulp/settings:/etc/pulp:Z
-  - /var/lib/pulp/pulp_storage:/var/lib/pulp:Z
-  - /var/lib/pulp/pgsql:/var/lib/pgsql:Z
-  - /var/lib/pulp/containers:/var/lib/containers:Z
-pulp_container_name: pulp
+  - /var/lib/pulp/pulp_storage:/var/lib/pulp
+  - /var/lib/pulp/containers:/var/lib/containers
```
ekohl, line 14: I don't think the containers mount is needed anymore with minimal containers and we only really need storage.

At that point, should we just expose `/var/lib/pulp` as `/var/lib/pulp`? And I still think you need the SELinux label.
```suggestion
  - /var/lib/pulp:/var/lib/pulp:Z
```

roles/pulp/templates/settings.py.j2
```diff
@@ -1,7 +1,21 @@
 CONTENT_ORIGIN="http://{{ ansible_hostname }}:8080"
+API_CONTENT_ORIGIN="http://{{ ansible_hostname }}:24817"
+CONTENT_SERVICE_ORIGIN="http://{{ ansible_hostname }}:24816"
```
ekohl, line 3: I don't think these settings exist. I only see `CONTENT_ORIGIN` (https://pulpproject.org/pulpcore/docs/admin/reference/settings/#content_origin) but not any of the others.

roles/pulp/templates/settings.py.j2
```diff
@@ -1,7 +1,21 @@
...
 CACHE_ENABLED=True
 REDIS_HOST="localhost"
 REDIS_PORT=6379
```
ekohl, line 6: Now that it's no longer using the included Redis, but the system one I think it may use the default DB. Foreman may do the same. In puppet-pulpcore we explicitly chose a DB number that we don't use elsewhere and I think we should mirror that.

roles/pulp/defaults/main.yaml
evgeni: > And I still think you need the SELinux label.

I wish. It didn't work with :Z in my tests :/
That's we set `label=disable` on the containers.

roles/pulp/defaults/main.yaml
evgeni: Technically? Sure. It just feels weird (in a container context, I'd scale by just adding more containers).
We have the same problem with Foreman/Puma.

But luckily we don't have to solve it here and now.

roles/pulp/defaults/main.yaml
evgeni: And yes, you're right, we don't need the containers mount, see https://discourse.pulpproject.org/t/what-is-var-lib-containers-used-for-when-deploing-oci/1782

roles/pulp/defaults/main.yaml
```diff
@@ -1,10 +1,18 @@
 ---
-pulp_image: quay.io/pulp/pulp:stable
-pulp_ports:
-  - "8080:80"
+pulp_api_image: quay.io/pulp/pulp-minimal:stable
+pulp_content_image: quay.io/pulp/pulp-minimal:stable
+pulp_worker_image: quay.io/pulp/pulp-minimal:stable
```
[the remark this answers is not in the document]
evgeni: or should we do:
```
pulp_image: quay.io/pulp/pulp-minimal:stable
pulp_api_image: "{{ pulp_image }}"
pulp_content_image: "{{ pulp_image }}"
pulp_worker_image: "{{ pulp_image }}"
```
so it can be overridden when necessary?

roles/pulp/tasks/main.yaml
```diff
@@ -34,20 +115,43 @@
...
-- name: Start the Pulp Service
+- name: Start the Pulp API services
   ansible.builtin.systemd:
-    name: pulp
+    name: pulp-api
     enabled: true
-    state: restarted
+    state: started
+
+- name: Migrate the Pylp database
```
evgeni, line 124: Eh, I typo'ed that :D
```suggestion
- name: Migrate the Pulp database
```

roles/pulp/tasks/main.yaml
```diff
@@ -34,20 +115,43 @@
...
+  containers.podman.podman_container_exec:
+    name: "{{ pulp_api_container_name }}"
+    command: pulpcore-manager migrate --noinput
```
evgeni, line 127: I wonder if this should be an own command (like now), an own service (like https://github.com/pulp/pulp-oci-images/blob/latest/images/compose/compose.yml does it) or whether we should use `pulpcore-manager migrate --noinput && pulp-api` as the command for the "normal" API container startup (like we do for Foreman)?

roles/pulp/tasks/main.yaml
```diff
@@ -10,21 +20,92 @@
...
+    network: host
+    volumes: "{{ pulp_volumes }}"
+    security_opt:
+      - "label=disable"
+    secrets:
+      - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
+      - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
     quadlet_options:
       - |
         [Install]
```
evgeni, line 111: Pulp workers (and maybe others, I didn't check), don't properly survive DB disconnects and just die.
Pulp's compose has `restart=always` (https://github.com/pulp/pulp-oci-images/blob/97a31eb775c51bd3080380afbe6cd7ec11145007/images/compose/compose.yml#L159) and so should we:

```suggestion
        [Service]
        # This provides reconnect support for PostgreSQL and Redis. Without reconnect support, if either
        # is not available at startup or becomes disconnected, this process will die and not respawn.
        Restart=always
        RestartSec=3
        [Install]
```

evgeni: And if you wonder why we have DB disconnects: `geerlinguy.postgresql` has a "restart postgresql" handler, which runs at the very end of the play… And Ansible is being Ansible.

tests/pulp_test.py
```diff
@@ -22,17 +22,38 @@ def test_pulp_service(host):
...
+    assert pulp_api.is_enabled
+
+def test_pulp_content_service(host):
+    pulp_content = host.service("pulp-content")
+    assert pulp_content.is_running
+    assert pulp_content.is_enabled
+
+def test_pulp_worker_services(host):
+    for i in range(1, 3):
+        pulp_worker = host.service(f"pulp-worker@{i}")
```
evgeni, line 37: We currently do not deploy the worker service as a template, so this test fails.

roles/pulp/defaults/main.yaml
ekohl: In a quadlet context you can probably share a systemd activated socket and make it `pulp-api@` so you can scale it up but I think that's premature

roles/pulp/templates/settings.py.j2
evgeni: https://github.com/theforeman/puppet-pulpcore/blob/26254db23ee9456520049a4651845fecc6e2cbe9/manifests/init.pp#L241 says we use `8`
so:
```suggestion
REDIS_HOST="localhost"
REDIS_PORT=6379
REDIS_DB=8
```

roles/pulp/defaults/main.yaml
evgeni: opened https://github.com/theforeman/foreman-quadlet/issues/51 to track this further

roles/pulp/tasks/main.yaml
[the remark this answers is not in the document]
evgeni: This would be not idempotent, yeah :(

evgeni:
I combined this PR, with #43 and the worker-related comments I had above, and it's GREEN! :green_apple: :tada: 

https://github.com/theforeman/foreman-quadlet/pull/50

roles/pulp/tasks/main.yaml
archanaserver: I guess while combining the migration and API startup commands seems convenient, but it will create challenges in error handling and clearly not flexible. 
And I think having a dedicated migration service would be a better approach for us. It allows for more flexible scaling of the migration process independently from the API and other services, making everything easier to manage in the long run.

roles/pulp/defaults/main.yaml
```diff
@@ -1,10 +1,18 @@
...
+  - "24816:80"
+pulp_worker_count: 2
+
 pulp_volumes:
-  - /var/lib/pulp/settings:/etc/pulp:Z
-  - /var/lib/pulp/pulp_storage:/var/lib/pulp:Z
-  - /var/lib/pulp/pgsql:/var/lib/pgsql:Z
-  - /var/lib/pulp/containers:/var/lib/containers:Z
-pulp_container_name: pulp
+  - /var/lib/pulp/pulp_storage:/var/lib/pulp
```
ekohl, line 14: As I noted before, can this simply be
```suggestion
  - /var/lib/pulp:/var/lib/pulp
```

roles/pulp/templates/settings.py.j2
```diff
@@ -1,7 +1,20 @@
-CONTENT_ORIGIN="http://{{ ansible_hostname }}:8080"
+CONTENT_ORIGIN="http://{{ ansible_hostname }}:24816"
```
ekohl, line 1: AFAIK this should be wherever Apache (httpd) presents the content, so end user clients talk to. They shouldn't talk to the internal endpoint.

roles/pulp/tasks/main.yaml
ekohl: I see some benefits in having tighter control, but you must migrate the DB before starting any of the Pulp services.

tests/pulp_test.py
```diff
@@ -1,37 +1,40 @@
...
+    assert pulp_worker.is_enabled
 
-def test_pulp_port(host):
-    pulp = host.addr(PULP_HOST)
-    assert pulp.port(PULP_PORT).is_reachable
-
-
-def test_pulp_status(pulp_status_curl):
-    assert pulp_status_curl.succeeded
-    assert pulp_status_curl.stderr == '200'
```
ekohl, line 33: It would be nice to keep this check in to verify Pulp actually thinks it's healthy.

roles/pulp/templates/settings.py.j2
archanaserver: I don't understand, would you mind expanding a little more?

tests/pulp_test.py
archanaserver: Definitely! I've reintroduced this.

roles/pulp/templates/settings.py.j2
ekohl: Quoting https://pulpproject.org/pulpcore/docs/admin/reference/settings/#content_origin

> A required string containing the protocol, fqdn, and port where the content app is reachable by users. This is used by pulpcore and various plugins when referring users to the content app. For example if the API should refer users to content at using http to pulp.example.com on port 24816, (the content default port), you would set: https://pulp.example.com:24816.

In our deployment we have httpd in front of Pulp which presents it at on `https://{{ ansible_hostname }}`. In other words, this was always wrong, but I'm noticing it now.

archanaserver: @ekohl so we basically have to set it up without port? if we want user to point at the right place? to this `CONTENT_ORIGIN="https://{{ ansible_hostname }}"
` (sorry pinging on this again, i'm not good at this part)

evgeni: That's what we do in puppet-pulpcore, yeah:
https://github.com/theforeman/puppet-pulpcore/blob/master/templates/settings.py.erb#L20

You could also argue, that this sort-of blurs the lines between the roles (the plain, port 443, HTTPS is not available w/o the `httpd` role), and the `pulp` role should have a default for the content origin being `http://{{ ansible_fqdn }}:port`, and then we override this in `deploy.yaml` with `https://{{ ansible_fqdn }}` as that's the value the overall deployment uses

roles/pulp/tasks/main.yaml
```diff
@@ -10,44 +20,156 @@
     mode: "0755"
   loop: "{{ pulp_volumes }}"
 
+- name: Create Pulp storage subdirs
+  ansible.builtin.file:
+    path: "/var/lib/pulp/pulp_storage/{{ item }}"
```
evgeni, line 25:
```suggestion
    path: "/var/lib/pulp/{{ item }}"
```

And then you don't need the "Create Pulp temporary directory" task after this.
(You'll need `assets` once you try to upload stuff to pulp)

roles/pulp/templates/settings.py.j2
ekohl: So it should be a variable on the role that the playbook sets (because that knows both httpd and pulp)?

evgeni: s/variable/default/, but yeah, that's how I'd do it.

roles/pulp/tasks/main.yaml
```diff
@@ -10,44 +20,156 @@
...
+- name: Create Pulp storage subdirs
+  ansible.builtin.file:
+    path: "/var/lib/pulp/{{ item }}"
+    state: directory
+    mode: "0755"
+  loop:
+    - tmp
+    - assets
+
+- name: Create Pulp temporary directory
```
ehelms, line 32: This is repetitive - `tmp` is covered by line 23

roles/pulp/tasks/main.yaml
ehelms: Fine for now but I think we should understand and decide how we handle this in a universal way given we have multiple applications having to do the same workflow: wait on database, apply outstanding migrations. Candlepin does this via the application startup procedure and so this will happen within the container itself.

Pulp's images have a wait for database and wait for migrations script that they currenly rely upon before startup: https://github.com/pulp/pulp-oci-images/blob/d74b7bbb9550152c0de1e45f16eec73f674e8f15/images/assets/pulp-api#L3-L4

The Foreman container being used takes a similar approach: https://github.com/theforeman/foreman-quadlet/blob/98f6a4110998d2f36257b9b18415b76571a2f6c4/container-images/foreman/Containerfile#L18

roles/pulp/tasks/main.yaml
```diff
@@ -10,44 +20,156 @@
...
-- name: Start the Pulp Service
+- name: Start the Pulp API services
   ansible.builtin.systemd:
-    name: pulp
+    name: pulp-api
     enabled: true
-    state: restarted
+    state: started
+
+- name: Migrate the Pulp database
```
ehelms, line 142: I would think we want to migrate the database before we start the API service?

archanaserver: Typically, we do this but when I try doing it here, I see migration command could not find the pulp-api container, i'm not very sure why this fails but maybe due to container isn't fully registered or maybe we can add another task for the pulp-api container creation little timeout. (is it due to some delay?)

the change: 974e8e3c59ddd8c8146a73deef58fde5b8446362 and the error: https://github.com/theforeman/foreman-quadlet/actions/runs/11823986613/job/32944500958?pr=34

roles/pulp/tasks/main.yaml
```diff
@@ -10,44 +20,150 @@
...
+        Restart=always
+        RestartSec=3
 
 - name: Run daemon reload to make Quadlet create the service files
   ansible.builtin.systemd:
     daemon_reload: true
 
-- name: Start the Pulp Service
+- name: Migrate the Pulp database
+  containers.podman.podman_container_exec:
```
ehelms, line 131: `podman_container_exec` makes use of the `exec` function in podman which assumes a running container already exists. This worked when this statement was after the start of the Pulp API service, but it coupled it's operation to the existence and functionality of that service. This instead needs to get switched to running the image but with the specific migration command. This diff allowed the install to pass for me:

```
diff --git a/roles/pulp/tasks/main.yaml b/roles/pulp/tasks/main.yaml
index bc51e00..297e738 100644
--- a/roles/pulp/tasks/main.yaml
+++ b/roles/pulp/tasks/main.yaml
@@ -128,9 +128,14 @@
     daemon_reload: true
 
 - name: Migrate the Pulp database
-  containers.podman.podman_container_exec:
-    name: "{{ pulp_api_container_name }}"
+    name: pulpcore-manager-migrate
+    image: "{{ pulp_api_image }}"
     command: pulpcore-manager migrate --noinput
+    network: host
+    secrets:
+      - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
+      - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
 
 - name: Start the Pulp API services
   ansible.builtin.systemd:
```

archanaserver: Make sense, thanks @ehelms, and it's green again.

ehelms (approved):

roles/pulp/templates/settings.py.j2
```diff
@@ -1,7 +1,20 @@
...
+REDIS_DB=8
+
+DATABASES = {
+	'default': {
+		'ENGINE': 'django.db.backends.postgresql',
+		'NAME': 'pulp',
+		'USER': 'pulp',
+		'PASSWORD': '{{ pulp_db_password }}',
+		'HOST': 'localhost',
+		'PORT': '5432',
```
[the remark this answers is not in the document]
ekohl: My suggestion was to remove the line.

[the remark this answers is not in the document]
archanaserver: ah, i understand now! sorry, updating it in another PR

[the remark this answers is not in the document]
archanaserver: @ekohl, https://github.com/theforeman/foreman-quadlet/pull/54

PR #37, opened 2024-10-23
add container build test

evgeni:


.github/workflows/container.yml
```diff
@@ -0,0 +1,36 @@
...
+      fail-fast: false
+      matrix:
+        container:
+          - candlepin
+          - foreman
+          - foreman-proxy
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build ${{ matrix.container }} container
+        run: make -C container-images/${{ matrix.container }} build VERSION=ci
```
evgeni, line 36: `VERSION=ci` is only needed because the variable would be otherwise unset for Candlepin:
https://github.com/theforeman/foreman-quadlet/blob/2df4338f70f9bc1903fc6becb077a21034c82638/container-images/candlepin/Makefile#L2
It's otherwise unused (the `Containerfile` has a hard-coded 4.4.14 in it right now), so I wonder if we should instead fix up the `Makefile`?

ehelms: > `VERSION=ci` is only needed because the variable would be otherwise unset for Candlepin:
> 
> https://github.com/theforeman/foreman-quadlet/blob/2df4338f70f9bc1903fc6becb077a21034c82638/container-images/candlepin/Makefile#L2
> 
> 
> It's otherwise unused (the `Containerfile` has a hard-coded 4.4.14 in it right now), so I wonder if we should instead fix up the `Makefile`?

Makes sense to me to keep this the same across each build.

evgeni: "across each Makefile" you mean? ack

ehelms: yep

evgeni: done

ehelms (approved):

container-images/candlepin/Makefile
```diff
@@ -1,5 +1,5 @@
 build:
-	podman build -t quay.io/ehelms/candlepin:${VERSION} .
+	podman build -t quay.io/ehelms/candlepin:nightly .
```
ehelms, line 2: Maybe this is for a follow on to figure out our tagging structure -- cause I think versioned tags alongside nightly tag would be useful.

evgeni: one issue closed, one opened, that's how it goes? :)

evgeni: in some totally unrelated project, I have a `make version` that I then re-use to generate the tag: https://github.com/evgeni/unifi-controller/blob/devel/.github/workflows/release.yml

PR #38, opened 2024-10-23
add dynflow hosts_queue worker

evgeni:


ehelms:
Can you add some tests for that worker existing?

evgeni:
I don't think there is an API for that, but I'll try to find out

ehelms:
> I don't think there is an API for that, but I'll try to find out

At the systemd level would be sufficient

evgeni:
Ah, that's fair!

PR #39, opened 2024-10-23
add ansible lint test

evgeni:


evgeni:
:green_apple:

playbooks/setup.yaml
```diff
@@ -3,45 +3,45 @@
...
-    ansible.posix.firewalld:
-      port: "{{ item }}"
-      permanent: true
-      state: enabled
-    loop:
-    - 23443/tcp
-    when:
-    - services['firewalld.service'] is defined
-    - services['firewalld.service']['state'] == 'running'
+    - name: Open Ports in firewalld
```
evgeni, line 38: @ehelms is this even needed? candlepin doesn't need to be open to the world.

ehelms: If tests pass.. then nope!

ehelms (approved):

playbooks/setup.yaml
evgeni: https://github.com/theforeman/foreman-quadlet/pull/40

PR #41, opened 2024-10-24
also mark dynflow containers as auto-started

evgeni:
Fixes: 371727b767dd7b012566138069ea08e54b00906b

damn template services

### DIFF
--- a/roles/foreman/tasks/main.yaml
+++ b/roles/foreman/tasks/main.yaml
@@ -87,6 +87,20 @@
       DYNFLOW_REDIS_URL: "redis://localhost:6379/6"
       REDIS_PROVIDER: "DYNFLOW_REDIS_URL"
     command: "/usr/libexec/foreman/sidekiq-selinux -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/%i.yml"
+    quadlet_options:
+      - |
+        [Install]
+        WantedBy=default.target
+
+- name: Create Dynflow Container instances
+  ansible.builtin.file:
+    state: link
+    src: "/etc/containers/systemd/dynflow-sidekiq@.container"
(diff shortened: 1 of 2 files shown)


roles/foreman/tasks/main.yaml
```diff
@@ -87,6 +87,20 @@
...
+
+- name: Create Dynflow Container instances
+  ansible.builtin.file:
+    state: link
+    src: "/etc/containers/systemd/dynflow-sidekiq@.container"
+    dest: "/etc/containers/systemd/dynflow-sidekiq@{{ item }}.container"
+  loop:
+    - orchestrator
+    - worker
+    - worker-hosts-queue
```
ehelms, line 103: Tests for each of these? 

Does the container startup command derive what persona to be based on the name of the service?

roles/foreman/tasks/main.yaml
```diff
@@ -87,6 +87,20 @@
       DYNFLOW_REDIS_URL: "redis://localhost:6379/6"
       REDIS_PROVIDER: "DYNFLOW_REDIS_URL"
     command: "/usr/libexec/foreman/sidekiq-selinux -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/%i.yml"
```
ehelms, line 89: I wonder if the `sidekiq-selinux` start up command will be needed inside a container.

roles/foreman/tasks/main.yaml
evgeni: Test that `/etc/containers/systemd/dynflow-sidekiq@{{ item }}.container` exists? Can add.

roles/foreman/tasks/main.yaml
evgeni: Maybe. Maybe not. Probably not.

PR #42, opened 2024-10-25
add settings.yaml to dynflow containers

evgeni:
otherwise they don't know which certs to use


evgeni:
this patch was brought to you by trying to write actual tests for Katello :D

ekohl (approved):

PR #43, opened 2024-10-25
use the api to create a set of reusable fixtures

evgeni:


tests/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+
+import apypie
+import pytest
+
+FOREMAN_PORT=3000
+
+def to_native(thing):
+    return thing
+
+class ForemanApi(apypie.Api):
```
evgeni, line 11: ignore this class for now, lots of :spaghetti: from foreman-ansible-modules

tests/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+        return task
+
+    def ping(self):
+        return self.resource('ping').call('ping')
+
+
+@pytest.fixture(scope="module")
+def foremanapi():
+    return ForemanApi(
+        uri=f'http://quadlet.tanso.example.com:{FOREMAN_PORT}',
```
evgeni, line 80: `pytest` runs on the "outer" system (my laptop), so it needs to be able to access Foreman like that.

probably clenest would be to have a *second* VM and run pytest (for the remote tests) there?

tests/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+    yield prod
+    foremanapi.resource_action('products', 'destroy', params={'id': prod['id']})
+
+@pytest.fixture
+def repository(product, organization, foremanapi):
+    # This is broken as Katello currently does not use the right CA when talking to Pulp
+    # https://github.com/Katello/katello/blob/master/app/models/katello/concerns/smart_proxy_extensions.rb#L254-L265
+    repo = foremanapi.resource_action('repositories', 'create', params={'name': 'Test Repo', 'product_id': product['id'], 'content_type': 'yum'})
+    yield repo
+    foremanapi.resource_action('repositories', 'destroy', params={'id': repo['id']})
```
evgeni, line 104: This found #42 and then stuck at `SSL_connect returned=1 errno=0 state=error: certificate verify failed (self-signed certificate in certificate chain)` when trying to talk to Pulp because Katello doesn't inject the CA into Faraday.

tests/conftest.py
ehelms: Why is that?

ekohl: > `pytest` runs on the "outer" system (my laptop), so it needs to be able to access Foreman like that.

In smoker we use pytest-variables: https://github.com/theforeman/smoker?tab=readme-ov-file#variables. These are then used in https://github.com/theforeman/smoker/blob/d30ff41ca4931d7db61e8495c226f778b563a744/tests/conftest.py#L65-L73 to return an API instance.

> probably clenest would be to have a _second_ VM and run pytest (for the remote tests) there?

It's what we do in our pipelines with smoker.

tests/conftest.py
ekohl: It'd be interesting to see what we could add to apypie to make it easier to consume.

evgeni: that's on my plate for next week :)

tests/conftest.py
ehelms: Look at that! Adding some basic tests are already paying dividends.

evgeni: I just don't like that the dividends are coming out of my account.

tests/conftest.py
evgeni: > Why is that?

Why is what? :)
Why I need to access from the "outside"?



> In smoker we use pytest-variables

Yeah, but someone/something needs to file in those variables.

I used paramiko to parse the ssh config we generate for testinfra for now.

tests/conftest.py
evgeni: https://github.com/theforeman/foreman-ansible-modules/pull/1789 has a reasonable first stab at that

tests/conftest.py
evgeni: https://github.com/Katello/katello/pull/8832 was the better fix, now done

tests/foreman_api_test.py
```diff
@@ -0,0 +1,8 @@
+def test_foreman_organization(organization):
+    assert organization
+
+#def test_foreman_product(product):
+#    assert product
```
evgeni, line 5: for some reason this test fails, even tho the repository test below uses (internally) the same fixture and works fine :thinking:

requirements.txt
```diff
@@ -1,2 +1,3 @@
 pytest-testinfra
 paramiko
+apypie @ git+https://github.com/Apipie/apypie.git@foreman
```
evgeni, line 3: https://github.com/Apipie/apypie/pull/175

tests/foreman_api_test.py
```diff
@@ -0,0 +1,28 @@
+import urllib.parse
+
+import requests
+
+
+def _repo_url(repo, ssh_config):
+    return urllib.parse.urlunparse(urllib.parse.urlparse(repo['full_path'])._replace(netloc=ssh_config['hostname']))
```
ekohl, line 7: I recall having a discussion with @maximiliankolb about exposing the repo's full path URLs in Katello. I think we should file an issue to expose this in the API (and UI) for a specific content source. Then add a TODO here to use that API when it becomes available.

Any thoughts on that?

evgeni: What do you mean by "full path"? This attribute is the full path in the form of `https://foreman.example.com/pulp/content/blah/blah/blah`. It's just that I need to fuck around with the fact that `foreman.example.com` doesn't exist as a DNS name when the tests run and thus have this ugly replacement here.

ekohl: Ah. Should we just fix that instead then? I can take a stab at that, though probably better to capture it in an issue first and link it here as a todo

evgeni: I think we can "configure" it via https://github.com/theforeman/foreman-quadlet/blob/master/roles/foreman_proxy/templates/pulpcore.yaml.j2#L4

tests/zzz_test.py
```diff
@@ -0,0 +1,8 @@
+def test_collect_report(host):
+    host.run('mkdir -p logs')
+    for container, filename in [('foreman', '/var/log/foreman/production.log')]:
+        localfile = filename.replace('/', '_')
+        host.run(f'podman cp {container}:{filename} logs/{container}-{localfile}')
+    host.run('tar caf logs.tar.gz logs/')
+    with open('logs.tar.gz', 'wb') as logstar:
+        logstar.write(host.file('logs.tar.gz').content)
```
evgeni, line 8: Crazy? Yes.
Works? Yes.
Did it help? No, for some reason the tests didn't want to fail since I've added this.

tests/foreman_api_test.py
evgeni: pytest, today, runs on the GH runner directly (or on my laptop)

ekohl: Hmm ok. Perhaps leave a comment on why the hack is needed then?

tests/zzz_test.py
ekohl: There are [reporting hooks](https://docs.pytest.org/en/stable/reference/reference.html#reporting-hooks) that can be used. https://docs.pytest.org/en/stable/example/simple.html#post-process-test-reports-failures has an example that looks useful and better than to capture it in a test.

But I think it raises an interesting topic: how do we deal with logging. It's often considered a good practice to log to stdout in containers and then we can simply capture the journal for our services. Should we capture that in another issue?

evgeni: other than https://github.com/theforeman/foreman-quadlet/issues/21? :)

ekohl: Well, I could have checked for issues first.

tests/foreman_api_test.py
maximiliankolb: RE exposing the URLs to foreman-katello.example.com/pulp/content/...: Yes please. It's necessary if you use F+K as content source when registering non-RHEL hosts using global host registration because non-RHEL OS do not have (access to) `subscription-manager`. We are also (someday) working on a button in the WebUI to allow users to copy the Katello API endpoint for the associated GPG pub key.

See https://github.com/theforeman/foreman-documentation/pull/3380 and https://docs.theforeman.org/nightly/Managing_Content/index-katello.html#Content-Views-in-Foreman_content-management

Vagrantfile
```diff
@@ -9,7 +9,8 @@ Vagrant.configure("2") do |config|
     override.vm.hostname = "quadlet.example.com"
 
     override.vm.provider "libvirt" do |libvirt, provider|
-      libvirt.memory = "4096"
+      libvirt.memory = 10240
+      libvirt.cpus = 4
```
evgeni, line 13: I've split this out to https://github.com/theforeman/foreman-quadlet/pull/48, as it's "technically" unrelated

tests/conftest.py
```diff
@@ -0,0 +1,64 @@
...
+
+@pytest.fixture
+def container_repository(product, organization, foremanapi):
+    repo = foremanapi.create('repositories', {'name': str(uuid.uuid4()), 'product_id': product['id'], 'content_type': 'docker', 'url': 'https://quay.io/', 'docker_upstream_name': 'foreman/busybox-test'})
+    yield repo
+    foremanapi.delete('repositories', repo)
+
+@pytest.fixture
+def lifecycle_environment(organization, foremanapi):
+    library = foremanapi.list('lifecycle_environments', 'name=Library', {'organization_id': organization['id']})[0]
```
ekohl, line 55: Same here perhaps?
```suggestion
def lifecycle_environment(organization, foremanapi, prior='Library'):
    library = foremanapi.list('lifecycle_environments', f'name={prior}', {'organization_id': organization['id']})[0]
```

Though perhaps it should just accept an instance where `None` is automatically resolve to the library:
```suggestion
def lifecycle_environment(organization, foremanapi, prior=None):
    if not prior:
        prior = foremanapi.list('lifecycle_environments', 'name=Library', {'organization_id': organization['id']})[0]
```

Which also makes me question what the Katello API does if no prior ID is given. It's not listed as required and mandating this, but would it make sense to find the library server side and make it an optional parameter?

evgeni: It does not find "Library" automatically and errors out:
https://github.com/Katello/katello/blob/db89c59031e34994805faaf3c6314b0e412b56c3/app/controllers/katello/api/v2/environments_controller.rb#L39
https://github.com/Katello/katello/blob/db89c59031e34994805faaf3c6314b0e412b56c3/app/controllers/katello/api/v2/environments_controller.rb#L172-L177

ekohl: I meant that last part as an RFE.

tests/conftest.py
```diff
@@ -0,0 +1,64 @@
...
+
+@pytest.fixture
+def product(organization, foremanapi):
+    prod = foremanapi.create('products', {'name': str(uuid.uuid4()), 'organization_id': organization['id']})
+    yield prod
+    foremanapi.delete('products', prod)
+
+@pytest.fixture
+def yum_repository(product, organization, foremanapi):
+    repo = foremanapi.create('repositories', {'name': str(uuid.uuid4()), 'product_id': product['id'], 'content_type': 'yum', 'url': 'https://fixtures.pulpproject.org/rpm-no-comps/'})
```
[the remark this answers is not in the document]
evgeni: I don't think fixtures take regular parameters like this (there are fixture factories, but those are more complicated)

tests/conftest.py
```diff
@@ -0,0 +1,64 @@
...
+def foremanapi(ssh_config):
+    return apypie.ForemanApi(
+        uri=f'https://{ssh_config['hostname']}',
+        username='admin',
+        password='changeme',
+        verify_ssl=False,
+    )
+
+@pytest.fixture
+def organization(foremanapi):
```
evgeni, line 24: Right now all these tests have an (implicit) `scope=function` (https://docs.pytest.org/en/stable/how-to/fixtures.html#fixture-scopes) which means that they are destroyed at the end of every test.

Conceptually, this is good, of course. But on the flip side this means that every repository test *also* first creates a new org (also in Candlepin), a new product, etc. Those times quickly add up, and make the test tiresome to execute.

We can speed this up with `scope=module`, but this means that any failure in a test might have cascading effects on later tests (which is the same behaviour as we have today with bats, FWIW).

WDYT?

tests/conftest.py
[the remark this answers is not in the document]
ekohl: https://docs.pytest.org/en/7.1.x/example/parametrize.html#indirect-parametrization says you can do it, but you need to use `request.param`. Feel free to leave it out for now, but let's keep it in mind for the future.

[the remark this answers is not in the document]
evgeni: Oh via indirect… Mhh, interesting. Maybe. I'll think about it.

.github/workflows/test.yml
```diff
@@ -49,8 +49,10 @@ jobs:
         run: pip install -r requirements.txt
       - name: Install collection dependencies
         run: ansible-galaxy install -r requirements.yml
-      - name: Start VM
-        run: vagrant up quadlet
+      - name: Start VMs
+        run: |
+          vagrant up quadlet
+          vagrant up client
```
ekohl, line 55: Do you intentionally start them up sequentially instead of parallel? Is that the issue with fetching the same box concurrently giving issues?

evgeni: I honestly haven't tried 😅

run_tests
```diff
@@ -2,3 +2,4 @@
 
 vagrant ssh-config > .vagrant/ssh-config
 py.test -vv --hosts=quadlet --sudo --ssh-config=.vagrant/ssh-config tests/*_test.py
+py.test -vv --hosts=client --sudo --ssh-config=.vagrant/ssh-config tests/foreman_client.py
```
ekohl, line 5: This made me wonder if we could somehow tag hosts, but I don't think we can.

Perhaps a better design is to introduce our own fixtures for certain roles so we can run a single pytest and not accidentally do the wrong stuff?

https://testinfra.readthedocs.io/en/latest/examples.html#test-docker-images also has an interesting pattern where it spins up containers. https://github.com/theforeman/smoker/blob/master/tests/test_content_client.py also has some some code to spin up clients, but now I wonder if we even run that given it mentions katello-agent a lot.

tests/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+    foremanapi.resource_action('repositories', 'sync', {'id': yum_repository['id']})
+    foremanapi.update('content_views', {'id': content_view['id'], 'repository_ids': [yum_repository['id']]})
+    foremanapi.resource_action('content_views', 'publish', {'id': content_view['id']})
+
+    library = foremanapi.list('lifecycle_environments', 'name=Library', {'organization_id': organization['id']})[0]
+    foremanapi.update('activation_keys', {'id': activation_key['id'], 'organization_id': organization['id'], 'environment_id': library['id'], 'content_view_id': content_view['id']})
+
+    yield activation_key
+
+    foremanapi.update('activation_keys', {'id': activation_key['id'], 'organization_id': organization['id'], 'environment_id': None, 'content_view_id': None})
```
evgeni, line 83: This needs https://github.com/Katello/katello/pull/11215 which is not yet in the used image

run_tests
evgeni: So I ended up creating custom fixtures calling `testinfra.get_host`, which seems to work fine.

And no, we don't run the content client tests in smoker today.

evgeni:
> Generally looks good.. what's needed to get it out of draft?

- [ ] https://github.com/Apipie/apypie/pull/175 being merged and released
- [ ] cleanup of the git history here

ansible.cfg
```diff
@@ -3,4 +3,4 @@ host_key_checking = False
 inventory = vagrant.py
 stdout_callback=debug
 stderr_callback=debug
-roles_path = ~/.ansible/roles:./roles
+roles_path = ~/.ansible/roles:./roles:./forklift/roles
```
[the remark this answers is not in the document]
evgeni: `etc_hosts` role, so that the `client` box can resolve `quadlet`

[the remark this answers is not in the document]
ehelms: Can you add that as a requirement in the README or just copy the role here or as something that can be installed via `ansible-galaxy` ?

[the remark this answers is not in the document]
evgeni: I can, yes

requirements.yml
```diff
@@ -6,6 +6,9 @@ collections:
   - name: containers.podman
     version: ">=1.14.0"
   - name: theforeman.foreman
+  - name: https://github.com/evgeni/forklift
+    type: git
+    version: collection
```
evgeni, line 11: https://github.com/theforeman/forklift/pull/1879

ansible.cfg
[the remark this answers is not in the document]
evgeni: https://github.com/theforeman/forklift/pull/1879

requirements.yml
evgeni:
```suggestion
  - name: https://github.com/theforeman/forklift
    type: git
```

Once merged :)

playbooks/setup.yaml
```diff
@@ -1,7 +1,14 @@
 ---
-- name: Setup quadlet demo machine
+- name: Setup basic stuff
   hosts: all
   become: true
+  roles:
+    - theforeman.forklift.etc_hosts
```
ehelms, line 6: :fireworks:

PR #47, opened 2024-10-28
Move the plugin specific layers to the bottom

ekohl:
This allows reusing the base layers in case you build multiple variations.

Currently untested and submitting for early review. If this is a good idea, we should also apply it to foreman-proxy.

### DIFF
--- a/container-images/foreman/Containerfile
+++ b/container-images/foreman/Containerfile
@@ -2,14 +2,9 @@ FROM quay.io/centos/centos:stream9
 
 ARG FOREMAN_VERSION=nightly
 ARG KATELLO_VERSION=nightly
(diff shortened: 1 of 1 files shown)


container-images/foreman/Containerfile
```diff
@@ -17,3 +13,8 @@ ENV RAILS_SERVE_STATIC_FILES=true
 ENV RAILS_ENV=production
 CMD /usr/share/foreman/bin/rails db:migrate && /usr/share/foreman/bin/rails db:seed && /usr/share/foreman/bin/rails server --environment production
 EXPOSE 3000/tcp
+
```
ehelms, line 16: Note line 8 where it switches to the `foreman` user and thus breaks the ability to run dnf.

ehelms (approved):

PR #56, opened 2024-11-18
use a dedicated "server" fixture, instead of the generic "host"

evgeni:
this allows differentiating different types of hosts which we will have soon


ehelms:
@evgeni please to rebase now

PR #59, opened 2024-11-20
migrate httpd installation from rpm to container-based

archanaserver:
This PR transitions the `httpd` setup from an RPM-based installation to a containerized deployment using the [ubi9/httpd-24](https://catalog.redhat.com/software/containers/ubi9/httpd-24/61a60c3e3e9240fca360f74a) image.

### DIFF
--- a/src/roles/httpd/tasks/main.yml
+++ b/src/roles/httpd/tasks/main.yml
@@ -34,8 +34,13 @@
 - name: Configure foreman-ssl vhost
   ansible.builtin.template:
     src: foreman-ssl-vhost.conf.j2
-    dest: /etc/httpd/conf.d/foreman-ssl.conf
+    dest: "{{ httpd_conf_dir }}/foreman-ssl.conf"
     mode: "0644"
+
+- name: Disable welcome page in container
+  ansible.builtin.file:
+    path: "{{ httpd_conf_dir }}/welcome.conf"
+    state: absent
 
 - name: Run Apache as a container
   containers.podman.podman_container:
@@ -49,8 +54,3 @@
     ports:
       - "80:80"
       - "443:443"
-
-- name: Disable welcome page in container
-  containers.podman.podman_container_exec:
-    name: httpd
-    command: rm -f /etc/httpd/conf.d/welcome.conf
--- a/src/roles/httpd/tasks/main.yml
+++ b/src/roles/httpd/tasks/main.yml
@@ -4,10 +4,12 @@
     name: "{{ httpd_container_image }}"
     state: present
 
-- name: Create cert directories
+- name: Create required directories
   ansible.builtin.file:
     path: "{{ item }}"
     state: directory
+    owner: root
+    group: root
     mode: "0750"
   loop:
     - "{{ httpd_ssl_dir }}/certs"
@@ -54,3 +56,5 @@
     ports:
       - "80:80"
       - "443:443"
+    env:
+      HTTP_LOG_TO_VOLUME: "1"
--- a/src/roles/httpd/tasks/main.yml
+++ b/src/roles/httpd/tasks/main.yml
@@ -4,11 +4,6 @@
     name: "{{ httpd_container_image }}"
     state: present
 
-- name: Disable welcome page in container
-  containers.podman.podman_container_exec:
-    container: httpd
-    command: rm -f /etc/httpd/conf.d/welcome.conf
-
 - name: Create cert directories
   ansible.builtin.file:
     path: "{{ item }}"
@@ -54,3 +49,8 @@
     ports:
       - "80:80"
       - "443:443"
+
+- name: Disable welcome page in container
+  containers.podman.podman_container_exec:
+    name: httpd
+    command: rm -f /etc/httpd/conf.d/welcome.conf
--- a/src/roles/httpd/tasks/main.yml
+++ b/src/roles/httpd/tasks/main.yml
@@ -1,8 +1,13 @@
 ---
 - name: Pull Apache container image
   containers.podman.podman_image:
-    name: registry.access.redhat.com/ubi9/httpd-24:latest
+    name: "{{ httpd_container_image }}"
     state: present
+
+- name: Disable welcome page in container
+  containers.podman.podman_container_exec:
+    container: httpd
+    command: rm -f /etc/httpd/conf.d/welcome.conf
 
 - name: Create cert directories
   ansible.builtin.file:
@@ -12,6 +17,8 @@
   loop:
     - "{{ httpd_ssl_dir }}/certs"
     - "{{ httpd_ssl_dir }}/private"
+    - "{{ httpd_conf_dir }}"
+    - "{{ httpd_log_dir }}"
 
 - name: Deploy certificates
   ansible.builtin.copy:
@@ -37,13 +44,13 @@
 
 - name: Run Apache as a container
   containers.podman.podman_container:
-    name: apache
-    image: registry.access.redhat.com/ubi9/httpd-24:latest
+    name: httpd
+    image: "{{ httpd_container_image }}"
     state: started
     volumes:
       - "{{ httpd_ssl_dir }}:/etc/httpd/ssl"
-      - "/etc/httpd/conf.d:/usr/local/apache2/conf.d"
-      - "/var/log/httpd:/usr/local/apache2/logs"
+      - "{{ httpd_conf_dir }}:/etc/httpd/conf.d"
+      - "{{ httpd_log_dir }}:/usr/local/apache2/logs"
     ports:
       - "80:80"
       - "443:443"
--- a/src/roles/httpd/tasks/main.yml
+++ b/src/roles/httpd/tasks/main.yml
@@ -1,21 +1,8 @@
 ---
-- name: Install Apache httpd
-  ansible.builtin.package:
-    name:
-      - httpd
-      - mod_ssl
+- name: Pull Apache container image
+  containers.podman.podman_image:
+    name: registry.access.redhat.com/ubi9/httpd-24:latest
     state: present
-
-- name: Set httpd_can_network_connect so Apache can connect to Puma and Gunicorn
-  ansible.posix.seboolean:
-    name: httpd_can_network_connect
-    state: true
-    persistent: true
-
-- name: Disable welcome page
-  ansible.builtin.file:
-    path: /etc/httpd/conf.d/welcome.conf
-    state: absent
 
 - name: Create cert directories
   ansible.builtin.file:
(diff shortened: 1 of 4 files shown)


roles/httpd/templates/foreman-ssl-vhost.conf.j2
ekohl, line 1: Rather than changing the paths to hardcoded paths, can't you modify the value of the variable to match?

roles/httpd/tasks/main.yml
```diff
@@ -1,22 +1,9 @@
 ---
-- name: Install Apache httpd
-  ansible.builtin.package:
-    name:
-      - httpd
-      - mod_ssl
+- name: Pull Apache container image
+  containers.podman.podman_image:
+    name: httpd:latest
```
ehelms, line 4: What image is this? I'd prefer we use fully qualified image names to know where from and what image we are targeting. Is this from docker hub? Quay?

Currently our containers are all CentOS 9 Stream based and I think that's a good standardization that we should aim for rather than mixing and matching.

archanaserver: That was just a draft i wrote, I just realized, the code was not updated(though things are still breaking so it is WIP) But I'm using this ` name: registry.access.redhat.com/ubi9/httpd-24:latest` https://catalog.redhat.com/software/containers/ubi9/httpd-24/61a60c3e3e9240fca360f74a

roles/httpd/tasks/main.yml
```diff
@@ -1,22 +1,9 @@
 ---
-- name: Install Apache httpd
-  ansible.builtin.package:
-    name:
-      - httpd
-      - mod_ssl
+- name: Pull Apache container image
+  containers.podman.podman_image:
+    name: registry.access.redhat.com/ubi9/httpd-24:latest
```
ehelms, line 4: Given this is used in two places, a variable in `defaults.yml` would be helpful.

roles/httpd/tasks/main.yml
```diff
@@ -1,22 +1,9 @@
...
     state: present
 
-- name: Set httpd_can_network_connect so Apache can connect to Puma and Gunicorn
-  ansible.posix.seboolean:
-    name: httpd_can_network_connect
-    state: true
-    persistent: true
-
-- name: Disable welcome page
```
ehelms, line 15: Is the welcome page disabled in the container we would be using?

roles/httpd/tasks/main.yml
```diff
@@ -48,8 +35,15 @@
     dest: /etc/httpd/conf.d/foreman-ssl.conf
     mode: "0644"
 
-- name: Start Apache httpd
-  ansible.builtin.service:
-    name: httpd
+- name: Run Apache as a container
+  containers.podman.podman_container:
+    name: apache
```
ehelms, line 40: I think we want to stick with `httpd` naming schema?

roles/httpd/tasks/main.yml
archanaserver: it's not, i see it still exists, so added a task to remove it explicitly.

roles/httpd/tasks/main.yml
```diff
@@ -1,21 +1,13 @@
...
-    persistent: true
-
-- name: Disable welcome page
-  ansible.builtin.file:
-    path: /etc/httpd/conf.d/welcome.conf
-    state: absent
+- name: Disable welcome page in container
+  containers.podman.podman_container_exec:
+    container: httpd
+    command: rm -f /etc/httpd/conf.d/welcome.conf
```
ekohl, line 10: Looking at the container definition it actually uses `/usr/local/apache2/conf.d` so this shouldn't make a difference. I also wonder if such an exec persist or whether it's needed at all. We mount a volume over `conf.d` so that volume should determine `welcome.conf`.

archanaserver: After testing, I can confirm that container uses `/etc/httpd/conf.d` for its configuration. The mounted volume does not exclude `welcome.conf`, so removing it explicitly is necessary.
However, I understand that the removal via podman exec will not persist across restarts. Would modifying the host-mounted volume to exclude the file entirely be a better approach here?

ekohl: I think it highlights a potential problem with mounting those directories: we don't know what's in there. In this case you probably installed httpd which created the file.

This makes me wonder if we should create our own directory in `/etc`, like `/etc/foreman-quadlet/httpd` for Apache

archanaserver: > This makes me wonder if we should create our own directory in /etc, like /etc/foreman-quadlet/httpd for Apache

I tried doing it, you can see my new changes but the problem I'm facing is how can we make sure that container state persist. I see Podman containers do not automatically persist their state.

roles/httpd/tasks/main.yml
```diff
@@ -45,11 +36,23 @@
...
+    state: absent
+
+- name: Run Apache as a container
+  containers.podman.podman_container:
     name: httpd
+    image: "{{ httpd_container_image }}"
     state: started
-    enabled: true
+    volumes:
+      - "{{ httpd_ssl_dir }}:/etc/httpd/ssl"
```
ehelms, line 53: I was figuring we'd want to switch TLS certificate handling to using secrets to normalize it with how we do other services that have certificates.

roles/httpd/tasks/main.yml
ekohl: I meant to create the directory on the host, not in the container itself.

roles/httpd/tasks/main.yml
```diff
@@ -1,30 +1,21 @@
...
     path: "{{ item }}"
     state: directory
+    owner: root
+    group: root
     mode: "0750"
   loop:
     - "{{ httpd_ssl_dir }}/certs"
     - "{{ httpd_ssl_dir }}/private"
+    - "{{ httpd_conf_dir }}"
+    - "/var/log/httpd"
```
ekohl, line 18: Rather than copying this path in multiple places, can you introduce `httpd_log_dir`?

roles/httpd/tasks/main.yml
```diff
@@ -45,11 +36,23 @@
...
+
+- name: Run Apache as a container
+  containers.podman.podman_container:
     name: httpd
+    image: "{{ httpd_container_image }}"
     state: started
-    enabled: true
+    volumes:
+      - "{{ httpd_ssl_dir }}:/etc/httpd/ssl"
+      - "{{ httpd_conf_dir }}:/etc/httpd/conf.d"
```
ekohl, line 54: Is this how it's intended to be used? Looking at https://catalog.redhat.com/software/containers/ubi9/httpd-24/61a60c3e3e9240fca360f74a this isn't how they think about it.

roles/httpd/tasks/main.yml
```diff
@@ -45,11 +36,23 @@
...
+      - "/var/log/httpd:/usr/local/apache2/logs"
```
ekohl, line 55: Quoting https://catalog.redhat.com/software/containers/ubi9/httpd-24/61a60c3e3e9240fca360f74a
>  By default, httpd logs into standard output, so the logs are accessible by using the podman logs command. When HTTPD_LOG_TO_VOLUME is set, httpd logs into /var/log/httpd24, which can be mounted to host system using the container volumes. This option is only allowed when container is run as UID 0.

ekohl (commented):
Today we rely on inheriting the openssl policy from the host in terms of ciphers etc. With containers we no longer have that, so do we need to start providing settings for that again?

I think using this prebuilt httpd container is questionable. It really looks designed for static content hosting and not a reverse proxy setup, though it contains pretty much all httpd modules for maximum compatibility. This is another regression compared to our currently tuned httpd.

Additionally, it doesn't contain modules we require for external auth so I question how we're going to deliver that feature.

ehelms:
> Today we rely on inheriting the openssl policy from the host in terms of ciphers etc. With containers we no longer have that, so do we need to start providing settings for that again?

I poked a bit at this and here's what I could conclude. Podman appears to enforce containers inheriting FIPS if the underlying system has it enabled.  Related issue https://github.com/containers/podman/discussions/23213

I believe this implies that if we mount `/etc/crypto-policies` into our containers they will inherit the system policy (minus FIPS modifiers which I do not know how common that is in our space).

ekohl:
> I believe this implies that if we mount `/etc/crypto-policies` into our containers they will inherit the system policy (minus FIPS modifiers which I do not know how common that is in our space).

How do you know crypto policies are compatible between OS versions? If the goal is to decouple from the OS version, then I worry about this.

For example, on Fedora my current policy:
```console
$ cat /etc/crypto-policies/state/current
FEDORA40
```

I'm unsure what a good solution is.

If we look forward to deployments in OpenShift then I'd prefer to use an ingress controller instead of some httpd container, so perhaps we should view httpd in the same way? Something the platform provides.

archanaserver:
> > I believe this implies that if we mount `/etc/crypto-policies` into our containers they will inherit the system policy (minus FIPS modifiers which I do not know how common that is in our space).
> 
> How do you know crypto policies are compatible between OS versions? If the goal is to decouple from the OS version, then I worry about this.
> 
I am not sure if this is feasible here but this just occured to me- can we define crypto policies explicitly within the container (like through OpenSSL configurations) to avoid inheriting them from the host and ensure consistency across environments?

However, I see the challenge would be maintaining and updating the crypto settings inside the container. and I'm not entirely sure how we would handle that effectively.

ehelms:
It can at least be fruitful to get this into state that passes tests but leave in draft for now, and focus more on the Postgresql and Redis changes while we discuss this one more.

archanaserver:
@ehelms Apologies for responding late on this, I just revisited it!

I noticed a job failure in the job, but is unrelated to my changes I believe. Could you please take a look, if there's something I need to look into?

src/roles/httpd/tasks/main.yml
```diff
@@ -45,11 +36,25 @@
...
+- name: Run Apache as a container
+  containers.podman.podman_container:
     name: httpd
+    image: "{{ httpd_container_image }}"
     state: started
-    enabled: true
+    volumes:
+      - "{{ httpd_ssl_dir }}:/etc/httpd/ssl"
+      - "{{ httpd_conf_dir }}:/etc/httpd/conf.d"
+      - "{{ httpd_log_dir }}:/usr/local/apache2/logs"
```
ekohl, line 55: This does not match the real log directory: it uses `ServerRoot` and that is by default `/etc/httpd` in the container image you chose.

src/roles/httpd/tasks/main.yml
```diff
@@ -45,11 +36,25 @@
...
+
+- name: Run Apache as a container
+  containers.podman.podman_container:
     name: httpd
+    image: "{{ httpd_container_image }}"
     state: started
-    enabled: true
+    volumes:
+      - "{{ httpd_ssl_dir }}:/etc/httpd/ssl"
+      - "{{ httpd_conf_dir }}:/etc/httpd/conf.d"
```
ekohl, line 54: I think mapping the who configuration directory will overwrite other files and you're mixing host and container configs this way.

ekohl (changes requested):
I would like to close this because at least initially we shouldn't focus on containerizing httpd, for the reasons outlined earlier:
* It does not respect crypto-policies
* It's unclear how to integrate external authentication
* This needs a lot of work to actually finish
* We have more important work to focus on now

archanaserver:
> I would like to close this because at least initially we shouldn't focus on containerizing httpd, for the reasons outlined earlier:
> 
>     * It does not respect crypto-policies
> 
>     * It's unclear how to integrate external authentication
> 
>     * This needs a lot of work to actually finish
> 
>     * We have more important work to focus on now

I’m closing this PR for now. We can revisit containerizing httpd later. Thanks for the feedback!

PR #62, opened 2024-11-27
migrate pgsql installation from rpm to container-based

archanaserver:
This PR transitions the pgsql installation from an RPM-based installation to a containerized based.

### DIFF
--- a/playbooks/deploy.yaml
+++ b/playbooks/deploy.yaml
@@ -54,13 +54,6 @@
         password: "{{ foreman_db_password }}"
       - name: pulp
         password: "{{ pulp_db_password }}"
-    postgresql_hba_entries:
-      - { type: local, database: all, user: postgres, auth_method: ident }
-      - { type: local, database: all, user: all, auth_method: ident }
-      - { type: host, database: all, user: postgres, address: '127.0.0.1/32', auth_method: md5 }
-      - { type: host, database: all, user: postgres, address: '0.0.0.0/0', auth_method: reject }
-      - { type: host, database: all, user: all, address: '127.0.0.1/32', auth_method: md5 }
-      - { type: host, database: all, user: all, address: '::1/128', auth_method: md5 }
   roles:
     - role: certificates
       when: "certificate_source == 'default'"
@@ -71,7 +64,7 @@
         certificate_checks_certificate: "{{ server_certificate }}"
         certificate_checks_key: "{{ server_key }}"
         certificate_checks_ca: "{{ ca_certificate }}"
-    - geerlingguy.postgresql
+    - postgresql
     - redis
     - candlepin
     - httpd
--- a/roles/postgresql/defaults/main.yml
+++ b/roles/postgresql/defaults/main.yml
@@ -1,0 +1,9 @@
+---
+postgresql_image: quay.io/sclorg/postgresql-13-c9s
+postgresql_container_name: postgresql
+postgresql_network: host
+postgresql_restart_policy: always
+
+postgresql_data_dir: /var/lib/pgsql/data
+
+postgresql_admin_password: "CHANGEME"
--- a/roles/postgresql/tasks/main.yml
+++ b/roles/postgresql/tasks/main.yml
@@ -1,0 +1,65 @@
+---
+- name: Pull PostgreSQL container image
+  containers.podman.podman_image:
+    name: "{{ postgresql_image }}"
+    state: present
+
+- name: Create PostgreSQL storage directory
+  ansible.builtin.file:
+    path: "{{ postgresql_data_dir }}"
+    state: directory
+    mode: "0700"
+    owner: 26
+    group: 26
+
+- name: Create Podman secret for PostgreSQL admin password
+  containers.podman.podman_secret:
+    name: postgresql_admin_password
+    data: "{{ postgresql_admin_password }}"
+
+- name: Deploy PostgreSQL container
+  containers.podman.podman_container:
+    name: "{{ postgresql_container_name }}"
+    image: "{{ postgresql_image }}"
+    state: quadlet
+    healthcheck: pg_isready
+    sdnotify: healthy
+    network: host
+    volumes:
+      - "{{ postgresql_data_dir }}:/var/lib/pgsql/data:Z"
+    secrets:
+      - 'postgresql_admin_password,target=POSTGRESQL_ADMIN_PASSWORD,type=env'
+    quadlet_options:
+      - |
+        [Install]
+        WantedBy=default.target
+
+- name: Run daemon reload
+  ansible.builtin.systemd:
+    daemon_reload: true
+
+- name: Start the PostgreSQL Service
+  ansible.builtin.systemd:
+    name: "{{ postgresql_container_name }}"
+    enabled: true
+    state: started
+
+- name: Create PostgreSQL users
+  community.postgresql.postgresql_user:
+    name: "{{ item.name }}"
+    password: "{{ item.password }}"
+    login_user: postgres
+    login_password: "{{ postgresql_admin_password }}"
+    login_host: localhost
+    state: present
+  loop: "{{ postgresql_users }}"
+
+- name: Create PostgreSQL databases
+  community.postgresql.postgresql_db:
+    name: "{{ item.name }}"
+    owner: "{{ item.owner }}"
+    login_user: postgres
+    login_password: "{{ postgresql_admin_password }}"
+    login_host: localhost
+    state: present
+  loop: "{{ postgresql_databases }}"
--- a/tests/postgresql_test.py
+++ b/tests/postgresql_test.py
@@ -3,7 +3,20 @@ def test_postgresql_service(server):
     assert postgresql.is_running
     assert postgresql.is_enabled
 
-
 def test_postgresql_port(server):
     postgresql = server.addr("localhost")
     assert postgresql.port("5432").is_reachable
+
+
+def test_postgresql_databases(server):
+    result = server.run("podman exec postgresql psql -U postgres -c '\\l'")
+    assert "foreman" in result.stdout
+    assert "candlepin" in result.stdout
+    assert "pulp" in result.stdout
+
+
+def test_postgresql_users(server):
+    result = server.run("podman exec postgresql psql -U postgres -c '\\du'")
+    assert "foreman" in result.stdout
+    assert "candlepin" in result.stdout
+    assert "pulp" in result.stdout
--- a/playbooks/setup.yaml
+++ b/playbooks/setup.yaml
@@ -19,3 +19,4 @@
           - python3-requests
           - bash-completion
           - nmap
+          - python3-psycopg2
--- a/requirements.yml
+++ b/requirements.yml
@@ -9,6 +9,3 @@ collections:
   - name: theforeman.operations
   - name: https://github.com/theforeman/forklift
     type: git
-
-roles:
-  - name: geerlingguy.postgresql


archanaserver:
idk what is wrong with image, it is failing everytime reading manifest in `registry.access.redhat.io/rhel9/postgresql-16:latest`?

archanaserver:
also i do not like doing this way, should i go with creating a separate role for postgresql and do the containerization work there? would it be good? any thoughts?

ehelms:
> idk what is wrong with image, it is failing everytime reading manifest in `registry.access.redhat.io/rhel9/postgresql-16:latest`?

This is an official Red Hat image locked behind subscriptions, only `ubi` is available for wide usage but I think we've realized it does not come in flavors like the `rhel` images.

roles/postgresql/defaults/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+postgresql_image: docker.io/library/postgres:16
```
ehelms, line 2: We should stick with 13 for now to match our nightly deployment.

ehelms, line 2: We should stick with 13 for now to match our nightly deployment.

roles/postgresql/defaults/main.yml
```diff
@@ -0,0 +1,41 @@
...
+postgresql_data_dir: /var/lib/pgsql/data
+postgresql_volumes:
+  - "{{ postgresql_data_dir }}:/var/lib/postgresql/data:Z"
+  - "/etc/postgresql/:/etc/postgresql:Z"
+
+postgresql_admin_user: postgres
+postgresql_admin_password: "CHANGEME"
+postgresql_port: 5432
+
+postgresql_databases:
```
ehelms, line 16: I think we should define this at the playbook level, or a wrapper role rather than hardcoding this role to knowledge of foreman/candlepin/pulp.

roles/postgresql/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+
+- name: Deploy PostgreSQL container
+  containers.podman.podman_container:
+    name: "{{ postgresql_container_name }}"
+    image: "{{ postgresql_image }}"
+    state: quadlet
+    command:
+      - "postgres"
+      - "-c"
+      - "config_file=/etc/postgresql/postgresql.conf"
```
ehelms, line 50: I am surprised we have to specify this, as postgresql on EL should have sane defaults set.

roles/postgresql/defaults/main.yml
ehelms: This image is debian based, which means we would mix-n-match base OS which I think we said we do not want to do. Give this a try -- https://quay.io/repository/sclorg/postgresql-13-c9s

roles/postgresql/tasks/main.yml
archanaserver: Yeah, now I changed the way I did before, made it so complex earlier which wasn't necessary do dropped it for now.

archanaserver:
@ehelms how can we re run the workflow?

ehelms:
With Candlepin, the thing to usually check is the `candlepin.log` file:

`vi /var/log/candlepin/candlepin.log`

And the output I see is:
```
2025-02-06 13:59:51,400 [thread=main] [=, org=, csid=] ERROR org.candlepin.database.DatabaseConnectionManager - Connection to localhost:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
org.postgresql.util.PSQLException: Connection to localhost:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
```

roles/postgresql/templates/init.sh.j2
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+
+
+{% for db in postgresql_databases %}
+psql -v ON_ERROR_STOP=1 --username "$POSTGRESQL_USER" <<-EOSQL
```
evgeni, line 6: As debugged together, the user variable is unset, so the script fails, and once we fix that the db creation needs to happen after the user creation, so this patch makes it work:

```
diff --git a/roles/postgresql/templates/init.sh.j2 b/roles/postgresql/templates/init.sh.j2
index d4c464a..afb706d 100644
--- a/roles/postgresql/templates/init.sh.j2
+++ b/roles/postgresql/templates/init.sh.j2
@@ -2,16 +2,16 @@
 set -e
 
 
+{% for user in postgresql_users %}
+psql -v ON_ERROR_STOP=1 --username "postgres" <<-EOSQL
+    CREATE USER {{ user.name }} WITH PASSWORD '{{ user.password }}';
+EOSQL
+{% endfor %}
+
 {% for db in postgresql_databases %}
-psql -v ON_ERROR_STOP=1 --username "$POSTGRESQL_USER" <<-EOSQL
+psql -v ON_ERROR_STOP=1 --username "postgres" <<-EOSQL
     CREATE DATABASE {{ db.name }};
     ALTER DATABASE {{ db.name }} OWNER TO {{ db.owner }};
 EOSQL
 {% endfor %}
 
-{% for user in postgresql_users %}
-psql -v ON_ERROR_STOP=1 --username "$POSTGRESQL_USER" <<-EOSQL
-    CREATE USER {{ user.name }} WITH PASSWORD '{{ user.password }}';
-    GRANT ALL PRIVILEGES ON DATABASE {{ user.databases | first }} TO {{ user.name }};
-EOSQL
-{% endfor %}
```

roles/postgresql/templates/init.sh.j2
```diff
@@ -0,0 +1,15 @@
...
+{% for user in postgresql_users %}
+psql -v ON_ERROR_STOP=1 --username "postgres" <<-EOSQL
+    CREATE USER {{ user.name }} WITH PASSWORD '{{ user.password }}';
+EOSQL
+{% endfor %}
+
+{% for db in postgresql_databases %}
+psql -v ON_ERROR_STOP=1 --username "postgres" <<-EOSQL
+    CREATE DATABASE {{ db.name }};
+    ALTER DATABASE {{ db.name }} OWNER TO {{ db.owner }};
```
ekohl, line 13: Per https://www.postgresql.org/docs/current/sql-createdatabase.html you can set the owner on creation
```suggestion
    CREATE DATABASE {{ db.name }} OWNER {{ db.owner }};
```

roles/postgresql/templates/init.sh.j2
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e
+
+{% for user in postgresql_users %}
+psql -v ON_ERROR_STOP=1 --username "postgres" <<-EOSQL
+    CREATE USER {{ user.name }} WITH PASSWORD '{{ user.password }}';
+EOSQL
+{% endfor %}
```
ekohl, line 8: I was debating using [createuser](https://www.postgresql.org/docs/current/app-createuser.html) but reading the documentation I wonder how to properly deal with passing in the password.

Then my other thought was whether if it's better to create the users in a single query:
```suggestion
psql -v ON_ERROR_STOP=1 --username "postgres" <<-EOSQL
{% for user in postgresql_users %}
    CREATE USER {{ user.name }} WITH PASSWORD '{{ user.password }}';
{% endfor %}
EOSQL
```

And if you take that latter step to the extreme: should there just be a single script that creates the users and the databases.

roles/postgresql/templates/postgresql.container.j2
```diff
@@ -0,0 +1,11 @@
+[Container]
+ContainerName={{ postgresql_container_name }}
+Image={{ postgresql_image }}
+Network=host
+Volume={{ postgresql_data_dir }}:/var/lib/pgsql/data:Z
+Volume={{ playbook_dir }}/postgresql-init:/opt/app-root/src/postgresql-init:Z
+Environment=POSTGRESQL_ADMIN_PASSWORD={{ postgresql_admin_password }}
```
ekohl, line 7: These files will be world readable, meaning any user on the system can read the admin password. This should be supplied via a secret.

Having said that: why do we need an admin password? https://github.com/sclorg/postgresql-container/tree/master/13#postgresql-admin-account says you can still use local connections. Or is that insecure because we use the host network?

roles/postgresql/templates/init.sh.j2
ekohl, line 1: By using `init` you don't guarantee the password is correct. Perhaps not relevant now, but for the long term: do we need to provide a mechanism to rotate passwords?

Another thing I'm worried about: this will contain plain text passwords in the container. Will that be a security risk?

By the default the container already contains code to create a single user (`POSTGRESQL_USER`, ` POSTGRESQL_PASSWORD`, `POSTGRESQL_DATABASE`) which could be provided via secrets. It makes me wonder what would be a good pattern. Should we have a PostgreSQL container for every database?

roles/postgresql/templates/postgresql.container.j2
archanaserver: > These files will be world readable, meaning any user on the system can read the admin password. This should be supplied via a secret.

That's right! To address this i've moved the password storage to a dedicated secrets file and restrict file permission to `0600` so only container process can now access it. But I'm trying to do this with Podman secrets.

> 
> Having said that: why do we need an admin password? https://github.com/sclorg/postgresql-container/tree/master/13#postgresql-admin-account says you can still use local connections. Or is that insecure because we use the host network?

While the container allows local connections without a password here by default, but host networking exposes PostgreSQL to all network interfaces I believe(not just localhost). So without a password, anyone on the network could access the database with postgres privileges. And admin password is that is gonna adds a layer of security here, even if local connections are allowed. If we switch to a non-host network later, maybe the password isn't needed, but for now, it's safer.

roles/postgresql/templates/postgresql.container.j2
```diff
@@ -0,0 +1,11 @@
+[Container]
+ContainerName={{ postgresql_container_name }}
+Image={{ postgresql_image }}
+Network=host
+Volume={{ postgresql_data_dir }}:/var/lib/pgsql/data:Z
+Volume={{ playbook_dir }}/postgresql-init:/opt/app-root/src/postgresql-init:Z
+EnvironmentFile=/etc/secrets/postgresql_admin_password
+PodmanArgs=--security-opt label=disable
```
ekohl, line 8: Why is this needed?

roles/postgresql/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+    path: "{{ postgresql_data_dir }}"
+    state: directory
+    mode: "0700"
+    owner: 26
+    group: 26
+
+- name: Configure host based authentication
+  ansible.builtin.template:
+    src: pg_hba.conf.j2
+    dest: "/var/lib/pgsql/data/pg_hba.conf"
```
ekohl, line 13:
```suggestion
    dest: "{{ postgresql_data_dir }}/pg_hba.conf"
```

roles/postgresql/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+  ansible.builtin.copy:
+    dest: /etc/secrets/postgresql_admin_password
+    content: |
+      POSTGRESQL_ADMIN_PASSWORD={{ postgresql_admin_password }}
+    owner: root
+    group: root
+    mode: "0600"
+  no_log: true
+
+- name: Deploy Quadlet configuration
```
ekohl, line 46: In other places we use `containers.podman.podman_container` so why the direct implementation this time?

roles/postgresql/templates/postgresql.container.j2
```diff
@@ -0,0 +1,11 @@
+[Container]
+ContainerName={{ postgresql_container_name }}
+Image={{ postgresql_image }}
+Network=host
+Volume={{ postgresql_data_dir }}:/var/lib/pgsql/data:Z
+Volume={{ playbook_dir }}/postgresql-init:/opt/app-root/src/postgresql-init:Z
```
ekohl, line 6: I'm not a fan of mounting something from the playbook directory. In general I'd assume you can delete the "installer" and this would break that assumption. We've already made that mistake in the past (looking at `/root/ssl-build`) and I'd prefer to avoid it.

roles/postgresql/templates/postgresql.container.j2
```diff
@@ -0,0 +1,11 @@
...
+EnvironmentFile=/etc/secrets/postgresql_admin_password
```
ekohl, line 7: I'd prefer a podman secret that you expose via an environment variable. The `/etc/secrets` directory is non-standard and I'd like to avoid creating those.

ekohl (changes requested):

playbooks/deploy.yaml
```diff
@@ -44,10 +44,13 @@
     postgresql_users:
       - name: candlepin
         password: "{{ candlepin_db_password }}"
+        databases: [candlepin]
```
ekohl, line 47: This `databases` entry doesn't appear to be used.
```suggestion
```

roles/postgresql/tasks/main.yml
```diff
@@ -0,0 +1,71 @@
...
+  ansible.builtin.systemd:
+    name: "{{ postgresql_container_name }}"
+    enabled: true
+    state: started
+
+- name: Wait for PostgreSQL to be ready
+  ansible.builtin.wait_for:
+    port: 5432
+    host: "localhost"
+    timeout: 300
```
ekohl, line 71: Taking lessons from https://github.com/theforeman/foreman-quadlet/pull/114, I'd really prefer if you used `sdnotify: true` on the PostgreSQL container definition so that it uses the built in notify support when the service is ready.

That way after a reboot you are also certain that anything that needs PostgreSQL can actually talk to PostgreSQL.

evgeni: `sdnotify: true` doesn't seem to work with the postgres container, but you can do
```
healthcheck: pg_isready
sdnotify: healthy
```

roles/postgresql/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+- name: Create PostgreSQL users
+  community.postgresql.postgresql_user:
+    name: "{{ item.name }}"
+    password: "{{ item.password }}"
+    login_user: postgres
+    login_password: "{{ postgresql_admin_password }}"
+    login_host: localhost
+    state: present
+  loop: "{{ postgresql_users }}"
+  when: postgresql_users | length > 0
```
evgeni, line 56: You can omit the `when` here, if the array is empty, the loop will not be executed :)

roles/postgresql/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+      POSTGRESQL_USER: "postgres"
+      POSTGRESQL_PASSWORD: "{{ postgresql_admin_password }}"
+      POSTGRESQL_DATABASE: "{{ postgresql_databases[0].name | default('foreman') }}"
+      POSTGRESQL_ADMIN_PASSWORD: "{{ postgresql_admin_password }}"
+    quadlet_options:
+      - |
+        [Install]
+        WantedBy=default.target
+
+- name: Create PostgreSQL users
```
evgeni, line 47: For this, the DB must be running already, so this (and the db creation) needs to happen after "Start the PostgreSQL Service"

roles/postgresql/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+    state: directory
+    mode: "0700"
+    owner: 26
+    group: 26
+
+- name: Configure host based authentication
+  ansible.builtin.template:
+    src: pg_hba.conf.j2
+    dest: "{{ postgresql_data_dir }}/pg_hba.conf"
+    mode: "0644"
```
evgeni, line 19: If I am reading https://github.com/sclorg/postgresql-container/blob/master/13/root/usr/share/container-scripts/postgresql/common.sh#L228-L240 correctly, we don't need to configure hba in the container case, as the container is already configured as "everyone can connect"

tests/postgresql_test.py
```diff
@@ -1,9 +1,30 @@
...
+def test_postgresql_users(server):
+    result = server.run("podman exec postgresql psql -U postgres -c '\\du'")
+    assert "foreman" in result.stdout
+    assert "candlepin" in result.stdout
+    assert "pulp" in result.stdout
+
+
+def test_postgresql_hba_config(server):
+    result = server.run("podman exec postgresql cat /var/lib/pgsql/data/pg_hba.conf")
+    assert re.search(r"host\s+all\s+all\s+127\.0\.0\.1/32\s+md5", result.stdout)
```
evgeni, line 30: You dropped the hba editing, so I think this test can be removed (and then the `re` import too)

roles/postgresql/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+    name: "{{ postgresql_container_name }}"
+    image: "{{ postgresql_image }}"
+    state: quadlet
+    healthcheck: pg_isready
+    sdnotify: healthy
+    network: host
+    volumes:
+      - "{{ postgresql_data_dir }}:/var/lib/pgsql/data:Z"
+    env:
+      POSTGRESQL_ADMIN_PASSWORD: "{{ postgresql_admin_password }}"
```
evgeni, line 31: You created a secret for that above, so let's use it
```suggestion
    secrets:
      - 'postgresql_admin_password,target=POSTGRESQL_ADMIN_PASSWORD,type=env'
```

tests/postgresql_test.py
archanaserver: Done!

roles/postgresql/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
...
+postgresql_network: host
+postgresql_restart_policy: always
+
+postgresql_data_dir: /var/lib/pgsql/data
+
+postgresql_admin_password: "CHANGEME"
+
+postgresql_databases: []
+postgresql_users: []
+postgresql_hba_entries: []
```
evgeni, line 13: This can be removed now (and so can the setting of it in `deploy.yaml`)

roles/postgresql/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+postgresql_image: quay.io/sclorg/postgresql-13-c9s
+postgresql_container_name: postgresql
+postgresql_network: host
+postgresql_restart_policy: always
```
evgeni, line 5: These two variables are unused right now. Do you have plans to use them?
(I don't think we *need* that flexibility right now, but it surely can be handy in the future)

archanaserver: this was added with the earlier setup, but i guess we can drop it for now.

ehelms:
Great to see this working! How much data do we want on how upgrades work (https://github.com/theforeman/foreman-quadlet/issues/66) before we merge this change?

ehelms:
I'll do a final review of this PR then. @evgeni could you add those upgrade notes to https://github.com/theforeman/foreman-quadlet/issues/66 ?

ehelms:
@archanaserver Do you know if this container setup satisfies https://github.com/theforeman/foreman-quadlet/issues/106 ?

archanaserver:
> @archanaserver Do you know if this container setup satisfies #106 ?

No it doesn't, currently container setup by default uses `md5`. So I can take look into adding explici settings for `scram-sha-256`.

ehelms (approved):
The SCRAM update can be a follow up.

PR #63, opened 2024-11-27
migrate redis installation from rpm to container-based

archanaserver:


roles/redis/tasks/main.yaml
```diff
@@ -1,14 +1,28 @@
...
+- name: Pull Redis container image
+  containers.podman.podman_image:
+    name: "{{ redis_image }}"
     state: present
 
-- name: Start the Redis Service
-  ansible.builtin.systemd:
+- name: Create directory for Redis data
+  ansible.builtin.file:
+    path: /data/redis
```
ehelms, line 9: What is `/data` ? I am surprised we are not using `/var/lib/redis`.

roles/redis/defaults/main.yml
```diff
@@ -0,0 +1 @@
+redis_image: docker.io/redis:6
```
ehelms, line 1: I think we should change this to a centos/stream9 container even if we have to build it ourselves. That could be logged as an issue and done as a follow up change.

tests/redis_test.py
```diff
@@ -1,16 +1,13 @@
 import pytest
 
-
```
ehelms, line 3: We use two blank lines in all our other files, so let's drop this change.

roles/redis/defaults/main.yml
archanaserver: https://github.com/theforeman/foreman-quadlet/issues/94

roles/redis/tasks/main.yaml
```diff
@@ -1,14 +1,39 @@
...
+    quadlet_options:
+      - |
+        [Install]
+        WantedBy=default.target
+
+- name: Run daemon reload
+  ansible.builtin.systemd:
+    daemon_reload: true
+
+- name: Start the Quadlet Demo Service
```
ekohl, line 33: Please correct the name here

roles/redis/defaults/main.yml
```diff
@@ -0,0 +1 @@
+redis_image: quay.io/akumari/redis-rpm:nightly
```
ekohl, line 1: This depends on the outcome of https://github.com/theforeman/foreman-quadlet/pull/97

roles/redis/tasks/main.yaml
```diff
@@ -1,14 +1,39 @@
...
+        WantedBy=default.target
+
+- name: Run daemon reload
+  ansible.builtin.systemd:
+    daemon_reload: true
+
 - name: Start the Redis Service
   ansible.builtin.systemd:
     name: redis
-    enabled: true
```
ekohl, line 10: Any reason `enabled` is no longer needed? Normally that's still needed if you want it to start on reboot.

archanaserver:
Is this test failure is related to my change? I just checked it is the same failure coming in postgresql PR as well. 
https://github.com/theforeman/foreman-quadlet/actions/runs/13651740488/job/38161679546?pr=63#step:10:475

ehelms:
> Isn't that the race condition we've seen more often?

I believe so.

archanaserver:
> Isn't that the race condition we've seen more often?

What does that mean?

ekohl:
> > Isn't that the race condition we've seen more often?
> 
> What does that mean?

I thought it was the same thing as https://github.com/theforeman/foreman-quadlet/pull/65#issuecomment-2663729977.

PR #65, opened 2025-01-07
disable pulpcore analytics

archanaserver:
Fixes https://github.com/theforeman/foreman-quadlet/issues/58

### DIFF
--- a/roles/pulp/templates/settings.py.j2
+++ b/roles/pulp/templates/settings.py.j2
@@ -17,3 +17,4 @@ DATABASES = {
 AUTHENTICATION_BACKENDS=['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
 REMOTE_USER_ENVIRON_NAME="HTTP_REMOTE_USER"
 REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES=('rest_framework.authentication.SessionAuthentication', 'pulpcore.app.authentication.PulpRemoteUserAuthentication')
+ANALYTICS = {{ pulp_enable_analytics | ternary('True', 'False') }}
--- a/roles/pulp/defaults/main.yaml
+++ b/roles/pulp/defaults/main.yaml
@@ -14,3 +14,5 @@ pulp_content_container_name: pulp-content
 pulp_worker_container_name: pulp-worker
 
 pulp_content_origin: "http://{{ ansible_fqdn }}:24816"
+
+pulp_enable_analytics: false


archanaserver:
Can anyone help with, how can I ensure here that the setting I'm trying to make is updated in our `setting.py` file? Any thoughts? Also does it require any other config update?

roles/pulp/templates/settings.py.j2
```diff
@@ -17,3 +17,4 @@ DATABASES = {
 AUTHENTICATION_BACKENDS=['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
 REMOTE_USER_ENVIRON_NAME="HTTP_REMOTE_USER"
 REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES=('rest_framework.authentication.SessionAuthentication', 'pulpcore.app.authentication.PulpRemoteUserAuthentication')
+ENABLE_ANALYTICS = {{ pulp_enable_analytics | default(false) }}
```
ekohl, line 20: Without looking too closely, I think this ends up as `true` or `false` but Python uses `True` and `False`

archanaserver: I've updated it to explicitly cast the value as a boolean using `| bool`, ensuring it correctly translates to `True` or `False`.

archanaserver:
@ekohl @ehelms can you help me understand what is this ansible lint failure here?
https://github.com/theforeman/foreman-quadlet/actions/runs/13130019605/job/36633198924?pr=65

ehelms:
Trying to fix that over here -- https://github.com/theforeman/foreman-quadlet/pull/74

roles/pulp/templates/settings.py.j2
```diff
@@ -17,3 +17,4 @@ DATABASES = {
 AUTHENTICATION_BACKENDS=['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
 REMOTE_USER_ENVIRON_NAME="HTTP_REMOTE_USER"
 REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES=('rest_framework.authentication.SessionAuthentication', 'pulpcore.app.authentication.PulpRemoteUserAuthentication')
+ENABLE_ANALYTICS = {{ pulp_enable_analytics | default(false) | bool | ternary('True', 'False') }}
```
ehelms, line 20: I think this should be `ANALYTICS` - https://github.com/theforeman/puppet-pulpcore/blob/master/templates/settings.py.erb#L17

ehelms:
Also, I am not sure why the tests are failing right now. This appears unrelated to your change.

roles/pulp/templates/settings.py.j2
```diff
@@ -17,3 +17,4 @@ DATABASES = {
 AUTHENTICATION_BACKENDS=['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
 REMOTE_USER_ENVIRON_NAME="HTTP_REMOTE_USER"
 REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES=('rest_framework.authentication.SessionAuthentication', 'pulpcore.app.authentication.PulpRemoteUserAuthentication')
+ANALYTICS = {{ pulp_enable_analytics | default(false) | bool | ternary('True', 'False') }}
```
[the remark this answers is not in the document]
archanaserver: You're right! I initially used `default(false)` to ensure that if `pulp_enable_analytics` was `None` or `undefined`, so it would default to `false`. And the `bool` filter to explicitly cast the value to a boolean, so the ternary filter would behave as expected. But, after looking at the doc you mentioned the ternary filter, I realize it already handling boolean evaluation, so the default and bool aren't needed. 

Thanks! I've simplified the expression now.

archanaserver:
How the failure is coming from different test every time? Any idea, why this is failing? @ehelms @ekohl

ekohl:
That looks like a race condition:
```
Pending tasks detected in repositories of this content view. Please wait for the tasks: - http://quadlet.example.com/foreman_tasks/tasks/0c2d6155-ad28-4864-baa2-17b4c0120229 before publishing.
```

I suspect the sync action is returning a task that sometimes completes on time, sometimes it doesn't. That probably needs a `wait_for_task`, but I'm not that familiar with the content side.

evgeni (approved):

PR #71, opened 2025-01-16
Configuring Candlepin with environment variables

ehelms:
I took one configuration option for `candlepin.conf` to see what it would look like to have an option we configure to be deployed as an environment variable.

### DIFF
--- a/roles/candlepin/defaults/main.yml
+++ b/roles/candlepin/defaults/main.yml
@@ -15,6 +15,7 @@ candlepin_ciphers:
 candlepin_container_image: quay.io/ehelms/candlepin
 candlepin_container_tag: 4.4.14
 
+candlepin_db_user: candlepin
 candlepin_db_host: localhost
 candlepin_db_port: 5432
 candlepin_db_ssl: false
--- a/roles/candlepin/templates/candlepin.conf.j2
+++ b/roles/candlepin/templates/candlepin.conf.j2
@@ -9,7 +9,7 @@ candlepin.audit.hornetq.config_path=/etc/candlepin/broker.xml
 candlepin.db.database_manage_on_startup=Manage
 
 candlepin.auth.oauth.enable=true
-candlepin.auth.oauth.consumer.katello.secret={{ candlepin_oauth_secret }}
+candlepin.auth.oauth.consumer.katello.secret="NOTTHEREALSECERT"
 
 module.config.adapter_module=org.candlepin.katello.KatelloModule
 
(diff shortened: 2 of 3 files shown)


roles/candlepin/defaults/main.yml
```diff
@@ -15,6 +15,7 @@ candlepin_ciphers:
 candlepin_container_image: quay.io/ehelms/candlepin
 candlepin_container_tag: 4.4.14
 
+candlepin_db_user: candlepin
```
ekohl, line 18: This looks like you were experimenting with something else and ended up committing it too.

roles/candlepin/templates/candlepin.conf.j2
```diff
@@ -9,7 +9,7 @@ candlepin.audit.hornetq.config_path=/etc/candlepin/broker.xml
 candlepin.db.database_manage_on_startup=Manage
 
 candlepin.auth.oauth.enable=true
-candlepin.auth.oauth.consumer.katello.secret={{ candlepin_oauth_secret }}
+candlepin.auth.oauth.consumer.katello.secret="NOTTHEREALSECERT"
```
ekohl, line 12: Do we even need to include this in the config file if we're not passing it? Right now I can see a failure mode where it somehow isn't provided/respected and the password ends up being `NOTTHEREALSECERT`. If you don't provide at all I'd hope that Candlepin has a sane default that refuses to work.

roles/candlepin/defaults/main.yml
ehelms: Correct - I started doing all of them and then decided to just do one to see what it looked and felt like.

roles/candlepin/templates/candlepin.conf.j2
ehelms: I have it here to test whether the environment variables overwrite what's in the config. To see if that provides a way to ship defaults but know that we can provide a configuration option to users if need be by adding the environment variable.

If we went this route as a whole, the question would be do we include this in the `.conf` file available in the container. I could see for anything that is a "secret" we opt *not* to include it. For other elements, we include a sane default that can be overridden.

ehelms:
The programmatic configuration options detailed below do not appear to be able to be set via environment variables in the combinations I have tried. This would mean we'd still end up with having to deploy a config file.

```
org.quartz.jobStore.dataSource=myDS

org.quartz.dataSource.myDS.driver=org.postgresql.Driver
org.quartz.dataSource.myDS.user=candlepin
org.quartz.dataSource.myDS.password={{ candlepin_db_password }}
org.quartz.dataSource.myDS.maxConnections=5
```

PR #74, opened 2025-02-04
Ignore .ansible directory for linting

ehelms:


.ansible-lint
```diff
@@ -1,3 +1,4 @@
 ---
 exclude_paths:
   - .github/
+  - .ansible/
```
ekohl, line 4: I always like to sort things alphabetically

ekohl (commented):
I  suppose this relates to https://github.com/theforeman/foreman-quadlet/blob/177348fc149e74a4adb6680cb239181f16cf8476/ansible.cfg#L6 so it makes sense we don't lint external roles.

ekohl (approved):

PR #76, opened 2025-02-12
Allow running a single test file

ehelms:


run_tests
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
 
 vagrant ssh-config > .vagrant/ssh-config
-py.test -vv tests/*_test.py
+
+if [ -z "$1" ]; then
+  py.test -vv tests/*_test.py
+else
+  py.test -vv $1
+fi
```
ekohl, line 9: https://docs.pytest.org/en/latest/how-to/usage.html#specifying-tests-selecting-tests says:
> In general, pytest is invoked with the command pytest (see below for [other ways to invoke pytest](https://docs.pytest.org/en/latest/how-to/usage.html#invoke-other)). This will execute all tests in all files whose names follow the form test_*.py or \*_test.py in the current directory and its subdirectories. More generally, pytest follows [standard test discovery rules](https://docs.pytest.org/en/latest/explanation/goodpractices.html#test-discovery).

So really, do we even need to specify `tests/*_test.py`? In other words, can we use:
```suggestion
py.test -vv "$@"
```

ehelms: Excellent, added.

ekohl (approved):

PR #81, opened 2025-02-18
use pulp_smart_proxy not smart_proxy_pulp

evgeni:


roles/pulp/defaults/main.yaml
```diff
@@ -14,5 +14,7 @@ pulp_content_container_name: pulp-content
 pulp_worker_container_name: pulp-worker
 
 pulp_content_origin: "http://{{ ansible_fqdn }}:24816"
+pulp_rhsm_url: "https://{{ ansible_fqdn }}/rhsm"
+pulp_pulp_url: "https://{{ ansible_fqdn }}"
```
ekohl, line 18: Keeping in line with the origin that's on port 24816, should this match the default of the all-in-one container?
```suggestion
pulp_pulp_url: "http://{{ ansible_fqdn }}:24817"
```

evgeni: Yeah

roles/pulp/defaults/main.yaml
```diff
@@ -14,5 +14,7 @@ pulp_content_container_name: pulp-content
 pulp_worker_container_name: pulp-worker
 
 pulp_content_origin: "http://{{ ansible_fqdn }}:24816"
+pulp_rhsm_url: "https://{{ ansible_fqdn }}/rhsm"
```
ekohl, line 17: I was debating this: if we don't expose the setting (or it has no value), how does Katello interpret this? It's defined [here](https://github.com/Katello/katello/blob/83498c4f1e027faf730f6d40d27dcdf309d0af1e/app/models/katello/concerns/smart_proxy_extensions.rb#L605-L615).

What if we enhance this that if it isn't set, Katello will show its own RHSM API URL? That way you can present a Pulp server without also the RHSM API. Because essentially this is really the `{{ foreman_url }}/rhsm` value unless it's a content proxy (which we don't support yet).

ekohl: https://github.com/Katello/katello/pull/11328 takes a stab at that.

src/roles/pulp/defaults/main.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-pulp_container_image: quay.io/foreman/pulp
+pulp_container_image: quay.io/evgeni/pulp-smart
```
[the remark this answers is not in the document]
evgeni: Once it is merged, we can drop the `[TMP] use evgenis fork` commit and then this is ready for merging.

src/playbooks/deploy/deploy.yaml
```diff
@@ -66,5 +62,4 @@
     - candlepin
     - httpd
     - pulp
-    - foreman_proxy
```
evgeni, line 69: should the role (and its tests) be completely deleted?

ehelms: Yes, we can find it again in history if we need.

evgeni: done

PR #85, opened 2025-02-18
Print the 10 slowest tests

ehelms:


ekohl (approved):
I thought the first test to use the client environment (which syncs in https://fixtures.pulpproject.org/rpm-no-comps/) always going to be the slowest, but it separates out the fixture duration so that's nice. That the `test_foreman_content_view` teardown takes so long does surprise me. Why is that so slow?

PR #89, opened 2025-02-24
Refactor certificates role to normalize server and client certificate…

ehelms:
… creation


ekohl:
Again, this also makes me think about how we could implement an ACME client to get what we used to call custom certificates

roles/certificates/tasks/ca.yml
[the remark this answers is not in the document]
ehelms: It could be worth a re-factor later on, as most of what I used was already written in Forklift, it was easy to just grab it and go. The certificate generation code is not super important in my opinion - we'll need a reference / default implementation but I am from the outset trying to ensure that clear separation we never achieved before.

ehelms:
> Again, this also makes me think about how we could implement an ACME client to get what we used to call custom certificates

I am more focused on the "input", the interface for providing certificates and so think of this certificate role as a reference implementation.

ekohl:
I often feel Ansible with all of its variables doesn't make a clear distinction between (user) input and (intermediate) output. It's late so I'm not going to thoroughly review it now, but perhaps it's already something to think about

ehelms:
> I often feel Ansible with all of its variables doesn't make a clear distinction between (user) input and (intermediate) output. It's late so I'm not going to thoroughly review it now, but perhaps it's already something to think about

Oh I have :) But baby steps.. first this re-factor, then my take on that part as this helps pave the way.

roles/certificates/templates/openssl.cnf.j2
```diff
@@ -28,44 +28,32 @@ unique_subject  = no
...
 nsCertType              = sslCA
 keyUsage                = cRLSign, keyCertSign
 extendedKeyUsage        = serverAuth, clientAuth
-subjectAltName          = @alt_names
 
 [ v3_req ]
 basicConstraints     = CA:FALSE
 subjectKeyIdentifier = hash
 extendedKeyUsage     = serverAuth, clientAuth
-subjectAltName       = @alt_names
```
ekohl, line 65: Don't we need a SAN on the certs?

ehelms: Added back, via CLI since it needs to be dynamic.

ehelms:
> This may be taking it a step too far in this PR, but I'd be tempted to work towards getting rid of the separate server and client certs. In vanilla Foreman we (by default) always used the same certificates to be the server and the client.

I think this is a fair design goal that we can work to support. The current certificates implementation for upgrades will be the `puppet-certs` implementation that does split server and client certificates and indicates their purpose. That's one implementation. This reference implementation attempts to not stray far from that basic design principle. I think, at the start, this also helps make sure we consider that there are places where client certificates are expected and places where server certificates are expected. Early on, I don't want to lose site of that and have a gap.

The place I'd like to end up with is to have the following:

 1. A reference implementation for certificate generation, that is used by default and by developers.
 2. Documentation that lays out what the required certificates are and any requirements.
 3. Eliminate the server vs. default concept that we created - either use the defaults or bring them all.
 
 Then for example, we can document it was a client certificate is needed here and a server certificate need there. You the user can choose to have a single certificate for both, that has both client and server purpose. Or you can choose to split them into dedicated purposes if you want that level of separation of concern.

ekohl:
I was wondering how to verify this. Could we somehow run `katello-certs-check` against the result of this in CI? Do we have any molecule tests already that we could leverage?

ehelms:
> I was wondering how to verify this. Could we somehow run `katello-certs-check` against the result of this in CI? Do we have any molecule tests already that we could leverage?

I think adding some tests to ensure this role does what we expect and vets the generated certificates against `katello-certs-check` is a good idea. 

This started as a re-factor to better be in service of `https://github.com/theforeman/foreman-quadlet/pull/90` to help out with our installation research. 

I'll look at adding some tests and verification.

ehelms:
I added a role to run certificate check using katello-certs-check. And, for this iteration, pulled out the certificate variables generated into a vars fie. This draws the line between user input and "internal" input parameters. 

@evgeni I am thinking then for the installer split work, instead of defining them directly (https://github.com/theforeman/foreman-quadlet/commit/f42585699b4b6dc36626c83542da41bd8bf4af53) you can provide a vars file that defines the certificates coming from `puppet-certs` and then just include that file conditionally rather than the one from the certificates role.

playbooks/deploy.yaml
```diff
@@ -3,35 +3,38 @@
   hosts:
     - quadlet
   become: true
+  vars_files:
+    - "../vars/{{ certificate_source }}_certificates.yml"
```
evgeni, line 7:
```suggestion
    - "../vars/{{ certificate_source | default('default') }}_certificates.yml"
```
Then you don't have to pass the default (default).

playbooks/deploy.yaml
```diff
@@ -57,6 +60,11 @@
       - { type: host, database: all, user: all, address: '::1/128', auth_method: md5 }
   roles:
     - certificates
```
evgeni, line 62:
```suggestion
    - role: certificates
      when:
        - "certificate_source | default('default')  == 'default'"
```
or something like that (untested)?

playbooks/deploy.yaml
evgeni: Better, probably:
```suggestion
    - "../vars/{{ certificate_source }}_certificates.yml"
  vars:
    certificate_source: default
```

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable says `-e certificate_source=installer` would still win, if set

playbooks/deploy.yaml
ehelms: Excellent ideas, added.

playbooks/deploy.yaml
```diff
@@ -56,7 +60,13 @@
       - { type: host, database: all, user: all, address: '127.0.0.1/32', auth_method: md5 }
       - { type: host, database: all, user: all, address: '::1/128', auth_method: md5 }
   roles:
-    - certificates
+    - role: certificates
+      when: "certificate_source | default('default') == 'default'"
```
evgeni, line 64:
```suggestion
      when: "certificate_source == 'default'"
```
(this should work with the latest changes)

roles/certificates/tasks/main.yml
```diff
@@ -1,132 +1,7 @@
...
-      -extensions ssl_client
-      -passin "file:{{ certificates_ca_directory_keys }}/ca.pwd"
-      -config "{{ certificates_ca_directory }}/openssl.cnf"
-      -in "{{ certificates_ca_directory_requests }}/{{ certificates_client }}.csr"
-      -out "{{ certificates_ca_directory_certs }}/{{ certificates_client }}.crt"
-  args:
-    creates: "{{ certificates_ca_directory_certs }}/{{ certificates_client }}.crt"
+- name: 'Generate CA certificate'
+  ansible.builtin.include_tasks: ca.yml
+  when: certificates_ca
```
evgeni, line 4: Under which circumstances would this be false?

evgeni:
I like this, especially as it makes my "use puppet-certs certs with quadlet" frankenstein easier. And tests are also passing, so :tada: 
So unless @ekohl has objections, I'd love to merge this, as it's an improvement and we can layer on further ones on top.

roles/certificates/tasks/main.yml
ehelms: When generating a certificate bundle for use on proxies - https://github.com/theforeman/foreman-quadlet/pull/90/files#diff-b1491adbed08c8b00b570245e695c525e8dbf2a9ef94181d292272ffae52b5a5R6

evgeni: Ah, so you ran this once with `certificates_ca: true` to create a CA, and then you run it with `false` (on the same machine) to create the bundle (with a new hostname -- for a different machine) without regenerating the CA?

Does that imply ca.yml is not idempotent? Or is that pure speed optimization?

ehelms: Speed and safety optimization I suppose. I was also not sure how the design would fall out where something else provided the certificates yet (e.g. puppet-certs). I am still not 100% and this will get re-factored as we iterate.

roles/certificate_checks/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
+---
+- name: Fetch katello_certs_check
+  ansible.builtin.get_url:
+    url: https://raw.githubusercontent.com/theforeman/foreman-installer/refs/heads/develop/bin/katello-certs-check
+    dest: /usr/bin/katello-certs-check
+    mode: '0744'
```
ekohl, line 6:
```suggestion
    mode: '0755'
```

ekohl (approved):
I like the change where a lot of the fields on the cert become optional. It always bothered me we had those weird defaults that probably nobody changed and served no purpose.

PR #90, opened 2025-02-24
Generate certificate bundle

ehelms:

There is no foremanctl equivalent of `foreman-proxy-certs-generate`. Proxy and secondary systems need certificate bundles to establish trust with the Foreman server, and currently there is no way to generate them.

* Add `foremanctl certificate-bundle <hostname>` command that generates a certificate tarball for a proxy/secondary system
* New `certificate_bundle` role that packages certificates into `/var/lib/foremanctl/certs/bundles/<hostname>.tar.gz`
* Proxy certificates are generated in the same `certificates_ca_directory` as server certificates, relying on hostname-differentiated filenames to avoid conflicts
* Two modes supported: default (all certs from internal CA) and custom (user-supplied server cert/key/CA with internally-generated client certs)
* Custom server certificate flags (`--certificate-server-certificate`, `--certificate-server-key`, `--certificate-server-ca-certificate`) must be provided together
* Add directory creation to `certificates` role `issue.yml` so it works standalone without prior CA generation
* Update `development/playbooks/custom-certs` to support generating certs for arbitrary hostnames via `--hostname`
* Update `docs/user/certificates.md` with certificate bundle usage, tarball contents, and internal design

Steps to reproduce:

* Deploy Foreman: `foremanctl deploy --foreman-initial-admin-password=changeme`
* Generate a certificate bundle: `foremanctl certificate-bundle proxy.example.com`
* Verify tarball at `/var/lib/foremanctl/certs/bundles/proxy.example.com.tar.gz` contains `certs/certs/ca.crt`, `certs/certs/server-ca.crt`, `certs/certs/proxy.example.com.crt`, `certs/certs/proxy.example.com-client.crt`, `certs/private/proxy.example.com.key`, `certs/private/proxy.example.com-client.key`
* Test with custom server certs: deploy with `--certificate-source=custom_server`, then run `foremanctl certificate-bundle --certificate-server-certificate /path/to/proxy.crt --certificate-server-key /path/to/proxy.key --certificate-server-ca-certificate /path/to/ca.crt proxy.example.com`
* Run tests: `python -m pytest tests/certificate_bundle_test.py -vv`

#### Checklist


evgeni:
This is sort-of, but not exactly, foreman-proxy-certs-generate?

src/roles/certificate_bundle/tasks/main.yml
```diff
@@ -0,0 +1,90 @@
...
+      ansible.builtin.copy:
+        src: "{{ certificate_bundle_server_ca_certificate }}"
+        dest: "{{ certificate_bundle_build_directory.path }}/ssl-build/katello-server-ca.crt"
+        remote_src: true
+        mode: '0444'
+
+    - name: Copy server certificate
+      ansible.builtin.copy:
+        src: "{{ certificate_bundle_server_certificate }}"
+        dest: "{{ certificate_bundle_build_directory.path }}/ssl-build/{{ certificate_bundle_hostname }}/{{ certificate_bundle_hostname }}-{{ item }}"
```
evgeni: style nit: I'd not put the file extension in the `loop` but here

```suggestion
        dest: "{{ certificate_bundle_build_directory.path }}/ssl-build/{{ certificate_bundle_hostname }}/{{ certificate_bundle_hostname }}-{{ item }}.crt"
```

(and similarly for the later invocations of the same pattern)

src/playbooks/certificate-bundle/certificate-bundle.yaml
```diff
@@ -0,0 +1,43 @@
...
+    certificates_ca_directory: "{{ _certificates_root }}/hosts/{{ hostname }}"
+    certificates_hostnames:
+      - "{{ hostname }}"
+    _ca_crt: "{{ _certificates_root }}/certs/ca.crt"
+    _host_crt: "{{ certificates_ca_directory }}/certs/{{ hostname }}.crt"
+    _host_key: "{{ certificates_ca_directory }}/private/{{ hostname }}.key"
+  pre_tasks:
+    - name: Read CA password from remote host
+      ansible.builtin.slurp:
+        src: "{{ _certificates_root }}/private/ca.pwd"
```
evgeni: Shouldn't you be able to get this via
```
src/vars/base.yaml:certificates_ca_password_file: "{{ obsah_state_path }}/certificates-ca-password"
src/vars/base.yaml:certificates_ca_password: "{{ lookup('ansible.builtin.password', certificates_ca_password_file, chars=['ascii_letters', 'digits']) }}"
```

development/playbooks/custom-certs/custom-certs.yaml
```diff
@@ -4,10 +4,23 @@
...
+        certificates_ca_directory: "{{ _custom_certs_root }}"
+        certificates_hostnames: []
+
+    - name: Generate host certificates
+      ansible.builtin.include_role:
+        name: certificates
+      vars:
+        certificates_ca: false
+        certificates_signing_ca_directory: "{{ _custom_certs_root }}"
+        certificates_ca_directory: "{{ _custom_certs_root }}/{{ hostname | default(ansible_facts['fqdn']) }}"
```
evgeni: I don't understand why you had to split these and then override 'em

evgeni: oooh, because dev setup is special?

src/playbooks/certificate-bundle/certificate-bundle.yaml
evgeni: but loading base.yaml also sets certificates_hostnames, and vars_files wins over vars. :exploding_head: 

https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

evgeni:
I was able to register a (puppet installed) proxy to a foremanctl setup with this \o/

tests/certificate_bundle_test.py
```diff
@@ -0,0 +1,124 @@
...
+    f'ssl-build/{HOSTNAME}/{HOSTNAME}-puppet-client.crt',
+    f'ssl-build/{HOSTNAME}/{HOSTNAME}-puppet-client.key',
+]
+
+
+@pytest.fixture(scope="module")
+def generate_custom_proxy_certs(server, certificate_source):
+    if certificate_source != 'custom_server':
+        yield
+        return
```
evgeni:
```suggestion
        return
```

this should be sufficient, there is nothing to yield here

tests/certificate_bundle_test.py
```diff
@@ -0,0 +1,124 @@
...
+
+    result = subprocess.run(
+        ['./forge', 'custom-certs', '--hostname', HOSTNAME],
+        capture_output=True, text=True,
+    )
+    assert result.returncode == 0, f'forge custom-certs failed: {result.stdout}\n{result.stderr}'
+    yield
```
evgeni:
```suggestion
```

same here, nothing to yield, IMHO

development/playbooks/custom-certs/custom-certs.yaml
```diff
@@ -4,10 +4,23 @@
...
   vars:
-    certificates_ca_directory: /root/custom-certificates
+    _custom_certs_root: /root/custom-certificates
     certificates_ca_password: "CUSTOMCA"
     certificates_ca_subject: 'Custom Test CA'
-    certificates_hostnames:
-      - "{{ ansible_facts['fqdn'] }}"
-  roles:
-    - role: certificates
+  pre_tasks:
```
evgeni: this can be `tasks` now?

evgeni:
How do we want to handle compatibility with N-1 installs where we might need to deploy new foreman-installer based proxies connected to a foremanctl based Foreman?

One of the previous iterations of this PR had re-used the old ssl-build layout for the bundle, making it "just work" with older code. The latest does not anymore, and while I think the new layout is more streamlined, I think we should have a compatibility layer somewhere?

src/playbooks/certificate-bundle/certificate-bundle.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - "../../vars/base.yaml"
+    - "../../vars/certificates.yml"
+  pre_tasks:
+    - name: Verify CA certificate exists
+      ansible.builtin.stat:
+        path: "{{ ca_certificate }}"
+      register: _ca_cert_check
+      failed_when: not _ca_cert_check.stat.exists
+      vars:
+        ansible_failed_msg: "CA certificate not found at {{ ca_certificate }}. Run 'foremanctl deploy' first to create the CA."
```
evgeni: What is this magic, and how should it work?

src/playbooks/certificate-bundle/certificate-bundle.yaml
```diff
@@ -0,0 +1,43 @@
...
+        certificate_bundle_ca_certificate: "{{ ca_certificate }}"
+        certificate_bundle_server_ca_certificate: >-
+          {{ certificates_custom_server_ca_certificate | default(ca_certificate) }}
+        certificate_bundle_server_certificate: >-
+          {{ certificates_custom_server_certificate | default(certificates_ca_directory ~ '/certs/' ~ hostname ~ '.crt') }}
+        certificate_bundle_server_key: >-
+          {{ certificates_custom_server_key | default(certificates_ca_directory ~ '/private/' ~ hostname ~ '.key') }}
+        certificate_bundle_client_certificate: "{{ certificates_ca_directory }}/certs/{{ hostname }}-client.crt"
+        certificate_bundle_client_key: "{{ certificates_ca_directory }}/private/{{ hostname }}-client.key"
+
```
evgeni:
```suggestion
```

src/roles/certificates/tasks/issue.yml
```diff
@@ -1,4 +1,14 @@
 ---
+- name: Ensure certificate directories exist
```
evgeni: isn't this already done in main.yml?

src/roles/certificates/tasks/issue.yml
```diff
@@ -29,8 +39,8 @@
         path: "{{ certificates_ca_directory_certs }}/{{ certificates_hostname }}.crt"
         csr_path: "{{ certificates_ca_directory_requests }}/{{ certificates_hostname }}.csr"
         provider: ownca
-        ownca_path: "{{ certificates_ca_directory_certs }}/ca.crt"
-        ownca_privatekey_path: "{{ certificates_ca_directory_keys }}/ca.key"
+        ownca_path: "{{ certificates_ca_directory }}/certs/ca.crt"
+        ownca_privatekey_path: "{{ certificates_ca_directory }}/private/ca.key"
```
evgeni: what's the difference here?

tests/certificate_bundle_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+def test_server_ca_and_default_ca_differ_for_custom_deployment(server, generate_bundle, custom_certificates):
+    """For custom server cert deployments, server-ca.crt (custom) differs from ca.crt (internal)."""
+    server_ca = server.run(f'tar xzf {TARBALL} -O certs/certs/server-ca.crt')
+    assert server_ca.succeeded
+    default_ca = server.run(f'tar xzf {TARBALL} -O certs/certs/ca.crt')
+    assert default_ca.succeeded
+    assert server_ca.stdout != default_ca.stdout
+
+
```
evgeni:
```suggestion
```

src/roles/certificates/tasks/issue.yml
ehelms: I did move things and forgot to clean this up.

src/roles/certificates/tasks/issue.yml
ehelms: Nothing, hold over from testing different ideas to get the scoping right.

ehelms:
> How do we want to handle compatibility with N-1 installs where we might need to deploy new foreman-installer based proxies connected to a foremanctl based Foreman?

Fair point. how about I tackle this as a follow up?

evgeni:
> Fair point. how about I tackle this as a follow up?

That sounds good.

evgeni (approved):

src/playbooks/certificate-bundle/metadata.obsah.yaml
```diff
@@ -0,0 +1,32 @@
+---
+help: |
+  Generate a certificate bundle
+
+variables:
+  hostname:
+    parameter: hostname
+    help: Hostname to generate a certificate bundle for that will be the common name.
+  certificates_custom_server_certificate:
```
evgeni, line 9: Should we run `certificate_checks` (or `foreman-certificate-check` in some other way) when the user passes custom stuff?

tests/certificate_bundle_test.py
```diff
@@ -0,0 +1,100 @@
...
+@pytest.mark.parametrize("expected_file", EXPECTED_CLIENT_FILES)
+def test_tarball_contains_client_certificate(tarball_members, expected_file):
+    assert expected_file in tarball_members
+
+
+def test_proxy_certs_stored_on_disk(server, generate_bundle):
+    """Verify proxy certificates are generated in the main certs directory."""
+    proxy_cert = server.file(f'/var/lib/foremanctl/certs/certs/{HOSTNAME}.crt')
```
evgeni: this path doesn't exist for custom certs. should it?

src/playbooks/certificate-bundle/metadata.obsah.yaml
```diff
@@ -0,0 +1,16 @@
...
+help: |
+  Generate a certificate bundle
+
+variables:
+  hostname:
+    parameter: hostname
+    help: Hostname to generate a certificate bundle for that will be the common name.
+  certificates_renew:
+    help: Regenerate certificates for this hostname.
+    parameter: --certificate-renew
```
amolpati30, line 21: If I'm not mistaken, `--certificate-renew` can renew certificates for both the `default` and `custom` certificate paths.

If I want to renew the certificate, which of the following commands is correct?

1.

```bash
foremanctl certificate-bundle \
  --certificate-server-ca-certificate /var/lib/foremanctl/certs/certs/ca.crt \
  --certificate-server-certificate /var/lib/foremanctl/certs/hosts/capsule.com/certs/capsule.com.crt \
  --certificate-server-key /var/lib/foremanctl/certs/hosts/capsule.com/private/capsule.com.key \
  capsule.com \
  --certificate-renew
```

or

2.

```bash
foremanctl certificate-bundle \
  --certificate-renew \
  capsule.com
```

I think it would only renew the Capsule server certificate. Is that correct? Just wanted to understand how certificate renewal will work for the Capsule.

ehelms: Both are correct, depending on if you are supplying custom or using the self-signed certificates.

ehelms:
Two additional elements that I updated this to enforce:

1. This now enforces that if the server is deployed with custom server certificates, the certificate bundle must also be supplied with custom certificates.
2. If custom certificates were used for the server, the certificate bundle includes the `server-ca.crt` in the bundle, thus the user only needs to input the cert and key for the bundle.

src/playbooks/certificate-bundle/certificate-bundle.yaml
```diff
@@ -0,0 +1,43 @@
...
+        certificate_checks_ca: "{{ server_ca_certificate }}"
+    - role: certificates
+      vars:
+        certificates_ca: false
+        certificates_output_directory: "{{ certificates_ca_directory }}/hosts/{{ hostname }}"
+        certificates_hostnames:
+          - "{{ hostname }}"
+  tasks:
+    - name: Generate certificate bundle
+      ansible.builtin.include_role:
```
evgeni: why using `roles` key above, but then the `include_role` module here? IMHO it makes the flow rather messy to read, why not using `include_role` always, then we can even fold pre_tasks/roles/tasks into one.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -57,11 +42,8 @@ variables:
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
 
-constraints:
-  required_together:
-    - [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]
-
 include:
+  - _certificate_custom
```
evgeni: Now that the bundle doesn't use the CA parameter, it seems pointless to move the custom certs params to an include? Unless you only move out cert/key and keep CA here?

src/roles/certificate_bundle/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
...
+    path: "{{ certificate_bundle_output_directory }}/{{ certificate_bundle_hostname }}.tar.gz"
+  register: _certificate_bundle_tarball
+
+- name: Build certificate bundle
+  when: not _certificate_bundle_tarball.stat.exists or (certificates_renew | default(false) | bool)
+  block:
+    - name: Create temporary directory
+      ansible.builtin.tempfile:
+        state: directory
+        suffix: certificate-build
```
evgeni:
```suggestion
        suffix: certificate-bundle-build
```
to match the rest of the naming here?

src/playbooks/certificate-bundle/certificate-bundle.yaml
```diff
@@ -0,0 +1,43 @@
+---
+- name: Generate a certificate bundle for a hostname
+  hosts:
+    - quadlet
+  become: true
+  vars_files:
+    - "../../vars/base.yaml"
+    - "../../vars/certificates.yml"
+  pre_tasks:
+    - name: Verify custom certificates are not provided with default certificate source
```
evgeni: Should that be a `forbidden_if` in the constraints of the command?

ehelms:
@evgeni Updated, and since I changed the design a bit I'll wait for your re-review.

evgeni (approved):

evgeni:
The tests in here were green, but the merge to master went red, and has been red since:
https://github.com/theforeman/foremanctl/actions/runs/27758710274/job/82127121859

WTH

evgeni:
AHAHAHA. The breakage is from https://github.com/theforeman/obsah/pull/118

amolpati30:
**QE Ack :** :green_apple:
Tested on build : foremanctl-3.0.0.develop-1.20260617132312908686.pr90.40.g0753dfe.el9.noarch
Result : Success

**Scenarios :**
1 Default -
**Generate a certificate bundle:** `foremanctl certificate-bundle proxy.example.com`

**Verify tarball at:**` /var/lib/foremanctl/certs/bundles/proxy.example.com.tar.gz` contains `certs/certs/ca.crt`, `certs/certs/server-ca.crt`, `certs/certs/proxy.example.com.crt`, `certs/certs/proxy.example.com-client.crt`, `certs/private/proxy.example.com.key`, `certs/private/proxy.example.com-client.key`

**Renew certificate:** `foremanctl certificate-bundle --certificate-renew proxy.example.com` or `foremanctl certificate-bundle --certificate-server-certificate certs/hosts/proxy.example.com/certs/proxy.example.com.crt --certificate-server-key certs/hosts/proxy.example.com/private/proxy.example.com.key --certificate-renew proxy.example.com`

2 Custom -
**Custom server set-up:** `foremanctl deploy --certificate-source=custom_server`

**Generate a certificate bundle:** `foremanctl certificate-bundle --certificate-server-certificate /path/to/proxy.crt --certificate-server-key /path/to/proxy.key proxy.example.com`

**Verify tarball at:**` /var/lib/foremanctl/certs/bundles/proxy.example.com.tar.gz` contains `certs/certs/ca.crt`, `certs/certs/server-ca.crt`, `certs/certs/proxy.example.com.crt`, `certs/certs/proxy.example.com-client.crt`, `certs/private/proxy.example.com.key`, `certs/private/proxy.example.com-client.key`

**Renew certificate:** `foremanctl certificate-bundle --certificate-renew proxy.example.com` or `foremanctl certificate-bundle --certificate-server-certificate certs/hosts/proxy.example.com/certs/proxy.example.com.crt --certificate-server-key 'certs/hosts/proxy.example.com/private/proxy.example.com.key' --certificate-renew proxy.example.com`

**Observation :**
1 Default -
foremanctl certificate-bundle successfully created the `tar.gz` file for the proxy. 
The `.tar.gz` file contains a bundle of certificates.
I checked the validity of the Foreman and proxy certificates, which are set to the default value of `7300 days`.
Then I updated the certificate validity for Foreman to `10 days` for both `-client.crt and .crt`. After that, I ran `foremanctl certificate-bundle --certificate-renew proxy.example.com` to update the proxy certificates as well. The -`client.crt and .crt` for the proxy were also updated with a validity of `10 days`.
Checked the fingerprints to track certificate identity across renewal, and noticed that they are changing after renewal for Foreman and the proxy.

2 Custom -
Create a custom certificate on Foreman.
Set the certificate source to `custom_server`.
foremanctl certificate-bundle successfully created the `tar.gz` file for the proxy. 
The `.tar.gz` file contains a bundle of certificates.
I checked the validity of the Foreman and proxy certificates, which are set to the default value of `7300 days`.
Then I updated the certificate validity for Foreman to `10 days` for both `-client.crt and .crt`. After that, I ran `foremanctl certificate-bundle --certificate-renew proxy.example.com` to update the proxy certificates as well. The -`client.crt and .crt` for the proxy were also updated with a validity of `10 days`.
Checked the fingerprints to track certificate identity across renewal, and noticed that they are changing after renewal for Foreman and the proxy.

PR #91, opened 2025-02-26
protect the CA using a password

evgeni:


roles/certificates/tasks/issue.yml
```diff
@@ -17,6 +17,7 @@
         -batch
         -policy signing_policy
         -extensions signing_req
+        -passin "pass:{{ certificates_ca_password }}"
```
ekohl, line 20: Hoe future proof do you want to make this? This will show in the process list, but that may be ok if this is a throw away poc

evgeni: it is ok, but it's also not expensive to fix it, thanks! done :)

roles/certificates/tasks/main.yml
```diff
@@ -47,15 +47,24 @@
...
+    owner: root
+    group: root
+    mode: '0600'
+  no_log: true
+
 - name: 'Creating CA certificate and key'
   ansible.builtin.command: >
     openssl req -new
       -x509
-      -nodes
```
evgeni, line 54: @ehelms why was this here?

ekohl (approved):
Looking forward: is there a good replacement for the "generate a random password and cache it" pattern we use with Puppet?

roles/certificates/tasks/main.yml
ehelms: I don't remember, this role is basically a lift of https://github.com/theforeman/forklift/blob/master/roles/custom_certificates/tasks/main.yml

ehelms:
> Looking forward: is there a good replacement for the "generate a random password and cache it" pattern we use with Puppet?

For the case of "if a password is not passed in, generate one as a default" ? I haven't seen that pattern. I think the Ansible approach is to always give it as input, "caching" it in a variables file or using vault.

ehelms (approved):

PR #95, opened 2025-03-03
enable oauth

evgeni:


ekohl (commented):
From a security point of view: would it make sense to only enable it if both are provided? Then not provide it by default?

Or is it better to open an issue for generating secure passwords?

evgeni:
You mean something like `oauth_active: {{ key is defined and secret is defined }}`? Yeah, probably a good idea.
And generating is another beast we'd have to cover at some point.

ekohl:
Indeed something like that

ekohl (approved):

PR #96, opened 2025-03-03
rename the pulp-only proxy

evgeni:


roles/foreman/tasks/main.yaml
```diff
@@ -143,7 +143,7 @@
 
 - name: Configure Foreman Proxy
   theforeman.foreman.smart_proxy:
-    name: "{{ ansible_fqdn }}"
+    name: "{{ ansible_fqdn }}-pulp"
     url: "https://{{ ansible_fqdn }}:9090"
```
ekohl, line 147: Pulp doesn't run on port 9090 so shouldn't you rename the other entry?

ekohl: Is there a point to this change before #81 then?

evgeni: mostly "splitting things up in smaller chunks", but I can add it to 81 if you prefer

PR #97, opened 2025-03-03
Build redis container image based on CentOS Stream 9

archanaserver:
Fixes https://github.com/theforeman/foreman-quadlet/issues/94

Uses [quay.io/akumari/redis-rpm:nightly](https://quay.io/repository/akumari/redis-rpm) as the image name and tag.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -30,6 +30,7 @@ jobs:
           - candlepin
           - foreman
           - foreman-proxy
+          - redis
     steps:
       - uses: actions/checkout@v4
       - name: Build ${{ matrix.container }} container
--- a/container-images/redis/Containerfile
+++ b/container-images/redis/Containerfile
@@ -1,0 +1,6 @@
+FROM quay.io/centos/centos:stream9
+RUN dnf -y update && \
+    dnf -y install redis && \
+    dnf clean all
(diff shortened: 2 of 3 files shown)


ekohl (commented):
When I look for CentOS Stream 9 container based images I find https://github.com/sclorg/redis-container so any reason we can't use https://quay.io/repository/sclorg/redis-6-c9s?

ehelms:
> When I look for CentOS Stream 9 container based images I find https://github.com/sclorg/redis-container so any reason we can't use https://quay.io/repository/sclorg/redis-6-c9s?

I didn't realize "sclorg" was still producing content -I guess they have evolved from the technology of `scls` to a more purposeful project. I think we can use their images as well.

archanaserver:
> > When I look for CentOS Stream 9 container based images I find https://github.com/sclorg/redis-container so any reason we can't use https://quay.io/repository/sclorg/redis-6-c9s?
> 
> I didn't realize "sclorg" was still producing content -I guess they have evolved from the technology of `scls` to a more purposeful project. I think we can use their images as well.

Yes updated https://github.com/theforeman/foreman-quadlet/pull/63 PR to use `quay.io/sclorg/redis-6-c9s:latest` instead of building our own Redis image and so closing this PR. Thanks!

PR #101, opened 2025-03-06
Add playbook to pre-pull images

ehelms:
Pre-pulling the images isn't necessary, but can help differentiate the runtime of the actual installer part vs. image pull.

### DIFF
--- a/playbooks/images.yaml
+++ b/playbooks/images.yaml
@@ -1,0 +1,12 @@
+---
+- name: Pull images
+  hosts:
+    - quadlet
+  vars_files:
+    - "../vars/images.yml"
+  become: true
+  tasks:
+    - name: Pull an image
+      containers.podman.podman_image:
+        name: "{{ item }}"
+      loop: "{{ images }}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,6 +56,9 @@ jobs:
       - name: Run setup
         run: |
           ansible-playbook playbooks/setup.yaml
+      - name: Run image pull
+        run: |
+          ansible-playbook playbooks/images.yaml
       - name: Run deployment
         run: |
           ansible-playbook playbooks/deploy.yaml
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -4,3 +4,4 @@ inventory = vagrant.py
 stdout_callback=debug
 stderr_callback=debug
 roles_path = ~/.ansible/roles:./roles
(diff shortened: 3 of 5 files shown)


playbooks/images.yaml
```diff
@@ -0,0 +1,17 @@
...
+    - quadlet
+  vars_files:
+    - "../vars/images.yml"
+  become: true
+  tasks:
+    - name: Pull an image
+      containers.podman.podman_image:
+        name: "{{ item }}"
+      loop:
+        - quay.io/ehelms/candlepin:4.4.14
```
ekohl, line 13: Can we reuse variables? I think we should make sure it doesn't live in 2 places or it will go out of sync

ekohl (changes requested):

playbooks/images.yaml
ehelms: :facepalm: Yea, that was the whole reason I created the vars file with images in it. Fixed.

ekohl (approved):
This makes me think: could we build on this to also build the container images locally? Out of scope for this PR, but it made me think.

ekohl:
Test failure looks unrelated.

ehelms:
> This makes me think: could we build on this to also build the container images locally? Out of scope for this PR, but it made me think.

I agree, I have some thoughts there and this could be a useful debugging option.

ehelms:
> Test failure looks unrelated.

Yea... this failure keeps popping up and is intermittent. I am going to see if I can figure out why.

ekohl:
> I agree, I have some thoughts there and this could be a useful debugging option.

There is `podman image scp` perhaps (at least in CI) we should build all containers locally, push them into the managed VM and then continue with the next steps.

ekohl (commented):
Thinking more: we have `podman_image` in various roles so you could clearly see when that fails, but systemd will pull on start up anyway and this separate step makes it redundant. Should we drop those instructions from the roles?

Should we also document this in `README` to recommend users to run it as well?

ehelms:
> Thinking more: we have `podman_image` in various roles so you could clearly see when that fails, but systemd will pull on start up anyway and this separate step makes it redundant. Should we drop those instructions from the roles?

I am not sure yet, and I think this should depend on our workflow strategy and update strategy. For example, do we want to treat pre-pulling as an optimization or a hard requirement? Do we want a fall back mechanism? How should we expect image updates to happen?

I would guess for updates, like we do today, we want to control and orchestrate that and not leave it up to the whims of a systemd service.

PR #102, opened 2025-03-06
Move config files into the container

ekohl:
This is an experiment to move configuration files into the container and only replace environment variables.

It creates a file to start the container where it creates all the needed files based on templates and environment variables. If the file already exists, it won't be overwritten. This allows bypassing the template.

This is a potential first step towards https://github.com/theforeman/foreman-quadlet/issues/68.


container-images/candlepin/Containerfile
```diff
@@ -2,8 +2,13 @@ FROM quay.io/centos/centos:stream9
 RUN dnf -y update && \
     dnf clean all
 RUN dnf -y --nodocs install \
+    gettext \
```
ekohl, line 5: This provides `envsubst`. On Fedora 41 it's split off into `gettext-envsubst` so I think it'll be the same on EL10.

container-images/candlepin/broker.xml.tpl
```diff
@@ -10,7 +10,7 @@
 
         <acceptors>
             <acceptor name="in-vm">vm://0</acceptor>
-            <acceptor name="stomp">tcp://localhost:61613?protocols=STOMP;useEpoll=false;sslEnabled=true;trustStorePath=/etc/candlepin/certs/truststore;trustStorePassword={{ candlepin_keystore_password }};keyStorePath=/etc/candlepin/certs/keystore;keyStorePassword={{ candlepin_keystore_password }};needClientAuth=true</acceptor>
+            <acceptor name="stomp">tcp://localhost:61613?protocols=STOMP;useEpoll=false;sslEnabled=true;trustStorePath=/etc/candlepin/certs/truststore;trustStorePassword=${CANDLEPIN_KEYSTORE_PASSWORD};keyStorePath=/etc/candlepin/certs/keystore;keyStorePassword=${CANDLEPIN_KEYSTORE_PASSWORD};needClientAuth=true</acceptor>
```
ekohl, line 13: I think https://activemq.apache.org/components/artemis/documentation/latest/configuring-transports.html#configuring-netty-ssl is the relevant documentation and I don't see a `keystorePasswordFile` option (or similar).

It does mention you can provide the `javax.net.ssl.keyStorePassword` system property as well. I wonder if Candlepin can reuse that same value. Perhaps we can end up with providing one big properties file with all secrets. Even if we don't directly expose that but still template it, it could be interesting to separate it out.

container-images/candlepin/README.md
ekohl, line 1: I'd like to expand this file to also document expected file paths or perhaps also introduce `CANDLEPIN_TRUSTSTORE_PATH` which can get a default value. Then it becomes introspectable.

container-images/candlepin/start
```diff
@@ -0,0 +1,17 @@
...
+set -e
+
+if [[ -z $CANDLEPIN_TRUSTSTORE_PASSWORD ]] ; then
+	CANDLEPIN_KEYSTORE_PASSWORD=$CANDLEPIN_TRUSTSTORE_PASSWORD
+	export CANDLEPIN_KEYSTORE_PASSWORD
+fi
+
+for template in /etc/candlepin/*.tpl ; do
+	destination=$(basename "$template" .tpl)
+	if [[ ! -f "$destination" ]] ; then
```
ekohl, line 12: This was intended to allow overrides but now I realize it may be created during container creation

ekohl:
This is failing because I'm making changes to the `Containerfile` and those aren't in the published images. Will continue to experiment locally.

container-images/candlepin/Containerfile
```diff
@@ -2,8 +2,19 @@ FROM quay.io/centos/centos:stream9
...
     https://yum.theforeman.org/candlepin/4.4/el9/x86_64/candlepin-4.4.14-1.el9.noarch.rpm \
     https://yum.theforeman.org/candlepin/4.4/el9/x86_64/candlepin-selinux-4.4.14-1.el9.noarch.rpm && \ 
     dnf clean all
 
-CMD ["/usr/libexec/tomcat/server", "start"]
+ENV CANDLEPIN_KEYSTORE_PATH=/etc/candlepin/certs/truststore
+ENV CANDLEPIN_KEYSTORE_PASSWORD=
+ENV CANDLEPIN_TRUSTSTORE_PATH=/etc/candlepin/certs/truststore
+# Optional - if unset, CANDLEPIN_KEYSTORE_PASSWORD is used
+ENV CANDLEPIN_TRUSTSTORE_PASSWORD=
```
ekohl, line 14: I noticed every line is another layer so perhaps it's better to use the multi-line mode of `ENV`.

container-images/candlepin/Containerfile
```diff
@@ -2,8 +2,19 @@ FROM quay.io/centos/centos:stream9
...
+
+ADD broker.xml.tpl /etc/candlepin/
```
ekohl, line 16: I thought about creating a `candlepin` directory locally and use `ADD candlepin/* /etc/candlepin/` so make it easier to scale but haven't experimented with that yet.

nikosmoum (approved):

ehelms:
Worth noting that when I played around with this, there was one scenario that wasn't working for me using Candlepin's built in environment variable evaluation (https://github.com/theforeman/foreman-quadlet/pull/71#issuecomment-2634875502). I think this method of substitution would work around that.

ehelms:
I think this idea would mean we cannot use the official Candlepin images directly. We could possibly use them as base containers, but we'd have to layer in our baseline configuration files on top of it.

ekohl:
> I think this idea would mean we cannot use the official Candlepin images directly. We could possibly use them as base containers, but we'd have to layer in our baseline configuration files on top of it.

That's correct, so we'd be on the hook to rebuild.

ehelms:
> Worth noting that when I played around with this, there was one scenario that wasn't working for me using Candlepin's built in environment variable evaluation ([#71 (comment)](https://github.com/theforeman/foreman-quadlet/pull/71#issuecomment-2634875502)). I think this method of substitution would work around that.

Could you next try this out with some of the values in `candlepin.conf`? Specifically:

```
org.quartz.dataSource.myDS.password={{ candlepin_db_password }}
```

ekohl:
Moved over to https://github.com/theforeman/candlepin-oci-images/pull/3.

PR #104, opened 2025-03-07
Use Candlepin project official image

ehelms:
Attempting to use the Candlepin project's official container, but running into errors I do not understand:

```
2025-03-07 02:22:21,154 [thread=Thread-0 (ActiveMQ-server-ActiveMQServerImpl::name=localhost)] [=, org=, csid=] WARN  org.apache.activemq.artemis.core.server - AMQ222216: Security problem while authenticating: AMQ229031: Unable to validate user from invm:0. Username: null; SSL certificate subject DN: unavailable
2025-03-07 02:22:21,156 [thread=main] [=, org=, csid=] WARN  org.candlepin.audit.ArtemisMessageSource - Unable to reconnect message listeners. Messages will not be received: event.default
org.apache.activemq.artemis.api.core.ActiveMQSecurityException: AMQ229031: Unable to validate user from invm:0. Username: null; SSL certificate subject DN: unavailable
```

### DIFF
--- a/roles/candlepin/defaults/main.yml
+++ b/roles/candlepin/defaults/main.yml
@@ -12,11 +12,13 @@ candlepin_ciphers:
   - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
   - TLS_ECDH_ECDSA_WITH_AES_128_GCM_SHA256
   - TLS_ECDH_RSA_WITH_AES_128_GCM_SHA256
-candlepin_container_image: quay.io/ehelms/candlepin
-candlepin_container_tag: 4.4.14
+candlepin_container_image: quay.io/candlepin/candlepin
+candlepin_container_tag: 4.4.20-1
 
 candlepin_db_host: localhost
 candlepin_db_port: 5432
 candlepin_db_ssl: false
 candlepin_db_ssl_ca: None
 candlepin_db_ssl_verify: true
+
+candlepin_tomcat_home: /opt/tomcat
--- a/roles/candlepin/tasks/artemis.yml
+++ b/roles/candlepin/tasks/artemis.yml
@@ -26,13 +26,12 @@
       filename: login.config
       app: artemis
 
-- name: Create Tomcat jaas.conf
+- name: Create the podman secret for the catalina_opts
   containers.podman.podman_secret:
     state: present
-    name: candlepin-artemis-jaas-conf
-    data: "{{ lookup('ansible.builtin.template', 'jaas.conf') }}"
+    name: candlepin-artemis-catalina-opts
+    data: "-Djava.security.auth.login.config=/opt/tomcat/conf/login.config"
     labels:
-      filename: jaas.conf
       app: artemis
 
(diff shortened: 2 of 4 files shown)


roles/candlepin/defaults/main.yml
```diff
@@ -12,11 +12,13 @@ candlepin_ciphers:
...
+candlepin_container_image: quay.io/candlepin/candlepin
+candlepin_container_tag: 4.4.20-1
 
 candlepin_db_host: localhost
 candlepin_db_port: 5432
 candlepin_db_ssl: false
 candlepin_db_ssl_ca: None
 candlepin_db_ssl_verify: true
+
+candlepin_tomcat_home: /opt/tomcat/conf
```
nikosmoum, line 24: what was the reason for this change? Once I undid this change and let the config files be under `/etc/tomcat` then it worked

ehelms: I think I ran into errors originally with this path. And then I assumed since tomcat is in /opt/tomcat and there is a `conf` directory in there that it should use that. Since it wouldn't want to clash with system tomcat from package.

What is the expectation for where configuration files should live with this image?

ehelms: I tested this locally and got a green result. Pushed update to this PR and we'll see what automation says.

ehelms:
@ekohl How do you see your config in containers work (https://github.com/theforeman/foreman-quadlet/pull/102) squaring with this kind of change?

roles/candlepin/defaults/main.yml
nikosmoum: I see in the candlepin image we build we use /opt/tomcat/conf, so I have no idea why that doesn't work. I'll try to tinker a bit more on monday to see why

ekohl:
> @ekohl How do you see your config in containers work (#102) squaring with this kind of change?

We can still create a container that is `FROM quay.io/candlepin/candlepin:4.4.20-1` and adds the config files & expose env vars. Maintenance wise we'll have to see what's best. I think quay can rebuild containers automatically if the base layer changes so that's something to investigate IMHO.

roles/candlepin/defaults/main.yml
nikosmoum: @ehelms so it looks like the reason it worked previously by using /etc/tomcat/conf was because the wrong container was still being used (the one with the rpms). I had to do a few changes to get it working. Should I push a commit to this PR?

ehelms:
Sure, or post a diff.

roles/candlepin/defaults/main.yml
nikosmoum: You can find the diff here: https://github.com/nikosmoum/foreman-quadlet/commit/94d127d73401d30aef6eb3d6c656c6682834ab71

ehelms:
@nikosmoum Updated based on your diff

ehelms:
Moved: https://github.com/theforeman/candlepin-oci-images/pull/2

PR #108, opened 2025-03-07
Use health_startup_cmd to determine if Candlepin is up

ekohl:
This moves the command into systemd/podman which means it can reliably report if the service is actually up.

### DIFF
--- a/roles/candlepin/tasks/main.yml
+++ b/roles/candlepin/tasks/main.yml
@@ -77,6 +77,8 @@
       - |
         [Install]
         WantedBy=default.target
+    healthcheck: curl --fail --insecure https://localhost:23443/candlepin/status
+    sdnotify: healthy
 
 - name: Run daemon reload to make Quadlet create the service files
   ansible.builtin.systemd:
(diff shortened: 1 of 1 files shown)


roles/candlepin/tasks/main.yml
```diff
@@ -77,6 +77,8 @@
       - |
         [Install]
         WantedBy=default.target
+    healthcheck: curl --fail --insecure https://localhost:23443/candlepin/status
+    sdnotify: healthy
```
ekohl, line 81: https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-sdnotify doesn't document this option, but `sdnotify` maps directly to `Notify=` and https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#notify-defaults-to-false does document that it uses the `HealthCmd` option (which is what the above `healthcheck` is for).

ekohl: https://github.com/containers/ansible-podman-collections/pull/911

ekohl:
This is an interesting error that I haven't seen before:
```
ERROR tests/client_test.py::test_foreman_content_view - apypie.foreman.ForemanApiException: Error while performing create on organizations: 500 Server Error: Internal Server Error for url: https://192.168.121.133/katello/api/organizations - {'message': 'resource have no errors'}
```

PR #114, opened 2025-03-07
Use sdnotify: true for services that support it

ekohl:
This offloads the readiness check to systemd instead of polling for it.


ekohl:
Related to https://github.com/theforeman/foreman-quadlet/pull/108.

roles/redis/tasks/main.yaml
```diff
@@ -17,6 +17,7 @@
     name: redis
     image: "{{ redis_image }}"
     state: quadlet
+    sdnotify: true
```
ekohl, line 20: Needs `--supervised systemd` added to the startup command.

ekohl:
```
django.contrib.auth.models.User.DoesNotExist: User matching query does not exist.
 psycopg.errors.NotNullViolation: null value in column "last_login" of relation "auth_user" violates not-null constraint
DETAIL:  Failing row contains (1, , null, t, admin, , , , t, t, 2025-03-07 17:20:02.468875+00).
The above exception was the direct cause of the following exception:
```

ehelms:
> django.contrib.auth.models.User.DoesNotExist: User matching query does not exist.
>  psycopg.errors.NotNullViolation: null value in column "last_login" of relation "auth_user" violates not-null constraint
> DETAIL:  Failing row contains (1, , null, t, admin, , , , t, t, 2025-03-07 17:20:02.468875+00).
> The above exception was the direct cause of the following exception:

Does this mean it's not waiting on the right condition to be true?

ekohl:
The Django model has a migration to make it allow `null` (from 2014, so certainly present) and the DB migrations ran. This isn't supposed to happen, but I don't have the logs to see what exactly happened.

ekohl:
It now fails with the `auth_user` table not existing so I suspect this is starting the services before migrations are complete. I expected the `Migrate the Pulp database` to complete before starting services. Perhaps something with `async` that I missed.

ekohl:
I've changed the implementation to be simpler: call `ansible.builtin.command` with `systemctl start` explicitly and then use `ansible.builtin.systemd` to still ensure each service is in a good state.

ansible-lint does not like this though.

ekohl:
New approach: pass `no_block: true` to each service so they can start in the background. Then call do it again in a loop to ensure they started.

ekohl:
For Foreman it appears to fail to start up. I think that's because it runs migrations within the container but that likely messes up when it happens in parallel. I think it would be better to have a migration service that you depend on.

ekohl:
When looking at the journal I could see Foreman and Pulp output so I certainly no longer trust Ansible to do anything async

ehelms:
> For Foreman it appears to fail to start up. I think that's because it runs migrations within the container but that likely messes up when it happens in parallel. I think it would be better to have a migration service that you depend on.

My experience has shown me there are three approaches to this:

### 1 - External orchestration

Run migrations as an explicit action that you manage with an orchestration tool (e.g. Ansible, systemd) and then serially execute the service management after the fact. In a Kubernetes world, this would be handled by logic in the operator. And thus would require writing the logic for different environments.

Example in pseudo-code:

 1. Ansible code to execute podman run container to run migration
 2. Start up systemd services
 3. Check that each service is healthy

### 2 - Independent containers

Perform migrations as part of container start-up, and include scripts as the entrypoint that check for the state of other services. This puts the logic into the containers and allows them to each be independently managed. In a Kubernetes world, this would work as is. 

In the Foreman case, the Foreman container would have the `wait_on_postgres` and then perform migrations, while the dynflow containers would have both `wait_on_postgres` and `wait_on_migrations`. They could then all be started or restarted together without additional orchestration.

Example entrypoint:

```
#!/bin/bash

set -x

/usr/bin/wait_on_postgres.py
/usr/bin/wait_on_migrations.py

exec "$@"
```

### 3 - Pods w/ Init Containers

When grouping containers into a pod and specifying initcontainer(s) then these will be run prior to the restart of the containers in the pod starting up. This exists in [Kubernetes](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) and has been implemented also in [podman via the kube](https://docs.podman.io/en/v5.0.3/markdown/podman-kube-play.1.html).

ekohl:
> 3 - Pods w/ Init Containers

I think something like this is the only feasible solution. Imagine someone pulls an updated image and that runs migrations. Suddenly the existing containers also need to be restarted because they will be incompatible with the newly migrated database. If you're not careful, they will restart concurrently and all trying to migrate.

My suggestion would be to create a oneshot service that runs the migrations. Then make sure the various services (foreman & various dynflow instances) depend on it.

Later on we can look at making sure they also restart.

As for the specific container: it may be reasonable to implement `pg_isready` on start up, but I wouldn't go too far in it since you can also let the container restart on failure.

ekohl:
I found the issue:
https://github.com/theforeman/foreman-quadlet/blob/88e7587ed285bafdf768b64a2d5a3344f47f0bd5/src/roles/pulp/tasks/main.yaml#L126-L135

This does nothing. It's actually the container itself that migrates on start up. If we start all services in parallel they will all migrate in parallel.

ehelms:
> I found the issue:
> 
> https://github.com/theforeman/foreman-quadlet/blob/88e7587ed285bafdf768b64a2d5a3344f47f0bd5/src/roles/pulp/tasks/main.yaml#L126-L135
> 
> This does nothing. It's actually the container itself that migrates on start up. If we start all services in parallel they will all migrate in parallel.

This does execute the migration based on the logs, but I do think `detach: false` is needed to ensure that Ansible runs this in the foreground and therefore waits. The output can be monitored by executing `./rop deploy -vv`. Then you will see for example:

```
Access policy for content/ansible/roles created.
Access policy for content/container/blobs created.
Access policy for distributions/container/container created.
Access policy for pulp_container/namespaces created.
Access policy for distributions/container/pull-through created.
Access policy for remotes/container/pull-through created.
Access policy for repositories/container/container-push/versions created.
Access policy for repositories/container/container/versions created.
Access policy for repositories/container/container-push created.
Access policy for remotes/container/container created.
Access policy for repositories/container/container created.
Access policy for content/container/signatures created.
Access policy for content/container/manifests created.
Access policy for content/container/tags created.
Access policy for content/rpm/distribution_trees created.
Access policy for content/rpm/modulemd_defaults created.
Access policy for content/rpm/modulemd_obsoletes created.
Access policy for content/rpm/modulemds created.
Access policy for content/rpm/packagecategories created.
Access policy for content/rpm/packageenvironments created.
Access policy for content/rpm/packagegroups created.
Access policy for content/rpm/packagelangpacks created.
Access policy for content/rpm/packages created.
Access policy for content/rpm/repo_metadata_files created.
Access policy for acs/rpm/rpm created.
Access policy for distributions/rpm/rpm created.
Access policy for publications/rpm/rpm created.
Access policy for remotes/rpm/rpm created.
Access policy for repositories/rpm/rpm/versions created.
Access policy for repositories/rpm/rpm created.
Access policy for remotes/rpm/uln created.
Access policy for content/rpm/advisories created.
Access policy for acs/file/file created.
Access policy for content/file/files created.
Access policy for distributions/file/file created.
Access policy for publications/file/file created.
Access policy for remotes/file/file created.
Access policy for repositories/file/file/versions created.
Access policy for repositories/file/file created.
Access policy for content/ostree/commits created.
Access policy for content/ostree/configs created.
Access policy for content/ostree/content created.
Access policy for distributions/ostree/ostree created.
Access policy for content/ostree/objects created.
Access policy for content/ostree/refs created.
Access policy for remotes/ostree/ostree created.
Access policy for repositories/ostree/ostree/versions created.
Access policy for repositories/ostree/ostree created.
Access policy for content/ostree/summaries created.



STDERR:

/usr/local/lib/python3.9/site-packages/pulp_container/app/migrations/0038_add_manifest_metadata_fields.py:8: UserWarning: Run 'pulpcore-manager container-handle-image-data' to initialize and expose metadata (i.e., annotations and labels) for all manifests.
  warnings.warn(
/usr/local/lib/python3.9/site-packages/pulp_container/app/migrations/0039_manifest_data.py:8: UserWarning: Run 'pulpcore-manager container-handle-image-data' to move the manifests' data from artifacts to the new 'data' database field.
  warnings.warn(
/usr/local/lib/python3.9/site-packages/pulp_container/app/migrations/0043_add_os_arch_image_size_manifest_fields.py:7: UserWarning: Run 'pulpcore-manager container-handle-image-data' to update the manifests' os, architecture, and compressed_image_size fields.
  warnings.warn(
pulp [None]: pulp_deb.app.migrations.0025_merge_colliding_structure_content:INFO: 2025-05-20 16:54:45.487306: Starting data migration!
pulp [None]: pulp_deb.app.migrations.0025_merge_colliding_structure_content:INFO: 2025-05-20 16:54:45.490536: Data migration completed!
```

ekohl:
> This does execute the migration based on the logs, but I do think `detach: false` is needed to ensure that Ansible runs this in the foreground and therefore waits.

Done. Also reverted back to using native Ansible parallelization for service management.

ekohl:
Foreman fails to be ready, probably because it's running migrations in parallel and that breaks things. I'm going to split this PR up into separate PRs.

ekohl:
Let's continue in https://github.com/theforeman/foreman-quadlet/pull/158, https://github.com/theforeman/foreman-quadlet/pull/159, https://github.com/theforeman/foreman-quadlet/pull/160 & https://github.com/theforeman/foreman-quadlet/pull/161.

PR #115, opened 2025-03-07
Add restart on failure to dynflow services

ehelms:
Mirrors configuration from RPM based deployment (https://github.com/theforeman/foreman/blob/develop/extras/systemd/dynflow-sidekiq%40.service#L25-L27) and maybe solves https://github.com/theforeman/foreman-quadlet/issues/112

### DIFF
--- a/roles/foreman/tasks/main.yaml
+++ b/roles/foreman/tasks/main.yaml
@@ -92,6 +92,10 @@
       - |
         [Install]
         WantedBy=default.target
+      - |
+        [Service]
+        RestartPolicy="on-failure"
+        Restart=1
 
 - name: Create Dynflow Container instances
   ansible.builtin.file:


roles/foreman/tasks/main.yaml
```diff
@@ -75,6 +75,8 @@
     state: quadlet
     network: host
     hostname: "{{ ansible_fqdn }}"
+    restart_policy: "on-failure"
+    restart_sec: 1
```
ekohl, line 79: If I read https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#ansible-collections-containers-podman-podman-container-module right then it needs to be under `generate_systemd`. I think that's what lint is complaining about.

ehelms: You are right.. my window was too tiny to see the options were nested.

ekohl (approved):

ehelms:
Welp, this did not solve the intermittent failure issue :/

roles/foreman/tasks/main.yaml
```diff
@@ -88,6 +88,9 @@
       DYNFLOW_REDIS_URL: "redis://localhost:6379/6"
       REDIS_PROVIDER: "DYNFLOW_REDIS_URL"
     command: "/usr/libexec/foreman/sidekiq-selinux -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/%i.yml"
+    generate_systemd:
```
evgeni, line 91: is that really in use when we do `state: quadlet`?

PR #117, opened 2025-03-11
Set postgres handler to started to prevent restarts

ehelms:
I noticed this output in the logs:

```
Mar 11 16:15:41 quadlet.example.com python3[54896]: ansible-ansible.legacy.systemd Invoked with name=postgresql state=restarted daemon_reload=False daemon_reexec=False scope=system no_block=False enabled=None force=None masked=None
Mar 11 16:15:41 quadlet.example.com systemd[1]: Stopping PostgreSQL database server...
Mar 11 16:15:41 quadlet.example.com foreman[51658]: FATAL:  terminating connection due to administrator command
Mar 11 16:15:41 quadlet.example.com foreman[51658]: #<Thread:0x00007f8f301e39a0 /usr/share/gems/gems/logging-2.4.0/lib/logging/diagnostic_context.rb:471 run> terminated with exception (report_on_exception is true):
Mar 11 16:15:41 quadlet.example.com foreman[51658]: /usr/share/gems/gems/sequel-5.89.0/lib/sequel/adapters/postgres.rb:171:in `exec': PG::ConnectionBad: PQsocket() can't get socket descriptor (Sequel::DatabaseDisconnectError)
```

I believe this is happening due to the defined [handler](https://github.com/geerlingguy/ansible-role-postgresql/blob/master/handlers/main.yml).

### DIFF
--- a/playbooks/deploy.yaml
+++ b/playbooks/deploy.yaml
@@ -39,6 +39,7 @@
     httpd_server_key: "{{ server_key }}"
     pulp_db_password: "CHANGEME"
     pulp_content_origin: "https://{{ ansible_fqdn }}"
+    postgresql_restarted_state: started
     postgresql_databases:
       - name: candlepin
         owner: candlepin
--- a/tests/foreman_api_test.py
+++ b/tests/foreman_api_test.py
@@ -1,6 +1,7 @@
 import urllib.parse
 
 import requests
+import time
 
 
 def _repo_url(repo, ssh_config):
@@ -38,6 +39,7 @@ def test_foreman_lifecycle_environment(lifecycle_environment):
 
 
 def test_foreman_content_view(content_view, yum_repository, foremanapi):
+    time.sleep(10)
     assert content_view
     foremanapi.update('content_views', {'id': content_view['id'], 'repository_ids': [yum_repository['id']]})
     foremanapi.resource_action('content_views', 'publish', {'id': content_view['id']})


ehelms:
The failure seen here now, which I have also seen sometimes, almost more seems like we are not properly waiting for tasks to finish within the tests:

```
FAILED tests/foreman_api_test.py::test_foreman_content_view - apypie.foreman.ForemanApiException: Error while performing publish on content_views: 500 Server Error: Internal Server Error for url: https://192.168.121.191/katello/api/content_views/11/publish - {'displayMessage': 'Pending tasks detected in repositories of this content view. Please wait for the tasks: - http://quadlet.example.com/foreman_tasks/tasks/6f77453f-42e1-45a3-80db-bb0a8fb81747 before publishing.', 'errors': ['Pending tasks detected in repositories of this content view. Please wait for the tasks: - http://quadlet.example.com/foreman_tasks/tasks/6f77453f-42e1-45a3-80db-bb0a8fb81747 before publishing.']}
```

This implies, and from my investigations, indicates the repositories are not done actually syncing when the content view is attempted to be published. @evgeni is there some tuning on if it waits for a success? or only tries so many times?

ehelms:
It appears that the repository sync finishes, but it queues up a "Metadata generate repository" task that then clashes with the content view publish. I am wondering if this is a race condition and one that we always win with bats tests by having enough other actions in between the sync and publish.

tests/foreman_api_test.py
```diff
@@ -38,6 +38,7 @@ def test_foreman_lifecycle_environment(lifecycle_environment):
 
 
 def test_foreman_content_view(content_view, yum_repository, foremanapi):
+    sleep(10)
```
ehelms, line 41: Just trying this out.

playbooks/deploy.yaml
```diff
@@ -39,6 +39,7 @@
     httpd_server_key: "{{ server_key }}"
     pulp_db_password: "CHANGEME"
     pulp_content_origin: "https://{{ ansible_fqdn }}"
+    postgresql_restarted_state: started
```
ehelms, line 42: The PostgreSQL / dynflow errors appear to be gone.

tests/foreman_api_test.py
```diff
@@ -38,6 +39,7 @@ def test_foreman_lifecycle_environment(lifecycle_environment):
 
 
 def test_foreman_content_view(content_view, yum_repository, foremanapi):
+    time.sleep(10)
```
ehelms, line 42: This sleep helped. This is not a deterministic solution, and I am investigating a better way to properly find and wait on the right set of tasks. In the meantime I think this is an acceptable solution to avoid constant CI re-runs.

evgeni (approved):

PR #118, opened 2025-03-11
Use systemd socket activation for Foreman

ekohl:
@ehelms as you requested, an example of using systemd socket activation an unix sockets.

It changes the connections to go through Apache, which may or may not be right.

### DIFF
--- a/src/roles/foreman/defaults/main.yaml
+++ b/src/roles/foreman/defaults/main.yaml
@@ -12,7 +12,8 @@ foreman_database_ssl_mode: disable
 foreman_database_ssl_ca:
 foreman_database_ssl_ca_path: /etc/foreman/db-ca.crt
 
-foreman_url: "http://{{ ansible_facts['fqdn'] }}:3000"
+foreman_listen_stream: localhost:3000
+foreman_url: "http://{{ foreman_listen_stream }}"
 
 # Puma threads calculation:
 # Max calculation is based on the default for MRI https://puma.io/puma6/#thread-pool
--- a/src/roles/foreman/defaults/main.yaml
+++ b/src/roles/foreman/defaults/main.yaml
@@ -12,6 +12,7 @@ foreman_database_ssl_ca: # noqa: no-empty-defaults
 foreman_database_ssl_ca_path: /etc/foreman/db-ca.crt
 
 foreman_name: "{{ ansible_facts['fqdn'] }}"
+foreman_listen_stream: 127.0.0.1:3000
 foreman_url: "http://{{ ansible_facts['fqdn'] }}:3000"
 
 # Puma threads calculation:
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -95,6 +95,12 @@
     - Restart foreman
     - Restart dynflow-sidekiq@
 
+- name: Deploy Foreman socket
+  ansible.builtin.template:
+    src: foreman.socket.j2
+    dest: /etc/systemd/system/foreman.socket
+    mode: '0644'
+
 - name: Deploy Foreman Container
   containers.podman.podman_container:
     name: "foreman"
@@ -122,6 +128,8 @@
       FOREMAN_ENABLED_PLUGINS: "{{ foreman_plugins | join(' ') }}"
     quadlet_options:
       - |
+        [Unit]
+        Requires=foreman.socket
         [Install]
         WantedBy=default.target foreman.target
         [Unit]
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -268,7 +268,7 @@
 - name: Wait for Foreman service to be accessible
   ansible.builtin.uri:
     url: '{{ foreman_url }}/api/v2/ping'
-    validate_certs: false
+    ca_path: '{{ ca_certificate }}'
   until: foreman_status.status == 200
   retries: 60
   delay: 5
@@ -287,7 +287,7 @@
 - name: Wait for Foreman tasks to be ready
   ansible.builtin.uri:
     url: '{{ foreman_url }}/api/v2/ping'
-    validate_certs: false
+    ca_path: '{{ ca_certificate }}'
   until:
     - foreman_tasks_status.status == 200
     - foreman_tasks_status.json['results']['katello']['services']['foreman_tasks']['status'] == 'ok'
@@ -304,4 +304,4 @@
     server_url: "{{ foreman_url }}"
     username: "{{ foreman_initial_admin_username }}"
     password: "{{ foreman_initial_admin_password }}"
-    validate_certs: false
+    ca_path: '{{ ca_certificate }}'
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -100,6 +100,12 @@
     - Restart foreman
     - Restart dynflow-sidekiq@
 
+- name: Deploy Foreman socket
+  ansible.builtin.template:
+    src: foreman.socket.j2
+    dest: /etc/systemd/system/foreman.socket
+    mode: '0644'
+
 - name: Deploy Foreman Container
   containers.podman.podman_container:
     name: "foreman"
@@ -130,6 +136,8 @@
       FOREMAN_ENABLED_PLUGINS: "{{ foreman_plugins | join(' ') }}"
     quadlet_options:
       - |
+        [Unit]
+        Requires=foreman.socket
         [Install]
         WantedBy=default.target foreman.target
         [Unit]
--- a/src/roles/foreman/templates/foreman.socket.j2
+++ b/src/roles/foreman/templates/foreman.socket.j2
@@ -1,0 +1,8 @@
+[Unit]
+Description=Foreman socket
+
+[Socket]
+ListenStream={{ foreman_listen_stream }}
+
+[Install]
+WantedBy=sockets.target
--- a/src/roles/foreman/templates/foreman.socket.j2
+++ b/src/roles/foreman/templates/foreman.socket.j2
@@ -3,6 +3,12 @@ Description=Foreman socket
 
 [Socket]
 ListenStream={{ foreman_listen_stream }}
+SocketUser=apache
+SocketMode=0600
+
+NoDelay=false
+ReusePort=true
+Backlog=1024
 
 [Install]
 WantedBy=sockets.target
--- a/src/roles/httpd/defaults/main.yml
+++ b/src/roles/httpd/defaults/main.yml
@@ -1,7 +1,7 @@
 httpd_ssl_dir: /etc/pki/httpd
 httpd_pulp_api_backend: http://localhost:24817
 httpd_pulp_content_backend: http://localhost:24816
-httpd_foreman_backend: http://localhost:3000
+httpd_foreman_backend: http://localhost:3000/
 httpd_pub_dir: /var/www/html/pub
 
 # External authentication configuration
--- a/src/roles/httpd/tasks/main.yml
+++ b/src/roles/httpd/tasks/main.yml
@@ -13,6 +13,13 @@
     persistent: true
   when: ansible_facts['selinux']['status'] == "enabled"
 
+# TODO: probably not the right boolean
+- name: Set daemons_enable_cluster_mode so Apache can connect to unix sockets
+  ansible.posix.seboolean:
+    name: daemons_enable_cluster_mode
+    state: true
+    persistent: true
+
 - name: Disable welcome page
   ansible.builtin.file:
     path: /etc/httpd/conf.d/welcome.conf
--- a/src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
+++ b/src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
@@ -95,8 +95,8 @@
   ProxyPass /icons !
   ProxyPass /images !
   ProxyPass /server-status !
-  ProxyPass / {{ httpd_foreman_backend }}/ retry=0 timeout=900 upgrade=websocket
-  ProxyPassReverse / {{ httpd_foreman_backend }}/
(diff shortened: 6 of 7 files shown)


ehelms:
Instead of enforcing Apache to be present and an ordering, could you use the [`unix_socket` parameter](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/uri_module.html#parameter-unix_socket) of the `uri` module to do the service up test?

ekohl:
> Instead of enforcing Apache to be present and an ordering, could you use the [`unix_socket` parameter](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/uri_module.html#parameter-unix_socket) of the `uri` module to do the service up test?

I had that for a moment, but the type of the result changes to a `dict` and didn't want to deal with it. Also didn't know if the `smart_proxy` module from FAM supports unix sockets. Then I took the easy route and relied on Apache though I kind of agree with you that it'd be nicer if it was more standalone.

On the other hand, perhaps we should rely more on health checks for ready detection. Then the Smart Proxy registration also feels like it doesn't belong in the Foreman role. Perhaps we need to swap the order of Foreman and Foreman Proxy. That stuff is easier in Puppet IMHO.

ehelms:
> Then the Smart Proxy registration also feels like it doesn't belong in the Foreman role.

Not that you say that, I am surprised it's in that role as well due to what I would have thought are order of operations. I suppose it's there as the last action with the way the playbook is organized. I think in past iterations where we played with Ansible we made registration it's own role to control it better. Ahh yes, [back in the day, we had a spark of an idea](https://github.com/ares/spark/tree/master/roles/foreman_proxy_register).

Then you have https://github.com/theforeman/foreman-quadlet/pull/81 which changes things and negates this being here all together I think.

roles/foreman/tasks/main.yaml
```diff
@@ -133,7 +141,8 @@
 
 - name: Wait for Foreman tasks to be ready
   ansible.builtin.uri:
-    url: 'http://{{ ansible_hostname }}:3000/api/v2/ping'
+    url: '{{ foreman_url }}/api/v2/ping'
+    validate_certs: false # TODO
```
evgeni, line 145:
```suggestion
    ca_path: '{{ ca_certificate }}'
```

roles/foreman/tasks/main.yaml
```diff
@@ -145,6 +154,7 @@
   theforeman.foreman.smart_proxy:
     name: "{{ ansible_fqdn }}"
     url: "https://{{ ansible_fqdn }}:9090"
-    server_url: "http://{{ ansible_fqdn }}:3000"
+    server_url: "{{ foreman_url }}"
+    validate_certs: false # TODO
```
evgeni, line 158: We don't support `ca_path` today in FAM. Should we?

roles/foreman/tasks/main.yaml
```diff
@@ -46,6 +46,11 @@
     name: foreman-client-key
     path: "{{ foreman_client_key }}"
 
+- name: Deploy Foreman socket
+  ansible.builtin.template:
+    src: foreman.socket.j2
+    dest: /etc/systemd/system/foreman.socket
```
evgeni, line 52:
```suggestion
    dest: /etc/systemd/system/foreman.socket
    mode: '0644'
```
to make Ansible happy

roles/foreman/tasks/main.yaml
```diff
@@ -64,6 +69,8 @@
       SEED_ADMIN_PASSWORD: changeme
     quadlet_options:
       - |
+        [Unit]
+        Requires=foreman.socket
```
evgeni, line 73: This is the only thing needed for it to do socket activation? :exploding_head:

evgeni:
> Also didn't know if the `smart_proxy` module from FAM supports unix sockets.

It does not, because `requests` can't. There is https://github.com/msabramo/requests-unixsocket but I never tried it.

roles/foreman/tasks/main.yaml
evgeni: Does the socket "just" need to be started too?

ekohl: > This is the only thing needed for it to do socket activation? 🤯

Pretty much yes. It ends up setting the right env vars and Puma picks that up because we tell it to (https://github.com/theforeman/foreman/blob/75dc40095546966477a927911e715263c42d1bea/config/puma/production.rb#L35-L38).

> Mhh, but the tests are failing with error 500, so maybe it isn't?

I think that's because of `async`. Looking at the Pulp failures I get the impression the service is started before migrations ran. That's why I'll split this off into 2 PRs.

> Does the socket "just" need to be started too?

Starting `foreman.service` (as we do) will activate the socket so it shouldn't be needed to enable it, but we could. In theory we could only start the socket and let `foreman.service` start on demand but I'm not sure we want to go that route.

roles/foreman/tasks/main.yaml
ekohl: I was looking for it and now I'm wondering if `validate_certs: '{{ ca_certificate }}'` works. https://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification says `verify='/path/to/certfile'` is supported and looking at https://github.com/theforeman/foreman-ansible-modules/blob/2343e66adf65aac42f2d26a076d555102320bcc6/plugins/inventory/foreman.py#L266 I'd think that works. Not sure about https://github.com/theforeman/foreman-ansible-modules/blob/2343e66adf65aac42f2d26a076d555102320bcc6/plugins/module_utils/foreman_helper.py#L617 but if we go that route, it'd make sense to me to copy requests' API because it saves a parameter.

evgeni: It saves one, yes, but diverges from `ansible.builtin.uri` API, which is the one Ansible users are more familiar with.

The problem (today) is https://github.com/theforeman/foreman-ansible-modules/blob/2343e66adf65aac42f2d26a076d555102320bcc6/plugins/module_utils/foreman_helper.py#L373 which makes it a boolean, not a string.

roles/foreman/tasks/main.yaml
ekohl: When I try this I get:
```
The conditional check 'foreman_status.status == 200' failed. The error was: error while evaluating conditional (foreman_status.status == 200): 'dict object' has no attribute 'status'. 'dict object' has no attribute 'status'
```

roles/foreman/tasks/main.yaml
ekohl: Sorry, async was a different patch. Apache fails with:

```
[Thu Mar 13 18:02:39.887738 2025] [proxy:warn] [pid 707:tid 891] [client 192.168.122.14:52218] AH01144: No protocol handler was valid for the URL / (scheme 'unix'). If you are using a DSO version of mod_proxy, make sure the proxy submodules are included in the configuration using LoadModule.
```

Turns out the template had a trailing slash and then it couldn't find the right socket anymore.

roles/foreman/tasks/main.yaml
evgeni: Weird. In Line 126 you mean?

roles/foreman/tasks/main.yaml
ekohl: > Starting `foreman.service` (as we do) will activate the socket so it shouldn't be needed to enable it, but we could. In theory we could only start the socket and let `foreman.service` start on demand but I'm not sure we want to go that route.

There is a good reason to start from Ansible: we should restart it if a change is made.

ekohl:
Next issue:
```
type=AVC msg=audit(1741889614.482:20207): avc:  denied  { write } for  pid=58362 comm="httpd" name="foreman.sock" dev="tmpfs" ino=2284 scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:var_run_t:s0 tclass=sock_file permissive=0
```

We handled this in our foreman-selinux package but that's no longer installed on the host.

playbooks/deploy.yaml
```diff
@@ -33,6 +33,9 @@
     foreman_db_password: "CHANGEME"
     foreman_oauth_consumer_key: abcdefghijklmnopqrstuvwxyz123456
     foreman_oauth_consumer_secret: abcdefghijklmnopqrstuvwxyz123456
+    foreman_listen_stream: /run/foreman.sock
+    foreman_url: "https://{{ ansible_fqdn }}"
+    httpd_foreman_backend: "unix://{{ foreman_listen_stream }}|http://%{HTTP_HOST/"
```
evgeni, line 38:
```suggestion
    httpd_foreman_backend: "unix://{{ foreman_listen_stream }}|http://%{HTTP_HOST}/"
```

roles/foreman/tasks/main.yaml
```diff
@@ -115,7 +123,8 @@
 
 - name: Wait for Foreman service to be accessible
   ansible.builtin.uri:
-    url: 'http://{{ ansible_hostname }}:3000/api/v2/ping'
+    url: '{{ foreman_url }}/api/v2/ping'
+    validate_certs: '{{ ca_certificate }}'
```
evgeni, line 127: With that, I too get "The conditional check 'foreman_status.status == 200' failed. The error was: error while evaluating conditional (foreman_status.status == 200): 'dict object' has no attribute 'status'. 'dict object' has no attribute 'status'"
But not with:
```suggestion
    ca_path: '{{ ca_certificate }}'
```

roles/foreman/tasks/main.yaml
```diff
@@ -133,7 +142,8 @@
 
 - name: Wait for Foreman tasks to be ready
   ansible.builtin.uri:
-    url: 'http://{{ ansible_hostname }}:3000/api/v2/ping'
+    url: '{{ foreman_url }}/api/v2/ping'
+    validate_certs: '{{ ca_certificate }}'
```
evgeni, line 146:
```suggestion
    ca_path: '{{ ca_certificate }}'
```

roles/foreman/tasks/main.yaml
```diff
@@ -145,6 +155,7 @@
   theforeman.foreman.smart_proxy:
     name: "{{ ansible_fqdn }}"
     url: "https://{{ ansible_fqdn }}:9090"
-    server_url: "http://{{ ansible_fqdn }}:3000"
+    server_url: "{{ foreman_url }}"
+    validate_certs: '{{ ca_certificate }}'
```
evgeni, line 159: As mentioned, you can't do that right now :)

evgeni: https://github.com/theforeman/foreman-ansible-modules/pull/1831

evgeni:
> Next issue:
> 
> ```
> type=AVC msg=audit(1741889614.482:20207): avc:  denied  { write } for  pid=58362 comm="httpd" name="foreman.sock" dev="tmpfs" ino=2284 scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:var_run_t:s0 tclass=sock_file permissive=0
> ```
> 
> We handled this in our foreman-selinux package but that's no longer installed on the host.

Nothing prevents us from installing it again.

Or we extend/adjust
https://github.com/theforeman/foreman-quadlet/blob/3c0ec9ca701dc196c693f812a265f558c1e3d511/roles/httpd/tasks/main.yml#L9-L13

ekohl:
> audit2allow says `daemons_enable_cluster_mode` could help? -- Edit: tested. nope it does not

If you use `/run/httpd.*` it gets the `httpd_var_run_t` type and then the boolean does work. Still doesn't feel right, but if it helps move this forward it can be a decent hack.

ehelms:
@ekohl Would you mind rebasing when you get a moment?

src/playbooks/deploy/deploy.yaml
```diff
@@ -32,6 +32,9 @@
     foreman_client_certificate: "{{ client_certificate }}"
     foreman_oauth_consumer_key: abcdefghijklmnopqrstuvwxyz123456
     foreman_oauth_consumer_secret: abcdefghijklmnopqrstuvwxyz123456
+    foreman_listen_stream: /run/httpd.foreman.sock
+    foreman_url: "https://{{ ansible_fqdn }}"
+    httpd_foreman_backend: "unix://{{ foreman_listen_stream }}|http://%{HTTP_HOST}/"
```
ehelms, line 37: I'm surprised all three of these are not the role defaults.

ekohl: Because they're connected: my thought was that the roles should be dependent and only on the playbook level you can relate them.

src/roles/foreman/defaults/main.yaml
```diff
@@ -1,6 +1,8 @@
 ---
 foreman_container_image: "quay.io/evgeni/foreman-rpm"
 foreman_container_tag: "nightly"
+foreman_listen_stream: localhost:3000
```
ehelms, line 4: I think what I don't know enough about is why we wouldn't default this to ` /run/httpd.foreman.sock` if our intent is to use systemd sock activation.

ekohl:
Will rebase once https://github.com/theforeman/foremanctl/pull/463 is merged.

pablomh:
Hi @ekohl. Will this be used for inter-container communication or only for activation?

ekohl:
Just for Apache -> Foreman. Technically that's crossing the host -> container boundary. Other services talk to Foreman via Apache. No services other than Apache will talk to the unix socket.

ekohl:
Rebased to resolve conflicts now that #463 is merged.

ehelms:
I think this needs some test fixes, the tests at `tests/foreman_test.py:5-6` hardcode:

FOREMAN_HOST = 'localhost'
FOREMAN_PORT = 3000

ekohl:
I've reduced the scope to just socket activation now, still using the TCP socket. This may already solve the issue IPv6-only has. Then in https://github.com/theforeman/foremanctl/pull/593 we can deal with the unix socket.

ekohl:
Closing in favor of https://github.com/theforeman/foremanctl/pull/608

PR #119, opened 2025-03-13
Split installation into three distinct phases

ehelms:
Another look at splitting the installation process into three distinct phases. This does use Ansible, but that's just to help orchestrate and automate for testing purposes. The playbook is split into three sections:

1. Generate default certificates
2. Install the server w/ quadlets using foreman-quadlets
3. Install foreman-proxy using the foreman-installer based on https://github.com/theforeman/foreman-installer/pull/1016

~~The foreman-proxy is not optional at this point. That would require https://github.com/theforeman/foreman-quadlet/pull/81~~

I rebased this on https://github.com/theforeman/foreman-quadlet/pull/105 to have the Pulp smart-proxy and thus make the stand-alone smart-proxy optional.

### DIFF
--- a/playbooks/3-stage-deploy.yaml
+++ b/playbooks/3-stage-deploy.yaml
@@ -1,0 +1,143 @@
+- name: Generate certificates
+  become: true
+  hosts:
+    - quadlet
+  vars_files:
+    - "../vars/{{ certificate_source }}_certificates.yml"
+  vars:
+    certificate_source: default
+    certificates_hostnames:
+      - "{{ ansible_fqdn }}"
+      - localhost
+    certificates_ca_password: "CHANGEME"
+  roles:
+    - role: certificates
+      when: "certificate_source == 'default'"
+    - role: certificate_checks
+      vars:
+        certificate_checks_certificate: "{{ server_certificate }}"
+        certificate_checks_key: "{{ server_key }}"
+        certificate_checks_ca: "{{ ca_certificate }}"
+
+- name: Install server
+  become: true
+  hosts:
+    - quadlet
+  vars_files:
+    - "../vars/{{ certificate_source }}_certificates.yml"
+    - "../vars/images.yml"
+  vars:
+    certificate_source: default
+    candlepin_db_password: "CHANGEME"
+    candlepin_keystore_password: "CHANGEME"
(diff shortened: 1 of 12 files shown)


playbooks/3-stage-deploy.yaml
```diff
@@ -0,0 +1,143 @@
...
+        name: theforeman.operations.foreman_repositories
+      vars:
+        foreman_repositories_version: nightly
+        foreman_repositories_katello_version: nightly
+
+    - name: Install foreman-proxy
+      package:
+        name: foreman-proxy
+
+    - name: Deploy default certificates
```
evgeni, line 117: Assuming this is temporarily, given the current installer can't deal well with "proxy only certs for the internal proxy" scenario, what would be the most elegant way for us to pass those certs on?
Part of me wants to say "the tarball we have for external proxies is nice", but given that we don't need to transfer it across systems, it's not strictly required to be a single file (which adds complexity).
We could set the API to be "a folder with a specified structure" (think like the one in ssl-build), but put it soemwhere like `/var/lib/foreman-certs` and it either gets fed by `foreman-certs` cli or the `certificates` role, and then everyone can just pick up things from there.
For the external proxy we'd then tar things up that resemble the same structure and off we go.

ehelms: This is tricky because there is this unfun orchestration that has to happen as shown here. The certificates need to be deployed to a location with the correct ownership (user/group/mode) access by foreman-proxy. Which means the foreman-proxy RPM needs to be installed or something needs to ensure creation of them.

For foreman-installer based certificates, I proposed fixing that with this class https://github.com/theforeman/puppet-certs/pull/449/files#diff-e6327449f108163297d1ce69ca08f3a4d8013b01d772a4be3aba628b223bd8c5 and adding it to the answers file. This wouldn't solve it for the new certificate generation though.

I was also thinking of adding certificate "management" to puppet-foreman_proxy based on a parameter so that the module could ensure the right user/group and permissions are ensured on the certificates. I tried this idea out like 4 years ago -- https://github.com/theforeman/puppet-foreman_proxy/pull/599

ehelms: I updated https://github.com/theforeman/puppet-foreman_proxy/pull/599 idea to our current certificates approach.

PR #120, opened 2025-03-14
Ability to use foreman-installer certificates

ehelms:
@evgeni How would we test different installation procedures such as each certificate scenario?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,12 @@ jobs:
           requirements_file: requirements.yml
 
   tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        certificate_source:
+          - default
+          - installer
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
@@ -61,6 +67,6 @@ jobs:
           ansible-playbook playbooks/images.yaml
       - name: Run deployment
         run: |
-          ansible-playbook playbooks/deploy.yaml
+          ansible-playbook playbooks/deploy.yaml -e certificate_source=${{ matrix.certificate_source }}
       - name: Run tests
-        run: ./run_tests
+        run: ./run_tests --certificate-source=${{ matrix.certificate_source }}
--- a/playbooks/deploy.yaml
+++ b/playbooks/deploy.yaml
@@ -64,6 +64,8 @@
   roles:
     - role: certificates
       when: "certificate_source == 'default'"
+    - role: foreman_installer_certs
+      when: "certificate_source == 'installer'"
     - role: certificate_checks
       vars:
         certificate_checks_certificate: "{{ server_certificate }}"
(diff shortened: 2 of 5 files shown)


evgeni:
Well, we could add a matrix to https://github.com/theforeman/foreman-quadlet/blob/3c0ec9ca701dc196c693f812a265f558c1e3d511/.github/workflows/test.yml#L32
and then do something like:
```
matrix:
  certificate_source:
    - default
    - installer
…
ansible-playbook … -e certificate_source=${{ matrix.certificate_source }}
```

not beautiful, but would work?

.github/workflows/test.yml
```diff
@@ -30,6 +30,10 @@ jobs:
           requirements_file: requirements.yml
 
   tests:
+    matrix:
+      certificate_source:
+        - default
+        - installer
```
evgeni, line 36:
```suggestion
    strategy:
      matrix:
        certificate_source:
          - default
          - installer
```

.github/workflows/test.yml
```diff
@@ -30,6 +30,11 @@ jobs:
           requirements_file: requirements.yml
 
   tests:
+    strategy:
```
evgeni, line 33: you probably also want:
```suggestion
    strategy:
      fail-fast: false
```
otherwise it will cancel "the other" branch when one fails, which in most case is not desired?

evgeni:
You also need to pass `--certificate-source={{ matrix.certificate_source }}` to pytest.
Try (untested):
```patch
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,3 +64,5 @@ jobs:
           ansible-playbook playbooks/deploy.yaml
       - name: Run tests
         run: ./run_tests
+        env:
+          PYTEST_ADDOPTS: '--certificate-source={{ matrix.certificate_source }}'
```

ehelms:
I did already, it's failing because we need to make at least one of the tests dynamic since it depends on the certificate being used:

```
FAILED tests/candlepin_test.py::test_certs_users_file - AssertionError: assert 'katelloUser=CN=quadlet.example.com, OU=PUPPET, O=FOREMAN, ST=North Carolina, C=US' == 'katelloUser=CN=quadlet.example.com'
```

.github/workflows/test.yml
```diff
@@ -61,6 +67,6 @@ jobs:
           ansible-playbook playbooks/images.yaml
       - name: Run deployment
         run: |
-          ansible-playbook playbooks/deploy.yaml
+          ansible-playbook playbooks/deploy.yaml -e certificate_source=${{ matrix.certificate_source }}
       - name: Run tests
-        run: ./run_tests
+        run: ./run_tests --certificate-source=${{ matrix.certificate_source }}
```
evgeni, line 72: Mhh, this could have worked too :thinking:

evgeni (approved):

PR #125, opened 2025-04-02
configure pulp using env vars, not a file

evgeni:


roles/pulp/tasks/main.yaml
```diff
@@ -138,8 +132,20 @@
...
+      PULP_ANALYTICS: "{{ pulp_enable_analytics | ternary('true', 'false') }}"
+      PULP_AUTHENTICATION_BACKENDS: "['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']"
+      PULP_CACHE_ENABLED: "true"
+      PULP_CONTENT_ORIGIN: "{{ pulp_content_origin }}"
+      PULP_REDIS_URL: "redis://localhost:6379/8"
+      PULP_REMOTE_USER_ENVIRON_NAME: "HTTP_REMOTE_USER"
+      PULP_REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES: "['rest_framework.authentication.SessionAuthentication', 'pulpcore.app.authentication.PulpRemoteUserAuthentication']"
+      PULP_DATABASES__default__HOST: "localhost"
+      PULP_DATABASES__default__NAME: "pulp"
+      PULP_DATABASES__default__PASSWORD: "{{ pulp_db_password }}"
```
ekohl, line 146: I think with normal systemd services the env vars are world readable. Is that also the case here?

evgeni: yes, there are reasons this PR is still a draft ;-)

roles/pulp/tasks/main.yaml
```diff
@@ -138,8 +141,21 @@
     command: pulpcore-manager migrate --noinput
     network: host
     secrets:
-      - 'pulp-settings-py,type=mount,target=/etc/pulp/settings.py'
       - 'pulp-symmetric-key,type=mount,target=/etc/pulp/certs/database_fields.symmetric.key'
+      - 'pulp-db-password,type=env,target=PULP_DATABASES__default__PASSWORD'
+    env:
+      PULP_ANALYTICS: "{{ pulp_enable_analytics | ternary('true', 'false') }}"
```
ekohl, line 147: Until the Ansible issue is fixed, can we deduplicate where they are the same values by using `pulp_settings_env` as a dict?
```suggestion
      PULP_ANALYTICS: "{{ pulp_settings_env['PULP_ANALYTICS'] }}"
```

evgeni: I've reduced the options to the ones needed for migrations

ekohl (commented):
Another thought on the db config: https://docs.djangoproject.com/en/5.1/ref/databases/#postgresql-notes mentions a service file. That way we can mount in a file (secret) while hardcoding the DB config. Perhaps even in the base container. Any thoughts on this approach?

evgeni:
I looked at that, and was not impressed, mostly because you *still* need to pass things into `DATABASES['default']` and get two files to manage.

I'd rather use https://pypi.org/project/dj-database-url/ instead, but that'd require support on the Pulp side :shrug:

ekohl:
I also didn't like the 2 files approach

ekohl (approved):
If we're going to tag Pulp into theforeman namespace we should consider adding some of our own settings into the container (as an extra layer) so we don't need to override them. For now I think this an improvement since we manage fewer files

ehelms:
There is something about this I am not understanding and the best way I can express that is through a question. How do I set one of these variables? For example, how would I disable the Pulp cache?

evgeni:
> There is something about this I am not understanding and the best way I can express that is through a question. How do I set one of these variables? For example, how would I disable the Pulp cache?

The same way you did before this PR: not at all :)

PR #128, opened 2025-04-04
Add documentation for container builds

ehelms:
This may not be the long term place we want to store these documents. In the short term, this feels like the right place.

This a lightweight copying of container build strategy that came out of some in person conversations.

### DIFF
--- a/docs/container-image-builds.md
+++ b/docs/container-image-builds.md
@@ -1,0 +1,39 @@
+# Container Builds
+
+All container images Foreman builds will live at `quay.io/foreman/$service:$tag`.
+Dependent services that we do not build ourselves will have a copy of containers stored in Foreman’s quay.io to allow tagging with the Foreman version.
+
+There will be three flavors of Foreman containers:
+
+  * Vanilla Foreman
+  * Foreman + Katello
+  * Foreman + all plugins
+
+Stage versions of containers will have stage in the name and be published in the Foreman's quay repository:
+
+  * `quay.io/foreman/$service-stage:$tag`
+
+The base image will be `quay.io/centos/centos:stream9`.
+
+## Containerfile Artifacts
+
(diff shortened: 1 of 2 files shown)


docs/container-image-builds.md
```diff
@@ -0,0 +1,37 @@
+# Container Builds
+
+All container images Foreman builds Will live at `quay.io/foreman/$service:$tag`.
```
pablomh, line 3: s/Will/will/

pablomh (changes requested):

docs/container-image-builds.md
```diff
@@ -0,0 +1,37 @@
...
+Dependent services that we do not build ourselves will have a copy of containers stored in Foreman’s quay.io to allow tagging with the Foreman version.
+
+There will be three flavors of Foreman containers:
+
+  * Vanilla Foreman
+  * Foreman + Katello
+  * Foreman + all plugins
+
+Stage versions of containers will have stage in the name: `$service-stage:$tag`
```
pablomh, line 12: Will those images be published on the same registry?

docs/container-image-builds.md
```diff
@@ -0,0 +1,39 @@
...
+
+The base image will be `quay.io/centos/centos:stream9`.
+
+## Containerfile Artifacts
+
+Container files will be stored in repositories for each service that mimic packaging.
+
+  * foreman-oci-images
+  * pulpcore-oci-images
+  * candlepin-oci-images
```
ekohl, line 24: Should we create the repos already, even if they're empty? Then update this to be links to them

ekohl (approved):

docs/container-image-builds.md
ehelms: Ehh, we can update when we are ready to start that.

PR #131, opened 2025-04-11
Prototyping CLI based on Obsah

ehelms:
Simple as I could get CLI using Obsah based on the `data-types` branch (https://github.com/theforeman/obsah/pull/63). I'll follow up with adding some additional variables that make use of types to play around with the idea more.


rop
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+# PYTHON_ARGCOMPLETE_OK
+
+"""
+An installer for Foreman.
+"""
+
+from __future__ import print_function
```
ekohl, line 8: You don't need this on Python 3

cli/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        return os.environ.get('OBSAH_INVENTORY', os.path.join(os.getcwd(), 'inventories'))
+
+
+def main(cliargs=['--help'], application_config=ApplicationConfig):  # pylint: disable=R0914
+    """
+    Main command
+    """
+    if len(sys.argv) > 1:
+        cliargs = sys.argv[1:]
+
```
ekohl, line 56: It's a bad practice in Python to have lists as defaults because they're shared between invocations. In this particular case it's not too bad but if we end up adding linters they may flag it so it's best to just avoid it:
```suggestion
def main(cliargs=None, application_config=ApplicationConfig):  # pylint: disable=R0914
    """
    Main command
    """
    if not cliargs:
        cliargs = sys.argv[1:]
        if not cliargs:
            cliargs = ['--help']
```

Perhaps this is a bit safer?

Though I really wonder if this shouldn't be the default obsah behavior. From my understanding of argparse and how obsah calls it I'd think this should work:
```suggestion
def main(cliargs=None, application_config=ApplicationConfig):  # pylint: disable=R0914
    """
    Main command
    """
```

cli/__main__.py
```diff
@@ -0,0 +1,6 @@
+"""
+Main module file. This allows you to run python -m obal
```
ekohl, line 2: This isn't obal ;)

playbooks/deploy/deploy.yaml
```diff
@@ -4,15 +4,16 @@
...
+    - "../../vars/images.yml"
+    - "../../vars/database.yml"
   vars:
     certificate_source: default
     certificates_hostnames:
       - "{{ ansible_fqdn }}"
       - localhost
     certificates_ca_password: "CHANGEME"
-    candlepin_db_password: "CHANGEME"
+    candlepin_db_password: "{{ database_password }}"
```
ekohl, line 16: We should really discourage people from reusing database passwords between applications.

rop
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+python -m cli $@
```
ekohl, line 3:
```suggestion
python -m cli "$@"
```

cli/__init__.py
ehelms: Should work in theory, that's how Obal works. But for my environment and testing, it does not work unless I have what I have currently.

roles/foreman/tasks/main.yaml
```diff
@@ -76,7 +76,7 @@
     network: host
     hostname: "{{ ansible_fqdn }}"
     secrets:
-      - 'foreman-database-url,type=env,target=DATABASE_URL'
+      - 'foreman-database-yaml,type=mount,target=/etc/foreman/database.yml'
```
evgeni, line 79: the change from env to a file is because you want to be able to set ssl mode etc?

ehelms: That and I couldn't see a way to set DB pools without this file.

evgeni: https://guides.rubyonrails.org/configuring.html#configuring-a-database says you add `?pool=5` to the URL.

cli/__init__.py
evgeni: Can you explain a bit better what you mean by "does not work"?

I have applied the following patch here locally:
```
diff --git cli/__init__.py cli/__init__.py
index 7b57828..9af357c 100755
--- cli/__init__.py
+++ cli/__init__.py
@@ -47,12 +47,10 @@ class ApplicationConfig(obsah.ApplicationConfig):
         return os.environ.get('OBSAH_INVENTORY', os.path.join(os.getcwd(), 'inventories'))
 
 
-def main(cliargs=['--help'], application_config=ApplicationConfig):  # pylint: disable=R0914
+def main(cliargs=None, application_config=ApplicationConfig):  # pylint: disable=R0914
     """
     Main command
     """
-    if len(sys.argv) > 1:
-        cliargs = sys.argv[1:]
 
     obsah.main(cliargs=cliargs, application_config=application_config)
```
and `rop` seems to work fine?

```
% ./rop              
usage: rop [-h] action ...
rop: error: the following arguments are required: action

% ./rop -h
usage: rop [-h] action ...

positional arguments:
  action         which action to execute
    deploy       Install
    pull-images  Pull necessary container images
    setup        Setup up a host to be installed.

options:
  -h, --help     show this help message and exit

% ./rop deploy -h
usage: rop deploy [-h] [-v] [-e EXTRA_VARS] [--candlepin-database-name CANDLEPIN_DATABASE_NAME] [--candlepin-database-password CANDLEPIN_DATABASE_PASSWORD] [--candlepin-database-user CANDLEPIN_DATABASE_USER]
                  [--certificate-source {default,installer}] [--database-host DATABASE_HOST] [--database-mode {internal,external}] [--database-port DATABASE_PORT] [--database-ssl-ca DATABASE_SSL_CA]
                  [--database-ssl-mode DATABASE_SSL_MODE] [--foreman-database-name FOREMAN_DATABASE_NAME] [--foreman-database-password FOREMAN_DATABASE_PASSWORD] [--foreman-database-user FOREMAN_DATABASE_USER]
                  [--pulp-database-name PULP_DATABASE_NAME] [--pulp-database-password PULP_DATABASE_PASSWORD] [--pulp-database-user PULP_DATABASE_USER]

Install

options:
  -h, --help            show this help message and exit
  -v, --verbose         verbose output
  --candlepin-database-name CANDLEPIN_DATABASE_NAME
                        Name of the Candlepin database.
…
```

setup.py
```diff
@@ -0,0 +1,80 @@
...
+    packages=find_packages(exclude=['contrib', 'docs', 'tests']),
+
+    install_requires=['obsah >= 1.0.0'],
+
+    extras_require={
+        'argcomplete': ['argcomplete'],
+    },
+
+    package_data={
+        'cli': find_package_data('cli'),
```
ekohl, line 72: This is missing a second argument so it can't work.

setup.py
```diff
@@ -0,0 +1,80 @@
...
+    },
+
+    entry_points={
+        'console_scripts': [
+            'cli=cli:main',
```
ekohl, line 77: This creates a command `cli` but don't you want to name this `rop`? Then it would also remove the need for a wrapper script.
```suggestion
            'rop=cli:main',
```

cli/__init__.py
evgeni: But obal doesn't do that either

(I agree it would be cool, but it works as the other thing right now)

ekohl (commented):
I took a stab at making it really shippable and https://github.com/ekohl/foreman-quadlet/commit/55664e853c8e174216261cdb1da88f49f7302aba is the result. Probably still incomplete and I didn't look at the README yet.

I had an idea to also use this as an input to an RPM and then add packit, but this is what I could do before heading back into a meeting.

cli/__init__.py
ehelms: Yea, obal doesn't work that way either but that doesn't mean it shouldn't :) And right now this version of it does work that way... what's the downside of this implementation?

evgeni: Well, you wrote "Should work in theory, that's how Obal works." which led me to assume "it works in obal but not in rop".

The downside is mostly "it's ugly" :)

As an alternative (and that would benefit all consumers), we could adjust it in obsah, like this:
```
diff --git obsah/__init__.py obsah/__init__.py
index 08b919d..52546da 100755
--- obsah/__init__.py
+++ obsah/__init__.py
@@ -316,6 +316,8 @@ def obsah_argument_parser(application_config=ApplicationConfig, playbooks=None,
     if argcomplete:
         argcomplete.autocomplete(parser)
 
+    parser.format_usage = parser.format_help
+
     return parser
```
(this forces "usage" which is what's used here, to use the longer "help" output instead)

I poked a few Pythonistas if there is a nicer way.

ekohl: What @evgeni suggested is also where I was heading, just didn't have time to look at the code just yet.

setup.py
ekohl: Right, but I wanted to get rid of the wrapper script you have now. The `console_scripts` entry point will work nicely once we start packaging things up in RPMs etc.

cli/__init__.py
ehelms: That sounds like a winner!

setup.py
ehelms: How do I call it without a wrapper script from git checkout?

ekohl: The workflow becomes
```console
git clone https://github.com/theforeman/foreman-quadlet
cd foreman-quadlet
# create a virtualenv and activate it
pip install -e .
pip install -r requirements.txt # we should merge these 2 together
```

Now in your virtualenv you have the wrapper script available just like it would be in production, calling the correct Python.

uv is also a new tool that makes packaging single scripts easier: https://docs.astral.sh/uv/guides/scripts/#running-a-script-with-dependencies. Perhaps we should investigate that approach too.

ehelms: My view was the native python packaging can happen after we've pioneered the start of the CLI as that can be a re-factoring on top of a working model. I'd like to explore the question of:

>  "Related to this is discussions about if this helps us provide an equivalent to the [operations collection](https://github.com/theforeman/foreman-operations-collection) more natively."

As I wonder about shipping the CLI bits as Python and the Ansible bits as a collection. Is it possible for it to work that way? can Obsah consume some Ansible bits as a collection? I'd assume so since we use modules that come from collections in Obsah.

ekohl: The way I view it we'll still have some CLI bits (likely as Python), but we can ship the actual Ansible bits separate. That effectively ends up in a similar situation like foreman-installer with separate Puppet modules.

Big question is: do we vendor the collections in packaging like we do with Puppet or ship them separate.

ekohl: Also, having some artifact we can install in a production like setup can make testing with tmt easier. That was my angle.

cli/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+    def data_path():
+        path = os.environ.get('ROP_DATA')
+        if path is None:
+            with resources.path(__name__, 'data') as fspath:
+                path = fspath.as_posix()
+
+        return path
+
+    @staticmethod
+    def inventory_path():
```
ehelms, line 35: @evgeni I think this concept get's kinda weird in this use case compared to Obal's. The inventory will be dynamic for this use case, and therefore shouldn't live in the Python package. For now, here in the git repository, the `inventories/` top level path works. But if I were taking it a step further and installing the Python package on a fresh machine, I would need an inventory defined and specified for using this against remote targets. For running this on the same machine (local) then having a default inventory works just fine. I was thinking of:

 1. Default to localhost so that the default mode is installation on the machine the installer exists.
 2. A user parameter that allows specifying the location of an inventory file.
 
 I wondered about defaulting to also checking `/etc/rop` for an inventory file for the standard packaged case.

ehelms:
This is now rebased on latest, and converted to a Python package along with updates to testing and Github actions. This also introduces a script to setup the virtualenv easily for local development and use in CI.

requirements.txt
```diff
@@ -4,3 +4,4 @@ paramiko
 pytest-testinfra
 pytest-durations
 pyyaml
+obsah @ git+https://github.com/theforeman/obsah
```
ekohl, line 7: Let's create a release so we don't need to and `setup.py` can simply point to the desired version.

setup-environment
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+python -m venv .venv
+source .venv/bin/activate
+pip install -e .
+pip install -r requirements.txt # we should merge these 2 together
```
ekohl, line 6: You can add `-e .` to `requirements.txt` to merge it.

.github/workflows/test.yml
```diff
@@ -51,23 +52,24 @@ jobs:
...
-      - name: Install dependencies
-        run: pip install -r requirements.txt
-      - name: Install collection dependencies
-        run: ansible-galaxy install -r requirements.yml
+      - name: Setup environment
+        run: |
+          ./setup-environment
+          source .venv/bin/activate
+          echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
+          echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> $GITHUB_ENV
```
ekohl, line 60: I don't think you need to setup a virtualenv with `setup-python`. Also, see https://github.com/actions/setup-python?tab=readme-ov-file#caching-packages-dependencies for how to use the caching.

ehelms: I'm not using `setup-python` unless I'm missing something?

evgeni: line 44:
```
      - name: Set up Python
        uses: actions/setup-python@v5
```

ehelms:
This is ready for another round of review. The failing test is a known frequent intermittent test failure. I'll worry about re-running it after I've addressed any updated comments.

I've not added any new tests yet to this. The current tests are geared toward testing the output of an installation. I'd appreciate opinion on the best way to think of testing the cli aspect (parameter inputs for example).

setup.py
```diff
@@ -0,0 +1,63 @@
...
+    url='https://github.com/theforeman/foreman-quadlet',
+    author='The Foreman Project',
+    author_email='foreman-dev@googlegroups.com',
+    zip_safe=False,
+
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Intended Audience :: Developers',
+        'Topic :: Software Development :: Build Tools',
+        'License :: OSI Approved :: GNU General Public License v2 (GPLv2)',
```
ekohl, line 38: Can you add `license='GPL-2.0-only'` to `setup()` and drop this line to avoid the deprecation warning?
```suggestion
```

I did the same in obsah: https://github.com/theforeman/obsah/commit/3c65e0ac0f8f75c31cb1159fcc13f7f54b538d12 added the `license` and https://github.com/theforeman/obsah/commit/52808215031233b63c081e953582cd7ac6661d30 dropped the classifier

setup.py
```diff
@@ -0,0 +1,63 @@
...
+        'Programming Language :: Python :: 3.12',
+        'Programming Language :: Python :: 3.13',
+    ],
+
+    keywords='foreman',
+
+    packages=find_namespace_packages(include=['cli']),
+    include_package_data=True,
+
+    install_requires=['obsah >= 1.0.0'],
```
ekohl, line 52: I just released 1.1.0 which includes the typing we need:
```suggestion
    install_requires=['obsah >= 1.1.0'],
```

setup-environment
ekohl: Since you included `-e .` I'd expect this to work and properly resolve all dependencies
```suggestion.
pip install -r requirements.txt
```

requirements.txt
ekohl: 1.1.0 was released
```suggestion
```

ekohl:
I've built obsah 1.1.0 in https://copr.fedorainfracloud.org/coprs/ekohl/obal/build/8984539/ and am tempted to build an RPM for this CLI using it.

cli/data/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -0,0 +1,54 @@
...
+    help: Name of the Pulp database.
+  pulp_database_user:
+    help: User of the Pulp database.
+  pulp_database_password:
+    help: Password for the Pulp database.
+constraints:
+  required_together:
+    - [database_ssl_mode, database_ssl_ca]
+  required_if:
+    - ['database_mode', 'external', ['database_host', 'database_port']]
```
evgeni, line 54: Why is the port required? Shouldn't it be 5432 in 99.9% of the cases (aka: it can have a sensible default)?

setup.py
```diff
@@ -0,0 +1,63 @@
...
+        'Programming Language :: Python :: 3.12',
+        'Programming Language :: Python :: 3.13',
+    ],
+
+    keywords='foreman',
+
+    packages=find_namespace_packages(include=['cli']),
+    include_package_data=True,
+
+    install_requires=['obsah >= 1.1.0'],
```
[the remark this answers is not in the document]
ehelms: Is it being used indirectly by the Ansible modules from other collections?

cli/__init__.py
evgeni: Yeah, I think for this case, `/etc/rop/inventory.yml` (or whatever) is the cleanest solution.

setup.py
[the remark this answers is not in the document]
ekohl: Yes. It's in the setup playbook now and by adding it here I think you can drop it.

cli/data/playbooks/setup/metadata.obsah.yaml
```diff
@@ -0,0 +1,3 @@
+---
+help: |
+  Setup up a host to be installed.
```
ekohl, line 3: https://grammarist.com/spelling/set-up-vs-setup/

```suggestion
  Set up a host to be installed.
```

ekohl (changes requested):

setup.py
```diff
@@ -0,0 +1,63 @@
...
+        'Programming Language :: Python :: 3.12',
+        'Programming Language :: Python :: 3.13',
+    ],
+
+    keywords='foreman',
+
+    packages=find_namespace_packages(include=['cli']),
+    include_package_data=True,
+
+    install_requires=['obsah >= 1.1.0', 'cryptography'],
```
ekohl, line 52: Nit: I think we should write this 1 line per dependency

cli/data/playbooks/setup/setup.yaml
```diff
@@ -14,7 +16,6 @@
       ansible.builtin.package:
         name:
           - podman
-          - python3-cryptography
           - python3-libsemanage
           - python3-requests
           - bash-completion
```
ekohl, line 21: 2 lower is also `python3-psycopg2` which needs to be moved into `setup.py` as a proper runtime dependency because `community.postgresql` depends on it

setup.py
```diff
@@ -0,0 +1,66 @@
...
+    ],
+
+    keywords='foreman',
+
+    packages=find_namespace_packages(include=['cli']),
+    include_package_data=True,
+
+    install_requires=[
+        'obsah >= 1.1.0',
+        'cryptography'
```
ekohl, line 54: Nit: I'm a fan of alphabetically sorted lists

ekohl (changes requested):

setup.py
ehelms: I was spicing it up for you -- reverse alphabetical :) I'll fix

.github/workflows/test.yml
ekohl, line 1: This file also installs Ansible, but we now pull that in via `obsah` so doesn't have to mentioned either.

ekohl (changes requested):

ehelms:
I opted for a change in strategy to drop the `setup` command/playbook and move those dependencies around. Two notable things:

 1. This moves the etc_host setup useful for testing into a small playbook that is run by Vagrant at provisioning time.
 2. This moves (in an ugly-ish way) the "what do we do about these dependencies" into the deploy playbook for now.

.gitignore
ekohl, line 1: Can we also ignore `dist/` which is created on `python setup.py sdist`?

fixtures/etc_host.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Set up etc_hosts for each host
+  become: true
+  hosts:
+    - quadlet
+    - client
```
ekohl, line 6: Should this apply on all hosts and let Vagrant figure it out?
```suggestion
  hosts: all
```

cli/data/playbooks/deploy/deploy.yaml
```diff
@@ -41,6 +41,17 @@
...
+      ansible.builtin.package:
+        name:
+          - podman
+    - name: Install other dependencies
+      ansible.builtin.package:
+        name:
+          - bash-completion
+          - nmap
+          - python3-libsemanage
+          - python3-psycopg2
```
ekohl, line 54: This is already listed in `install_requires` so we can drop this
```suggestion
```

However, now I need to be kept honest. https://galaxy.ansible.com/ui/repo/published/community/postgresql/docs/ lists external requirements but it doesn't list where. Is it the control node or the node it's really running on? That may be important for external database support.

ehelms: Exactly, otherwise you get https://github.com/theforeman/foreman-quadlet/actions/runs/14798416218/job/41551130603#step:10:251

ekohl: > On the controlled node (so the external db)

So the installation really needs to be in the PostgreSQL role? And we can drop it from `setup.py`?

Then we'd need to do the same in the place where we use the crypto role.

ekohl: We can add them as extras in setup.py. In the final system we can then add a Suggests on the extra. By default dnf will install that, reducing the transactions. I'd hope that ansible ensuring an already installed package is cheap

ekohl (commented):
Random shower thought: can we use a simple shell script wrapper instead?

My idea is to configure obsah via environment variables. Something like:
```sh
#!/usr/bin/sh

OBSAH_NAME=rop
OBSAH_DATA=data
OBSAH_INVENTORY=
export OBSAH_NAME OBSAH_DATA OBSAH_INVENTORY

exec obsah "$@"
```

Right now `OBSAH_NAME` doesn't exist but we can enhance it to. I had a thought to enhance obsah where you symlink `/usr/bin/obsah` and it looks at `__name__` to derive it, but since you need to set `OBSAH_DATA` anyway I think you need a wrapper anyway.

How I started to think about this was development. Can we install production `rop` (or whatever), ship all data somewhere and set `ROP_DATA` / `OBSAH_DATA` to that location? Then if you take that to a more logical extreme: do we even need `rop` then?

If we go that route then it may be feasible to package it up as a true Ansible collection while in the RPM we only add the trivial wrapper and package dependencies.

Thinking more about that, perhaps we can still use the symlink `/usr/bin/obsah` idea but add a `/usr/libexec/obsah-collection` (or something) that will look at an ansible collection with the same name in `/usr/share/ansible/collections/ansible_collections` (though that may be too hard with namespaces). Or we can introduce `/usr/share/obsah` for that purpose.

cli/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def main(cliargs=['--help'], application_config=ApplicationConfig):  # pylint: disable=R0914
+    """
+    Main command
+    """
+    if len(sys.argv) > 1:
+        cliargs = sys.argv[1:]
+
+    obsah.main(cliargs=cliargs, application_config=application_config)
```
evgeni, line 49:
```suggestion
def main(cliargs=None, application_config=ApplicationConfig):  # pylint: disable=R0914
    """
    Main command
    """

    obsah.main(cliargs=cliargs, application_config=application_config)
```

https://github.com/theforeman/obsah/commit/583442c65a87f998d06d069653fe5a78e6661c21 is in

setup.py
```diff
@@ -0,0 +1,67 @@
...
+setup(
+    name='cli',
+    version='0.0.1',
+    license='GPL-2.0-only',
+    description='packaging wrapper using ansible',
+    long_description=get_long_description(),
+    long_description_content_type='text/markdown',
+    url='https://github.com/theforeman/foreman-quadlet',
+    author='The Foreman Project',
+    author_email='foreman-dev@googlegroups.com',
```
evgeni, line 32: does that even still exist? :)

cli/__init__.py
ehelms: Weird, I thought I had already put this back.

cli/__init__.py
evgeni: To elaborate a bit what I wanted to say: Ansible implies `localhost` when the given intentory is empty or non-parse-able, so we can just point this to the "production" inventory (which the user will edit, IF they want to add the external DB) that we ship in the RPM, and the dev setup should still work?

```
% ansible -i /etc/lol/what -m ping localhost
[WARNING]: Unable to parse /etc/lol/what as an inventory source
[WARNING]: No inventory was parsed, only implicit localhost is available
localhost | SUCCESS => {
    "changed": false,
    "ping": "pong"
}
```

evgeni:
> Random shower thought: can we use a simple shell script wrapper instead?
> 
> My idea is to configure obsah via environment variables. Something like:
> 
> ```shell
> #!/usr/bin/sh
> 
> OBSAH_NAME=rop
> OBSAH_DATA=data
> OBSAH_INVENTORY=
> export OBSAH_NAME OBSAH_DATA OBSAH_INVENTORY
> 
> exec obsah "$@"
> ```

Most probably: yes. We originally went this route, as the current `ApplicationConfig` in `obal` contains executable code (`os.getcwd()` for the inventory etc), which would need more logic in this wrapper. `rop` probably won't have that, so it can be easier.

That said, this is an improvement we can do later. Or are you hinting at "then we don't need a python package after all"?

> Right now `OBSAH_NAME` doesn't exist but we can enhance it to. I had a thought to enhance obsah where you symlink `/usr/bin/obsah` and it looks at `__name__` to derive it, but since you need to set `OBSAH_DATA` anyway I think you need a wrapper anyway.
> 
> How I started to think about this was development. Can we install production `rop` (or whatever), ship all data somewhere and set `ROP_DATA` / `OBSAH_DATA` to that location? Then if you take that to a more logical extreme: do we even need `rop` then?
> 
> If we go that route then it may be feasible to package it up as a true Ansible collection while in the RPM we only add the trivial wrapper and package dependencies.
> 
> Thinking more about that, perhaps we can still use the symlink `/usr/bin/obsah` idea but add a `/usr/libexec/obsah-collection` (or something) that will look at an ansible collection with the same name in `/usr/share/ansible/collections/ansible_collections` (though that may be too hard with namespaces). Or we can introduce `/usr/share/obsah` for that purpose.

We should not put (or require! or even accept!) things in `/usr/share/ansible/collections/ansible_collections`. People will have all sorts of weird stuff there. Incompatible versions etc.

But yes, we should place things in `/usr/share/something` and point the vars there for prod

setup-environment
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+python -m venv .venv
+source .venv/bin/activate
+pip install -r requirements.txt
+ansible-galaxy install -r requirements.yml
+
+echo "\n\nSetup complete. Activate the virtualenv 'source .venv/bin/activate' and then run the cli 'rop'"
```
evgeni, line 8: `bash`'s `echo` does not interpret escape chars by default. you need to either use `echo -e` (non POSIX, afaik) or `printf`

ehelms:
> > Random shower thought: can we use a simple shell script wrapper instead?
> > My idea is to configure obsah via environment variables. Something like:
> > ```shell
> > #!/usr/bin/sh
> > 
> > OBSAH_NAME=rop
> > OBSAH_DATA=data
> > OBSAH_INVENTORY=
> > export OBSAH_NAME OBSAH_DATA OBSAH_INVENTORY
> > 
> > exec obsah "$@"
> > ```
> 
> Most probably: yes. We originally went this route, as the current `ApplicationConfig` in `obal` contains executable code (`os.getcwd()` for the inventory etc), which would need more logic in this wrapper. `rop` probably won't have that, so it can be easier.
> 
> That said, this is an improvement we can do later. Or are you hinting at "then we don't need a python package after all"?

I think it's an interesting idea, that we should capture but I'd like to see us get further and through more of the complex bits before we decide that it can be that simple.

We also have some non-collection things in our tool like the metadata. I guess we could put it in there but would the collection validators complain?

setup.py
```diff
@@ -0,0 +1,67 @@
...
+
+    with codecs.open(os.path.join(here, 'README.md'), encoding='utf-8') as readme:
+        return readme.read()
+
+
+setup(
+    name='cli',
+    version='0.0.1',
+    license='GPL-2.0-only',
+    description='packaging wrapper using ansible',
```
evgeni, line 27:
```suggestion
    description='foreman deployment using ansible',
```

ekohl:
> That said, this is an improvement we can do later. Or are you hinting at "then we don't need a python package after all"?

I was thinking about that. At least for the prototype it would keep things simple if we don't need a Python package.

Then when we get closer to an actual installer (and perhaps tool to replace foreman-maintain) we can look at how to build it. For example, `foreman-maintain configure` could be the new installer entry point (with an `foreman-maintain install` as an alias).

evgeni:
https://github.com/theforeman/obsah/pull/75

evgeni:
I like the "simple shell script" idea, as it not only reduces the amount of (boilerplate) code, but also simplifies the directory structure in the repo. OTOH the current state in the PR works fine and would allow us to continue iterating.

Read this comment as: ACK, merge it, unless you want to rewrite it again right now :)

> We also have some non-collection things in our tool like the metadata. I guess we could put it in there but would the collection validators complain?

Collection validators don't care much about additional files

evgeni:
> [theforeman/obsah#75](https://github.com/theforeman/obsah/pull/75)

that's merged, and the following works after a `git mv cli/data .`:
```
#!/bin/bash

OBSAH_NAME=rop
OBSAH_DATA=data
OBSAH_INVENTORY=./inventories/
export OBSAH_NAME OBSAH_DATA OBSAH_INVENTORY

exec obsah "$@"
```

do whatever you want with that info :)

ehelms:
> Then when we get closer to an actual installer (and perhaps tool to replace foreman-maintain) we can look at how to build it. For example, `foreman-maintain configure` could be the new installer entry point (with an `foreman-maintain install` as an alias).

Are you thinking this would be recreated in Ansible and follow the same CLI model? Or that it would require some real code in python that would be handled?

ekohl:
> Are you thinking this would be recreated in Ansible and follow the same CLI model? Or that it would require some real code in python that would be handled?

That will depend on how much of the functionality we keep. I'd love to keep it minimal but at this point I'm unclear on where we'd end up and if it's realistic.

ehelms:
Closing in favor of https://github.com/theforeman/foreman-quadlet/pull/143

PR #132, opened 2025-04-15
Abstract away Vagrant as the VM provider to speed up subsequent calls

ehelms:
Introduces a script to start and stop VMs and produce a static inventory. This speeds up each call after the fact and creates less reliance on Vagrant.

### DIFF
--- a/start-vms
+++ b/start-vms
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+mkdir -p inventories
+vagrant up $@
+
+./vagrant.py --yaml > inventories/local_vagrant
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,8 +57,7 @@ jobs:
         run: ansible-galaxy install -r requirements.yml
       - name: Start VMs
         run: |
-          vagrant up quadlet
-          vagrant up client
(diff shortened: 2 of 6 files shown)


start-vms
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+mkdir -p inventories
+vagrant up $@
+
+./vagrant.py --yaml > inventories/local_vagrant
```
archanaserver, line 6: `start-vms` script currently overwrites `inventories/local_vagrant` without any checks, should we preserve the existing inventory file when regenerating?

archanaserver (commented):
Since this PR introduces new scripts (`start-vms`/`stop-vms`) and changes inventory handling, could we update the `README.md` file.

start-vms
ehelms: Since the intent is that this script controls the creation of these VMs, I do not see a reason to preserve the inventory in any way.

ehelms:
> Since this PR introduces new scripts (`start-vms`/`stop-vms`) and changes inventory handling, could we update the `README.md` file.

Updated

archanaserver (approved):
LGTM!

PR #134, opened 2025-04-15
Add all database variables for Foreman, Candlepin and Pulp

ehelms:
This does include SSL variables, but not SSL testing for PostgreSQL access. I think that will require quite a bit extra to spin up a variation of testing that include those. This PR is instead intended to centralize the database variables to a single set of inputs, that I can then use with Obsah (https://github.com/theforeman/obsah/pull/63) to do variable combining and mutual requirement testing on top of https://github.com/theforeman/foreman-quadlet/pull/13.

### DIFF
--- a/playbooks/deploy.yaml
+++ b/playbooks/deploy.yaml
@@ -6,13 +6,13 @@
   vars_files:
     - "../vars/{{ certificate_source }}_certificates.yml"
     - "../vars/images.yml"
+    - "../vars/database.yml"
   vars:
     certificate_source: default
     certificates_hostnames:
       - "{{ ansible_fqdn }}"
       - localhost
     certificates_ca_password: "CHANGEME"
-    candlepin_db_password: "CHANGEME"
     candlepin_keystore_password: "CHANGEME"
     candlepin_oauth_secret: "CHANGEME"
     candlepin_ca_key_password: "{{ ca_key_password }}"
@@ -30,30 +30,13 @@
     foreman_ca_certificate: "{{ ca_certificate }}"
     foreman_client_key: "{{ client_key }}"
     foreman_client_certificate: "{{ client_certificate }}"
-    foreman_db_password: "CHANGEME"
     foreman_oauth_consumer_key: abcdefghijklmnopqrstuvwxyz123456
     foreman_oauth_consumer_secret: abcdefghijklmnopqrstuvwxyz123456
     httpd_server_ca_certificate: "{{ ca_certificate }}"
     httpd_client_ca_certificate: "{{ ca_certificate }}"
     httpd_server_certificate: "{{ server_certificate }}"
     httpd_server_key: "{{ server_key }}"
-    pulp_db_password: "CHANGEME"
     pulp_content_origin: "https://{{ ansible_fqdn }}"
-    postgresql_restarted_state: started
-    postgresql_databases:
-      - name: candlepin
-        owner: candlepin
-      - name: foreman
-        owner: foreman
-      - name: pulp
-        owner: pulp
-    postgresql_users:
-      - name: candlepin
-        password: "{{ candlepin_db_password }}"
-      - name: foreman
-        password: "{{ foreman_db_password }}"
-      - name: pulp
-        password: "{{ pulp_db_password }}"
   pre_tasks:
     - name: Deploy debug_tools
       ansible.builtin.include_role:
--- a/roles/candlepin/templates/candlepin.conf.j2
+++ b/roles/candlepin/templates/candlepin.conf.j2
@@ -20,10 +20,10 @@ candlepin.async.jobs.ExpiredPoolsCleanupJob.schedule=0 0 0 * * ?
 
 jpa.config.hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect
 jpa.config.hibernate.hbm2ddl.auto=validate
-jpa.config.hibernate.connection.username=candlepin
-jpa.config.hibernate.connection.password={{ candlepin_db_password }}
+jpa.config.hibernate.connection.username={{ candlepin_database_user }}
+jpa.config.hibernate.connection.password={{ candlepin_database_password }}
 jpa.config.hibernate.connection.driver_class=org.postgresql.Driver
-jpa.config.hibernate.connection.url=jdbc:postgresql://{{ candlepin_db_host }}:{{ candlepin_db_port }}/candlepin{% if candlepin_db_ssl %}?ssl=true{% endif %}{% if candlepin_db_ssl and candlepin_db_ssl_ca is defined %}&sslrootcert={{ candlepin_db_ssl_ca }}{% endif %}{% if not candlepin_db_ssl_verify and candlepin_db_ssl %}&sslfactory=org.postgresql.ssl.NonValidatingFactory{% endif %}
+jpa.config.hibernate.connection.url=jdbc:postgresql://{{ candlepin_database_host }}:{{ candlepin_database_port }}/{{ candlepin_database_name }}?sslmode={{ candlepin_database_ssl_mode }}{% if candlepin_database_ssl_ca is defined %}&sslrootcert={{ candlepin_database_ssl_ca }}{% endif %}
 
 
 org.quartz.jobStore.misfireThreshold=60000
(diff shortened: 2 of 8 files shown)


ekohl (commented):
As stated elsewhere, I don't like reusing passwords between applications. It's insecure and I think we should avoid, even if it's not so nice for users that they need to specify multiple parameters

ehelms:
> As stated elsewhere, I don't like reusing passwords between applications. It's insecure and I think we should avoid, even if it's not so nice for users that they need to specify multiple parameters

I can update to that model.

I've also started documenting parameters, so another view of this is https://github.com/theforeman/foreman-quadlet/pull/136

vars/database.yml
```diff
@@ -0,0 +1,38 @@
...
+
+candlepin_db_user: "{{ candlepin_database_user }}"
+candlepin_db_name: "{{ candlepin_database_name }}"
+candlepin_db_password: "{{ candlepin_database_password }}"
+candlepin_db_host: "{{ database_host }}"
+candlepin_db_port: "{{ database_port }}"
+candlepin_db_ssl_mode: "{{ database_ssl_mode }}"
+candlepin_db_ssl_ca: "{{ database_ssl_ca }}"
+
+pulp_settings_db_env:
```
ehelms, line 26: @evgeni Am I doing this right here? Or do you expect each of the entries to be their own variable mapped to a secret that creates an environment variable?

vars/database.yml
```diff
@@ -0,0 +1,38 @@
...
+
+foreman_database_name: foreman
+foreman_database_user: foreman
+foreman_database_password: CHANGEME
+candlepin_database_user: candlepin
+candlepin_database_name: candlepin
+candlepin_database_password: CHANGEME
+pulp_database_user: pulp
+pulp_database_name: pulp
+pulp_database_password: CHANGEME
```
ehelms, line 16: For these I think we have two options. 

Option 1: These serve as the role variables and we expose them externally to the user as the inputs.

Option 2: These server as the external inputs for the user and those map to internal role variables (e.g. foreman_db_name)

roles/foreman/templates/database.yaml.j2
```diff
@@ -0,0 +1,21 @@
+---
```
ehelms, line 1: This file will be removed on next push.

playbooks/deploy.yaml
```diff
@@ -51,9 +49,9 @@
       - name: candlepin
         password: "{{ candlepin_db_password }}"
```
ekohl, line 50: If you're changing things, can you make this consistent
```suggestion
        password: "{{ candlepin_database_password }}"
```

playbooks/deploy.yaml
```diff
@@ -51,9 +49,9 @@
...
       - name: foreman
```
ekohl, line 51: We now have a variable for this:
```suggestion
      - name: "{{ foreman_database_user }}"
```

playbooks/deploy.yaml
```diff
@@ -51,9 +49,9 @@
...
-        password: "{{ foreman_db_password }}"
+        password: "{{ foreman_database_password }}"
       - name: pulp
```
ekohl, line 53: We now have a variable for this:
```suggestion
      - name: "{{ pulp_database_user }}"
```

playbooks/deploy.yaml
```diff
@@ -51,9 +49,9 @@
       - name: candlepin
```
ekohl, line 49: We now have a variable for this:
```suggestion
      - name: "{{ candlepin_database_user }}"
```

roles/foreman/tasks/main.yaml
```diff
@@ -8,7 +8,7 @@
   containers.podman.podman_secret:
     state: present
     name: foreman-database-url
-    data: "postgresql://foreman:{{ foreman_db_password }}@localhost/foreman"
+    data: "postgresql://{{ foreman_database_user }}:{{ foreman_database_password }}@{{ foreman_database_host }}:{{ foreman_database_port }}/{{ foreman_database_name }}?pool={{ foreman_database_pool }}{% if foreman_database_sslmode %}&sslmode={{ foreman_database_sslmode }}{% endif %}{% if foreman_database_ssl_ca is defined %}&sslrootcert={{ foreman_database_ssl_ca }}{% endif %}"
```
ekohl, line 11: There is a default for SSL mode so can we simplifyy this?
```suggestion
    data: "postgresql://{{ foreman_database_user }}:{{ foreman_database_password }}@{{ foreman_database_host }}:{{ foreman_database_port }}/{{ foreman_database_name }}?pool={{ foreman_database_pool }}&sslmode={{ foreman_database_sslmode }}{% if foreman_database_ssl_ca is defined %}&sslrootcert={{ foreman_database_ssl_ca }}{% endif %}"
```

playbooks/deploy.yaml
ehelms: Yea - I want to settle https://github.com/theforeman/foreman-quadlet/pull/134#discussion_r2056242787 before I update them as that will likely dictate the naming scheme.

vars/database.yml
ekohl: > Option 1: These serve as the role variables and we expose them externally to the user as the inputs.

I like this best.

I also lean to `X_database_Y` if you also have `database_Y`. More concrete: `--foreman-database-password` and `--database-host`. I'll also accept `--foreman-db-password` and `--db-host` but think writing out `database` isn't that bad.

Random thought: do we want to accept aliases? Probably not, but this is a good example where it could make sense.

roles/candlepin/templates/candlepin.conf.j2
```diff
@@ -20,10 +20,10 @@ candlepin.async.jobs.ExpiredPoolsCleanupJob.schedule=0 0 0 * * ?
 
 jpa.config.hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect
 jpa.config.hibernate.hbm2ddl.auto=validate
-jpa.config.hibernate.connection.username=candlepin
+jpa.config.hibernate.connection.username={{ candlepin_db_user }}
 jpa.config.hibernate.connection.password={{ candlepin_db_password }}
 jpa.config.hibernate.connection.driver_class=org.postgresql.Driver
-jpa.config.hibernate.connection.url=jdbc:postgresql://{{ candlepin_db_host }}:{{ candlepin_db_port }}/candlepin{% if candlepin_db_ssl %}?ssl=true{% endif %}{% if candlepin_db_ssl and candlepin_db_ssl_ca is defined %}&sslrootcert={{ candlepin_db_ssl_ca }}{% endif %}{% if not candlepin_db_ssl_verify and candlepin_db_ssl %}&sslfactory=org.postgresql.ssl.NonValidatingFactory{% endif %}
+jpa.config.hibernate.connection.url=jdbc:postgresql://{{ candlepin_db_host }}:{{ candlepin_db_port }}/{{ candlepin_db_name }}{% if candlepin_db_ssl_mode %}?sslmode={{ candlepin_db_ssl_mode }}{% endif %}{% if candlepin_db_ssl_ca is defined %}&sslrootcert={{ candlepin_db_ssl_ca }}{% endif %}
```
ekohl, line 26: We also have a default and can simplify this. That also resolves the issue that if `candlepin_db_ssl_ca` is defined but `candlepin_db_ssl_mode` isn't the URL is invalid. And I wonder if `candlepin_db_ssl_ca` isn't always defined since it's mentioned in the defaults. Perhaps like this?
```suggestion
jpa.config.hibernate.connection.url=jdbc:postgresql://{{ candlepin_db_host }}:{{ candlepin_db_port }}/{{ candlepin_db_name }}?sslmode={{ candlepin_db_ssl_mode }}{% if candlepin_db_ssl_ca %}&sslrootcert={{ candlepin_db_ssl_ca }}{% endif %}
```

The same can be applied below for `org.quartz.dataSource.myDS.URL`.

vars/database.yml
ehelms: > Random thought: do we want to accept aliases? Probably not, but this is a good example where it could make sense.

Can you define aliases in this context?

vars/database.yml
```diff
@@ -0,0 +1,50 @@
...
+
+candlepin_database_host: "{{ database_host }}"
+candlepin_database_port: "{{ database_port }}"
+candlepin_database_ssl_mode: "{{ database_ssl_mode }}"
+candlepin_database_ssl_ca: "{{ database_ssl_ca }}"
+
+pulp_settings_database_env:
+  PULP_DATABASES__default__NAME: "{{ pulp_database_name }}"
+  PULP_DATABASES__default__USER: "{{ pulp_database_user }}"
+  PULP_DATABASES__default__PASSWORD: "{{ pulp_database_password }}"
```
evgeni, line 26: this needs to be a secret, the others can remain in the main "env dict"

evgeni: And because that's already done in `roles/pulp/tasks/main.yaml`, you can just
```suggestion
```

evgeni: but at the same time, I don't think you need to define `pulp_settings_database_env` here -- `pulp_database_*` is the input to the `pulp` role, which is all we should care about in the vars file.

ehelms: I don't see for example `pulp_database_host` in the Pulp role anywhere. What am I missing about the design?

ehelms: Should I be adding podman secret blocks for each? Or should I be adding these to defaults? Or changing https://github.com/theforeman/foreman-quadlet/blob/master/roles/pulp/defaults/main.yaml#L21 to not be hardcoded to `localhost` but instead to `pulp_database_host` ?

evgeni: Well, it's not yet defined (as we only needed localhost), but the structure is here:
https://github.com/theforeman/foreman-quadlet/blob/a203f0170a23e5159028855bd8ffb3ab7211e2c7/roles/pulp/defaults/main.yaml#L20-L24

so you could change that to
```yaml
pulp_database_name: pulp
pulp_database_user: pulp
pulp_database_host: localhost
pulp_database_port: 5432

pulp_settings_db_env:
  PULP_DATABASES__default__HOST: "{{ pulp_database_host }}"
  PULP_DATABASES__default__NAME: "{{ pulp_database_name }}"
  PULP_DATABASES__default__PORT: "{{ pulp_database_port }}"
  PULP_DATABASES__default__USER: "{{ pulp_database_user }}"
```

and then this database.yml only needs to override `pulp_database_*`

playbooks/deploy.yaml
```diff
@@ -30,30 +30,14 @@
...
-    foreman_db_password: "CHANGEME"
     foreman_oauth_consumer_key: abcdefghijklmnopqrstuvwxyz123456
     foreman_oauth_consumer_secret: abcdefghijklmnopqrstuvwxyz123456
     httpd_server_ca_certificate: "{{ ca_certificate }}"
     httpd_client_ca_certificate: "{{ ca_certificate }}"
     httpd_server_certificate: "{{ server_certificate }}"
     httpd_server_key: "{{ server_key }}"
-    pulp_db_password: "CHANGEME"
     pulp_content_origin: "https://{{ ansible_fqdn }}"
     postgresql_restarted_state: started
```
evgeni, line 40: Unrelated, but this can be dropped since we moved to the containerized PostgreSQL

ekohl (approved):

PR #140, opened 2025-04-29
Trying out tmt testing locally

ehelms:


ehelms:
Current state is throwing error:

```
        prepare task #5: Pull images on default-0
        how: shell
        name: Pull images
        overview: 1 script found
        fail: Command '/var/tmp/tmt/run-004/plans/example/tree/tmt-prepare-wrapper.sh-Pull-images-default-0' returned 1.
    finish
    
        guest: stopped
        guest: removed
        summary: 0 tasks completed

plan failed

The exception was caused by 1 earlier exceptions

Cause number 1:

    prepare step failed

    The exception was caused by 1 earlier exceptions

    Cause number 1:

        Command '/var/tmp/tmt/run-004/plans/example/tree/tmt-prepare-wrapper.sh-Pull-images-default-0' returned 1.

        stderr (17/17 lines)
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        Traceback (most recent call last):
          File "/var/tmp/tmt/run-004/plans/example/tree/.venv/bin/rop", line 33, in <module>
            sys.exit(load_entry_point('cli', 'console_scripts', 'rop')())
          File "/var/tmp/tmt/run-004/plans/example/tree/cli/__init__.py", line 49, in main
            obsah.main(cliargs=cliargs, application_config=application_config)
          File "/var/tmp/tmt/run-004/plans/example/tree/.venv/lib64/python3.9/site-packages/obsah/__init__.py", line 349, in main
            cfg_path = application_config.ansible_config_path()
          File "/var/tmp/tmt/run-004/plans/example/tree/.venv/lib64/python3.9/site-packages/obsah/__init__.py", line 234, in ansible_config_path
            return os.environ.get('OBSAH_ANSIBLE_CFG', os.path.join(cls.data_path(), 'ansible.cfg'))
          File "/var/tmp/tmt/run-004/plans/example/tree/cli/__init__.py", line 29, in data_path
            with resources.path(__name__, 'data') as fspath:
          File "/usr/lib64/python3.9/contextlib.py", line 119, in __enter__
            return next(self.gen)
          File "/usr/lib64/python3.9/importlib/resources.py", line 175, in _path_from_reader
            opener_reader = reader.open_resource(norm_resource)
          File "<frozen importlib._bootstrap_external>", line 1056, in open_resource
        IsADirectoryError: [Errno 21] Is a directory: '/var/tmp/tmt/run-004/plans/example/tree/cli/data'
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ekohl:
My idea was to release obsah, build it in the obal copr I have and add that. Then also do rpm packaging here. I noticed you can install packages from a directory and in PRs we can use packit.

In doing so I got to the point: do we vendor in the ansible playbooks we depend on?

ehelms:
> My idea was to release obsah, build it in the obal copr I have and add that. Then also do rpm packaging here. I noticed you can install packages from a directory and in PRs we can use packit.

Do you see this working when doing local development? I started with local development because I think that's the trickier workflow to get right.



> In doing so I got to the point: do we vendor in the ansible playbooks we depend on?

Did you mean do we vendor the Ansible collections?

ekohl:
> > My idea was to release obsah, build it in the obal copr I have and add that. Then also do rpm packaging here. I noticed you can install packages from a directory and in PRs we can use packit.
> 
> Do you see this working when doing local development? I started with local development because I think that's the trickier workflow to get right.

Not sure, but you're right that the local workflow is the harder one.

I'm starting to get the impression tmt isn't well suited to development like we can with forklift. As you already noticed, it cleans up aggressively and getting bits in isn't also so easy.

https://tmt.readthedocs.io/en/stable/plugins/prepare.html#install does describe how to install packages from a local directory, so that's why I went with the packaging route in my head. Then I was hoping you could skip that step if it was initiated by packit since that already installs the built RPM.

> Did you mean do we vendor the Ansible collections?

Of course I meant that. First review of the day is never a good one.

plans/example.fmf
```diff
@@ -0,0 +1,42 @@
...
+      - podman
+      - python3-psycopg2
+  - name: Setup virtualenv
+    how: shell
+    script: ./setup-environment
+  - name: Create local inventory
+    how: shell
+    script: ./local_inventory
+  - name: Run setup
+    how: ansible
```
ekohl, line 26: This looks more like shell
```suggestion
    how: shell
```

PR #141, opened 2025-05-01
Test external database

ehelms:
This is using Vagrant to provide a way to do external database testing. This is rough, and based on the CLI PR. I bumbled my way through some pytest updates for testing postgresql.

1. Please review the pytest changes to see if I am doing things correct.
2. The addition of "test fixtures" that are playbooks worked but I am still not crazy about how it works or the layout. Maybe it's fine.
3. My goal was to try this with [tmt](https://github.com/theforeman/foreman-quadlet/pull/140) but hit problems with local deployment.


ehelms:
I updated this for now to be based off the other prototype (https://github.com/theforeman/foreman-quadlet/pull/143) that includes development CLI, that allows building in this deployment, e.g.

```
./forge vms start --vms database
./forge remote-database
```

Where the last command will configure a `database` host to be external for development and testing. This starts to feel like what Forklift could also be, or Forklift could evolve into this repo or a mix of the two.

ehelms:
This is passing now! This includes a commit that re-works how the `test` command generates the ssh-config to base it off the Ansible inventory instead of Vagrant. This addresses the error that Vagrant throws if not all boxes are provisioned when requesting ssh-config and makes the test command able to work with more environments.

PR #142, opened 2025-05-02
Add RPM spec and packit config

ekohl:
This builds on https://github.com/theforeman/foreman-quadlet/pull/131 and adds an RPM spec file with packit configuration.

<del>Right now it won't include any of the galaxy modules we depend on and we need to have a discussion on how to resolve that. Will we vendor those into the Python package or list them separate in the RPM spec file?</del>

### DIFF
--- a/rop.spec
+++ b/rop.spec
@@ -1,0 +1,56 @@
+Name:      rop
+Version:   0.0.1
+Release:   0%{?dist}
+Summary:   Install Foreman using containers
+
+License:   GPL-2-only
+URL:       https://github.com/theforeman/foreman-quadlet
+Source:    https://github.com/theforeman/foreman-quadlet/releases/download/%{version}/%{name}-%{version}.tar.gz
+
+BuildArch: noarch
+Requires:  ansible-collection-ansible-posix
+Requires:  ansible-collection-community-crypto
+Requires:  ansible-collection-community-general
+Requires:  ansible-collection-community-postgresql
+Requires:  ansible-collection-containers-podman >= 1.14.0
+Requires:  ansible-collection-theforeman-foreman
+Requires:  ansible-collection-theforeman-operations
+Requires:  python3-obsah >= 1.1
+
+# These are needed on the target host, which is usually localhost
+Recommends:  podman
+Recommends:  python3-libsemanage
+Recommends:  python3-psycopg2
+
+# Tab completion is nice to have
+Suggests:  bash-completion
+Suggests:  python3-argcomplete
+
+%description
+Install Foreman using containers. They are deployed as podman quadlets.
+
+%prep
+%setup -q -n %{name}-%{version}
+
+%build
+cat > inventories/quadlet <<INVENTORY
+[quadlet]
+localhost ansible_connection=local
+INVENTORY
+
+sed -i '/^OBSAH_BASE=/ s|=.\+|=%{_datadir}/%{name}|' rop
+
+%install
+install -d -m0755 %{buildroot}%{_datadir}/%{name}
+install -d -m0755 %{buildroot}%{_bindir}
+
+cp -r inventories src %{buildroot}%{_datadir}/%{name}
+cp -r rop %{buildroot}%{_bindir}/%{name}
+
+
+%files
+%{_bindir}/rop
+%{_datadir}/%{name}
+
+
+%changelog
--- a/rop.spec
+++ b/rop.spec
@@ -14,7 +14,6 @@ Requires:  ansible-collection-community-general
 Requires:  ansible-collection-community-postgresql
 Requires:  ansible-collection-containers-podman >= 1.14.0
 Requires:  ansible-collection-theforeman-foreman
-Requires:  ansible-collection-theforeman-operations
 Requires:  python3-obsah >= 1.1
 
 # These are needed on the target host, which is usually localhost
--- a/.packit.yml
+++ b/.packit.yml
@@ -1,0 +1,13 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+specfile_path: rop.spec
+
+upstream_package_name: rop
+downstream_package_name: rop
+
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    targets:
+      - epel-9
--- a/rop
+++ b/rop
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 OBSAH_NAME=rop
-OBSAH_DATA=src
-OBSAH_INVENTORY=./inventories/
+OBSAH_BASE=.
+OBSAH_DATA=${OBSAH_BASE}/src
+OBSAH_INVENTORY=${OBSAH_BASE}/inventories
 export OBSAH_NAME OBSAH_DATA OBSAH_INVENTORY
 
 exec obsah "$@"


ekohl:
After a lot of experimentation I think I'm blocked until https://github.com/packit/packit-service/issues/2787 moves forward.

plans/installation.fmf
```diff
@@ -0,0 +1,23 @@
+summary: Installation
+prepare:
+  # Work around packit builds being installed unconditionally
+  - how: install
+    exclude:
+      - ".*"
+  - how: install
+    copr: ekohl/obal
+    package: []
```
thrix, line 9:
```suggestion
    package: []
    # workaround TFT-3360
    exclude: []
```

plans/installation.fmf
```diff
@@ -0,0 +1,23 @@
...
+  - how: shell
+    script: |
+      if [[ -n $PACKIT_COPR_PROJECT ]] ; then
+        for copr in $PACKIT_COPR_PROJECT ; do
+          sudo dnf copr enable -y $PACKIT_COPR_PROJECT
+        done
+      fi
+  - how: install
+    package: python3-cli
```
thrix, line 18:
```suggestion
    package: python3-cli
    # workaround TFT-3360
    exclude: []
```

ekohl:
This is now rebased on https://github.com/theforeman/foreman-quadlet/pull/151 and includes the workaround. It now fails on:
```
Command 'dnf copr -y enable -y ekohl/obal' returned 1.
Repository 'epel-9-x86_64' does not exist in project 'ekohl/obal'.
Available repositories: 'fedora-40-x86_64', 'fedora-rawhide-x86_64', 'centos-stream-9-x86_64', 'fedora-41-x86_64', 'fedora-42-x86_64'
```

Which means it gets to actually executing tmt and that's great.

While I could easily work around this by using Fedora now, it'll fail on the next step which is to install the built package.

ekohl:
It now times out on `Wait for Foreman tasks to be ready` so that's great progress.

ekohl:
The waiting for foreman tasks to be ready is still the problem. Not sure what's up:
```
TASK [foreman : Wait for Foreman tasks to be ready] ****************************
Friday 16 May 2025  19:45:54 +0000 (0:00:01.906)       0:04:38.696 ************ 
Shared connection to 3.139.91.192 closed.
```

ekohl:
Locally it works btw. I've submitted https://github.com/theforeman/foreman-quadlet/pull/152 and https://github.com/theforeman/foreman-quadlet/pull/153 since they're at least useful improvements outside of this PR.

I now wonder if there's a time limit. In total this run took just under 25 minutes.

evgeni:
#152 is merged now :)

ekohl:
Rebased and I think this is now ready for a review.

Some further discussion would be nice. As a first step: how can we get the same coverage as our current testing? It would involve running pytest, but also deploy a client machine.

Another bigger discussion that would follow: can it replace our vagrant usage?

rop.spec
```diff
@@ -0,0 +1,56 @@
...
+Source:    https://github.com/theforeman/foreman-quadlet/releases/download/%{version}/%{name}-%{version}.tar.gz
+
+BuildArch: noarch
+Requires:  ansible-collection-ansible-posix
+Requires:  ansible-collection-community-crypto
+Requires:  ansible-collection-community-general
+Requires:  ansible-collection-community-postgresql
+Requires:  ansible-collection-containers-podman >= 1.14.0
+Requires:  ansible-collection-theforeman-foreman
+Requires:  ansible-collection-theforeman-operations
```
ehelms, line 17: This dependency can be dropped with https://github.com/theforeman/foreman-quadlet/pull/155 which I think should go in first.

rop.spec
```diff
@@ -0,0 +1,56 @@
...
+localhost ansible_connection=local
+INVENTORY
+
+sed -i '/^OBSAH_BASE=/ s|=.\+|=%{_datadir}/%{name}|' rop
+
+%install
+install -d -m0755 %{buildroot}%{_datadir}/%{name}
+install -d -m0755 %{buildroot}%{_bindir}
+
+cp -r inventories src %{buildroot}%{_datadir}/%{name}
```
evgeni, line 47: IMHO the inventory should live in `/etc` as it's something users might want to edit (ext db at some point?)

rop.spec
```diff
@@ -0,0 +1,56 @@
...
+Install Foreman using containers. They are deployed as podman quadlets.
+
+%prep
+%setup -q -n %{name}-%{version}
+
+%build
+cat > inventories/quadlet <<INVENTORY
+[quadlet]
+localhost ansible_connection=local
+INVENTORY
```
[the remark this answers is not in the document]
ekohl: > or is it to put things into a `quadlet` group, which the one above does not as it uses the `quadlet` _host_ that we generate from vagrant?

Yes, this is it. Because our playbooks use `hosts: quadlet` and not `hosts: all`.

> if so, I think we should not ship `inventories/localhost` at all.

Possibly. I wasn't quite sure.

rop.spec
ekohl: I thought to keep it simple now, but we should discuss the multi host story.

If you take it to the extreme, would you also deploy new (content) proxies with it?

evgeni:
> Some further discussion would be nice. As a first step: how can we get the same coverage as our current testing? It would involve running pytest, but also deploy a client machine.

I think this involves even more questions (but they also don't need to be answered right now in this PR):
- can we run multiple test scenarios¹ in parallel? that's both a technical one (is it just a matter of defining multiple test plans?), but also a "are we allowed" in terms of resources (no idea how many ">8G" VMs testing farm can offer us without getting angry calls from other users or sponsors).
- the testing requires quite a lot of (pytest related) deps that are not to be present on the final, production, system, would it be possible to get a third, "tester", machine, so we don't pollute the two systems that represent the production env

> Another bigger discussion that would follow: can it replace our vagrant usage?

ideally, yes. the CI and the local dev env should be as close as possible.

also, vagrants fate is not looking too good these days, so I'd rather not rely on it.

¹: today it's the two certificate sources, but I recall "ext db" coming soon", and I am sure we can come up with more if we want to XD

rop.spec
ehelms: This can be removed now

rop.spec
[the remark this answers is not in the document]
ehelms: > > if so, I think we should not ship `inventories/localhost` at all.
> 
> Possibly. I wasn't quite sure.

I would agree we shouldn't ship the localhost one.

rop.spec
ehelms: I tend to agree on putting it in etc. Or we ship this default one, and we support `/etc` as well where `/etc` has precedence. This could lead to unexpected behaviour though if we have multiple potential inventory sources of truth.

ehelms:
> > Another bigger discussion that would follow: can it replace our vagrant usage?
> 
> ideally, yes. the CI and the local dev env should be as close as possible.
> 
> also, vagrants fate is not looking too good these days, so I'd rather not rely on it.

I'd almost appreciate keeping both for a time. This might help re-enforce our keeping the provisioned machines separate from the actions we take on them mantra.



> * the testing requires quite a lot of (pytest related) deps that are not to be present on the final, production, system, would it be possible to get a third, "tester", machine, so we don't pollute the two systems that represent the production env

Could we put our pytests into a container? Then it could run on separate machine or same machine without polluting as well as have dependencies already in it that we need. Maybe that's overkill but it feels relevant.

evgeni:
keeping vagrant and tmt for a while sounds fair, yeah.

container you say? that's an interesting thought. given we already require a working container runtime to be present, that seems like a cool idea.

ekohl:
There is [multihost testing](https://tmt.readthedocs.io/en/stable/guide.html#multihost-testing) and I think we can run pytest from there. I'd tempted to rely on [smoker](https://github.com/theforeman/smoker) for that since it's already designed and used in exactly that use case in our current pipelines.

Then we can convert the simple checks like "is service X running" on the container host itself to [beakerlib](https://beakerlib.readthedocs.io/en/latest/) to avoid dependencies.

evgeni:
Not a fan of beakerlib, but that's a discussion for another day.

ekohl:
Any other suggestions for very simple verification?

rop.spec
```diff
@@ -0,0 +1,56 @@
...
+License:   GPL-2-only
+URL:       https://github.com/theforeman/foreman-quadlet
+Source:    https://github.com/theforeman/foreman-quadlet/releases/download/%{version}/%{name}-%{version}.tar.gz
+
+BuildArch: noarch
+Requires:  ansible-collection-ansible-posix
+Requires:  ansible-collection-community-crypto
+Requires:  ansible-collection-community-general
+Requires:  ansible-collection-community-postgresql
+Requires:  ansible-collection-containers-podman >= 1.14.0
```
evgeni, line 15: sidequest: neither of these is in EL-proper (they are in EPEL, but that doesn't really help us long term).

- should we vendor them at build time (like we do with puppet modules in foreman-installer)?
- I'd prefer not to package them "properly", as that would make them available *outside* rop, and people might think they can use them for other things and claim support

ekohl: I had similar concerns and I'm not sure yet. They just happened to be in EPEL and allowed me to ignore the problem at first.

evgeni: Absolutely, it's not a problem that needs solving right now.

rop.spec
evgeni: new content proxies, if they are deployed by a new (non existent) installation method, yes

new content proxies, deployed by the classical method… maybe

ehelms:
I'd be happy to merge in the packit and RPM part. The `tmt` part with the unknowns with local make me want to do more playing around. What about splitting this into packit and tmt PRs and we merge the packit part?

evgeni:
> I'd be happy to merge in the packit and RPM part. The `tmt` part with the unknowns with local make me want to do more playing around. What about splitting this into packit and tmt PRs and we merge the packit part?

That sounds good to me!

rop.spec
evgeni: Moved to https://github.com/theforeman/foreman-quadlet/issues/157

rop.spec
evgeni:
```suggestion
```

please :)

rop.spec
```diff
@@ -0,0 +1,56 @@
...
+%build
+cat > inventories/quadlet <<INVENTORY
+[quadlet]
+localhost ansible_connection=local
+INVENTORY
+
+sed -i '/^OBSAH_BASE=/ s|=.\+|=%{_datadir}/%{name}|' rop
+
+%install
+install -d -m0755 %{buildroot}%{_datadir}/%{name}
```
evgeni, line 44:
```suggestion
install -d -m0755 %{buildroot}%{_datadir}/%{name}
install -d -m0755 %{buildroot}%{%_sysconfdir}/%{name}
```

rop.spec
evgeni:
```suggestion
cp -r src %{buildroot}%{_datadir}/%{name}
cp inventories/quadlet %{buildroot}%{_sysconfdir}/%{name}/inventory
```

rop.spec
```diff
@@ -0,0 +1,56 @@
...
+%prep
+%setup -q -n %{name}-%{version}
+
+%build
+cat > inventories/quadlet <<INVENTORY
+[quadlet]
+localhost ansible_connection=local
+INVENTORY
+
+sed -i '/^OBSAH_BASE=/ s|=.\+|=%{_datadir}/%{name}|' rop
```
evgeni, line 41:
```suggestion
sed -i '/^OBSAH_BASE=/ s|=.\+|=%{_datadir}/%{name}|' rop
sed -i '/^OBSAH_INVENTORY=/ s|=.\+|=%{_sysconfdir}/%{name}/inventory|' rop
```

evgeni:
@ekohl we can also address the inventory placement in another PR and merge this as is, depends on your time etc.

ekohl:
> @ekohl we can also address the inventory placement in another PR and merge this as is, depends on your time etc.

I'd prefer to split that off and have a proper discussion on how we see the inventory

evgeni:
I am pretty sure the `installer certs` failure is unrelated.

evgeni (approved):

evgeni:
I've opened https://github.com/theforeman/foreman-quadlet/pull/166 and we can continue the inventory discussion there.

PR #143, opened 2025-05-07
Add a production and development CLI

ehelms:
This is an alternative to https://github.com/theforeman/foreman-quadlet/pull/131 where I use the method of a simple bash script outlined by https://github.com/theforeman/foreman-quadlet/pull/131#issuecomment-2854437644.

I have also used this method to create the idea of a "development CLI" instead of bash scripts to show how we can encapsulate the development tasks into a similar CLI style for developers. This is also a way we could onboard additional CLIs in production if we wanted to have separate commands rather than actions.

This only adds `./forge start-vms` but if we like the idea, I will adjust the other scripts also.

This aims for the following layout:

- `src` - production code
- `development` - development and testing tools
- `docs` - documentation
- `tests` - directory for tests and fixtures


development/playbooks/start-vms/start-vms.yaml
```diff
@@ -0,0 +1,21 @@
...
+  tasks:
+    - name: Run vagrant up
+      ansible.builtin.command:
+        cmd: "vagrant up {{ vms | default() }}"
+      args:
+        chdir: "{{ inventory_dir }}/../"
+
+    - name: Create local_vagrant inventory
+      ansible.builtin.command:
+        cmd: ../../scripts/vagrant.py --yaml
```
ehelms, line 15: This relative pathing feels kinda ugly but I couldn't find a good variable to use.

development/playbooks/start-vms/start-vms.yaml
```diff
@@ -0,0 +1,21 @@
...
+- name: Start VMs for development and testing
+  gather_facts: false
+  hosts:
+    - localhost
+  tasks:
+    - name: Run vagrant up
+      ansible.builtin.command:
+        cmd: "vagrant up {{ vms | default() }}"
+      args:
+        chdir: "{{ inventory_dir }}/../"
```
ehelms, line 11: This relative pathing feels kinda ugly but I couldn't find a good variable to use.

rop
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+OBSAH_NAME=rop
+OBSAH_DATA=src
+OBSAH_INVENTORY=./inventories/
+export OBSAH_NAME OBSAH_DATA OBSAH_INVENTORY
```
ehelms, line 6: We'll need to disable extra vars and add to this list, I opened https://github.com/theforeman/obsah/pull/76 to do so.

pyproject.toml
```diff
@@ -0,0 +1,48 @@
+[build-system]
```
ehelms, line 1: I went ahead and swapped from setup.yp to pyproject.toml -- is this still needed if we go this route?

forge
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
ehelms, line 1: This is the development CLI wrapper. To make the ring, you need a forge.

pyproject.toml
evgeni: I don't think so, no.
We need something that pulls in the obsah dep etc, but otherwise…

pyproject.toml
```diff
@@ -0,0 +1,48 @@
...
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+backend-path = ["."]
```
evgeni, line 4:
```suggestion
```

otherwise it fails for me

.github/workflows/test.yml
```diff
@@ -59,15 +60,13 @@ jobs:
...
         run: |
-          ./start-vms
-      - name: Run setup
-        run: |
-          ansible-playbook playbooks/setup.yaml
+          ./forge start-vms
       - name: Run image pull
         run: |
-          ansible-playbook playbooks/images.yaml
+          rop pull-images
```
evgeni, line 66:
```suggestion
          ./rop pull-images
```

development/playbooks/start-vms/start-vms.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+    - name: Create local_vagrant inventory
+      ansible.builtin.command:
+        cmd: ../../scripts/vagrant.py --yaml
+      register: inventory_output
+
+    - name: Write inventory
+      ansible.builtin.copy:
+        dest: "{{ inventory_dir }}/local_vagrant"
+        content: "{{ inventory_output.stdout }} + \n"
```
evgeni, line 21:
```suggestion
        content: "{{ inventory_output.stdout }}"
```

.github/workflows/test.yml
```diff
@@ -29,7 +29,8 @@ jobs:
       - name: Run ansible-lint
         uses: ansible/ansible-lint@main
         with:
-          requirements_file: requirements.yml
+          requirements_file: ../../requirements.yml
+          working_directory: cli/data
```
evgeni, line 33:
```suggestion
          requirements_file: ../requirements.yml
          working_directory: src
```

.github/workflows/test.yml
```diff
@@ -59,15 +60,13 @@ jobs:
...
-          ansible-playbook playbooks/images.yaml
+          ./rop pull-images
       - name: Run deployment
         run: |
-          ansible-playbook playbooks/deploy.yaml -e certificate_source=${{ matrix.certificate_source }}
+          ./rop deploy --certificate-source=${{ matrix.certificate_source }}
       - name: Run tests
-        run: ./run_tests --certificate-source=${{ matrix.certificate_source }}
+        run: |
+          ./run_tests --certificate-source=${{ matrix.certificate_source }}
```
evgeni, line 72: this should be `./forge test`, right? :)

.gitignore
```diff
@@ -39,6 +39,9 @@ packer/build.*
 .env
 .venv
 .python-version
+build/
+*.egg-info/
+dist/
```
ekohl, line 44: If we don't have a Python package then we won't create these files

.github/workflows/test.yml
```diff
@@ -29,7 +29,8 @@ jobs:
       - name: Run ansible-lint
         uses: ansible/ansible-lint@main
         with:
-          requirements_file: requirements.yml
+          requirements_file: ../requirements.yml
```
ekohl, line 32: I was wondering if we should move `requirements.yml` into `src` instead so I looked up how a collection is structured. https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_structure.html tells me we should have a `galaxy.yml` in the same directory as `roles` and `playbooks`.

We can ignore it for now but where I'd like to head to is that we can point obsah at a collection and it can just do the right thing.

pyproject.toml
ekohl, line 1: We can drop this

src/playbooks/deploy/deploy.yaml
```diff
@@ -41,6 +41,19 @@
...
         name: debug_tools
+    - name: Install podman
+      ansible.builtin.package:
+        name:
+          - podman
+    - name: Install other dependencies
+      ansible.builtin.package:
+        name:
+          - bash-completion
+          - nmap
```
ekohl, line 52: This is test- only

src/playbooks/deploy/deploy.yaml
```diff
@@ -41,6 +41,19 @@
...
+          - python3-cryptography
+          - python3-libsemanage
+          - python3-psycopg2
+          - python3-requests
```
ekohl, line 56: This is test-only

requirements.txt
```diff
@@ -4,3 +4,4 @@ paramiko
 pytest-testinfra
 pytest-durations
 pyyaml
+-e .
```
ekohl, line 7:
```suggestion
obsah >= 1.1
```

And it should be sorted

ekohl (commented):
I'm trying to wrap my head around what we all have.

* (semi-)transformation of our Ansible data into a collection for production
* Rewrite of development/testing tooling into a collection
* Add wrapper scripts

Did I get that right?

ehelms:
> 'm trying to wrap my head around what we all have.
> 
> * (semi-)transformation of our Ansible data into a collection for production
> * Rewrite of development/testing tooling into a collection
> * Add wrapper scripts
> 
> Did I get that right?

I think the "semi" is correct, as this keeps that option available without explicitly defining one yet. And as you note, the development parts are also taking the same shape now. Turning our development and testing actions into a developer focused CLI using wrapper scripts.

src/playbooks/deploy/deploy.yaml
ehelms: Nope! And you keep tricking me into thinking that :stuck_out_tongue: 

This is used by the smart_proxy module for registration:

```
TASK [foreman : Configure Foreman Proxy] ***************************************
Wednesday 07 May 2025  16:05:30 +0000 (0:00:22.288)       0:05:09.651 ********* 
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'requests'
fatal: [quadlet]: FAILED! => {
    "changed": false
}

MSG:

Failed to import the required Python library (requests) on quadlet.example.com's Python /usr/bin/python3.9. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter
```

ehelms:
The same test keeps failing, but doesn't fail for me locally:

```
=========================== short test summary info ============================
FAILED tests/client_test.py::test_foreman_content_view - AssertionError: assert False
 +  where False = <package bear>.is_installed
 +    where <package bear> = <class 'testinfra.modules.base.RpmPackage'>('bear')
 +      where <class 'testinfra.modules.base.RpmPackage'> = <testinfra.host.Host paramiko://client>.package
======= 1 failed, 68 passed, 2 xfailed, 9 warnings in 237.13s (0:03:57) ========
```

Vagrantfile
```diff
@@ -5,6 +5,10 @@ Vagrant.configure("2") do |config|
   config.vm.box = "centos/stream9"
   config.vm.synced_folder ".", "/vagrant"
 
+  config.vm.provision "ansible" do |ansible|
+    ansible.playbook = "development/playbooks/etc_host.yml"
```
evgeni, line 9:
```suggestion
    ansible.playbook = "development/playbooks/etc_host.yml"
    ansible.compatibility_mode = "2.0"
```

this fixes the network/dns issue

evgeni: The problem is that Vagrant does not properly detect the version of Ansible, and falls back to Ansible 1.8 compatibility mode. However, in that mode, the inventory ends up with only a single host and using 127.0.0.1 IPs.

ehelms: How'd you figure that out?

evgeni: I used https://github.com/marketplace/actions/debugging-with-ssh to ssh into the GitHub Actions runner and then ran `vagrant destroy client` and `vagrant up client`, saw the "Using Ansible 1.8 compat" warning, and was enlightened.

development/playbooks/test/metadata.obsah.yaml
```diff
@@ -0,0 +1,7 @@
+---
+help: |
+  Execute tests
+
+variables:
+  args:
```
evgeni, line 6: Ansible is unhappy:
```
Warning: : Found variable using reserved name: args
```

Probably better to call this `pytestargs` or something

evgeni (approved):
I am sure we can find enough things to improve, but let's get the current state in and iterate later.

Issue #92, opened 2025-02-27
Secret updates in podman need to cause a service restart


ehelms:
For example, if certificates change, this needs to trigger the service to restart.

ehelms:
@evgeni Is the correct pattern for this to use a handler that gets notified by the secret tasks?

evgeni:
In general, yes.
There are two "problems" with that (well, gotchas, really):
- handlers by default run at the end of the play, that means that if you updated a secret and want the app to use it "in the next step", you need to "flush handlers" yourself
- handlers often produce "needless" restarts on the initial deployment, as the service is started after a secret has been defined the first time and thus already uses it.

ehelms:
> * handlers by default run at the end of the play, that means that if you updated a secret and want the app to use it "in the next step", you need to "flush handlers" yourself

What defines the end of a play? The whole playbook? or the whole role?

evgeni:
Playbook (or maybe a play in a playbook, not sure 100%), but certainly not role

PR #147, opened 2025-05-13
Add foreman service handler for certificate changes

ehelms:
Taking an initial tactic on using handler to handle changes to certificate secrets. I know there are more secrets that should trigger a change. I am keeping this focused to start with as I think the "things changed and services need to restart" paradigm is the part to get right to start with.

@evgeni I recall you had some reservations about handlers.

I am also wondering what if anything happens with the initial service start we have and the "wait for" task in a handler based setup:

```
- name: Start the Foreman Service
  ansible.builtin.systemd:
    name: foreman
    enabled: true
    state: started

- name: Wait for Foreman service to be accessible
  ansible.builtin.uri:
    url: 'http://{{ ansible_hostname }}:3000/api/v2/ping'
  until: foreman_status.status == 200
  retries: 60
  delay: 5
  register: foreman_status
```

### DIFF
--- a/src/roles/foreman/handlers/main.yaml
+++ b/src/roles/foreman/handlers/main.yaml
@@ -1,0 +1,5 @@
+---
+- name: Restart foreman service
+  ansible.builtin.systemd:
+    name: foreman
+    state: restarted
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -33,18 +33,24 @@
     name: foreman-ca-cert
     path: "{{ foreman_ca_certificate }}"
     state: present
+  notify:
+    - Restart foreman service
 
 - name: Create the podman secret for Foreman client certificate
   containers.podman.podman_secret:
     state: present
     name: foreman-client-cert
     path: "{{ foreman_client_certificate }}"
+  notify:
+    - Restart foreman service
 
 - name: Create the podman secret for Foreman client key
   containers.podman.podman_secret:
     state: present
     name: foreman-client-key
     path: "{{ foreman_client_key }}"
+  notify:
+    - Restart foreman service
 
 - name: Deploy Foreman Container
   containers.podman.podman_container:


ekohl:
> I am also wondering what if anything happens with the initial service start we have and the "wait for" task in a handler based setup:

I still think https://github.com/theforeman/foreman-quadlet/pull/118 is a better solution for it. Then just drop the whole waiting.

evgeni:
I had to use GitHub search (it's better than Jira…) and found my comment in https://github.com/theforeman/foreman-quadlet/issues/92#issuecomment-2724820136, which still applies.

ehelms:
> I had to use GitHub search (it's better than Jira…) and found my comment in [#92 (comment)](https://github.com/theforeman/foreman-quadlet/issues/92#issuecomment-2724820136), which still applies.

Quoting that here:

> In general, yes.
There are two "problems" with that (well, gotchas, really):

handlers by default run at the end of the play, that means that if you updated a secret and want the app to use it "in the next step", you need to "flush handlers" yourself
handlers often produce "needless" restarts on the initial deployment, as the service is started after a secret has been defined the first time and thus already uses it.


That does not read to me as we shouldn't use them, just things to be mindful of as we build things out. The alternative to handlers would be a homegrown mechanism?

ekohl:
> That does not read to me as we shouldn't use them, just things to be mindful of as we build things out. The alternative to handlers would be a homegrown mechanism?

I had been thinking about using systemd to notify if something changes, but I don't think you can trigger that. I know systemd can trigger on files changing (https://www.freedesktop.org/software/systemd/man/latest/systemd.path.html) but that isn't a secret.

I would like a solution where you can easily swap out certificates without using the whole installer, but I struggle to see how to do that easily. For example, if we would adopt [certbot](https://certbot.eff.org/) or a similar tool to request certificates from a CA then they take ownership of rotation. They have hooks to restart services. Ansible doesn't feel like it's very good at having those triggers so I'm looking at simpler mechanisms.

PR #149, opened 2025-05-15
Add logging parameters support

ehelms:
This is the set of parameters from our docs around logging that are mentioned. I'm not 100% on whether we should include all of these so consider this a conversation starter.

This also raises a question -- what's the best strategy to test these kind of parameters? These take an input, change a settings file and then do have an affect on the log output.

### DIFF
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -60,6 +60,23 @@ There are multiple use cases from the users perspective that dictate what parame
 | foreman-installer Parameter | Description | Reason |
 | --------------------------- | ----------- | ------ |
 
+#### Logging
+
+##### Mapped
+
+| Parameter | Description | foreman-installer Parameter |
+| ----------| ----------- | --------------------------- |
+| `--foreman-logging-level` | Allows setting the level of the Foreman logs | `--foreman-logging-level` |
+| `--foreman-loggers` | Allows enabling specific loggers in Foreman | `--foreman-loggers` |
+| `--candlepin-loggers` | Allows enabling specific loggers in Candlepin | `--candlepin-loggers` |
+
+##### Unmapped
+
+| foreman-installer Parameter | Description | Reason |
+| --------------------------- | ----------- | ------ |
+| `--foreman-logging-type` | The type of logging - syslog, file, journald | For containers logging should be to stdout |
+| `--foreman-logging-layout` | Allow defining a different style of structured output for Foreman logs | For now this should be hardcoded to a single style |
+
 #### Undetermined
 
 | foreman-installer Parameter | Description | Module | Puppet Parameter | Keep |
@@ -69,7 +86,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-foreman-service-puma-threads-max` | | foreman | foreman_service_puma_threads_max | `--foreman-puma-threads-max` |
 | `--foreman-dynflow-worker-instances` | | foreman | dynflow_worker_instances |
 | `--foreman-dynflow-worker-concurrency` | | foreman | dynflow_worker_concurrency |
-| `--foreman-loggers` | | foreman | loggers |
 | `--foreman-plugin-tasks-cron-line` | | foreman::plugin::tasks | cron_line |
 | `--foreman-plugin-tasks-automatic-cleanup` | | foreman::plugin::tasks | automatic_cleanup |
 | `--tuning` | Sets the tuning profile | foreman-installer | |
@@ -90,9 +106,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-keycloak` | | |
 | `--foreman-keycloak-app-name` | | |
 | `--foreman-keycloak-realm` | | |
-| `--foreman-logging-layout` | | |
-| `--foreman-logging-level` | | |
-| `--foreman-logging-type` | | |
 | `--foreman-oauth-map-users` | | |
 | `--foreman-plugin-remote-execution-cockpit-ensure` | | |
 | `--foreman-telemetry-prometheus-enabled` | | |
--- a/src/playbooks/deploy/metadata.obsah.yaml
+++ b/src/playbooks/deploy/metadata.obsah.yaml
@@ -47,6 +47,18 @@ variables:
     help: User of the Pulp database.
   pulp_database_password:
     help: Password for the Pulp database.
+  foreman_logging_level:
+    help: Set the log level for Foreman.
+    choices:
+      - fatal
+      - error
+      - warn
+      - info
+      - debug
+  foreman_loggers:
+    help: Enable specific loggers for debugging Foreman.
+  candlepin_loggers:
+    help: Set the log level for Candlepin loggers.
 constraints:
   required_together:
     - [database_ssl_mode, database_ssl_ca]
--- a/src/roles/candlepin/templates/candlepin.conf.j2
+++ b/src/roles/candlepin/templates/candlepin.conf.j2
@@ -38,3 +38,9 @@ org.quartz.dataSource.myDS.user={{ candlepin_database_user }}
 org.quartz.dataSource.myDS.password={{ candlepin_database_password }}
 org.quartz.dataSource.myDS.maxConnections=5
 org.quartz.dataSource.myDS.URL=jdbc:postgresql://{{ candlepin_database_host }}:{{ candlepin_database_port }}/{{ candlepin_database_name }}?sslmode={{ candlepin_database_ssl_mode }}{% if candlepin_database_ssl_ca is defined %}&sslrootcert={{ candlepin_database_ssl_ca }}{% endif %}
+
+{% if candlepin_loggers %}
+  {% for logger, log_level in candlepin_loggers.items() %}
+log4j.logger.{{ logger }}= {{ log_level }}
+  {% endfor %}
+{% endif %}
--- a/src/roles/candlepin/defaults/main.yml
+++ b/src/roles/candlepin/defaults/main.yml
@@ -22,3 +22,5 @@ candlepin_database_ssl_mode: disable
 candlepin_database_ssl_ca:
 candlepin_database_ssl_cert:
 candlepin_database_ssl_key:
+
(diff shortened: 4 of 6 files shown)


docs/parameters.md
```diff
@@ -60,6 +60,23 @@ There are multiple use cases from the users perspective that dictate what parame
...
 | --------------------------- | ----------- | ------ |
 
+#### Logging
+
+##### Mapped
+
+| Parameter | Description | foreman-installer Parameter |
+| ----------| ----------- | --------------------------- |
+| `--foreman-logging-level` | Allows setting the level of the Foreman logs | `--foreman-logging-level` |
+| `--foreman-logging-layout` | Allow defining a different style of structured output for Foreman logs | `--foreman-logging-layout` |
```
ekohl, line 70: In our current module we derive the default layout on the logging type (https://github.com/theforeman/puppet-foreman/blob/86e73d267bb6e05eb607ae8fedc00a8ffa132334/manifests/config.pp#L38-L44). My feeling is that this should be good enough for most users. Not sure we need to expose this.

docs/parameters.md
```diff
@@ -60,6 +60,23 @@ There are multiple use cases from the users perspective that dictate what parame
...
+| `--foreman-loggers` | Allows enabling specific loggers in Foreman | `--foreman-loggers` |
+| `--candlepin-loggers` | Allows enabling specific loggers in Candlepin | `--candlepin-loggers` |
+
+##### Unmapped
+
+| foreman-installer Parameter | Description | Reason |
+| --------------------------- | ----------- | ------ |
+| `--foreman-logging-type` | The type of logging - syslog, file, journald | For containers logging should be to stdout |
```
ekohl, line 78: I agree this may not need to be mapped but structured logging to the journal is nice. If we can get that then that would be my preference. That way you can use the current `journalctl REQUEST_ID=$id` to get everything for a single request.

src/roles/candlepin/templates/candlepin.conf.j2
```diff
@@ -38,3 +38,9 @@ org.quartz.dataSource.myDS.user={{ candlepin_database_user }}
 org.quartz.dataSource.myDS.password={{ candlepin_database_password }}
 org.quartz.dataSource.myDS.maxConnections=5
 org.quartz.dataSource.myDS.URL=jdbc:postgresql://{{ candlepin_database_host }}:{{ candlepin_database_port }}/{{ candlepin_database_name }}?sslmode={{ candlepin_database_ssl_mode }}{% if candlepin_database_ssl_ca is defined %}&sslrootcert={{ candlepin_database_ssl_ca }}{% endif %}
+
+{% if candlepin_loggers %}
+  {% for logger, log_level in candlepin_loggers.items() %}
+log4j.logger.{{ logger }}= {{ log_level }}
```
ekohl, line 44: No idea if it matters
```suggestion
log4j.logger.{{ logger }}={{ log_level }}
```

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -47,6 +47,25 @@ variables:
...
+  foreman_logging_layout:
+    help: Set the structured layout type for Foreman logs.
+    choices:
+      - pattern
+      - multiline_pattern
+      - multiline_request_pattern
+      - json
+  foreman_loggers:
+    help: Enable specific loggers for debugging Foreman.
+  candlepin_loggers:
```
ekohl, line 67: Can we somehow tell obsah this is a dict? In kafo you can specify `--candlepin-loggers logger1:value1 --candlepin-loggers logger2:value2` and that results in a single hash with 2 keys.

Same above for `foreman_loggers`

docs/parameters.md
ehelms: You are thinking we should just hard-code this to `multiline_request_pattern` and not give users the opportunity to change it? I was figuring `json` might be useful for users that ship logs to a centralized logging system.

docs/parameters.md
ehelms: Isn't the idea that we log to stdout and that gets captured and put into the journal automagically?

src/playbooks/deploy/metadata.obsah.yaml
ehelms: I wondered the same thing. And wanted to see what @evgeni thought here on that as a type.

docs/parameters.md
ekohl: Then we may also need to makevthe type configureable. Something to investigate

docs/parameters.md
ehelms: My understanding is that podman has different log drivers. The default for quadlet being `journald` that stdout is sent to and thus you get this example output in the journal:

```
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [I|app|c73b53ae] Started GET "/rhsm/status" for 192.168.121.102 at 2025-05-16 01:05:02 +0000
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [I|app|c73b53ae] Processing by Katello::Api::Rhsm::CandlepinProxiesController#server_status as JSON
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [D|kat|c73b53ae] Resource GET request: /candlepin/status
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [D|kat|c73b53ae] Headers: null
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [D|kat|c73b53ae] Body: null
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [D|kat|c73b53ae] Candlepin request ddec0dd3-9fd9-46d3-832a-b86ef2a45741 returned with code 200
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [D|kat|c73b53ae] Processing response: 200
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [D|kat|c73b53ae] {"mode":"NORMAL","modeReason":null,"modeChangeTime":null,"result":true,"version":"4.4.14","release":"1","standalone":true,"timeUTC":"2025-05-16T01:05:02+0000","rulesSource":"default","rulesVersion":"5.44","managerCapabilities":["instance_multiplier","derived_product","vcpu","cert_v3","hypervisors_heartbeat","remove_by_pool_id","syspurpose","storage_band","cores","multi_environment","hypervisors_async","org_level_content_access","typed_environments","guest_limit","ram","batch_bind"],"keycloakRealm":null,"keycloakAuthUrl":null,"keycloakResource":null,"deviceAuthRealm":null,"deviceAuthUrl":null,"deviceAuthClientId":null,"deviceAuthScope":null}
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [I|app|c73b53ae] Completed 200 OK in 34ms (Views: 0.2ms | ActiveRecord: 14.3ms | Allocations: 4757)
May 16 01:05:02 quadlet.example.com foreman[194952]: 2025-05-16T01:05:02 [D|app|c73b53ae] With body: {"mode":"NORMAL","modeReason":null,"modeChangeTime":null,"result":true,"version":"4.4.14","release":"1","standalone":true,"timeUTC":"2025-05-16T01:05:02+0000","rulesSource":"default","rulesVersion":"5.44","managerCapabilities":["instance_multiplier","derived_product","vcpu","cert_v3","hypervisors_heartbeat","remove_by_pool_id","syspurpose","storage_band","cores","multi_environment","hypervisors_async","org_level_content_access","typed_environments","guest_limit","ram","batch_bind","combined_reporting"],"keycloakRealm":null,"keycloakAuthUrl":null,"keycloakResource":null,"deviceAuthRealm":null,"deviceAuthUrl":null,"deviceAuthClientId":null,"deviceAuthScope":null}
May 16 01:05:02 quadlet.example.com foreman[194952]:  c73b53ae | 
```

In Foreman there are two configuration values: where to send logs and how to structure them. The [logging config file](https://github.com/theforeman/foreman/blob/develop/config/logging.yaml) outlines much of this and in truth I don't think we use most of this. In RPM based installs, the default ends up as:

```
:logging:
  :level: info
  :production:
    :type: file
    :layout: multiline_request_pattern
```

src/playbooks/deploy/metadata.obsah.yaml
evgeni: What's the end data structure you expect here?
```
candlepin_loggers:
  logger1: value1
  logger2: value2
```
That?

With "regular ansible", you'd just pass that as JSON, so `--candlepin-loggers '{"logger1":"value1","logger2":"value2"}'`. Yes, not strictly userfriendly, I know.

Kafo (judging by Ewouds comment) does allow multiple params that then get joined to one big dict?

I am sure we can implement that, just not sure it's an actual improvement for the user (I guess for the "one entry case, the syntax is nicer, but once you have more… idk)

docs/parameters.md
ehelms: For now, I am going to put this into the unmapped category and leave it hard-coded. We can address stdout vs journald in the future and do so via https://github.com/theforeman/foreman-quadlet/issues/21

docs/parameters.md
ehelms: I think we can discuss journald vs stdout in https://github.com/theforeman/foreman-quadlet/issues/21 and leave this parameter unmapped for now.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -47,6 +47,18 @@ variables:
...
+  foreman_logging_level:
+    help: Set the log level for Foreman.
+    choices:
+      - fatal
+      - error
+      - warn
+      - info
+      - debug
+  foreman_loggers:
+    help: Enable specific loggers for debugging Foreman.
```
stejskalleos, line 59: Should we list the choices the same way as we do for the `foreman_logging_level`?

ehelms:
I am going to close this to be revisited more systematically at a later time.

PR #153, opened 2025-05-17
Stop setting a fixed password for Pulp admin user

ekohl:
This was needed before we had cert auth.

Fixes: 2ad1ca5343aa ("configure cert auth for pulp")

### DIFF
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -164,8 +164,7 @@
     enabled: true
     state: started
 
-# Only needed until we have cert auth configured
-- name: Set Pulp admin password
+- name: Ensure Pulp admin user exists
   containers.podman.podman_container_exec:
     name: "{{ pulp_api_container_name }}"
-    command: pulpcore-manager reset-admin-password --password CHANGEME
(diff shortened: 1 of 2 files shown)


ehelms:
Some of the tests relied on Pulp admin to be setup and will need to be updated.

ekohl:
I think we still need an admin user entry, just with a random password. We do the same in the current installer:
https://github.com/theforeman/puppet-pulpcore/blob/38bf668e76191aa314cbe8042f7db866ed3f615c/manifests/database.pp#L46-L50

evgeni:
> I think we still need an admin user entry, just with a random password. We do the same in the current installer: https://github.com/theforeman/puppet-pulpcore/blob/38bf668e76191aa314cbe8042f7db866ed3f615c/manifests/database.pp#L46-L50

I think this is correct.

The only test we have for the password is already `xfail` (and can be removed, I guess):
https://github.com/theforeman/foreman-quadlet/blob/338fd536799d2c222d677d6598c9fb509e34b542/tests/pulp_test.py#L62-L66

But we need the `admin` user, and it doesn't exist until you set some password for it, as far as I can tell

evgeni (approved):

PR #154, opened 2025-05-19
Add katello-certs-check as foreman-certificate-check

ehelms:
Add `katello-certs-check` directly to the repository. This renames it to `foreman-certificate-check` to be more explicit and focus on Foreman as the project. This adds the same test suite.

Currently the test suite fails for me, and I'm not clear why. Something about how subprocess is running the script I think and how it cannot see the cert files.


ehelms:
Update: The tests are passing. Turns out the script was fine, it was the tests being translated into pytest that made a few incorrect assumptions.

ekohl (changes requested):
I'd at least like to get rid of the branding.

This currently installs the file via Ansible, but I've been thinking if that's the right way. Would it be better to ship this in the installer itself? For example, with the RPM.

ekohl:
> How would you handle this for local development and testing?

Honestly, good question that I don't know the answer to now. I do think that random tools placing files in `/usr/bin` is a bad practice though because IMHO it's reserved for the package manager.

In this particular case it's not going to conflict to existing installations with `foreman-installer` so for upgrades that's not going to be a problem. However, long term I can imagine we want to support running on image mode EL and then writing to `/usr/bin` isn't going to be valid.

ehelms:
I think supporting image mode will require enough investigation and thinking through the process that we shouldn't optimize for it right now. 

Would you prefer we use `/usr/local/bin` ?

ehelms:
> Would you prefer we use `/usr/local/bin` ?

@ekohl Can you answer this question for me?

evgeni (approved):

PR #155, opened 2025-05-19
Move generating installer based certs to development

ehelms:
This removes the production requirement on the foreman-installer and makes testing against installer based certificates a development operation. This helps focus the production CLI and moves the existence of installer certificates to something we'd expect to be present but not generating during installation itself.

This does not yet resolve the how to manage these certificates after that point but it does help create a clearer distinction at present.


src/playbooks/deploy/deploy.yaml
```diff
@@ -56,8 +56,6 @@
   roles:
     - role: certificates
       when: "certificate_source == 'default'"
```
ekohl, line 58: This is the only remaining usage of `certificate_source` so with the focus on simplification I'd suggest to drop it.

What's the point of installer provided certificates at this point?

ehelms: This is also used here -- https://github.com/theforeman/foreman-quadlet/pull/155/files#diff-2fc200de05abc2d83ea118cae7634c213de69b3e8640146a18644935bfce14f2R7 

And the use of it in this spot is intentional even without installer based certs.

We have three scenarios:

 1. User provides no certificates and default certs are generated by this role
 2. User has upgraded from an rpm-based installation which has certificates that were generated by foreman-installer and thus need to be the inputs to the installer. There is a [vars file](https://github.com/theforeman/foreman-quadlet/blob/master/src/vars/installer_certificates.yml) that captures these.
 3. The user provides certificates for all things and needed to provide them as inputs.
 
 How are foreman-installer certificates managed after upgrade? That's a fair question that I am working on the answer for still.

evgeni:
I guess we should split up `requirements.yml` into `requirements.yml` and `requirements-dev.yml`? Or move them to the respective folders?

https://github.com/theforeman/foreman-quadlet/blob/7beb5b65a03f932cd354f92afed3cab76d1c6de2/requirements.yml#L9

ekohl:
> I guess we should split up `requirements.yml` into `requirements.yml` and `requirements-dev.yml`? Or move them to the respective folders?

@ehelms started https://github.com/theforeman/foreman-quadlet/pull/150 for that. In https://github.com/theforeman/foreman-quadlet/pull/142 I don't need that directly, but if would be relevant if we converted `requirements.yml` into RPM requirements automatically. I see `ansible-collection-ansible-posix` provides `ansible-collection(ansible.posix) = 1.5.4` so that should be fairly easy. Looks like this is automatically generated by `ansible-packaging` which has `/usr/lib/rpm/ansible-generator`. At least on Fedora. Doesn't look like this is in base EL9.

One thing I'd consider is making it a proper collection and relying on it that way.

ekohl:
> "it"? The contents of this repo?

At least the `src` directory. It may be sufficient to add galaxy metadata there because I think otherwise we're pretty close to a standard collection layout.

PR #158, opened 2025-05-21
Rely on sdnotify: true for foreman_proxy

ekohl:
Split off from https://github.com/theforeman/foreman-quadlet/pull/114.


ekohl:
I'm not sure how the 404 error is the result of this change.

evgeni (approved):

PR #159, opened 2025-05-21
Rely on sdnotify: true for pulp

ekohl:
This relies on using systemd notification support to signal a service is ready. That means we no longer need to wait for it to be listening and they can start in parallel.

For that we need to wait for the migration to complete, which is important. Otherwise steps after it may run that depend on a migrated database.

### DIFF
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -137,23 +137,27 @@
       - 'pulp-db-password,type=env,target=PULP_DATABASES__default__PASSWORD'
     env: "{{ pulp_settings_database_env }}"
 
-- name: Start the Pulp API services
+- name: Start Pulp services
   ansible.builtin.systemd:
-    name: pulp-api
+    name: "{{ item }}"
     enabled: true
     state: started
+  async: 60
+  poll: 0
+  loop:
+    - pulp-api
+    - pulp-content
+    - pulp-worker
+  register: pulp_services
 
-- name: Start the Pulp Content services
-  ansible.builtin.systemd:
-    name: pulp-content
-    enabled: true
-    state: started
-
-- name: Start the Pulp Worker service
-  ansible.builtin.systemd:
-    name: pulp-worker
-    enabled: true
-    state: started
+- name: Wait for Pulp services
+  ansible.builtin.async_status:
+    jid: "{{ item.ansible_job_id }}"
+  register: job_result
+  until: job_result is finished
+  retries: 100
+  delay: 1
+  loop: "{{ pulp_services.results }}"
 
 - name: Ensure Pulp admin user exists
   containers.podman.podman_container_exec:
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -52,6 +52,7 @@
     name: "{{ pulp_api_container_name }}"
     image: "{{ pulp_api_image }}"
     state: quadlet
+    sdnotify: true
     command: pulp-api
     network: host
     volumes: "{{ pulp_volumes }}"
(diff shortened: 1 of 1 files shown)


src/roles/pulp/tasks/main.yaml
```diff
@@ -128,41 +130,34 @@
...
+  async: 3
+  poll: 0
+  loop:
+    - pulp-api
+    - pulp-content
+    - pulp-worker
+  register: pulp_services
+
+- name: Wait for Pulp services
+  async_status:
```
evgeni, line 154: seems my suggestion offended the :policeman: and this needs to be
```suggestion
  ansible.builtin.async_status:
```

src/roles/pulp/tasks/main.yaml
```diff
@@ -128,41 +130,34 @@
...
+    jid: "{{ item.ansible_job_id }}"
+  register: job_result
+  until: job_result is finished
+  retries: 500
+  delay: 1
+  loop: "{{ pulp_services }}"
```
evgeni, line 160:
```suggestion
  loop: "{{ pulp_services.results }}"
```

this actually works :)

verified locally with a simple:
```
---
- hosts: localhost
  tasks:
    - name: sleep
      ansible.builtin.command: sleep 10
      async: 30
      poll: 0
      register: sleeper
      loop:
        - 1
        - 2
        - 3
    - name: await
      ansible.builtin.async_status:
        jid: "{{ item.ansible_job_id }}"
      register: job_result
      until: job_result is finished
      retries: 100
      delay: 10
      loop: "{{ sleeper.results }}"
```

ekohl:
Does this mean `pulp-content` failed to start? Debugging really is harder like this.
```
failed: [quadlet] (item={'failed': 0, 'started': 1, 'finished': 0, 'ansible_job_id': 'j774400269359.46862', 'results_file': '/root/.ansible_async/j774400269359.46862', 'changed': True, 'item': 'pulp-content', 'ansible_loop_var': 'item'}) => {
    "ansible_job_id": "j774400269359.46862",
    "ansible_loop_var": "item",
    "attempts": 3,
    "changed": false,
    "child_pid": 46894,
    "finished": 1,
    "item": {
        "ansible_job_id": "j774400269359.46862",
        "ansible_loop_var": "item",
        "changed": true,
        "failed": 0,
        "finished": 0,
        "item": "pulp-content",
        "results_file": "/root/.ansible_async/j774400269359.46862",
        "started": 1
    },
    "results_file": "/root/.ansible_async/j774400269359.46862",
    "started": 1
}
```

src/roles/pulp/tasks/main.yaml
```diff
@@ -128,41 +130,34 @@
...
-    port: 24816
-    timeout: 600
-
-- name: Start the Pulp Worker service
-  ansible.builtin.systemd:
-    name: pulp-worker
+    name: "{{ item }}"
     enabled: true
     state: started
+  async: 3
```
evgeni, line 145:
```suggestion
  async: 30
```

The task runs in a timeout if it's not completed after `async` seconds. Which is what happened here, or at least that's what I assume happened.

ekohl: So `async` is the number of seconds to wait, not the concurrency as I assumed? Interesting.

ekohl: I agree, the fundamentals in Ansible for configuration management just suck. I think Puppet with notifications solved restarts much better than handlers. Granted, it doesn't have anything for async. Many years ago I had hopes for mgmt to solve both well.

ekohl:
I've split this into 3 commits so we can easily revert the parallel stuff if we end up not liking it. I think the patches are now easier to follow.

src/roles/pulp/tasks/main.yaml
```diff
@@ -52,6 +52,7 @@
     name: "{{ pulp_api_container_name }}"
     image: "{{ pulp_api_image }}"
     state: quadlet
+    sdnotify: true
```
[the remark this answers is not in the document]
ekohl: I'd love more health checks but perhaps tackle that separate?

PR #160, opened 2025-05-21
Use sdnotify: true for redis

ekohl:


ekohl:
Again the same failure as in https://github.com/theforeman/foreman-quadlet/pull/158#issuecomment-2898200092 and it doesn't seem related to the changes in this PR.

PR #161, opened 2025-05-21
Use sdnotify: true for foreman

ekohl:
Currently fails because all services attempt to migrate and that fails.

Another open debate: should we keep waiting for tasks to come up safely, or should we drop that? Make it part of system verification later on. It feels wrong to make it a health check, because it's not that restarting one single container will solve it.

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -71,6 +71,7 @@
     name: "foreman"
     image: "{{ foreman_container_image }}:{{ foreman_container_tag }}"
     state: quadlet
+    sdnotify: true
     network: host
     hostname: "{{ ansible_fqdn }}"
     secrets:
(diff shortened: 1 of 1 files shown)


src/roles/foreman/tasks/main.yaml
```diff
@@ -153,11 +155,18 @@
...
 
-- name: Start the Foreman Service
+- name: Start services
   ansible.builtin.systemd:
-    name: foreman
+    name: "{{ item }}"
     enabled: true
     state: started
+  async: 4
+  poll: 0
```
evgeni: I always need to remind myself what this means (from [docs](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_async.html#run-tasks-concurrently-poll-0)):
- `poll: 0` -- don't wait to finish, just continue
- `async: 4` -- mark the task as "timed out" if not finished in 4 seconds (no idea how that's supposed to work if we don't poll)

For pulp we set `async` to `60` -- should we do the same here, just for consistency?

https://github.com/theforeman/foremanctl/blob/aecd4f0b196f43a9ea1a28ce7df4dab16fb4d685/src/roles/pulp/tasks/main.yaml#L182-L183

evgeni (approved):

PR #169, opened 2025-05-26
run upterm if things fail

evgeni:


evgeni:
Yes!

But I didn't see anything useful when doing so

ehelms (approved):

PR #172, opened 2025-05-28
Add sleep for file repo test

ehelms:
Seeing if this addresses https://github.com/theforeman/foreman-quadlet/issues/139

This appears to be the same issue as https://github.com/theforeman/foreman-quadlet/commit/d87681231b473251e19009af3e4c14cac4d3ec2a


ehelms:
I ran this twice with no failures. I know this is ugly  to add yet another one, however, my understanding is there is not a clear answer to what to do and that there is investigation on going. I have seen some details in https://issues.redhat.com/browse/SAT-31818

PR #175, opened 2025-06-04
Rename to foremanctl

ehelms:


evgeni (approved):

evgeni:
I have no idea why things fail, so I opened #177 to see if a sosreport would help us :D

evgeni:
it helped!

https://github.com/theforeman/foreman-quadlet/pull/178 is the fix

PR #176, opened 2025-06-04
Use named parameters style database URI

ehelms:
This switches Foreman and Candlepin to use named parameters for specifying database URI details. This can allow for advanced setups like connecting to PostgreSQL it's socket.


ehelms:
JDBC has it's own magic for how you have to specify when using a socket connection:

```
Simply add ?socketFactory=org.newsclub.net.unix.AFUNIXSocketFactory$FactoryArg&socketFactoryArg=[path-to-the-unix-socket] to the connection URL.
```

I am going to put this in draft and may revisit.

PR #179, opened 2025-06-05
always wait for MetadataGenerate when creating repos

evgeni:
this is better than random sleeps

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -63,18 +63,21 @@ def product(organization, foremanapi):
 @pytest.fixture
 def yum_repository(product, organization, foremanapi):
     repo = foremanapi.create('repositories', {'name': str(uuid.uuid4()), 'product_id': product['id'], 'content_type': 'yum', 'url': 'https://fixtures.pulpproject.org/rpm-no-comps/'})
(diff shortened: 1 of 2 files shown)


evgeni:
I'm confused… why does it not work now for the CV…

ekohl (commented):
I wonder: does the API create a task? If so, is that returned and can we poll for it to complete?

evgeni:
to be precise, CV create/update and Repo create/update use sync actions, so it shouldn't be happening here… yet…

ekohl:
Is that worth filing a bug? I'd expect an API consumer to know when the operation is done. Then link the issue with the sleep.

evgeni:
I think this is the more robust solution we wanted to have for https://github.com/theforeman/foreman-quadlet/issues/139

ekohl (approved):
I'm a bit tempted to have a `wait_for_metadata_generate` helper but this is fine.

ianballou:
> I recall @ianballou saying this is "by design", but I'd love to be proven wrong

We're planning on fixing this weirdness, but in the meantime this solution is definitely more concise.

PR #180, opened 2025-06-05
Switch Foreman images to images from foreman-oci-images

ehelms:


src/roles/foreman/defaults/main.yaml
```diff
@@ -1,6 +1,6 @@
 ---
-foreman_container_image: "quay.io/evgeni/foreman-rpm"
-foreman_container_tag: "nightly"
+foreman_container_image: "quay.io/theforeman/foreman"
+foreman_container_tag: "3.16"
```
ekohl, line 3: We shouldn't have 3.16 yet be ause it doesn't exist. Don't we have a nightly tag there?

PR #181, opened 2025-06-10
add a dedicated dist target

evgeni:


ehelms (approved):
Makefiles and I are not friends despite my best efforts to work with them. The only alternative I would have is straight up bash scripts.

PR #182, opened 2025-06-10
Derive the version from git tags

ekohl:
Rather than copying the version and keeping them in sync, this uses git describe to determine that. It still delivers the version in a VERSION file by letting git replace the version on export.

This will fail until a tag exists.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 NAME := foremanctl
-VERSION := $(shell cat VERSION)
+VERSION := $(shell git describe)
 REQUIREMENTS_YML := $(firstword $(wildcard src/requirements-lock.yml src/requirements.yml))
 
 dist: $(NAME)-$(VERSION).tar.gz
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+VERSION export-subst
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,12 +1,10 @@
 # Release
 
-To create a release, bump `VERSION`, create a commit and tag.
+To create a release, create a tag.
 It must follow the x.y.z pattern without any prefix.
 
 ```
 VERSION=x.y.z
-echo $VERSION > VERSION
-git commit -m "Release $VERSION" VERSION
 git tag -s "$VERSION" -m "Release $VERSION"
 git push --follow-tags
 ```
--- a/VERSION
+++ b/VERSION
@@ -1,1 +1,1 @@
-1.0.0
+$Format:%(describe)$


evgeni:
"This will fail until a tag exists."

Eric and I chatted earlier to go and tag a 0.0.1 as a very first thing to get release automation etc going.

ehelms:
I suppose I don't understand how this works exactly. How do I know when looking at the repository which version I am looking at?

evgeni:
> I suppose I don't understand how this works exactly. How do I know when looking at the repository which version I am looking at?

When you have a git checkout, you can use `git describe` to get the version.

In the release tarballs, `VERSION` will have the right value via https://git-scm.com/docs/git-archive/2.29.0#Documentation/git-archive.txt-export-subst

Makefile
```diff
@@ -1,5 +1,5 @@
 NAME := foremanctl
-VERSION := $(shell cat VERSION)
+VERSION := $(shell git describe)
```
evgeni, line 2: this implies tags are strictly `X.Y.Z`, right, no `v` prefix, no `foremanctl-` prefix?

ehelms:
What about when I'm looking at it in the Web UI? For example if I'm on a
release branch 2.x.

Eric D. Helms
Red Hat Engineering

Makefile
ekohl: That's a good point. It's also what we do with `VERSION` and I'm not sure if we can filter it out directly in `git describe`. Is it a problem to use plain `x.y.z` tags?

ekohl:
How about this?

ehelms:
That is looking good and now raises a related question. This would generate a tarball of the entire repository and that would be the "release". That seems overkill to me? I would not think we want the tests and development tooling in the release and instead limit it to the code that is intended for production usage.

ekohl:
It is common in many projects. We don't vendor in the development dependencies so I don't think it's a big deal, but we can look at [export-ignore](https://git-scm.com/docs/git-archive/2.29.0#Documentation/git-archive.txt-export-ignore).

ekohl:
> The reasons I am harping on this is to include only what's necessary to avoid shipping potential security risks into a production setup. This prevents having to then go and solve this within each build system instead of having a central, minimal artifact. This also helps with the "what's needed in each environment" question.

We can still filter things out in the RPM build. I still see the tarball as an input to the RPM build, not something we expect end users to really consume.

evgeni:
Would you mind splitting out the "Add release workflow" commit into an own PR?
It works regardless the exact workflow we'll be using, and I'd love to go and tag a 0.0.1 and have an artifact that can be used (and improved, if we decide the contents are not perfect)

ekohl:
I've opened #183 and modified to reflect you need to update `VERSION`.

ekohl:
Rebased.

ekohl:
I've rebased this to resolve conflicts, but there's still manual work to update `foremanctl.spec`. What's the long term plan with this? Move it to `foreman-packaging` or keep it in?

PR #184, opened 2025-06-13
add hammer setup on the host

evgeni:
- **add hammer role**
- **move setup-repositories to an own command**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,9 @@ jobs:
       - name: Start VMs
         run: |
           ./forge vms start
+      - name: Configure repositories
+        run: |
+          ./forge setup-repositories
       - name: Create installer certificates
         if: contains(matrix.certificate_source, 'installer')
         run: |
--- a/development/playbooks/setup-repositories/setup-repositories.yaml
+++ b/development/playbooks/setup-repositories/setup-repositories.yaml
@@ -1,0 +1,18 @@
+---
+- name: Configure Puppet and Foreman repositories
+  hosts:
+    - quadlet
+  become: true
+  tasks:
+    - name: Setup Puppet repositories
+      ansible.builtin.include_role:
+        name: theforeman.operations.puppet_repositories
+      vars:
+        foreman_puppet_repositories_version: "8"
+
+    - name: Setup Foreman repositories
+      ansible.builtin.include_role:
+        name: theforeman.operations.foreman_repositories
+      vars:
+        foreman_repositories_version: nightly
+        foreman_repositories_katello_version: nightly
--- a/development/roles/foreman_installer_certs/tasks/main.yml
+++ b/development/roles/foreman_installer_certs/tasks/main.yml
@@ -6,19 +6,6 @@
     name: packit/theforeman-foreman-installer-935
     chroot: rhel-9-x86_64
 
-- name: Setup Puppet repositories
-  ansible.builtin.include_role:
-    name: theforeman.operations.puppet_repositories
-  vars:
-    foreman_puppet_repositories_version: "8"
-
-- name: Setup Foreman repositories
-  ansible.builtin.include_role:
-    name: theforeman.operations.foreman_repositories
-  vars:
-    foreman_repositories_version: nightly
-    foreman_repositories_katello_version: nightly
-
 - name: Install foreman-installer package
   ansible.builtin.package:
     name: foreman-installer-katello


src/playbooks/deploy/deploy.yaml
```diff
@@ -68,3 +68,4 @@
     - pulp
     - foreman_proxy
     - foreman
+    - hammer
```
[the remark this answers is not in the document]
evgeni: you mean `./foremanctl setup-hammer`?

src/roles/hammer/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+hammer_foreman_server_url: "https://{{ ansible_fqdn }}"
+hammer_server_ca: "{{ server_ca_certificate }}"
```
ehelms, line 3: This bleeds the top level variable into the role.

ehelms: Is this setting required? Will it use the system trust if not set?

src/playbooks/deploy/deploy.yaml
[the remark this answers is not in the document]
evgeni: I wonder if that'll break persisting of params now. One way to find out!

.github/workflows/test.yml
```diff
@@ -74,6 +77,9 @@ jobs:
       - name: Run deployment
         run: |
           ./foremanctl deploy --certificate-source=${{ matrix.certificate_source }}
+      - name: Setup hammer
+        run: |
+          ./foremanctl setup-hammer
```
evgeni, line 82: no `--certificate-source`, as it's supposed to remember that from the step before!

src/playbooks/deploy/deploy.yaml
```diff
@@ -12,6 +12,7 @@
     certificates_hostnames:
       - "{{ ansible_fqdn }}"
       - localhost
+    certificates_ca_directory: /root/certificates
```
evgeni, line 15: this can also be set in src/vars/default_certificates.yml, which is probably nicer?

src/vars/default_certificates.yml
```diff
@@ -1,4 +1,5 @@
 ---
+certificates_ca_directory: /root/certificates
```
evgeni, line 2: this is needed as otherwise you can't reference the certs w/o including the certs role

src/roles/hammer/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+hammer_foreman_server_url: "https://{{ ansible_fqdn }}"
+hammer_ca_certificate: "/etc/pki/tls/certs/ca-bundle.trust.crt"
+hammer_packages:
+  - rubygem-hammer_cli
+  - rubygem-hammer_cli_foreman
+  - rubygem-hammer_cli_foreman_tasks
+  - rubygem-hammer_cli_foreman_remote_execution
+  - rubygem-hammer_cli_katello
```
ekohl, line 9: I thought we had implemented Provides for this, but https://github.com/theforeman/foreman-packaging/commit/10e0424442e2a8d9703d83e07801218676a480f6 hasn't propagated to the ones with nightly packages. I'd like to have it clean from the start in the next installer so can we update our packaging and use:
```suggestion
  - hammer-cli-plugin-foreman
  - hammer-cli-plugin-foreman_tasks
  - hammer-cli-plugin-foreman_remote_execution
  - hammer-cli-plugin-katello
```

Later we can also update our Debian packages to do the same.

evgeni: I stand corrected, I tried, and it works!

```
# ansible -m ansible.builtin.package -a name=hammer-cli-plugin-foreman_remote_execution -v localhost
Using /etc/ansible/ansible.cfg as config file
localhost | SUCCESS => {
    "changed": false,
    "msg": "Nothing to do",
    "rc": 0,
    "results": []
}
```

ekohl (approved):

src/roles/hammer/README.md
```diff
@@ -0,0 +1,17 @@
+hammer role
```
ehelms, line 1: Should we make an issue to add README to each role?

ekohl: This was the result of a discussion @evgeni and I had. It's an effort to capture a design decission: we rely on Hammer itself to do the right thing in the role but in the playbook we combine them since there we know what it has to be. That way the role could be used standalone on other systems.

evgeni: Yeah, probably.

@ekohl and me discussed it'd be a good idea to point out how the roles are useable standalone and how they are integrated into foremanctl, so people can re-use them.

ekohl: It could even be in the foreman-operations-collection

evgeni: I think at some point operations and `src/roles` will just merge.

ekohl: Also what I'd like to see

PR #186, opened 2025-06-13
Use Pulp images from quay.io/foreman

ehelms:


evgeni:
Looks like `/var/lib/pulp/tmp` doesn't exist, even if we've created it in 
https://github.com/theforeman/foremanctl/blob/5de54e3204c93227e0ec17781e82d907b5776880/src/roles/pulp/tasks/main.yaml#L23-L31

src/vars/images.yml
```diff
@@ -4,7 +4,7 @@ foreman_container_image: quay.io/foreman/foreman
 foreman_container_tag: nightly
 foreman_proxy_container_image: quay.io/foreman/foreman-proxy
 foreman_proxy_container_tag: nightly
-pulp_image: quay.io/pulp/pulp-minimal:3.73
+pulp_image: quay.io/foreman/pulp:3.73
```
ekohl, line 7: Do we want to adopt the Foreman tagging scheme already?
```suggestion
pulp_image: quay.io/foreman/pulp:foreman-nightly
```

Or what we really want long term:
```suggestion
pulp_image: quay.io/foreman/pulp:foreman-{{ foreman_container_tag }}
```

Now another question: why does everything have `_image` or `_tag` but Pulp basically has both combined into one?

ehelms: > Do we want to adopt the Foreman tagging scheme already?

I'll do that following this change.

ehelms:
As this indicates we always end up having to update the image tag in two places. This could easily get out of sync and I am considering to drop the tag in the role and force supplying it if using the role stand-alone.

ehelms:
The changes from https://github.com/theforeman/pulp-oci-images/pull/2 have been merged, built and pushed. This reflects those updates. And switching to nightly tags is a separate update tracked here -- https://github.com/theforeman/foremanctl/pull/191

ekohl (approved):

PR #189, opened 2025-06-16
Align container image and tag variables

ehelms:


src/vars/images.yml
ekohl, line 1: Should we also include PostgreSQL in this file? I also noticed that when we pull that we don't specify a tag at all.

ekohl (commented):
Sorry for the string of reviews.

While playing with in the past I noticed PostgreSQL isn't fetched during `pull`. Should it? It may not play well with the external DB support where you don't need the image so is that the reason?

src/vars/images.yml
ehelms: Correct - we would need a specific vars file for postgresql that is conditionally included. I don't want to hold up this change for that but let me think on it since I've been doing other remote db work.

PR #191, opened 2025-06-16
Use foreman nightly

ehelms:
Depends on (or includes https://github.com/theforeman/foremanctl/pull/186).

### DIFF
--- a/src/vars/images.yml
+++ b/src/vars/images.yml
@@ -1,11 +1,12 @@
+container_tag_stream: "nightly"
 candlepin_container_image: quay.io/foreman/candlepin
-candlepin_container_tag: "4.4.14"
+candlepin_container_tag: "foreman-{{ container_tag_stream }}"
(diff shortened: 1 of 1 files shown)


src/vars/images.yml
```diff
@@ -1,11 +1,11 @@
 candlepin_container_image: quay.io/foreman/candlepin
-candlepin_container_tag: "4.4.14"
+candlepin_container_tag: "foreman-nightly"
```
ekohl, line 2: I'd love to have 1 variable to update for every release. Something like:
```suggestion
candlepin_container_tag: "foreman-{{ foreman_container_tag }}"
```

Perhaps introduce something like `container_tag_stream: nightly` and then consistently use that for all the containers? That allows changing a single container easily while also making branching easier.

ekohl (approved):
:+1: for the principle. Inline a comment about a next step we should IMHO also make, but can do as a separate change.

evgeni (approved):

PR #200, opened 2025-06-25
Important commit about how to pronounce it's name

Odilhao:


ekohl (commented):
I think I suggested the name and I did not know this.

Odilhao:
> I think I suggested the name and I did not know this.

I've been poking Eric and Evgeni about this since it was changed to `foremanctl`. Cuddle is the only possible name.

PR #204, opened 2025-06-26
add copr command so one doesn't need to use the UI

evgeni:


ekohl (approved):
I had been thinking about setting up webhooks after a new tag is pushed. https://docs.pagure.org/copr.copr/user_documentation.html#triggerring-builds-by-tag-events is there. However, I'm not sure how well this takes into account that at some point we may have stable branches.

This is a decent workaround.

Issue #167, opened 2025-05-22
pull-images does't pull in PostgreSQL image


ekohl:
It does pull in Redis, but not PostgreSQL.

PR #206, opened 2025-06-26
Pull PostgreSQL image

evgeni:
Fixes: #167


Odilhao:
Can we use a boolean to define if it's internal or external db?

ehelms:
> Can we use a boolean to define if it's internal or external db?

Instead of a string? Or what are you thinking?

ekohl:
@evgeni what will the installation instructions look like when we apply this? Will this answer be stored for the deployment?

evgeni:
And maybe to re-iterate, because Eric said it wasn't too clear: you can omit passing `--database-mode` to `deploy` if you passed it to `pull-images` already (it's stored), but I put it explicitly in the doc example as `pull-images` is optional and if you omit that, you HAVE to pass `database-mode` to `deploy`.

Issue #138, opened 2025-04-23
The installer should provide checks as a stand-alone command and as part of the happy path


ehelms:
This issue is tracking the addition of checks to the installation.

- Relevant checks for prerequisite
- Certificate checks
- Database checks for remote database

ehelms:
Maybe? I think I would need to see it when multiple checks fail.

evgeni:
Aye. Will do once we have checks ;)

I think the yaml output above matches what `debug` would have done (formatting wise, debug also outputs more fields, but that's not the problem here)

archanaserver:
To track down all the checks:

- [x] https://github.com/theforeman/foremanctl/pull/154
- [x] https://github.com/theforeman/foremanctl/pull/207
- [x] https://github.com/theforeman/foremanctl/pull/209
- [ ] ~~https://github.com/theforeman/foremanctl/pull/208~~
- [ ] https://github.com/theforeman/foremanctl/pull/210
- [ ] https://github.com/theforeman/foremanctl/pull/218

PR #207, opened 2025-06-30
 Add hostname check

archanaserver:
Relates: #138


src/roles/hostname_check/files/foreman-hostname-check
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function error() {
+    echo -e "\n${RED}[FAIL]${RESET}\n"
+    CURRENT_EXIT_CODE=$1
+    EXIT_CODE=$((EXIT_CODE|CURRENT_EXIT_CODE))
+    echo -e "$2" >&2
+}
+
+function check-not-localhost() {
```
ekohl, line 20: I don't think you can use dashes in function names in bash. If you can, I don't think you should because it's too easy to confuse with subtraction.

src/roles/hostname_check/files/foreman-hostname-check
```diff
@@ -0,0 +1,81 @@
...
+    if [[ "$HOSTNAME" == "localhost" || "$HOSTNAME" == "localhost.localdomain" || -z "$HOSTNAME" ]]; then
+        error 1 "Hostname is invalid: '$HOSTNAME'"
+    else
+        success
+    fi
+}
+
+function check-fqdn-resolves() {
+    printf "Checking that FQDN resolves via DNS or /etc/hosts: "
+    FQDN=$(hostname -f)
```
ekohl, line 32: Puppet and Facter used `hostname -f` a lot. In our new installer we don't. Perhaps it's better to pass them as explicit inputs. We use `ansible_fqdn` everywhere so that's all you need to care about.

src/playbooks/check/check.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Run preflight checks
+  hosts: localhost
+  gather_facts: false
+  tasks:
+    - name: Run hostname check script
+      command: "{{ playbook_dir }}/../../roles/hostname_check/files/foreman-hostname-check"
+      register: hostname_check
+      failed_when: hostname_check.rc != 0
```
ekohl, line 9: Isn't failing on non-zero output default behavior of the `command` module?

src/playbooks/check/check.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Run preflight checks
+  hosts: localhost
```
ekohl, line 3: Needs to run on `quadlet` because there's where we really deploy.

src/roles/hostname_check/files/foreman-hostname-check
evgeni: You can. Bash does not do any substraction by default either.

```
evgeni@juuni:~$ 1-1
bash: 1-1: command not found...
evgeni@juuni:~$ $((1-1))
bash: 0: command not found...
```

src/playbooks/check/check.yaml
```diff
@@ -0,0 +1,9 @@
...
+  gather_facts: false
+  tasks:
+    - name: Run hostname check script
+      command: "{{ playbook_dir }}/../../roles/hostname_check/files/foreman-hostname-check"
```
evgeni, line 7: Instead of a full-blown script, could we use `assert` with the `ansible_fqdn` fact?

Something like:
```
- name: check hostname
  ansible.builtin.assert:
    that:
      - "'_' not in ansible_fqdn"
      - "'.' in ansible_fqdn"
```
etc?

archanaserver: Yes, I tried doing that. I can't clearly understand if it is working.

ehelms:
I am torn on introducing molecule testing alongside our pytest testing. I'd prefer to have one test framework unless there is a good argument for both. If we use both, I think the README or a `doc/testing.md` should indicate when to use which type of testing.

src/roles/hostname_check/tasks/main.yaml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Assert FQDN is not 'localhost' or 'localhost.localdomain'
+  assert:
+    that:
+      - ansible_fqdn != "localhost"
+      - ansible_fqdn != "localhost.localdomain"
+    fail_msg: "Invalid FQDN: {{ ansible_fqdn }}"
```
ehelms, line 7: This could be more descriptive that the FQDN cannot be localhost.

evgeni:
> I am torn on introducing molecule testing alongside our pytest testing. I'd prefer to have one test framework unless there is a good argument for both. If we use both, I think the README or a `doc/testing.md` should indicate when to use which type of testing.

Molecule can be ran from pytest :)

The benefit of using molecule would be that we could test behavior of a role without a full foreman installation run.

Like, I don't need to deploy foreman to verify the hostname check refuses "localhost"

ehelms:
> Molecule can be ran from pytest :)

But the tests and setup are still written in molecule?

> The benefit of using molecule would be that we could test behavior of a role without a full foreman installation run.

That can be true with our current setup as well. Molecule will spin up a container or VM for testing right vs traditional unit test?

src/roles/check_hostname/tasks/main.yaml
```diff
@@ -0,0 +1,25 @@
...
+  ansible.builtin.assert:
+    that:
+      - ansible_fqdn.count('.') >= 1
+    fail_msg: "FQDN {{ ansible_fqdn }} must contain at least one dot."
+
+- name: Assert FQDN has no underscores
+  ansible.builtin.assert:
+    that:
+      - "'_' not in ansible_fqdn"
+    fail_msg: "FQDN {{ ansible_fqdn }} must not contain underscores"
```
evgeni, line 19: Out of curiosity, why did you use `count >= 1` (and not `in`) for the dot, but `not in` (and not `count == 0`) for the underscore?
Obviously either works, but I am curious :)

evgeni (approved):

PR #208, opened 2025-06-30
Add IPv6 check

archanaserver:
Relates: #138


ekohl (commented):
I'm not sure we need the IPv6 check in a containerized world. It was there partly because of how we configured services in the past (at least Redis, possibly others). I think the new installer can drop it.

evgeni:
:+1: let's skip this one for now

archanaserver:
As per the discussion moving this PR to draft.

PR #209, opened 2025-06-30
Add Subuid/Subguid check

archanaserver:
Relates: #138


tests/unit/test_foreman_subuid_subgid_check.py
ekohl, line 1: It may be better to test Ansible roles via [molecule](https://ansible.readthedocs.io/projects/molecule/).

src/playbooks/checks/check.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Run preflight checks
+  hosts: localhost
+  gather_facts: false
+  tasks:
+    - name: Run Subuid/Subgid check script
+      command: "{{ playbook_dir }}/../../roles/subuid_subgid_check/files/foreman-subuid-subgid-check"
```
ekohl, line 7: This doesn't work because the file is on the control node, but the command runs on the target host. You need to transfer it to the target machine.

It may be better to write an Ansible module that performs the check and rely on Ansible to transfer the file there rather than a shell script.

src/playbooks/checks/check.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Run preflight checks
+  hosts: localhost
```
ekohl, line 3: This should be on the quadlet host because that's where we run podman.
```suggestion
  hosts: quadlet
```

src/playbooks/checks/check.yaml
evgeni: IMHO no need to write a module. You can `command` a `grep` call with `ignore_errors: true` and then form a nice error message with `assert`.

But yes, I would too prefer it being in "pure" Ansible, over an external script.

archanaserver: I’ve now removed the external shell script and implemented the check entirely in Ansible. Doing the same for the #207 one as well. Thanks for the suggestion!

src/playbooks/checks/check.yaml
evgeni, line 1: this file needs to be called `checks.yaml` or the folder needs to be `check`, I prefer the plural form.

src/roles/subuid_subgid_check/tasks/main.yaml
```diff
@@ -0,0 +1,26 @@
+---
+- name: Get current user
+  command: whoami
+  register: current_user
+
+- name: Check /etc/subuid for current user
+  command: grep "^{{ current_user.stdout }}:" /etc/subuid
```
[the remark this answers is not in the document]
evgeni: (ansible-lint complains about this)

evgeni:
Design question (which will also apply to other checks we write):
Currently, you have a dedicated `assert` for each "detail" (subuid, subgid). That means that when the first fails, it's reported to the user and the second assert is never executed and if that would have failed too, the user might fix the first issue, re-run checks just to be greeted with another error.

Ansible's `assert` module supports multiple entries to `that`, so we could write
```yaml
- name: Assert user has subuid/subgid
  assert:
    that:
      - subuid_check.rc == 0
      - subgid_check.rc == 0
    fail_msg: "{{ current_user.stdout }} is missing entries for subuid/subgid"
```
It'd still only report one of the assertions in the output what failed, but we can craft the fail message to hint at both?
```
fatal: [localhost]: FAILED! => {
    "assertion": "_subgid is success",
    "changed": false,
    "evaluated_to": false,
    "msg": "Your user has no subuid/subgid range defined"
}
```

src/roles/subuid_subgid_check/tasks/main.yaml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Check /etc/subuid for current user
+  ansible.builtin.command: grep "^{{ ansible_user_id }}:" /etc/subuid
```
ekohl, line 3: Should this use [lineinfile](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/lineinfile_module.html) in check mode instead?

evgeni:
@archanaserver the help text etc needs to go to `metadata.obsah.yaml`, not `checks.yaml`

evgeni:
```
bash-5.1# foremanctl checks

PLAY [Run preflight checks] ***********************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ****************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [subuid_subgid_check : Check /etc/subuid for current user] ***********************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": ["grep", "^root:", "/etc/subuid"], "delta": "0:00:00.004194", "end": "2025-07-10 12:27:08.247341", "msg": "non-zero return code", "rc": 1, "start": "2025-07-10 12:27:08.243147", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
...ignoring

TASK [subuid_subgid_check : Check /etc/subgid for current user] ***********************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": ["grep", "^root:", "/etc/subgid"], "delta": "0:00:00.003509", "end": "2025-07-10 12:27:08.382988", "msg": "non-zero return code", "rc": 1, "start": "2025-07-10 12:27:08.379479", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
...ignoring

TASK [subuid_subgid_check : Assert /etc/subuid and /etc/subgid has entry] *************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {
    "assertion": "subuid_check is success",
    "changed": false,
    "evaluated_to": false,
    "msg": "Entries for user root are missing in /etc/subuid or /etc/subgid"
}

PLAY RECAP ****************************************************************************************************************************************************************************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=2
```

fails as expected :)

src/playbooks/checks/checks.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Run preflight checks
+  hosts: quadlet
+  gather_facts: true
+  roles:
+    - subuid_subgid_check
```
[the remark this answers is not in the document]
evgeni: Another thought: should we have a meta `checks` role, that includes all the others. We can then add the same role to the `deploy` playbook and ensure all checks are executed in both cases.

(Would also be a good place to do the reporting I was proposing in #138 )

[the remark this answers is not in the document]
archanaserver: > Another thought: should we have a meta `checks` role, that includes all the others. We can then add the same role to the `deploy` playbook and ensure all checks are executed in both cases.
> 
> (Would also be a good place to do the reporting I was proposing in #138 )

Yes that is what I thought of and created this meta `checks` role and I'm gonna include others as well. But how we can capture the results of all included checks and present them together at the end?

evgeni:
ansible-lint failures are unrelated, fix in https://github.com/theforeman/foremanctl/pull/211

evgeni (approved):

src/playbooks/checks/checks.yaml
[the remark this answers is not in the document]
archanaserver: Yeah make sense.

PR #210, opened 2025-07-10
Add remote db check

archanaserver:
Relates: https://github.com/theforeman/foremanctl/issues/138

### DIFF
--- a/src/roles/check_database_connection/tasks/main.yaml
+++ b/src/roles/check_database_connection/tasks/main.yaml
@@ -5,8 +5,10 @@
     login_user: "{{ foreman_database_user }}"
     login_password: "{{ foreman_database_password }}"
     login_db: "{{ foreman_database_name }}"
+    ca_cert: "{{ foreman_database_sslrootcert | default(omit) }}"
+    ssl_mode: "{{ foreman_database_sslmode | default(omit) }}"
   register: check_database_connection_ping_result
-  when: (foreman_database_host | default("localhost")) not in ["localhost", "127.0.0.1", "::1"]
+  when: database_mode == 'external'
   ignore_errors: true
 
 - name: Assert Foreman database is reachable
@@ -16,4 +18,4 @@
     fail_msg: |
       Cannot connect to Foreman database '{{ foreman_database_name }}' at {{ foreman_database_host }}.
       Please verify the database host, port, name, user, and password.
-  when: (foreman_database_host | default("localhost")) not in ["localhost", "127.0.0.1", "::1"]
+  when: database_mode == 'external'
--- a/src/playbooks/_database_connection/metadata.obsah.yaml
+++ b/src/playbooks/_database_connection/metadata.obsah.yaml
@@ -1,0 +1,46 @@
+---
+variables:
+  database_host:
+    help: Host of the PostgreSQL database.
+  database_port:
+    help: Port of the PostgreSQL database.
+
+  database_ssl_mode:
+    help: If using an SSL connection to the database, define how it should be verified.
+    choices:
+      - disable
+      - allow
+      - prefer
+      - require
+      - verify-ca
+      - verify-full
+  database_ssl_ca:
+    help: Path to the database SSL CA certificate.
+    type: AbsolutePath
+
+  foreman_database_name:
+    help: Name of the Foreman database.
+  foreman_database_user:
+    help: User of the Foreman database.
+  foreman_database_password:
+    help: Password for the Foreman database.
+
+  candlepin_database_name:
+    help: Name of the Candlepin database.
+  candlepin_database_user:
+    help: User of the Candlepin database.
+  candlepin_database_password:
+    help: Password for the Candlepin database.
+
+  pulp_database_name:
+    help: Name of the Pulp database.
+  pulp_database_user:
+    help: User of the Pulp database.
+  pulp_database_password:
+    help: Password for the Pulp database.
+
+constraints:
+  required_together:
+    - [database_ssl_mode, database_ssl_ca]
+  required_if:
+    - ['database_mode', 'external', ['database_host']]
--- a/src/playbooks/checks/checks.yaml
+++ b/src/playbooks/checks/checks.yaml
@@ -2,5 +2,8 @@
 - name: Run preflight checks
   hosts: quadlet
   gather_facts: true
+  vars_files:
+    - "../../vars/defaults.yml"
+    - "../../vars/database.yml"
   roles:
     - checks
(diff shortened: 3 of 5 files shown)


ekohl:
Something to consider: `pg_isready` can be used to connect to a host but also verify credentials. That could be an interesting feature to validate all user parameters.

archanaserver:
@evgeni the workflow is currently failing, looks like launchpad is returning 503 for `ppa:evgeni/vagrant`

evgeni:
Whatever launchpad was doing, now it works again :woman_shrugging:

archanaserver:
it just work when you re-run it

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -0,0 +1,17 @@
...
+    login_user: "{{ foreman_database_user }}"
+    login_password: "{{ foreman_database_password }}"
+    login_db: "{{ foreman_database_name }}"
+  register: check_database_connection_ping_result
+  ignore_errors: true
+
+- name: Assert Foreman database is reachable
+  ansible.builtin.assert:
+    that:
+      - check_database_connection_ping_result.ping | default(false)
```
evgeni, line 14: Shouldn't that be 
```suggestion
      - check_database_connection_ping_result.is_available
```
at least that's how I read the code of the module

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Check Foreman database connectivity and credentials
+  community.postgresql.postgresql_ping:
+    login_host: "{{ foreman_database_host }}"
+    login_user: "{{ foreman_database_user }}"
+    login_password: "{{ foreman_database_password }}"
+    login_db: "{{ foreman_database_name }}"
```
evgeni, line 7: This currently gets executed for local DB too, but the user and the db don't exist yet -- they are created later by the `postgresql` role. So we probably should skip this check for local DB alltogether?

ekohl:
> > Something to consider: `pg_isready` can be used to connect to a host but also verify credentials. That could be an interesting feature to validate all user parameters.
> 
> 
> 
> I went with `community.postgresql.postgresql_ping` since it handles validation natively in Ansible, but also fits better into the role-based structure.

It is better, I just didn't know about it. Wouldn't surprise me if it uses pg_ready under the hood

src/roles/check_database_connection/tasks/main.yaml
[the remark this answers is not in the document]
archanaserver: @evgeni made the suggested changes does that looks good to you?

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -0,0 +1,19 @@
...
+  register: check_database_connection_ping_result
+  when: (foreman_database_host | default("localhost")) not in ["localhost", "127.0.0.1", "::1"]
```
evgeni, line 9: you can check `database_mode == 'external'` instead, similar to here:

https://github.com/theforeman/foremanctl/blob/a56ab1daa893df57fda45cbb8a2d5c788b9da420/src/playbooks/deploy/deploy.yaml#L61-L63

src/playbooks/_database_connection/metadata.obsah.yml
```diff
@@ -0,0 +1,10 @@
+---
+variables:
+  foreman_database_host:
+    help: Hostname of the PostgreSQL server.
+  foreman_database_user:
+    help: Username for the Foreman DB.
+  foreman_database_password:
+    help: Password for the Foreman DB.
+  foreman_database_name:
+    help: Name of the Foreman DB.
```
evgeni, line 10: can you move the definition from `deploy` here, instead?

https://github.com/theforeman/foremanctl/blob/a56ab1daa893df57fda45cbb8a2d5c788b9da420/src/playbooks/deploy/metadata.obsah.yaml#L6-L39

evgeni: and then you load them into the playbook via
https://github.com/theforeman/foremanctl/blob/a56ab1daa893df57fda45cbb8a2d5c788b9da420/src/playbooks/deploy/deploy.yaml#L6-L10

evgeni: (you only need defaults and database, certificates and images aren't needed for checks (yet))

src/playbooks/_database_connection/metadata.obsah.yml
```diff
@@ -0,0 +1,14 @@
+---
+variables:
+  database_host:
+    help: Host of the PostgreSQL database.
+  database_port:
+    help: Port of the PostgreSQL database.
+  foreman_database_host:
+    help: Hostname of the PostgreSQL server.
```
evgeni, line 8:
```suggestion
```

This doesn't need to be defined here, it will be created internally by `vars/database.yml`

evgeni:
Works perfectly once the last two suggested changes are made

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Check Foreman database connectivity and credentials
+  community.postgresql.postgresql_ping:
+    login_host: "{{ foreman_database_host }}"
+    login_user: "{{ foreman_database_user }}"
+    login_password: "{{ foreman_database_password }}"
+    login_db: "{{ foreman_database_name }}"
```
evgeni, line 7:
```suggestion
    login_db: "{{ foreman_database_name }}"
    ca_cert: "{{ database_ssl_ca | default(omit) }}"
    ssl_mode: "{{ database_ssl_mode | default(omit) }}"
```

archanaserver:
> Question: should we also test Pulp and Candlepin DBs?

I just realized this and yes it is worth considering here, as per the current db setup. I'll do it as a follow up.

archanaserver:
> > Question: should we also test Pulp and Candlepin DBs?
> 
> I just realized this and yes it is worth considering here, as per the current db setup. I'll do it as a follow up.

@ekohl as Evgeni is out, what do you think of this?

archanaserver:
> IMHO we should test also candlepin and pulp databases, especially if we require the credentials for them

This follows up on PR #221 and adds `postgresql_ping` checks for Candlepin and Pulp databases

archanaserver:
@evgeni can we get this in, if this looks good to you? i have followed up the rest of the work in different PR

> > IMHO we should test also candlepin and pulp databases, especially if we require the credentials for them
> 
> This follows up on PR #221 and adds `postgresql_ping` checks for Candlepin and Pulp databases

archanaserver:
> @archanaserver the ca_cert/ssl_mode part is still outstanding -- people will usually configure these on remote db setups. Otherwise LGTM!

@evgeni i don't understand, can you explain more please?

PR #211, opened 2025-07-11
make ansible-lint happy again

evgeni:


src/roles/pulp/tasks/main.yaml
```diff
@@ -154,8 +154,8 @@
 - name: Wait for Pulp services
   ansible.builtin.async_status:
     jid: "{{ item.ansible_job_id }}"
-  register: job_result
-  until: job_result is finished
+  register: _pulp_job_result
```
archanaserver, line 157: why we have `_pulp_job_result` instead of `pulp_job_result`?

evgeni: Mostly to indicate "it's not something that we need in other steps, it's a private variable".

`pulp_job_result` would make ansible-lint happy just fine too.

archanaserver (approved):
LGTM!

PR #213, opened 2025-07-12
Use yaml file extension consistently

ehelms:


ehelms:
> Obsah itself "forces" `.yaml`, we'd have to change that first.

Ahh. Should we just follow the `yaml.org` convention then instead of "Ansible's" ?

ehelms:
> What's the `yaml.org` convention? :)

To use `yaml` :grin:  And I feel like, using that extension is every other projects convention. Ansible is the special one here.

PR #216, opened 2025-07-24
build every commit into our copr

evgeni:
https://packit.dev/docs/configuration/upstream/copr_build#using-a-custom-copr-project has been followed, so this should "just work"

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -16,7 +16,14 @@ actions:
     - bash -c 'ls -1t foremanctl-*.tar.gz | head -n1'
 
 jobs:
-  - job: copr_build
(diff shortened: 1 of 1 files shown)


evgeni:
> So this gives us nightly builds? Should we update https://copr.fedorainfracloud.org/coprs/g/theforeman/foremanctl/ with that description?

Correct. This is something @ehelms showed me last week in https://github.com/theforeman/foreman_rh_cloud/pull/1027

evgeni:
hah, but the owner needs to be `@theforeman`, not `theforeman` :see_no_evil:

Issue #51, opened 2024-11-05
scaling backend services


evgeni:
Today we deploy various services (Foreman/Puma, Pulp/Gunicorn) in a default configuration which usually spans a couple (~2) workers (threads, processes, doesn't matter) inside the container.
We do not expose any way to change those numbers today, so it might become a bottleneck in setups with higher load.

On the other hand, it feels more natural for container scaling to spawn more containers instead of changing configuration *inside* the container.

pablomh:
Will there be any kind of load-balancer in front of those scaled containers? How the requests to those new containers will be "routed"?

PR #218, opened 2025-07-28
add system requirement check

archanaserver:
Relates: https://github.com/theforeman/foremanctl/issues/138

### DIFF
--- a/src/roles/check_system_requirements/tasks/main.yaml
+++ b/src/roles/check_system_requirements/tasks/main.yaml
@@ -1,6 +1,6 @@
 - name: Load tuning thresholds based on tuning_profile
   ansible.builtin.include_vars:
-    file: "../../vars/tuning/{{ tuning_profile | default('medium') }}.yaml"
+    file: "../../vars/tuning/{{ tuning_profile }}.yaml"
     name: tuning_vars
 
 - name: Check if system has enough CPU cores
--- a/src/vars/tuning/large.yaml
+++ b/src/vars/tuning/large.yaml
@@ -1,3 +1,0 @@
----
-min_cpu_cores: 8
-min_ram_mb: 16384
--- a/src/vars/tuning/medium.yaml
+++ b/src/vars/tuning/medium.yaml
@@ -1,3 +1,0 @@
----
-min_cpu_cores: 4
-min_ram_mb: 10240


archanaserver:
> These checks need to be more flexible. For a plain Foreman these requirements are way too high. We also have "development" setups which are known to be underpowered, but good enough.
> 
> Any idea how that would be implemented?

That make sense, I think we can do that by introducing a whole new mode like a var for switching to choices as per the user, and then tailor the system requirement checks based on that. And we can do something like this:
`foremanctl checks -e deployment_mode=development`

ShimShtein:
I second Ewoud's concern: we should probably make this parametrized. Also we need to have an option to skip the tests in case the user is certain that the system is good enough for his needs.
Having an aggregate option, like `deployment_mode` sounds nice, as long as it would be easy to use it, and it will have sane defaults.

ehelms:
I think we should just copy the tuning concept from `foreman-installer` (https://github.com/theforeman/foreman-installer/tree/develop/config/foreman.hiera/tuning) and allow the checks to be generic and based upon a set of variable inputs. This will give us flexibility and whatever granularity we choose.

archanaserver:
@evgeni any thoughts on the failure?

src/roles/check_system_requirements/tasks/main.yaml
```diff
@@ -0,0 +1,24 @@
...
+    file: "../../vars/tuning/{{ deployment_mode | default('medium') }}.yaml"
+    name: tuning_vars
+  when: not skip_system_requirements | default(false)
+
+- name: Check if system has enough CPU cores
+  ansible.builtin.assert:
+    that:
+      - ansible_facts.processor_cores >= tuning_vars.min_cpu_cores
+    fail_msg: "System must have at least {{ tuning_vars.min_cpu_cores }} CPU cores."
+  when: not skip_system_requirements | default(false)
```
evgeni, line 17: I think we should use `processor_vcpus` here.

I have a VM here, where I configured libvirt to give it 2 CPUs, and Ansible sees the following facts:
```
        "ansible_processor_cores": 1,
        "ansible_processor_count": 2,
        "ansible_processor_nproc": 2,
        "ansible_processor_threads_per_core": 1,
        "ansible_processor_vcpus": 2,
```

That's because libvirt (by default) exposes each "core" as a dedicated processor, even tho the physical machine outside only has one:
```
        "ansible_processor_cores": 12,
        "ansible_processor_count": 1,
        "ansible_processor_nproc": 14,
        "ansible_processor_threads_per_core": 1,
        "ansible_processor_vcpus": 12,
```

We can also use `processor_nproc`, not exactly sure what's better. The fact definition is at https://github.com/ansible/ansible/blob/d93f63ecfbd00712065db74586b6724fed8cb1b5/lib/ansible/module_utils/facts/hardware/linux.py#L291-L294 if you want to dive into that.

The system above has an Intel Ultra 7 165U CPU, which has 12 physical cores, and two of those are capable of hyperthreading, which is exposed as a "virtual" core in linux, so you get nproc=14, while cores and vcpus remains at 12.

evgeni: Mhh, yeah, looking at that code in Ansible, `nproc` is our safest bet here.

archanaserver: yeah `nproc` does seem like the safest default in this case

src/vars/defaults.yml
```diff
@@ -1,3 +1,4 @@
 ---
 certificate_source: default
 database_mode: internal
+deployment_mode: medium
```
evgeni, line 4: this needs to be exposed in metadata for `deploy` and `checks` (probably again with a dedicated `_tuning_mode` file, like we have for `database_mode` etc).

Also, I think this should be called `tuning`, not `deployment`.

@ehelms @ekohl WDYT?

archanaserver:
Looks like the image pull `quay.io/sclorg/postgresql-13-c9s:latest` fails with a 404.

src/vars/defaults.yml
evgeni: It will be a requirement for https://github.com/theforeman/foremanctl/issues/51

As today we don't do anything with those numbers, I wouldn't block this PR on it.

(I think I'd still call the variable `tuning` or `tuning_profile` and keep only one `default.yml` as the profile, we can then add the others once we have the code to actually deploy different configurations)

src/roles/check_system_requirements/tasks/main.yaml
```diff
@@ -0,0 +1,16 @@
+- name: Load tuning thresholds based on tuning_profile
+  ansible.builtin.include_vars:
+    file: "../../vars/tuning/{{ tuning_profile | default('medium') }}.yaml"
```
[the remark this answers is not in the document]
archanaserver: updated now!

src/roles/check_system_requirements/tasks/main.yaml
```diff
@@ -0,0 +1,16 @@
+- name: Load tuning thresholds based on tuning_profile
+  ansible.builtin.include_vars:
+    file: "../../vars/tuning/{{ tuning_profile ) }}.yaml"
```
evgeni, line 3:
```suggestion
    file: "../../vars/tuning/{{ tuning_profile }}.yaml"
```

PR #221, opened 2025-08-04
Add database connectivity checks for Candlepin and Pulp

archanaserver:
Depends on https://github.com/theforeman/foremanctl/pull/210

Extends the existing database connectivity checks to include Candlepin and Pulp when database_mode=external.

### DIFF
--- a/src/roles/check_database_connection/tasks/check.yaml
+++ b/src/roles/check_database_connection/tasks/check.yaml
@@ -1,0 +1,19 @@
+- name: Check database connectivity to {{ db_item.name }}
+  community.postgresql.postgresql_ping:
+    login_host: "{{ db_item.host }}"
+    login_user: "{{ db_item.user }}"
+    login_password: "{{ db_item.password }}"
+    login_db: "{{ db_item.dbname }}"
+    ca_cert: "{{ db_item.ca | default(omit) }}"
+    ssl_mode: "{{ db_item.sslmode | default(omit) }}"
+  register: check_database_connection_ping_result
+  ignore_errors: true
+
+- name: Assert database is reachable for  {{ db_item.name }}
+  ansible.builtin.assert:
+    that:
+      - check_database_connection_ping_result.is_available
+    fail_msg: >
+      Cannot connect to {{ db_item.name }} database '{{ db_item.dbname }}' at {{ db_item.host }}.
+      Please verify the database host, port, name, user, and password.
+      Error: {{ check_database_connection_ping_result.conn_err_msg | default('No error message available.') }}
--- a/src/roles/check_database_connection/tasks/main.yaml
+++ b/src/roles/check_database_connection/tasks/main.yaml
@@ -1,66 +1,31 @@
 ---
-- name: Check Foreman database connectivity and credentials
-  community.postgresql.postgresql_ping:
-    login_host: "{{ foreman_database_host }}"
-    login_user: "{{ foreman_database_user }}"
-    login_password: "{{ foreman_database_password }}"
-    login_db: "{{ foreman_database_name }}"
-    ca_cert: "{{ foreman_database_sslrootcert | default(omit) }}"
-    ssl_mode: "{{ foreman_database_sslmode | default(omit) }}"
-  register: check_database_connection_ping_result
-  when: database_mode == 'external'
-  ignore_errors: true
+- name: Check DB
+  ansible.builtin.include_tasks: check.yaml
+  no_log: true
+  loop:
+    - name: Foreman
+      host: "{{ foreman_database_host }}"
+      user: "{{ foreman_database_user }}"
+      password: "{{ foreman_database_password }}"
+      dbname: "{{ foreman_database_name }}"
+      ca_cert: "{{ foreman_database_sslrootcert | default(omit) }}"
+      sslmode: "{{ foreman_database_sslmode | default(omit) }}"
 
-- name: Assert Foreman database is reachable
-  ansible.builtin.assert:
-    that:
-      - check_database_connection_ping_result.is_available
-    fail_msg: >
-      Cannot connect to Foreman database '{{ foreman_database_name }}' at {{ foreman_database_host }}.
-      Please verify the database host, port, name, user, and password.
-      Error: {{ check_database_connection_ping_result.msg | default('No error message available.') }}
-  when: database_mode == 'external'
+    - name: Candlepin
+      host: "{{ candlepin_database_host }}"
+      user: "{{ candlepin_database_user }}"
+      password: "{{ candlepin_database_password }}"
+      dbname: "{{ candlepin_database_name }}"
+      ca_cert: "{{ candlepin_database_ssl_ca | default(omit) }}"
+      sslmode: "{{ candlepin_database_ssl_mode | default(omit) }}"
 
-- name: Check Candlepin database connectivity and credentials
-  community.postgresql.postgresql_ping:
-    login_host: "{{ candlepin_database_host }}"
-    login_user: "{{ candlepin_database_user }}"
-    login_password: "{{ candlepin_database_password }}"
-    login_db: "{{ candlepin_database_name }}"
-    ca_cert: "{{ candlepin_database_ssl_ca | default(omit) }}"
-    ssl_mode: "{{ candlepin_database_ssl_mode | default(omit) }}"
-  register: check_database_connection_candlepin_ping_result
-  when: database_mode == 'external'
-  ignore_errors: true
-
-- name: Assert Candlepin database is reachable
-  ansible.builtin.assert:
-    that:
-      - check_candlepin_db_ping_result.is_available
-    fail_msg: >
-      Cannot connect to Candlepin database '{{ candlepin_database_name }}' at {{ candlepin_database_host }}.
-      Please verify the database host, port, name, user, and password.
-      Error: {{ check_database_connection_candlepin_ping_result.msg | default('No error message available.') }}
-  when: database_mode == 'external'
-
-- name: Check Pulp database connectivity and credentials
-  community.postgresql.postgresql_ping:
-    login_host: "{{ pulp_database_host }}"
-    login_user: "{{ pulp_database_user }}"
-    login_password: "{{ pulp_database_password }}"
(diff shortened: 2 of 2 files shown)


src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -19,3 +19,45 @@
...
+  when: database_mode == 'external'
+  ignore_errors: true
+
+- name: Assert Candlepin database is reachable
+  ansible.builtin.assert:
+    that:
+      - check_candlepin_db_ping_result.is_available
+    fail_msg: >
+      Cannot connect to Candlepin database '{{ candlepin_database_name }}' at {{ candlepin_database_host }}.
+      Please verify the database host, port, name, user, and password.
```
ShimShtein, line 41: Can we get the error from the ping task too? It would be much appreciated.

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -0,0 +1,66 @@
+---
+- name: Check Foreman database connectivity and credentials
+  community.postgresql.postgresql_ping:
```
ShimShtein, line 3: Does it mean that we need to make sure this module is present on the machine when the command runs?

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -0,0 +1,66 @@
...
+  ignore_errors: true
+
+- name: Assert Pulp database is reachable
+  ansible.builtin.assert:
+    that:
+      - check_pulp_db_ping_result.is_available
+    fail_msg: >
+      Cannot connect to Pulp database '{{ pulp_database_name }}' at {{ pulp_database_host }}.
+      Please verify the database host, port, name, user, and password.
+      Error: {{ check_pulp_db_ping_result.msg | default('No error message available.') }}
```
ShimShtein, line 65: Shouldn't it be 
```suggestion
      Error: {{ check_database_connection_pulp_ping_result.msg | default('No error message available.') }}
```

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -0,0 +1,66 @@
...
+  ignore_errors: true
+
+- name: Assert Candlepin database is reachable
+  ansible.builtin.assert:
+    that:
+      - check_candlepin_db_ping_result.is_available
+    fail_msg: >
+      Cannot connect to Candlepin database '{{ candlepin_database_name }}' at {{ candlepin_database_host }}.
+      Please verify the database host, port, name, user, and password.
+      Error: {{ check_candlepin_db_ping_result.msg | default('No error message available.') }}
```
ShimShtein, line 43: Shouldn't it be:
```suggestion
      Error: {{ check_database_connection_candlepin_ping_result.msg | default('No error message available.') }}
```

src/roles/check_database_connection/tasks/main.yaml
archanaserver: Yes and `community.postgresql` collection is already included in our [src/requirements.yml](https://github.com/theforeman/foremanctl/blob/master/src/requirements.yml#L2), so it should be available when the checks run.

ShimShtein (commented):
One question, otherwise LGTM

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -0,0 +1,66 @@
...
+  when: database_mode == 'external'
+
+- name: Check Candlepin database connectivity and credentials
+  community.postgresql.postgresql_ping:
+    login_host: "{{ candlepin_database_host }}"
+    login_user: "{{ candlepin_database_user }}"
+    login_password: "{{ candlepin_database_password }}"
+    login_db: "{{ candlepin_database_name }}"
+    ca_cert: "{{ candlepin_database_ssl_ca | default(omit) }}"
+    ssl_mode: "{{ candlepin_database_ssl_mode | default(omit) }}"
```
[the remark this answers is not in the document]
evgeni: https://github.com/theforeman/foremanctl/blob/1db58d48c8e4cfea39b5fb8305e809a1fb92dc32/src/vars/database.yml#L19-L20
they are inherited from the main vars, as we do not want users to set *different* db servers for different services -- only different username/password

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -15,7 +15,52 @@
...
+
+- name: Assert Pulp database is reachable
+  ansible.builtin.assert:
+    that:
+      - check_pulp_db_ping_result.is_available
+    fail_msg: >
+      Cannot connect to Pulp database '{{ pulp_database_name }}' at {{ pulp_database_host }}.
+      Please verify the database host, port, name, user, and password.
+      Error: {{ check_database_connection_pulp_ping_result.msg | default('No error message available.') }}
+  when: database_mode == 'external'
```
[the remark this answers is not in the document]
archanaserver: on it, i can't able to push this change some git issue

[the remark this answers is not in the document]
archanaserver: @evgeni everytime it is saying everything is up to date, can you help?

archanaserver:
> But I wonder if that would _actually_ make it better -- shorter (in terms of lines) for sure, but better to read? WDYT?

Thanks for pointing this out @evgeni ! I really like the approach of using a loop with `include_tasks` to DRY up the repeated blocks. I was considering of doing this, once it works! so here it is, you can see current changes, I tried implementing it.

src/roles/check_database_connection/tasks/check.yaml
```diff
@@ -0,0 +1,19 @@
+- name: Check {{ db_item.name }} database connectivity
```
evgeni, line 1: Ansible lint doesn't like it if you use `{{ item }}` inside the `name` (at the end it's fine with it):
```
Jinja templates should only be at the end of 'name'
```

I must admit I am not a huge fan of this rule, so would be OK if we just disable it in our config.

OTOH we can rewrite this as:
```suggestion
- name: Check database connectivity to {{ db_item.name }}
```

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -1,21 +1,30 @@
...
-    login_password: "{{ foreman_database_password }}"
-    login_db: "{{ foreman_database_name }}"
-    ca_cert: "{{ foreman_database_sslrootcert | default(omit) }}"
-    ssl_mode: "{{ foreman_database_sslmode | default(omit) }}"
-  register: check_database_connection_ping_result
-  when: database_mode == 'external'
-  ignore_errors: true
+- name: Check DB
+  ansible.builtin.include_tasks: check.yml
+  loop:
```
evgeni, line 4:
```suggestion
  no_log: true
  loop:
```

the loop contains passwords (and we can't avoid that), so we should ensure the contents of the loop variable don't get logged.

before:
```
TASK [check_database_connection : Check DB] *******************************************************************************************************************************************************************************************
included: /usr/share/foremanctl/src/roles/check_database_connection/tasks/check.yaml for localhost => (item={'name': 'Foreman', 'host': '127.0.0.1', 'user': 'foreman', 'password': 'CHANGEME', 'db': 'foreman', 'ca_cert': 'None', 'ssl_mode': 'disable'})
included: /usr/share/foremanctl/src/roles/check_database_connection/tasks/check.yaml for localhost => (item={'name': 'Candlepin', 'host': '127.0.0.1', 'user': 'candlepin', 'password': 'CHANGEME', 'db': 'candlepin', 'ca_cert': 'None', 'ssl_mode': 'disable'})
included: /usr/share/foremanctl/src/roles/check_database_connection/tasks/check.yaml for localhost => (item={'name': 'Pulp', 'host': '127.0.0.1', 'user': 'pulp', 'password': 'CHANGEME', 'db': 'pulp', 'ca_cert': 'None', 'ssl_mode': 'disable'})
```

after:
```
TASK [check_database_connection : Check DB] *******************************************************************************************************************************************************************************************
included: /usr/share/foremanctl/src/roles/check_database_connection/tasks/check.yaml for localhost => (item=(censored due to no_log))
included: /usr/share/foremanctl/src/roles/check_database_connection/tasks/check.yaml for localhost => (item=(censored due to no_log))
included: /usr/share/foremanctl/src/roles/check_database_connection/tasks/check.yaml for localhost => (item=(censored due to no_log))
```

archanaserver:
i can't see checks, what happens?

src/roles/check_database_connection/tasks/main.yaml
```diff
@@ -1,21 +1,31 @@
...
-  ignore_errors: true
+- name: Check DB
+  ansible.builtin.include_tasks: check.yaml
+  no_log: true
+  loop:
+    - name: Foreman
+      host: "{{ foreman_database_host }}"
+      user: "{{ foreman_database_user }}"
+      password: "{{ foreman_database_password }}"
+      db: "{{ foreman_database_name }}"
```
evgeni, line 10: you changed this to `dbname` in`check.yaml`, but it's still `db` here

src/roles/check_database_connection/tasks/check.yaml
```diff
@@ -0,0 +1,19 @@
...
+  ignore_errors: true
+
+- name: Assert database is reachable for  {{ db_item.name }}
+  ansible.builtin.assert:
+    that:
+      - check_database_connection_ping_result.is_available
+    fail_msg: >
+      Cannot connect to {{ db_item.name }} database '{{ db_item.dbname }}' at {{ db_item.host }}.
+      Please verify the database host, port, name, user, and password.
+      Error: {{ check_database_connection_ping_result.msg | default('No error message available.') }}
```
evgeni, line 19:
```suggestion
      Error: {{ check_database_connection_ping_result.conn_err_msg | default('No error message available.') }}
```

According to https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_ping_module.html#return-values it should store the error (if any) in `conn_err_msg`

evgeni (approved):

evgeni:
@ShimShtein did you also want to do a review round, or were you already happy?

ShimShtein:
LGTM, I don't have permissions to merge though.

PR #225, opened 2025-08-15
Add initial admin username & password parameters

stejskalleos:
New foremanctl options: foreman-admin-username and foreman-admin password.
Default password for admin user is autogenerated.

### DIFF
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -83,8 +83,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--certs-server-ca-cert` | | certs | server_ca_cert |
 | `--certs-update-server` | Parameter to mark server certs for update | foreman-installer | No |
 | `--certs-reset` | Parameter to reset all certificates to default | foreman-installer | No |
-| `--foreman-initial-admin-password` | | |
-| `--foreman-initial-admin-username` | | |
 | `--foreman-initial-location` | | |
 | `--foreman-initial-organization` | | |
 | `--foreman-ipa-authentication` | | |
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -47,6 +47,8 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--pulp-database-name` | Name of the Pulp database | `--foreman-proxy-content-pulpcore-postgresql-db-name` |
 | `--pulp-database-user` | Owner of the Pulp database | `--foreman-proxy-content-pulpcore-postgresql-user` |
 | `--pulp-database-password` | Password for Pulp database | `--foreman-proxy-content-pulpcore-postgresql-password` |
+| `--foreman-initial-admin-username` | Initial username for the admin user | `--foreman-initial-admin-username` |
+| `--foreman-initial-admin-password` | Initial password for the admin user | `--foreman-initial-admin-password` |
 
 #### Certs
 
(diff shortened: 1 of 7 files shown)


src/playbooks/deploy/deploy.yaml
```diff
@@ -49,6 +51,10 @@
           - python3-libsemanage
           - python3-psycopg2
           - python3-requests
+  post_tasks:
```
ehelms, line 54: Can you move this after `roles:` so that it shows the execution order?

src/vars/foreman.yml
```diff
@@ -0,0 +1,3 @@
+---
+foreman_admin_username: admin
+foreman_admin_password: "{{ lookup('ansible.builtin.password', '/tmp/passwordfile', chars=['ascii_letters']) }}"
```
ehelms, line 3: I think we should make this the default in the role and then force users to of this to supply `changeme` to avoid accidents where an installation ends up with the simple well known password.

evgeni, line 3: We should not store the password in `/tmp`.

We could expose https://github.com/theforeman/obsah/blob/ef8e816ea27c5a36368011e89ffe89f00b20ce16/obsah/__init__.py#L251-L256 in an variable and let people write there?

evgeni: 1.4.0 now contains `obsah_state_path`

evgeni:
> 🤔 GitHub CI is changing `--foreman-initial-admin-password=changeme` to the `--foreman-initial-admin-***`.
> 
> @evgeni @ehelms any ideas on how to make it work?

It's not. That's just the output hiding the "password" part. See inline comment for the actual bug :)

stejskalleos:
... and it's :green_apple:

docs/parameters.md
```diff
@@ -47,6 +47,8 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--pulp-database-name` | Name of the Pulp database | `--foreman-proxy-content-pulpcore-postgresql-db-name` |
 | `--pulp-database-user` | Owner of the Pulp database | `--foreman-proxy-content-pulpcore-postgresql-user` |
 | `--pulp-database-password` | Password for Pulp database | `--foreman-proxy-content-pulpcore-postgresql-password` |
+| `--foreman-initial-admin-username` | Initial username for the admin user | `--foreman-initial-admin-username` |
+| `--foreman-initial-admin-password` | Initial password for the admin user | `--foreman-initial-admin-password` |
```
ehelms, line 51: Sorry I didn't catch this earlier, can you remove these entries from the Undetermined section?

https://github.com/theforeman/foremanctl/pull/225/files#diff-7ad93860b53d91cc603737c90612d6533ce53edf882eda504ddd476df97c9a2fR86-R87

src/vars/foreman.yml
```diff
@@ -0,0 +1,4 @@
+---
+foreman_admin_passwd_file: "{{ obsah_state_path }}/foreman-admin-init-passwd"
+foreman_initial_admin_username: admin
+foreman_initial_admin_password: "{{ lookup('ansible.builtin.password', lookup('ansible.builtin.vars', 'foreman_admin_passwd_file'), chars=['ascii_letters']) }}"
```
evgeni, line 4: Isn't this the same?
```suggestion
foreman_initial_admin_password: "{{ lookup('ansible.builtin.password', foreman_admin_passwd_file, chars=['ascii_letters']) }}"
```

PR #235, opened 2025-09-19
Debian 13

evgeni:


development/playbooks/setup-repositories/setup-repositories.yaml
```diff
@@ -9,10 +9,30 @@
...
+        name: epel-release
+        state: installed
+      when:
+        - ansible_distribution_major_version == '10'
+
+    - name: Enable hammer-el10 copr
+      community.general.copr:
+        host: copr.fedorainfracloud.org
+        state: enabled
+        name: evgeni/hammer-el10
```
evgeni, line 35: you may know this copr from fixes like https://github.com/theforeman/foreman-packaging/pull/12499 and https://github.com/theforeman/foreman-packaging/pull/12500

evgeni: (this needs to be read in the voice of Troy McClure)

development/playbooks/setup-repositories/setup-repositories.yaml
```diff
@@ -9,10 +9,23 @@
         name: theforeman.operations.puppet_repositories
       vars:
         foreman_puppet_repositories_version: "8"
+      when:
+        - ansible_distribution_major_version == '9'
```
ekohl, line 13: Should this be when the certificate source is the installer instead?

evgeni: Yepp, but this detail is not passed to the "setup repos" playbook, and I was lazy

src/roles/httpd/tasks/main.yml
```diff
@@ -1,20 +1,28 @@
...
 - name: Set httpd_can_network_connect so Apache can connect to Puma and Gunicorn
   ansible.posix.seboolean:
     name: httpd_can_network_connect
     state: true
     persistent: true
+  when:
+    - ansible_selinux.status == 'enabled'
+
+- name: Enable required modules
+  community.general.apache2_module:
```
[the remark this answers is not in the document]
ekohl: We may also want to enable the event MPM in a similar way.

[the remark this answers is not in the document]
evgeni: Mostly because it's a dependency we don't yet have

src/roles/httpd/tasks/main.yml
```diff
@@ -45,11 +53,11 @@
...
-    dest: /etc/httpd/conf.d/foreman-ssl.conf
+    dest: "{{ httpd_conf_path }}/foreman-ssl.conf"
     mode: "0644"
 
 - name: Start Apache httpd
   ansible.builtin.service:
-    name: httpd
-    state: started
+    name: "{{ httpd_service }}"
+    state: restarted
```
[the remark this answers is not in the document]
evgeni: There is https://github.com/theforeman/foremanctl/issues/92

But we could make it broader

src/roles/httpd/tasks/main.yml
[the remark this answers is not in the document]
ekohl: It also mentions compatibility:
> This does not work on RedHat-based distributions. It does work on Debian- and SuSE-based distributions. Whether it works on others depend on whether the C(a2enmod) and C(a2dismod) tools are available or not.

So configuring Apache with Ansible is certainly worse than what we had with our Puppet module.

jakduch:
Hi @evgeni, great work on the Debian 13 support! I built on top of your branch and found a few things that could be improved. I created PR #427 targeting your `debian` branch with the following commits:

**Bug fixes:**
- Fix duplicate `conf.d/` in `foreman-vhost.conf.j2` — `httpd_conf_path` already contains `conf.d` (RedHat) / `conf-enabled` (Debian), so the extra `conf.d/` segment in the `IncludeOptional` path on line 5 results in an invalid path. The SSL vhost was already correct.
- Fix spurious space in `cleanup.yml` line 4: `{{ httpd_conf_path}} /05-` should be `{{ httpd_conf_path }}/05-`

**Debian adaptations:**
- Make httpd IPA external auth OS-aware — use `libapache2-mod-*` packages and `community.general.apache2_module` (a2enmod/a2dismod) on Debian instead of manual LoadModule conf files
- Add OS-specific vars for hammer role (`ruby-hammer-cli-*` on Debian vs `rubygem-hammer_cli_*` on RedHat) and load them via `include_vars`
- Add Foreman repository setup for Debian in the development playbook (using `theforeman.operations.foreman_repositories`)
- Enable hammer deployment on Debian in CI (remove the centos-only condition)

Feel free to cherry-pick any of these commits directly into your branch, or merge #427 if you prefer. Happy to adjust anything!

evgeni:
Thanks, I've picked up the bugfixes, will have a look at the enhancements later.

bochi:
works very well, tested on Ubuntu 26 and Debian 13!

src/roles/httpd/tasks/main.yml
```diff
@@ -117,6 +123,6 @@
 
 - name: Start Apache httpd
   ansible.builtin.service:
-    name: httpd
+    name: "{{ httpd_service }}"
     state: started
     enabled: true
```
evgeni, line 128: Services are enabled by default on Debian, so this line doesn't trigger an "enable" action, and thus the "Add httpd to foreman.target" is never applied (as `systemctl enable` is where the symlinks are created for the targets). 😿

PR #237, opened 2025-09-19
Don't restart Candlepin every time a deployment runs

evgeni:


evgeni:
> Late review, but is there a trigger to restart the service if anything changes? On my phone so awkward to check

There is not, but also not for the others. I was working on that, and wanted a consistent baseline first.

PR #242, opened 2025-09-29
adjust checked memory inside the check, not in the profile

evgeni:


ekohl (approved):
This matches our current installer

PR #246, opened 2025-10-07
Set Pulp SECRET_KEY

arvind4501:
It sets SECRET_KEY for Django. used [this](https://pulpproject.org/pulpcore/docs/admin/reference/settings/?h=secret#secret_key) to generate the key

### DIFF
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -36,6 +36,26 @@
     name: pulp-db-password
     data: "{{ pulp_database_password }}"
 
+- name: Generate Django secret key
+  ansible.builtin.shell: |
+    python3 -c 'import secrets; chars = "abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)"; \
+    print("".join(secrets.choice(chars) for i in range(50)))' > /var/lib/pulp/secret_key
+  args:
+    creates: /var/lib/pulp/secret_key
+
+- name: Set secret key file permissions
+  ansible.builtin.file:
+    path: /var/lib/pulp/secret_key
+    owner: root
+    group: root
+    mode: '0600'
+
+- name: Create django pulp secret key secret
+  containers.podman.podman_secret:
+    state: present
+    name: pulp-secret-key
+    path: /var/lib/pulp/secret_key
+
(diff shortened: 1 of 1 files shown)


src/roles/pulp/tasks/main.yaml
```diff
@@ -36,6 +36,26 @@
...
+    path: /var/lib/pulp/secret_key
+    owner: root
+    group: root
+    mode: '0600'
+
+- name: Create django pulp secret key secret
+  containers.podman.podman_secret:
+    state: present
+    name: pulp-secret-key
+    path: /var/lib/pulp/secret_key
```
evgeni, line 57: bikeshedding question: would it make sense to call the secret `pulp-django-secret-key` (and the file `django_secret_key`), we did that in Puppet and `django` would be fitting for the "app" part that we define in https://github.com/theforeman/foremanctl/blob/master/DEVELOPMENT.md#naming-convention

(again, not blocking, but food for discussion)

arvind4501: sure, we should ideally follow the naming-convention. and that also has extension for the file, so it makes

my secret `pulp-django-secret-key`  and file as `django_secret_key.key` , does it make sense to add `.key` to the file?

src/roles/pulp/tasks/main.yaml
```diff
@@ -36,6 +36,26 @@
     name: pulp-db-password
     data: "{{ pulp_database_password }}"
 
+- name: Generate Django secret key
+  ansible.builtin.shell: |
+    python3 -c 'import secrets; chars = "abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)"; \
+    print("".join(secrets.choice(chars) for i in range(50)))' > /var/lib/pulp/secret_key
```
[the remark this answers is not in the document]
arvind4501: and `bash -c 'openssl rand -base64 50 | tr \"+/\" \"-_\"` this does create secret with `~68` chars, is it ok? the 50 here is for bytes(50 bytes of random data)?

src/roles/pulp/tasks/main.yaml
evgeni: I see no good reason to add `.key` here

PR #248, opened 2025-10-10
Share /var/run/foreman between Foreman and Dynflow containers so that Manifest can be imported

evgeni:


stejskalleos (approved):
🍏 LGTM

Manifest import works and tests are green,
let's get this in.

ehelms (approved):

PR #252, opened 2025-10-15
Add quadlet timer for foreman recurring task (reports:daily)

archanaserver:
This is for the prototype implementation of recurring task support in foremanctl using quadlet-based systemd timers

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -199,3 +199,28 @@
     username: "{{ foreman_initial_admin_username }}"
     password: "{{ foreman_initial_admin_password }}"
     validate_certs: false
+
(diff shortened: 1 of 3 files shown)


archanaserver:
> Is the general idea that we define one systemd timer per (old) cronjob and that timer launches a temporary container, runs the magic in it, and then finishes?

Hmm exactly, so with each legacy cron job maps to its own systemd timer that spins up a short-lived container, runs the rake task (like foreman-rake reports:daily), and exits. I believe it’s a simple one-to-one replacement for now, and later we can look into integrating this into foreman-tasks or a more HA-friendly model. But WDYT? @evgeni

archanaserver:
As per our discussion, I’ve opened a new PR #288 addressing this. Since this one served as a PoC, I’ll go ahead and close it.

PR #253, opened 2025-10-15
Serve static assets via apache running on host

arvind4501:
This contains a initial work for making apache(running on host) serve assets of foreman container.

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -80,7 +80,7 @@
 
 - name: Extract public assets
   ansible.builtin.shell:
-    cmd: podman exec foreman-extract-temp tar -chC /usr/share/foreman/public . | tar -xC {{ foreman_public_assets_dir }}/
+    cmd: set -o pipefail && podman exec foreman-extract-temp tar -chC /usr/share/foreman/public . | tar -xC {{ foreman_public_assets_dir }}/
   changed_when: false
 
 - name: Remove temporary container


ekohl:
Having them served dynamically isn't that bad IMHO. Especially if it sends the right headers. That way you guarantee the files are in sync. Not sure if we send those headers though.

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -63,10 +63,64 @@
...
+    RewriteCond %{REQUEST_FILENAME}.gz -s
+    RewriteRule ^(.+) $1.gz [L]
+  </LocationMatch>
+
   ## Proxy rules
   ProxyRequests Off
   ProxyPreserveHost On
   ProxyAddHeaders On
+  ProxyPass /assets !
+  ProxyPass /webpack !
```
ShimShtein: `/webpack` is basically a static assets folder. Why do we want to `ProxyPass` it? We can serve it directly from Apache too.
And the same thing goes for the assets folder.

evgeni: [`!` means "do not proxy"](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass): 
> The ! directive is useful in situations where you don't want to reverse-proxy a subdirectory, e.g.

src/roles/foreman/tasks/main.yaml
```diff
@@ -74,6 +96,8 @@
     sdnotify: true
     network: host
     hostname: "{{ ansible_fqdn }}"
+    volumes:
+      - "{{ foreman_public_assets_dir }}:/usr/share/foreman/public:z"
```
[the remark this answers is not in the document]
evgeni: No need to mount it, indeed.

ekohl:
> > Taking a step back and looking forward to where Rails is headed. Today Foreman is on Rails 7, but Rails 8 has started to include a [Dockerfile.tt](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt) template. We don't use that, but it's good to look at it. They use [thruster](https://github.com/basecamp/thruster) as a reverse proxy in the container. It wraps the Puma process and can serve the static assets efficiently. https://rubyonrails.org/2024/11/7/rails-8-no-paas-required has a bit more context.
> 
> When I've first seen Thruster, my initial reaction was "eww", and it still holds.
> 
> That starts with the fact that it's distributed as a Ruby gem, but in reality is a Go tool, so the gem just contains some rando pre-built binary, you're supposed to execute in prod.
> 
> But even ignoring that deployment weirdness and look at the actual code, it's basically a reverse proxy with a bit of caching, something that Apache, NGINX, Caddy and friends already can do, but at the expense of using Go's `net/http` and not having decades of experience doing HTTP (like the mentioned projects).
> 
> Sure, it makes "single command" deployments easier, but not at a cost I am willing to accept.

That's good feedback. But I still think keeping the static files within a container is a more robust design. If we run into performance problems then I'd rather include a webserver then I think it's better.

> > While it may require a bit more engineering in other parts of the stack, I think that is a much more sustainable model than what is proposed here. Since we already proxy all traffic to the Foreman container, it's a drop in replacement for letting Puma serve the static files.
> 
> I am sure we can configure `mod_cache` instead if we want to.

That is an interesting thought. We want to send long expire headers anyway (so browsers can cache them). Looks like you can set `config.public_file_server.headers` in `config/environments/production.rb`:
```ruby
config.public_file_server.headers = {
  'Cache-Control' => "public, max-age=#{1.year.to_i}, immutable",
}
```

That should be a pretty cheap win. https://appsintheopen.com/posts/64-rails-page-caching-cache-headers-and-thruster also points to a solution that limits it to just assets if we need to.

> > > The IOP services need a way to serve static assets. If Rails serves Foreman's assets, how would you expect those other services assets to be served?
> > 
> > 
> > I would expect those services to take care of their own assets. Now I don't know what is the best way for IoP since I haven't looked at that.
> > However, mixing Foreman and other assets into a common directory was a plan I wouldn't have liked in the first place. Each application should have its own namespace that can't collide. If it means we need to move Foreman's assets from `/assets` to `/assets/foreman` (or similar) then I'm sure we can make that work. Then you can have `/assets/advisor` and be sure it never tries to serve the same files.
> 
> That's roughly what the current IOP setup does, it serves them from `/assets/apps/<app_name>`, leaving the top-level `/assets` to Foreman, see my graphs above.

I did base it on your graphs. I just didn't like that Foreman is on the top level. Is it time to bring back Foreman sub-uri support again so it becomes `https://foreman.example.com/foreman` again? ;) (https://github.com/theforeman/puppet-foreman/blob/f69a26f3a3cffb7f525f91678afa297b8d6ea28b/manifests/config/apache.pp#L150-L165)

ShimShtein:
> But I still think keeping the static files within a container is a more robust design.

I would argue that for this exact reason, we should use an apache server in front of our assets - this will make sure we can serve each asset in the most suitable way for that specific asset. For example if tomorrow we will want to split the UI out of the Foreman Ruby container, we will be able to do so and it will be just an implementation detail.

src/roles/foreman/tasks/main.yaml
```diff
@@ -66,6 +66,28 @@
     - Restart foreman
     - Restart dynflow-sidekiq@
 
+- name: Create directory for Foreman public assets
```
stejskalleos: Can we extract it into a sub-task? The `main.yml` is getting bigger and bigger.

arvind4501: yes, we could, i think once we have green flag for this approch  it makes more sense to handle these. correct me if i am wrong?

PR #254, opened 2025-10-16
DEV - run dnf update for the quadlet VM

stejskalleos:


stejskalleos:
> annoying amount of time to local testing and CI

:thinking: I could remove the `provision` command from the `forge vms start` and leave it up to dev.

ekohl:
Should we recommend updating the base box before provisioning?

ehelms:
> Our docs state one is supposed to use an "uptodate" system. Shouldn't CI reflect that?

Our current docs do, but that is a good question if that is still required in a container based way since we are not installing all of these RPMs. 

The best answer for CI would be to have an image that is nearly always up to date so that we are not eating the update cost every single CI run.

PR #259, opened 2025-10-22
Add Foreman development environment

ehelms:
This adds a way to deploy a development environment similar to how the devel boxes in Forklift work. This deploys the backend services as containers to match production, and clones the source code directly to the VM and installs and configures.

In my testing, timing with this to get to a working development setup:

Forklift: 29 minutes
foremanctl: 16 minutes

I think this is a good starting baseline, as it provides a way to develop using our container-based installation that is similar to how developers work today. I'll be exploring other development setups that can be follow-ons to this method. The important bit for me is getting an environment that works within the context of this repository and our containers.

I did try running a development container and mounting the code into the development container. Within the VM environment this proved to be incredibly slow, likely due to the amount of files being shared and the I/O overhead.


development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -0,0 +1,58 @@
...
+            - Direct Rails: http://localhost:3000 (when running)
+
+          Credentials:
+            - Username: {{ foreman_development_admin_user | default('admin') }}
+            - Password: {{ foreman_development_admin_password | default('changeme') }}
+
+          Next steps:
+            - SSH into the VM
+            - Navigate to /home/vagrant/foreman
+            - Run: bundle exec rails server -p 3000 -b 0.0.0.0
```
adamruzicka: Should we also run the webpack watcher or is it handled somehow somewhere else?
```suggestion
            - Run: bundle exec foreman start webpack
            - In another terminal run: bundle exec rails server -p 3000 -b 0.0.0.0
```

ehelms: I thought `foreman start` did this inherently? If not, then I'll definitely add that.

ehelms:
Updates support broker hosts now, for example:

```
ANSIBLE_ASK_PASS=true ./forge deploy-dev --target-host <host> --foreman-development-user ehelms
```

I opened a change to Obsah to support the password ask naturally: https://github.com/theforeman/obsah/pull/91

development/playbooks/deploy-dev/metadata.obsah.yaml
```diff
@@ -0,0 +1,14 @@
+---
+help: |
+  Deploy and manage Foreman development environment with git-based Foreman and containerized backend services.
+
+variables:
+  foreman_development_enabled_plugins:
+    help: List of plugins to enable from registry (comma-separated)
+    default: katello,foreman_remote_execution
```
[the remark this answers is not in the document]
ehelms: Thanks for trying that out, I had not yet. I'll fix it up.

adamruzicka:
If anyone wants to follow along and doesn't have a properly set up machine (like me), here's a handy scriptlet to run this within a container to set up a remote machine (remote from the container's point of view, chances are it could be the machine hosting the container)

Spawn a fedora 42 container
```
podman run -ti --rm fedora:42 bash
```

Inside
```
USER=<remote_ssh_username>
HOST=<remote_hostname>

dnf install -y git python3 python3-virtualenv make

git clone https://github.com/theforeman/foremanctl
cd foremanctl
git fetch origin refs/pull/259/head
git checkout FETCH_HEAD

echo "$HOST ansible_user=$USER" >inventories/$HOST

./setup-environment
source .venv/bin/activate
ANSIBLE_ASK_PASS=true ./forge deploy-dev \
    --target-host $HOST \
    --foreman-development-enabled-plugins="katello,foreman_remote_execution,foreman_ansible,foreman_openscap" \
    --foreman-development-user $USER
```

.github/workflows/devel.yml
```diff
@@ -0,0 +1,63 @@
+---
+name: Devel Test
+
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - 'container-images/**'
+      - '.github/workflows/container.yml'
```
ekohl: Following https://github.com/theforeman/foremanctl/commit/ef188c7ab78743f8d4a0ac67540ff343c88eae4c
```suggestion
```

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -0,0 +1,57 @@
...
+            owner: "{{ foreman_database_user }}"
+          - name: "{{ pulp_database_name }}"
+            owner: "{{ pulp_database_user }}"
+        postgresql_users:
+          - name: "{{ candlepin_database_name }}"
+            password: "{{ candlepin_database_password }}"
+          - name: "{{ foreman_database_name }}"
+            password: "{{ foreman_database_password }}"
+          - name: "{{ pulp_database_name }}"
+            password: "{{ pulp_database_password }}"
```
evgeni:
```suggestion
          - name: "{{ candlepin_database_user }}"
            password: "{{ candlepin_database_password }}"
          - name: "{{ foreman_database_user }}"
            password: "{{ foreman_database_password }}"
          - name: "{{ pulp_database_user }}"
            password: "{{ pulp_database_password }}"
```

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -0,0 +1,57 @@
...
+          - name: "foreman_development"
+            owner: "{{ foreman_database_user }}"
+          - name: "foreman_development_test"
+            owner: "{{ foreman_database_user }}"
+          - name: "{{ pulp_database_name }}"
+            owner: "{{ pulp_database_user }}"
+        postgresql_users:
+          - name: "{{ candlepin_database_name }}"
+            password: "{{ candlepin_database_password }}"
+          - name: "{{ foreman_database_name }}"
```
ekohl: I've found that Foreman needs to be able to recreate the test DB so in practice it needed permissions to do so. Not sure if there's a better way to make them the owner.

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -0,0 +1,57 @@
...
+            - Foreman UI: https://{{ ansible_fqdn }}
+            - Direct Rails: http://localhost:3000 (when running)
+
+          Credentials:
+            - Username: {{ foreman_development_admin_user | default('admin') }}
+            - Password: {{ foreman_development_admin_password | default('changeme') }}
+
+          Next steps:
+            - SSH into the VM
+            - Navigate to /home/{{ foreman_development_user }}/foreman
```
ekohl:
```suggestion
            - Navigate to {{ foreman_development_foreman_dir }}
```

evgeni:
Use setup python action and explicitly use Python 3.12?

src/requirements.txt
```diff
@@ -1,2 +1,3 @@
 obsah >= 1.4
+passlib
```
evgeni: what is passlib used for in prod?

evgeni: but that's in development, not in prod?

ehelms: Oh, you are right, I didn't notice that.

development/playbooks/deploy-dev/metadata.obsah.yaml
```diff
@@ -0,0 +1,17 @@
...
```
evgeni: I wonder how this will play together with https://github.com/theforeman/foremanctl/pull/188 :)

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -0,0 +1,58 @@
...
+          - name: "{{ foreman_development_database_name }}_test"
+            owner: "{{ foreman_database_user }}"
+          - name: "{{ pulp_database_name }}"
+            owner: "{{ pulp_database_user }}"
+        postgresql_users:
+          - name: "{{ candlepin_database_user }}"
+            password: "{{ candlepin_database_password }}"
+          - name: "{{ foreman_database_user }}"
+            password: "{{ foreman_database_password }}"
+            role_attr_flags: SUPERUSER
```
evgeni: where is this applied?

docs/development-environment.md
```diff
@@ -0,0 +1,160 @@
...
+
+2. **Access the environment:**
+   - SSH into the VM: `vagrant ssh`
+   - Navigate to Foreman directory: `cd /home/vagrant/foreman`
+   - Start Rails server: `bundle exec foreman start`
+
+3. **Access URLs:**
+   - Foreman UI: `http://$(hostname -f):3000` (development server)
+   - Production-style UI: `https://$(hostname -f)` (via Apache proxy)
+   - Direct Rails server: `http://localhost:3000` (when running locally)
```
evgeni: What's "locally" in this case?

evgeni:
I think overall this makes sense. I posted a ton of comments, but nothing that screams "whoa, don't do this".

Couple of architectural questions that jumped at me:
- What targets do we want to support here? In the vagrant case we default to CentOS Stream 9 boxes, which "just work" for a lot of things. But you also add a `target-host` that basically allows any arbitrary host to become one. And then you add some RHEL-isms that we can't verify in CI here.
- What is the purpose of the plugin settings YAMLs you provide, without any way to adjust them? (I see why we need a katello.yml, but otherwise?)

development/roles/foreman_development/defaults/main.yaml
```diff
@@ -0,0 +1,82 @@
...
+    name: "theforeman/foreman_remote_execution"
+    scm_revision: "master"
+    manage_repo: true
+  foreman_ansible:
+    name: "theforeman/foreman_ansible"
+    settings_template: "foreman_ansible.yaml.j2"
+    scm_revision: "master"
+    manage_repo: true
+  foreman_rh_cloud:
+    name: "RedHatInsights/foreman_rh_cloud"
```
evgeni: This looks like a hallucination ;)

development/roles/foreman_development/defaults/main.yaml
```diff
@@ -0,0 +1,82 @@
...
+    scm_revision: "master"
+    manage_repo: true
```
evgeni: I think by default we should not set neither `scm_version` (this defaulting to whatever the repo has as default), nor `manage_repo` (and having `true` as a default somewhere else).

development/roles/foreman_development/tasks/main.yaml
```diff
@@ -0,0 +1,232 @@
...
+    name: "codeready-builder-for-rhel-{{ ansible_distribution_major_version }}-{{ ansible_architecture }}-rpms"
+    state: enabled
+  when: ansible_distribution == 'RedHat'
+
+- name: Install development packages
+  ansible.builtin.package:
+    name: "{{ foreman_development_packages }}"
+    state: present
+
+- name: Disable firewall for development environment
```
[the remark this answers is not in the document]
ehelms: With some VMs provided through my use of the broker tool, I run into this every time. Which means I think we either:

1. disable it all the time
2. make this support configuring the firewall if it's present

What do you prefer?

docs/development-environment.md
```diff
@@ -0,0 +1,160 @@
...
+- **Pulp Smart Proxy Registration**: Automatically configures Pulp integration during deployment
+
+### Certificates
+
+Development certificates are copied to `/home/vagrant/foreman-certs/`:
+- `proxy_ca.pem` - CA certificate
+- `client_cert.pem` - Client certificate
+- `client_key.pem` - Client private key
+
+## Customization
```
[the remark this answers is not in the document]
ehelms: That's a good question, and not a part I had solved yet. The CLI parameters vs providing a file with variables. This feels like it will lean towards the latter for most use cases to reproduce environments often.

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -0,0 +1,58 @@
...
+          Access URLs:
+            - Foreman UI: https://{{ foreman_development_url }}
+            - Direct Rails: http://localhost:3000 (when running)
+
+          Credentials:
+            - Username: {{ foreman_development_admin_user }}
+            - Password: {{ foreman_development_admin_password }}
+
+          Next steps:
+            - SSH into the VM: ssh {{ foreman_development_user }}@{{ ansible_fqdn }}
```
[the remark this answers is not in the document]
ehelms: I was torn because `vagrant` won't be used in every case. There is likely no good universal answer here.

development/playbooks/deploy-dev/metadata.obsah.yaml
ehelms: Good question, I can explore switching when we get there. I mostly just want to get people to using this setup.

development/playbooks/deploy-dev/deploy-dev.yaml
ehelms: From my testing, it was being applied magically... but I've made it more explicit.

development/playbooks/deploy-dev/deploy-dev.yaml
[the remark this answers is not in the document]
evgeni: You could do some magic like you do with the `hosts` above: `target_host if target_host is defined and target_host != '' else 'quadlet'`

Or you just do:
```
- SSH into the system
    when using Vagrant: vagrant ssh quadlet
    when using a remote system: ssh {{ foreman_development_user }}@{{ ansible_fqdn }}
```

development/roles/foreman_development/tasks/main.yaml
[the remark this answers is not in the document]
evgeni: Configuring firewalls is hard (and then someone will come and complain we do not support `ferm` or `ufw` or whatever).

Both this and the `codeready` enablement somehow screams for a `./forge make-my-system-less-annoying` (there was a different adjective here before), but this is not good UX so… idk.

Wont block on this, but please note my disliking ;-)

development/roles/foreman_development/defaults/main.yaml
evgeni: Oh, I meant to have `default(true)` in `foreman_development_plugin_manage_repo: "{{ foreman_development_plugin_config.manage_repo | default(true) }}"` in `development/roles/foreman_development/tasks/plugin.yaml`, which you do already!

development/roles/foreman_development/tasks/main.yaml
```diff
@@ -0,0 +1,232 @@
...
+  ansible.builtin.systemd:
+    name: firewalld
+    state: stopped
+    enabled: false
+  failed_when: false
+
+- name: Create development user
```
bherrin3: Since we would like this to work as root by default on a machine that has no users, I would suggest when the user is `root`:

```
when: foreman_development_user != "root"
```

To avoid: 

>      TASK [foreman_development : Create development user]
>     ***************************
>   - 'fatal: [localhost]: FAILED! => {'
>   - '    "changed": false,'
>   - '    "name": "root",'
>   - '    "rc": 8'
>   - '}'
>   - ''
>   - 'MSG:'
>   - ''
>   - 'usermod: user root is currently used by process 1'
>   - ''
>   - ''
>   - >-

development/playbooks/deploy-dev/metadata.obsah.yaml
```diff
@@ -0,0 +1,18 @@
+---
+help: |
+  Deploy and manage Foreman development environment with git-based Foreman and containerized backend services.
+
+variables:
+  foreman_development_enabled_plugin:
+    help: Plugin to enable (can be used multiple times)
+    action: append
+    default: [katello, foreman_remote_execution]
```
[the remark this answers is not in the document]
evgeni: unless you explicitly want the git version, not the gem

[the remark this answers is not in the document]
ehelms: This came from https://github.com/theforeman/puppet-katello_devel/blob/master/manifests/config.pp#L54 where it was default set up via git.

.github/workflows/test.yml
```diff
@@ -108,6 +108,40 @@ jobs:
...
+        run: ./setup-environment
+      - name: Start VMs
+        run: |
+          ./forge vms start
+      - name: Configure repositories
+        run: |
+          ./forge setup-repositories
+      - name: Run deployment
+        run: |
+          ./forge deploy-dev --foreman-development-enabled-plugin foreman_ansible --foreman-development-enabled-plugin katello
```
evgeni: `katello` is part of the `default` definition, so you don't need to repeat it here

evgeni (approved):
two nitpicks, but otherwise LGTM

.github/workflows/test.yml
ehelms: That's how it should work, but not what I was encountering. Aaand it's because we do not actually support default values at the argparse level:

https://github.com/theforeman/obsah/blob/master/obsah/__init__.py#L383

I am assuming this is because we wanted to forge this information into the Ansible layer.

That leaves me with two options I can think of:

1. Change this behaviour in Obsah
2. Change to having variable inside the role for defaults and separate one for external CLI setting and then append the two together

evgeni: Right, why did I think we can set `default` in obsah? :thinking:

PR #266, opened 2025-10-28
Rely on smoker for the content tests

ekohl:
To test out https://github.com/theforeman/smoker/pull/46.


ekohl:
> I've debated using https://gist.github.com/ekohl/e487422d4aaad8ab176c19ec713b700a in the Vagrant setup to have working DNS.

https://github.com/theforeman/foremanctl/pull/269 does this and passes. Upside is no hacks, downside is that developers may not have the same environment.

ehelms:
@ekohl Is this a strategy you still think we should pursue?

ekohl:
Yes, I do. Mostly because we also run smoker on our RPMs so we'd gate earlier. Overall I think we should aim the test suite at:
* linting, static analysis, etc (tools like ansible-lint)
* unit test / integration test for smaller components (optional, tools like molecule)
* test foremanctl specific glue (the tests directory)
* run complete end-to-end that has to work regardless of the deployment method (smoker)

But that's just my opinion. I'd like some broader discussion on this.

PR #269, opened 2025-10-29
Set up DNS for libvirt

ekohl:
This sets up a DNS networking hook that reconfigures systemd-resolved to resolve the domain, if the libvirt network has one set. For that it also needs systemd-resolved to be installed.

Currently a draft to try out changes. If this works, https://github.com/theforeman/foremanctl/pull/266 can proceed and in other places we can also avoid ugly hacks.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,16 +28,9 @@ jobs:
         with:
           python-version: '3.12'
       - name: Setup libvirt for Vagrant
-        uses: voxpupuli/setup-vagrant@v0
-      - name: Install systemd-resolved
-        run: |
-          sudo apt-get install -y --no-install-recommends systemd-resolved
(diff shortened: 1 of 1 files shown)


ekohl:
Connecting to `qemu:///system` worked so now it can ping `quadlet.example.com` via proper DNS. Obviously this PR right now is only a proof of concept, but I think it can easily be worked into a real thing.

Question is, where do we integrate this? voxpupuli/setup-vagrant?

ekohl:
This is now updated to set the management domain in `Vagrantfile` so you don't need to modify the network anymore. That makes it a fairly clean.

I've submitted it as https://github.com/voxpupuli/setup-vagrant/pull/1. I've also learned systemd-resolved is present by default (at least in 24.04) so that can be skipped.

PR #272, opened 2025-10-30
Add option to scale workers for puma service

arvind4501:
This PR adds ability to change puma workers for scaling. It is a initial workaround

### DIFF
--- a/src/vars/foreman.yml
+++ b/src/vars/foreman.yml
@@ -6,4 +6,4 @@ foreman_initial_admin_password: "{{ lookup('ansible.builtin.password', foreman_a
 # Puma workers calculation:
 # CPU based calculation is based on https://github.com/puma/puma/blob/master/docs/deployment.md#mri
 # Memory based calculation is based on https://docs.gitlab.com/ee/install/requirements.html#puma-settings
-foreman_puma_workers: "{{ [100, ansible_processor_nproc * 1.5, (ansible_memtotal_mb / 1024) - 1.5] | min | int }}"
+foreman_puma_workers: "{{ [32, ansible_processor_nproc * 1.5, (ansible_memtotal_mb / 1024) - 1.5] | min | int }}"


arvind4501:
I agree with the idea of handling both puma and pulp scaling seprately.

> For puma we also have a complex formula how to compute the "optimal" default number (unless overriden by the user) in https://github.com/theforeman/puppet-foreman/blob/f69a26f3a3cffb7f525f91678afa297b8d6ea28b/manifests/config.pp#L94-L103

looking at the calculation for getting worker count, it could be complex in a container based setup as we need to get the resources(processers and memory) allocated to foreman container only to set puma workers in the foreman container(maybe by inspecting cgroup config)
also we don't limit the resources for a container in the current deployment, is that intentional or could that harm us in any way?

src/vars/foreman.yml
```diff
@@ -2,3 +2,7 @@
 foreman_admin_passwd_file: "{{ obsah_state_path }}/foreman-admin-init-passwd"
 foreman_initial_admin_username: admin
 foreman_initial_admin_password: "{{ lookup('ansible.builtin.password', foreman_admin_passwd_file, chars=['ascii_letters']) }}"
+# Puma workers calculation:
+# CPU based calculation is based on https://github.com/puma/puma/blob/master/docs/deployment.md#mri
+# Memory based calculation is based on https://docs.gitlab.com/ee/install/requirements.html#puma-settings
+foreman_service_puma_workers: "{{ [100, (ansible_processor_vcpus * 1.5) | int, ((ansible_memtotal_mb / 1024) - 1.5) | int] | min }}"
```
evgeni: In https://github.com/theforeman/foremanctl/pull/218/files#diff-5fb78252acba9a0f6406f470d587d4d8c84771561e8101b67e23ea2463f9ca38 we ended up using `ansible_facts.processor_nproc`, not `_vcpus`

Probably should use the same fact in both places.

Also, looking at the Puppet code and here, would this be more readable:
```suggestion
foreman_service_puma_workers: "{{ [100, ansible_processor_vcpus * 1.5, (ansible_memtotal_mb / 1024) - 1.5] | min | int }}"
```
(only calls `int` once on the resulting smallest value)

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -7,6 +7,8 @@ variables:
     help: Initial username for the admin user.
   foreman_initial_admin_password:
     help: Initial password for the admin user.
+  foreman_service_puma_workers:
```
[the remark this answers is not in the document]
arvind4501: i don't mind making it as `--foreman-puma-workers`  as its more concise and clear

src/vars/foreman.yml
```diff
@@ -2,3 +2,7 @@
 foreman_admin_passwd_file: "{{ obsah_state_path }}/foreman-admin-init-passwd"
 foreman_initial_admin_username: admin
 foreman_initial_admin_password: "{{ lookup('ansible.builtin.password', foreman_admin_passwd_file, chars=['ascii_letters']) }}"
+# Puma workers calculation:
+# CPU based calculation is based on https://github.com/puma/puma/blob/master/docs/deployment.md#mri
+# Memory based calculation is based on https://docs.gitlab.com/ee/install/requirements.html#puma-settings
+foreman_service_puma_workers: "{{ [100, ansible_processor_nproc * 1.5, (ansible_memtotal_mb / 1024) - 1.5] | min | int }}"
```
evgeni: The 100 is taken from Puppet which was added by @ekohl in https://github.com/theforeman/puppet-foreman/pull/1197, but https://github.com/puma/puma/blob/main/docs/deployment.md says nobody should do more than 32…

ekohl: That 100 was just a number that didn't break the system. Guess I should have read the docs back then. Sometimes it's good to revisit all the values

evgeni:
> I'd prefer to get rid of a manual tuning guide for workers as we have it now. Perhaps this is a chance to finally do that.

Does that mean you'd prefer not to expose `--foreman-puma-workers` at all to the user, but only keep the computed number (and possible tuning profiles)?

src/vars/foreman.yml
evgeni: I think I'd go with 32 then, allowing the user to go higher on their own if they really want to.

ekohl: In Ruby you have `Etc.nproc` to get the number of effective processors. Should we put that in Foreman's `config/environments/production.rb` as a default, leaving it empty here? Then it should auto tune, also taking quotas into account.

evgeni: > In Ruby you have `Etc.nproc` to get the number of effective processors. Should we put that in Foreman's `config/environments/production.rb` as a default, leaving it empty here? Then it should auto tune, also taking quotas into account.

You say putting the whole calculation into `config/puma/production.rb`? As just using `Etc.nproc` would be wrong.
works for me!

> 100 is static here is it suppose to be dynamic for different tuning profiles. and yes we should keep it to 32 as per docs

The 100 is purely a "magic number" for "not too many". I think even in the tuning profiles we wouldn't go over 32 by default.

ekohl: A quick look shows me there are implementations but they (on Linux) parse `/proc/meminfo` or shell out. Not pretty. Either it's not portable or a lot of code. Keeping it here is fair.

evgeni: Yepp, that's also what I had found.

PR #279, opened 2025-10-31
Foreman Proxy

stejskalleos:


src/roles/foreman/tasks/main.yaml
```diff
@@ -194,12 +194,3 @@
   register: foreman_tasks_status
   when:
     - "'katello' in foreman_status.json['results']"
-
-- name: Configure Foreman Proxy
```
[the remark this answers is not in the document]
stejskalleos: But users/customers will now see two smart proxies in UI, instead of one. Both with pulp.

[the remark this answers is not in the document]
ehelms: They will see two smart-proxies, but only one should have Pulp. Your implementation should not include the Pulp plugin because that's handled by the smart-proxy pulp plugin. The end state should be:

1. Pulp deployed running pulp_smart_proxy as a plugin acting a smart-proxy just for Pulp
2. A smart-proxy deployed for other features, without the smart_proxy_pulp plugin enabled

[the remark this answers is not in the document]
ehelms: And I see some of the confusion, the current image we have defaults to having the Pulp smart-proxy enabled:

https://github.com/theforeman/foreman-oci-images/blob/master/images/foreman-proxy/Containerfile#L5

@evgeni should we go ahead and drop this from that image and get it rebuilt?

ehelms:
> Why did you go with a separate playbook instead of a adding a feature?

I know I suggested going this route to get started since the feature implementation is not flushed out, and having a dedicated command allows for easy testing and conceptualizing right now. And helps start off with a "clean" split between Foreman and smart-proxy.

ekohl:
Ok. As long as the Ansible role prepares for extension then we can later include it with proper features support into the main deploy playbook

src/roles/smart_proxy/tasks/main.yaml
```diff
@@ -0,0 +1,69 @@
+---
```
evgeni: We used to have a `foreman_proxy` role (removed in https://github.com/theforeman/foremanctl/pull/81), it also had tests, which we should bring back :)

ehelms: Reminds me of the whole should we lean into smart-proxy or foreman-proxy as the naming.

src/roles/smart_proxy/tasks/main.yaml
```diff
@@ -0,0 +1,65 @@
...
+    name: "{{ smart_proxy_container_image }}:{{ smart_proxy_container_tag }}"
+    state: present
+
+- name: Create config secrets
+  ansible.builtin.include_tasks: configs.yaml
+
+- name: Create certs secrets
+  ansible.builtin.include_tasks: certs.yaml
+
+- name: Deploy Foreman Container
```
arvind4501:
```suggestion
- name: Deploy Smart Proxy Container
```

stejskalleos: > this is used for the content part(pulp), so i am wondering is this playbook take care of only services part?

Right now, the playbook is really "dumb", just deploying the service and that's it.
You are correct; in the future, I will add more features, as well as an option to deploy it on an external server.

src/roles/smart_proxy/tasks/main.yaml
evgeni: So would I.

stejskalleos:
* Renamed to `Foreman Proxy`
* Added test

tests/foreman_proxy_test.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+def test_foreman_proxy_features(server, certificates, server_fqdn):
+    cmd = server.run(f"curl --cacert {certificates['ca_certificate']} --silent https://{server_fqdn}:8443/features")
+    assert cmd.succeeded
```
evgeni: This is failing.

Is it because of the missing `:` at the beginning of
```
https_port: {{ foreman_proxy_https_port }}
```

src/playbooks/foreman-proxy/foreman-proxy.yaml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Setup Foreman Proxy
```
ekohl: Nit: needs to be 2 words here. See https://grammarist.com/spelling/set-up-vs-setup/ for more details.
```suggestion
- name: Set up Foreman Proxy
```

src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -0,0 +1,66 @@
+---
+- name: Create Foreman Proxy directories
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: directory
+    owner: root
+    group: root
+    mode: '0755'
+  loop:
+    - /var/log/foreman-proxy
```
ekohl: We shouldn't create `/var/log/foreman-proxy` and instead configure it to log to stdout or journald

src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -0,0 +1,66 @@
...
+    state: started
+
+- name: Configure Foreman Proxy
+  theforeman.foreman.smart_proxy:
+    name: "{{ ansible_fqdn }}"
+    url: "{{ foreman_proxy_url }}"
+    server_url: "{{ foreman_url }}"
+    username: "{{ foreman_initial_admin_username }}"
+    password: "{{ foreman_initial_admin_password }}"
+    validate_certs: false
```
ekohl: IMHO we should always validate certificates. We should have them on the system already.

src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -0,0 +1,66 @@
...
+    name: foreman-proxy
+    state: started
+
+- name: Configure Foreman Proxy
+  theforeman.foreman.smart_proxy:
+    name: "{{ ansible_fqdn }}"
+    url: "{{ foreman_proxy_url }}"
+    server_url: "{{ foreman_url }}"
+    username: "{{ foreman_initial_admin_username }}"
+    password: "{{ foreman_initial_admin_password }}"
```
ekohl: This is unreliable, since the user may change username & password after installation. The oauth credentials should be reliable.

src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -0,0 +1,66 @@
...
+  ansible.builtin.meta: flush_handlers
+
+- name: Start the Foreman Proxy Service
+  ansible.builtin.systemd:
+    name: foreman-proxy
+    state: started
+
+- name: Configure Foreman Proxy
+  theforeman.foreman.smart_proxy:
+    name: "{{ ansible_fqdn }}"
```
ekohl: Please avoid using `ansible_fqdn` for anything except defaults.
```suggestion
    name: "{{ foreman_proxy_name }}"
```

src/roles/foreman_proxy/tasks/main.yaml
ekohl: Follow up issue for foremanctl 3.0?

src/roles/foreman_proxy/tasks/configs.yaml
```diff
@@ -0,0 +1,25 @@
...
+- name: Create logs config secret
+  containers.podman.podman_secret:
+    state: present
+    name: foreman-proxy-logs-yaml
+    data: "{{ lookup('ansible.builtin.template', 'settings.d/logs.yaml.j2') }}"
+  notify:
+    - Restart Foreman Proxy
+
+
+- name: Create pulpcore config secret
```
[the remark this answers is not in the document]
evgeni: https://github.com/theforeman/foreman-oci-images/commit/69724855dd0b2ee4c1b2848c5b405a51c4171de1 was never built/released to quay (queue Evgeni complaining about pipelines missing)

src/playbooks/foreman-proxy/foreman-proxy.yaml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Set up Foreman Proxy
+  hosts:
+    - quadlet
+  become: true
+  vars_files:
+    - "../../vars/defaults.yml"
+    - "../../vars/{{ certificate_source }}_certificates.yml"
+    - "../../vars/foreman.yml"
```
evgeni: please also use https://github.com/theforeman/foremanctl/blob/master/src/vars/images.yml

src/playbooks/foreman-proxy/metadata.obsah.yaml
```diff
@@ -0,0 +1,8 @@
+---
+help: |
+  Install Foreman Proxy
+
+include:
+  - _certificate_source
+  - _database_mode
+  - _database_connection
```
ekohl: The proxy itself doesn't have a database, so why are these needed?

src/vars/images.yml
```diff
@@ -3,6 +3,9 @@ candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "foreman-{{ container_tag_stream }}"
 foreman_container_image: quay.io/foreman/foreman
 foreman_container_tag: "{{ container_tag_stream }}"
+foreman_proxy_container_image: "quay.io/foreman/foreman-proxy"
+foreman_proxy_container_tag: "{{ container_tag_stream }}"
```
evgeni: I wanted to say "please add this to the `images` list below" but given it's an optional feature, I am not that sure.

@ehelms @ekohl opinions?

src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -0,0 +1,55 @@
...
+  containers.podman.podman_container:
+    name: "foreman-proxy"
+    image: "{{ foreman_proxy_container_image }}:{{ foreman_proxy_container_tag }}"
+    state: quadlet
+    sdnotify: true
+    network: host
+    hostname: "{{ ansible_fqdn }}"
+    secrets:
+      - 'foreman-proxy-settings-yaml,type=mount,target=/etc/foreman-proxy/settings.yml'
+      - 'foreman-proxy-logs-yaml,type=mount,target=/etc/foreman-proxy/settings.d/logs.yml'
```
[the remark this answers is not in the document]
evgeni: I think `/etc/containers/systemd/foreman-proxy.container.d/<feature>.conf` makes a lot of sense for this.

src/roles/foreman_proxy/tasks/main.yaml
stejskalleos: Clarification for the 3.0 ticket:
Do we want to support the `ca_path` parameter for the Ansible module, or do we want to add the CA to the trusted store?

src/roles/foreman_proxy/templates/settings.yaml.j2
```diff
@@ -0,0 +1,20 @@
...
+:ssl_private_key: /etc/foreman-proxy/ssl_key.pem
+
+:foreman_ssl_ca: /etc/foreman-proxy/foreman_ssl_ca.pem
+:foreman_ssl_cert: /etc/foreman-proxy/foreman_ssl_cert.pem
+:foreman_ssl_key: /etc/foreman-proxy/foreman_ssl_key.pem
+
+:bind_host: '*'
+
+:log_level: INFO
+:log_file: JOURNAL
```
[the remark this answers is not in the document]
stejskalleos: I don't know TBH, not really sure how this smart-proxy -> container -> journal relationship works under the hood, so I'll leave the final decision up to you

src/roles/foreman_proxy/tasks/main.yaml
stejskalleos: https://issues.redhat.com/browse/SAT-40212

src/playbooks/foreman-proxy/metadata.obsah.yaml
evgeni: I think I'd drop it for now

src/roles/foreman_proxy/tasks/main.yaml
stejskalleos: https://issues.redhat.com/browse/SAT-40221

src/roles/foreman_proxy/templates/settings.yaml.j2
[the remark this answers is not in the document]
ekohl: https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading

evgeni (approved):

src/vars/images.yml
evgeni: #188 is updated and makes "foreman-proxy" a feature, have a look :)

PR #280, opened 2025-11-02
Add IOP

ehelms:


.github/workflows/test.yml
```diff
@@ -46,6 +46,9 @@ jobs:
           - none
         database:
           - internal
+        iop:
+          - enabled
+          - disabled
```
evgeni: This will run IoP enabled tests for *every* combination, thus doubling the matrix. Do we *need* that?

See how we used `security` and `database` with a single entry in the matrix itself and then special additions via `include` (resulting in one additional job)

evgeni:
> @evgeni Could you check just the github action part of the commit ([eb1df00](https://github.com/theforeman/foremanctl/commit/eb1df00b5877ff0330b229e81cdac1d9c31bee9e)) to make sure it matches with how you think we should use matrix?

I think it can be optimized, posted comments how :)

.github/workflows/test.yml
evgeni: Want an explanation?

evgeni:
> @evgeni For now, I want to keep the same behavior of not supporting remote database yet. Looking at Obsah, I don't think that's possible but I wanted you to check me on that. I need to forbid:
> 
>     * database_mode == external and 'iop' in enabled_features
> 
> 
> `forbidden_if` looks like it only allows the first parameter to have a value. Do I read that correctly?

That's correct, today we can only forbid "if param P has value Y you can't set params A and B at all"

src/roles/iop_remediation/defaults/main.yaml
```diff
@@ -0,0 +1,9 @@
+---
+iop_remediation_container_image: "quay.io/iop/remediations"
+iop_remediation_container_tag: "foreman-3.16"
+
+iop_remediation_database_name: remediations_db
```
pablomh: All of these vars can be deleted as they are in `src/vars/database_iop.yml`.

src/roles/iop_vmaas/defaults/main.yaml
```diff
@@ -0,0 +1,9 @@
+---
+iop_vmaas_container_image: "quay.io/iop/vmaas"
+iop_vmaas_container_tag: "foreman-3.16"
+
+iop_vmaas_database_name: vmaas_db
```
pablomh: All of these vars can be deleted as they are in `src/vars/database_iop.yml`.

pablomh (commented):
Builds and creates running services :)

src/roles/iop_advisor/defaults/main.yaml
```diff
@@ -0,0 +1,9 @@
+---
+iop_advisor_container_image: "quay.io/iop/advisor-backend"
+iop_advisor_container_tag: "foreman-3.16"
+
+iop_advisor_database_name: advisor_db
```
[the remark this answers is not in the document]
ehelms: Role defaults are not the same as playbook defaults. We try to follow Ansible best practices by ensuring these are defined at the role level. We then define them at a higher level for our use case to provide coordination of values and advanced workflows.

src/roles/iop_fdw/tasks/main.yaml
```diff
@@ -0,0 +1,165 @@
...
+    db: "{{ iop_fdw_database_name }}"
+    login_user: postgres
+    login_host: "{{ iop_fdw_database_host }}"
+    query: "SELECT srvname FROM pg_foreign_server WHERE srvname = %s"
+    positional_args:
+      - "{{ iop_fdw_foreign_server_name }}"
+  register: iop_fdw_foreign_server_check
+  changed_when: false
+
+- name: Create foreign server for inventory database
```
pablomh: `s/inventory/target/` ?

src/roles/iop_remediation/tasks/main.yaml
```diff
@@ -0,0 +1,82 @@
...
+    name: iop-service-remediations-db-port
+    data: "{{ iop_remediation_database_port }}"
+  notify: restart remediation
+
+- name: Deploy Remediation API container
+  containers.podman.podman_container:
+    name: iop-service-remediations-api
+    image: "{{ iop_remediation_container_image }}:{{ iop_remediation_container_tag }}"
+    state: quadlet
+    network: host
```
pablomh: Shouldn't this be `iop-core-network` given that `iop_remediation_database_host: "host.containers.internal"`?

vkrizan (approved):
Thank you. It looks neat.

src/roles/iop_ingress/tasks/main.yaml
```diff
@@ -0,0 +1,44 @@
...
+    name: iop-core-ingress
+    image: "{{ iop_ingress_container_image }}:{{ iop_ingress_container_tag }}"
+    state: quadlet
+    env:
+      INGRESS_VALID_UPLOAD_TYPES: "advisor,compliance,qpc,rhv,tower,leapp-reporting,xavier,playbook,playbook-sat,malware-detection,tasks"
+      INGRESS_KAFKA_BROKERS: "iop-core-kafka:9092"
+      BOOTSTRAP_SERVERS: "iop-core-kafka:9092"
+      INGRESS_STAGERIMPLEMENTATION: "filebased"
+      INGRESS_STORAGEFILESYSTEMPATH: "/var/tmp"
+      INGRESS_SERVICEBASEURL: "http://localhost:8080"
```
jeremylenz:
```suggestion
      INGRESS_SERVICEBASEURL: "http://iop-core-ingress:8080"
```

jeremylenz:
Latest commit seems to fix #459 . :+1:

jeremylenz (approved):
🚀

PR #282, opened 2025-11-03
Scale pulp-api and pulp-content service workers.

arvind4501:


src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -7,6 +7,8 @@ variables:
     help: Initial username for the admin user.
   foreman_initial_admin_password:
     help: Initial password for the admin user.
+  pulp_worker_count:
+    help: Number of pulpcore workers. Defaults to 8 or the number of CPU cores, whichever is smaller.
```
ehelms:
```suggestion
    help: Number of Pulp workers. Defaults to 8 or the number of CPU cores, whichever is smaller.
```

src/roles/pulp/defaults/main.yaml
```diff
@@ -5,7 +5,9 @@ pulp_api_image: "{{ pulp_container_image }}:{{ pulp_container_tag }}"
 pulp_content_image: "{{ pulp_container_image }}:{{ pulp_container_tag }}"
 pulp_worker_image: "{{ pulp_container_image }}:{{ pulp_container_tag }}"
 
-pulp_worker_count: 2
+pulp_worker_count: "{{ [8, ansible_processor_nproc] | min }}"
+pulp_content_service_worker_count: "{{ (2 * ([8, ansible_processor_nproc] | min)) + 1 }}"
+pulp_api_service_worker_count: "{{ ([4, ansible_processor_nproc] | min) + 1 }}"
```
ehelms: I see you also have changes for api and content worker counts, was that intentional to have scaling for all three here?

arvind4501: yes, so i also one PR up for puma worker scaling and there evegni mentioned https://github.com/theforeman/foremanctl/pull/272#issuecomment-3467786802 , also i do not see directly exposed option in foreman-installer for overriding the default content and api worker configuration(https://github.com/theforeman/puppet-pulpcore/blob/0d73cc4299e0f2ac71c2e4b70dce472f6b7a015c/manifests/init.pp#L157C2-L164C2)  , so i kept the only option availabe to user to override worker(https://github.com/theforeman/puppet-pulpcore/blob/0d73cc4299e0f2ac71c2e4b70dce472f6b7a015c/manifests/init.pp#L139-L142)

ehelms: @arvind4501 is correct, those parameters are exposed on puppet-pulpcore, but only worker count is exposed on puppet-foreman_proxy_content.

evgeni:
api and content workers are properly deployed now, but i still only see one single "worker" :thinking:

Issue #283, opened 2025-11-03
foremanctl should have persistent logs


ehelms:
The `foreman-installer` generates log files on each run which has proven to be helpful for debugging. `foremanctl` should do the same and will need to consider how different action interact or are represented in log files.

PR #295, opened 2025-11-12
foremanctl.log

stejskalleos:
For #283

**Requires:**
* https://github.com/theforeman/obsah/pull/98
* https://github.com/theforeman/obsah/pull/100

### DIFF
--- a/foremanctl.spec
+++ b/foremanctl.spec
@@ -10,7 +10,7 @@ URL:       https://github.com/theforeman/foremanctl
 Source:    https://github.com/theforeman/foremanctl/releases/download/%{version}/%{name}-%{version}.tar.gz
 
 BuildArch: noarch
-Requires:  python3-obsah >= 1.4
+Requires:  python3-obsah >= 1.7.1
 
 # These are needed on the target host, which is usually localhost
 Recommends:  podman
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -1,2 +1,2 @@
-obsah >= 1.7.0
+obsah >= 1.7.1
 requests


src/ansible.cfg
```diff
@@ -2,3 +2,4 @@
 host_key_checking = False
 roles_path = ./roles
 callback_result_format = yaml
+log_path=/var/log/foremanctl/foremanctl.log
```
ehelms: Should this default to a relative path, `.var` and then be overriden in the RPM? Perhaps this should be an Obsah controlled element similar to how state storage( e.g. parameters) are handled.

stejskalleos: > Should this default to a relative path

I was thinking about it, but then I thought: what if the user runs `foremanctl` from a different directory?

The user can be in `~/` and for a different run, in the `/` directory, or somewhere else entirely. 

> Perhaps this should be an Obsah

I haven't thought about that one, but that could be the way as well.

foremanctl.spec
```diff
@@ -43,6 +43,7 @@ install -d -m0755 %{buildroot}%{_sysconfdir}/%{name}
 install -d -m0755 %{buildroot}%{_datadir}/%{name}
 install -d -m0755 %{buildroot}%{_bindir}
 install -d -m0750 %{buildroot}%{_sharedstatedir}/%{name}
+install -d -m0755 %{buildroot}%{_localstatedir}/log/%{name}
```
stejskalleos:
```suggestion
install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}
```

foremanctl.spec
evgeni: We should drop this file in favor of https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/foreman/foremanctl, otherwise we risk they getting out of sync too quickly.

(But that doesn't mean this PR is wrong, just noticing it here)

stejskalleos:
Yes, sorry, `obsah`. And yes, every time Obsah runs, it will make a copy of the log file before executing any Ansible role/playbook.

src/roles/pre_install/tasks/main.yaml
```diff
@@ -1,4 +1,20 @@
...
+    path: /var/log/foremanctl
+    state: directory
+    mode: '0750'
+    owner: root
+    group: root
+
+- name: Backup the latest log file
+  ansible.builtin.copy:
+    src: "{{ lookup('ansible.builtin.env', 'ANSIBLE_LOG_PATH') }}"
+    dest: "/var/log/foremanctl/foremanctl-{{ ansible_facts['date_time']['iso8601_basic_short'] }}.log"
```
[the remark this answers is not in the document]
stejskalleos: This is log for the foremanctl, not for the application log.
AFAIK there is no daily log rotation for installer logs, and I think it's not needed.

foremanctl
```diff
@@ -8,8 +8,9 @@ OBSAH_STATE=${OBSAH_BASE}/.var/lib/foremanctl
 OBSAH_PERSIST_PARAMS=true
 export OBSAH_NAME OBSAH_DATA OBSAH_INVENTORY OBSAH_STATE OBSAH_PERSIST_PARAMS
 
+ANSIBLE_LOG_PATH=/var/log/foremanctl/foremanctl.log
```
evgeni: CI fails because
```
Warning: : log file at '/var/log/foremanctl/foremanctl.log' is not writeable and we cannot create it, aborting
```

Probably should be `.var/…` by default?

evgeni (changes requested):

evgeni:
I think it resolves it.

PR #288, opened 2025-11-07
add quadlet recurring job runner and timers for core rake tasks

archanaserver:
- define templated Quadlet container `foreman-recurring@.container` via `containers.podman.podman_container` (state: quadlet)
- run rake tasks inside the foreman image with same secrets/mounts as `foreman.service`
- so far added per-instance timers for:
  - `reports:daily`
  - `db:sessions:clear`
  - `reports:expire`
  - `audits:expire`

Refs: https://github.com/theforeman/foremanctl/issues/61 and https://github.com/theforeman/foremanctl/pull/252

PR #299, opened 2025-11-13
add recurring task support for katello

archanaserver:
based on #288


ekohl (commented):
Is there an issue somewhere to rewrite it into Foreman's own recurring logic? Katello already depends on it so I'd think it could be done.

This is an architectural choice that I think we can drive. The foreman-tasks cleanup may be an exception because there is a bigger risk (cleaning up a task from a task can break if you aren't careful) but Katello can rely on it.

If that isn't too hard for the Katello devs to do then we'll have a better and more scalable design.

evgeni:
needs a rebase

PR #309, opened 2025-11-19
Proxy features: REX

evgeni:


src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -35,6 +34,20 @@
         PartOf=foreman.target
   notify: Restart Foreman Proxy
 
+- name: Create foreman-proxy.container.d folder
```
ekohl: With Puppet we also purged other entries. Do we still want to do that to disable features? Or go with an explicit `ensure => absent` equivalent on features?

adamlazik1:
https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/126 added new settings for SSH certs, here is a patch that adds the new settings:

```patch
diff --git a/src/roles/foreman_proxy/templates/settings.d/remote_execution_ssh.yml.j2 b/src/roles/foreman_proxy/templates/settings.d/remote_execution_ssh.yml.j2
index b7d6402..76ad344 100644
--- a/src/roles/foreman_proxy/templates/settings.d/remote_execution_ssh.yml.j2
+++ b/src/roles/foreman_proxy/templates/settings.d/remote_execution_ssh.yml.j2
@@ -11,6 +11,15 @@
 # Mode of operation, one of ssh, pull, pull-mqtt
 :mode: ssh
 
+# Enables the use of SSH certificate for smart proxy authentication
+# The file should contain an SSH CA public key that the SSH public key of smart proxy is signed by
+# :ssh_user_ca_public_key_file:
+
+# Enables the use of SSH host certificates for host authentication
+# The file should contain a list of trusted SSH CA authorities that the host certs can be signed by
+# Example file content: @cert-authority * <SSH CA public key>
+# :ssh_ca_known_hosts_file:
+
 # Defines how often (in seconds) should the runner check
 # for new data leave empty to use the runner's default
 # :runner_refresh_interval: 1
```

evgeni:
> What remains to be done here until this can be undrafted?

This PR was my PoC to figure out what we need to put into our feature metadata (see #372).

This PR also has two dependencies that can be reviewed/merged independently:
- https://github.com/theforeman/foreman-ansible-modules/pull/1930
- https://github.com/theforeman/foreman-oci-images/pull/13

Once those are settled (and the "temp" commits dropped), we can continue here.

src/filter_plugins/foremanctl.py
```diff
@@ -0,0 +1,66 @@
...
+    return compact_list(plugins)
+
+
+class FilterModule(object):
+    ''' foremanctl filters'''
+
+    def filters(self):
+        return {
+            'features_to_foreman_plugins': foreman_plugins,
+            'knonwn_foreman_plugins': known_foreman_plugins,
```
arvind4501:
```suggestion
            'known_foreman_plugins': known_foreman_plugins,
```

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -10,3 +10,10 @@ foreman_proxy_url: "https://{{ foreman_proxy_name }}:{{ foreman_proxy_https_port
 # Settings
 foreman_proxy_trusted_hosts:
   - "{{ foreman_proxy_name }}"
+
+foreman_proxy_base_features:
+  - logs
+foreman_proxy_plugins: []
+foreman_proxy_features: "{{ foreman_proxy_base_features + foreman_proxy_plugins }}"
+foreman_proxy_known_features: "{{ [] | known_foreman_proxy_plugins }}"
```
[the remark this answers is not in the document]
evgeni: known: the list of all features defined in the metadata, each of  can be either enabled or disabled.

this variable is an internal helper as the user input is only the list of "enabled" features, but we also need to have the list of disabled ones to be able to perform the disablement.

src/roles/foreman_proxy/tasks/feature/remote_execution_ssh.yaml
```diff
@@ -0,0 +1,36 @@
+---
+- name: Create SSH Key
+  community.crypto.openssh_keypair:
+    path: /root/foreman-proxy-ssh
+
+- name: Create SSH Key podman secret
+  containers.podman.podman_secret:
+    state: present
+    name: foreman-proxy-remote_execution_ssh-ssh-key
```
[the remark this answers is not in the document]
evgeni: I think the secrets in this PR should be:
- `foreman_proxy-remote_execution_ssh-key`
- `foreman_proxy-remote_execution_ssh-key-pub`

src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -57,3 +78,6 @@
     username: "{{ foreman_initial_admin_username }}"
     password: "{{ foreman_initial_admin_password }}"
     validate_certs: false
+
+- name: Flush handlers to restart services
+  ansible.builtin.meta: flush_handlers
```
ShimShtein: Why do you need an explicit run of handlers at this point? The handlers should be executed after all the roles have been executed: https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_handlers.html#controlling-when-handlers-run

PR #311, opened 2025-11-20
Expand .gitignore to ignore entire inventories/ directory

Gauravtalreja1:


Gauravtalreja1:
@evgeni When we deploy using `deploy-foreman-development` workflow, it adds a few inventories like `foreman.example.com`, broker.py, quadlet, and localhost inventory, so I feel it would be better to add entire inventories/ dir to gitignore would make more sense to avoid exposing this, wdyt?

PR #314, opened 2025-11-21
Remove misleading is_enabled checks for generated services

arvind4501:
## Problem ##

Quadlet services are dynamically generated by systemd-generator and return "generated" (exit code 0) when checking `systemctl is-enabled <service>`. This caused test assertions using `service.is_enabled` to always pass, making them misleading and not actually validating service enablement.

  Example:
```
  [root@quadlet ~]# systemctl is-enabled candlepin
  generated
  [root@quadlet ~]# echo $?
  0
```

pytest-testinfra's is_enabled property only checks the exit code (https://github.com/pytest-dev/pytest-testinfra/blob/main/testinfra/modules/service.py#L197), so it returns True for any service that exits
  with 0, including "generated" services.

 ## Solution ##

 Removed `is_enabled` checks for all Quadlet-generated services

### DIFF
--- a/tests/pulp_test.py
+++ b/tests/pulp_test.py
@@ -64,3 +64,8 @@ def test_pulp_status_workers(pulp_status):
 
 def test_pulp_volumes(server):
     assert server.file("/var/lib/pulp").is_directory
+
+def test_pulp_worker_target(server):
+    pulp_worker_target = server.service("pulp-worker.target")
+    assert pulp_worker_target.is_running
+    assert pulp_worker_target.is_enabled
--- a/tests/foreman_target_test.py
+++ b/tests/foreman_target_test.py
@@ -1,0 +1,4 @@
+def test_foreman_target(server):
+    foreman_target = server.service("foreman.target")
+    assert foreman_target.is_running
+    assert foreman_target.is_enabled


evgeni (approved):
Needs a rebase, but otherwise LGTM! Thanks

One idea (but extends the scope): we could instead test `foreman.target` and `pulp-worker.target` to be `is_enabled` (those are not generated and thus can be indeed enabled)

arvind4501:
> we could instead test foreman.target and pulp-worker.target to be is_enabled (those are not generated and thus can be indeed enabled)

yes, they can be enabled and we can test that, should it be included in the existing tests, like for foreman.target i add test in the `foreman_test.py` or should i create a new `systemd_target_test.py` file for specifically testing targets?

evgeni:
pulp-worker.target I'd test in test_pulp, as it's defined in the pulp role

foreman.target… probably own test file is more correct

tests/pulp_test.py
```diff
@@ -67,3 +64,8 @@ def test_pulp_status_workers(pulp_status):
 
 def test_pulp_volumes(server):
     assert server.file("/var/lib/pulp").is_directory
+
+def test_pulp_worker_target(server):
+    pulp_worker_target = server.service("pulp-worker.target")
+    assert pulp_worker_target.is_running
+    assert pulp_worker_target.is_enabled
```
amolpati30:
```suggestion
```

PR #320, opened 2025-11-24
Initial tuning code

pablomh:
First steps needed to implement tuning profile support.

### DIFF
--- a/src/playbooks/_tuning/metadata.obsah.yaml
+++ b/src/playbooks/_tuning/metadata.obsah.yaml
@@ -1,0 +1,11 @@
+---
+variables:
+  tuning:
+    help: Tuning profile to apply. Defaults to development. Set to a different value according to the available hardware.
+    choices:
+      - default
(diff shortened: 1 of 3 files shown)


src/playbooks/_tuning/metadata.obsah.yaml
```diff
@@ -0,0 +1,11 @@
+---
+variables:
+  tuning:
+    help: Tuning profile to apply. Defaults to development. Set to a different value according to the available hardware.
```
ehelms: Shouldn't this default to...default?

pablomh: At this point I didn't want to change the current default and impact developers or CI .

pablomh: Can we get this in and I'll change the default to `default` in a subsequent PR and `--tuning development` in the tests to not break them?

PR #326, opened 2025-11-25
Drop checks role from setup-foreman-proxy

ehelms:
In their current form, the checks only apply to the Foreman itself.


evgeni:
@ehelms which checks did you not want to run on a proxy? `check_system_requirements`?

I am asking because https://github.com/theforeman/foremanctl/pull/188 adds it back, effectively, so I want to follow up with a feature-based disablement.

PR #327, opened 2025-11-26
ensure collections are uptodate by depending on the requirement files

evgeni:


evgeni:
@ehelms this should avoid the issue you had yesterday

ehelms (approved):

PR #328, opened 2025-11-26
Change default Puma threads min and max values to match Foreman ones

pablomh:
The values are taken from:

https://github.com/theforeman/puppet-foreman/blob/master/manifests/init.pp

Because this is the first stab at this, they'll be hardcoded and will be configurable in a later commit.

### DIFF
--- a/src/roles/foreman/defaults/main.yaml
+++ b/src/roles/foreman/defaults/main.yaml
@@ -12,7 +12,13 @@ foreman_database_sslrootcert:
 
 foreman_url: "http://{{ ansible_facts['fqdn'] }}:3000"
 
+# Puma threads calculation:
+# Max calculation is based on the default for MRI https://puma.io/puma6/#thread-pool
+# Min calculation is based on testing that showed that having the same value as the max led to
+# lower memory usage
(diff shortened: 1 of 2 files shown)


ekohl:
We discussed it in https://github.com/theforeman/foremanctl/pull/272#issuecomment-3472057558 to reflect what we did in https://github.com/theforeman/puppet-foreman/pull/986. That recommendation to change it to 16 in the tuning guide is still something that I have an issue with because tuning is an anti-goal IMHO. We should have good defaults for most deployments.

We chose 5 because if you have multiple processes doing actual computation work it can create problems. I'm unsure which workflows cause high CPU usage (rather than waiting like on a DB or HTTP service), but with those you can easily create high latency on unrelated processes. Perhaps with registrations or fact uploads?

pablomh:
I'm fine with the default, then. This wasy it'll be easier to do an "apple-to-containerized_apple" comparison :)

If you don't mind, and if threads tuning is discarded, I'll keep a local patch to be able to run a test campaign and revisit the numbers later.

ekohl (approved):

PR #330, opened 2025-11-26
Change the default tuning profile to default

pablomh:
Change the default tuning profile to `default`


ehelms:
@evgeni Any concerns with merging this now? This will match reality, but mean a bit more of a burden on development having to remember to include this option. I was thinking of adding a forge command that deployed with a set of common options (since I find I am having to include a few `--add-feature` statements).

PR #332, opened 2025-11-27
update  defaults to define 4 generic timers

archanaserver:
Depends on https://github.com/theforeman/foreman/pull/10784
- Update `foreman_recurring_tasks` defaults to define 4 generic
  timers:
  * hourly → `cron:hourly`
  * daily → `cron:daily`
  * weekly → `cron:weekly`
  * monthly → `cron:monthly`
- Keep `ldap:refresh_usergroups` as a separate recurring job with
  its existing `OnCalendar` expression, since it runs every 30
  minutes and does not fit cleanly into the hourly/daily/weekly
  buckets yet.

### DIFF
--- a/src/roles/foreman/defaults/main.yaml
+++ b/src/roles/foreman/defaults/main.yaml
@@ -25,27 +25,15 @@ foreman_puma_workers: "{{ [32, ansible_facts['processor_nproc'] * 1.5, (ansible_
 
 foreman_recurring_tasks_enabled: true
 foreman_recurring_tasks:
-  - instance: reports-daily
-    rake: "reports:daily"
+  - instance: hourly
+    rake: "cron:hourly"
+    schedule: "hourly"
+  - instance: daily
+    rake: "cron:daily"
     schedule: "daily"
-  - instance: db-sessions-clear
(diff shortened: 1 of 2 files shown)


src/roles/foreman/defaults/main.yaml
```diff
@@ -19,27 +19,19 @@ foreman_puma_workers: "{{ [32, ansible_facts['processor_nproc'] * 1.5, (ansible_
...
+  - instance: monthly
+    rake: "cron:monthly"
     schedule: "monthly"
-  - instance: notifications-clean
-    rake: "notifications:clean"
-    schedule: "weekly"
+
   - instance: ldap-refresh_usergroups
     rake: "ldap:refresh_usergroups"
     schedule: "*-*-* *:00,30:00"
```
evgeni: would it be sufficient to run this `hourly` too?

:eyes: @adamruzicka

adamruzicka: I think so, yes.

evgeni:
The tests needs to be updated, and we need to wait for https://github.com/theforeman/foreman/pull/10784 to be merged before this can go in.

PR #333, opened 2025-11-27
don't set working_dir, it's already in the containerfile

evgeni:


archanaserver (approved):

src/roles/foreman/tasks/main.yaml
```diff
@@ -173,7 +173,6 @@
       - -c
       - bin/rails db:migrate && bin/rails db:seed
     detach: false
-    working_dir: /usr/share/foreman
```
Gauravtalreja1: Since we hardcode this path in `Deploy Dynflow Container` play as well, so could we replace it with "$HOME/extras/dynflow-sidekiq.rb"?

evgeni: I can look into this in a separate PR

Gauravtalreja1 (approved):

PR #334, opened 2025-11-27
use relative path to dynflow script

evgeni:


src/roles/foreman/tasks/main.yaml
```diff
@@ -133,10 +133,9 @@
       - 'foreman-client-key,type=mount,target=/etc/foreman/client_key.pem'
       - 'foreman-dynflow-worker-hosts-queue-yaml,type=mount,target=/etc/foreman/dynflow/worker-hosts-queue.yml'
     env:
-      DYNFLOW_SIDEKIQ_SCRIPT: "/usr/share/foreman/extras/dynflow-sidekiq.rb"
```
evgeni: I've split out this to https://github.com/theforeman/foremanctl/pull/335, as that part should be working fine

PR #343, opened 2025-12-02
don't skip foreman-tasks test

evgeni:


evgeni:
We probably can do that now, yes.

(we initially did that to avoid the separate migrate/seed step, but as we've seen later that didn't work well enough)

PR #344, opened 2025-12-02
SAT-40191 Compute resources & plugins

stejskalleos:
Allow enabling and installing of two features:
foreman_azure_rm  & foreman_google

This includes foreman plugins and hammer-cli plugins.

How to test:
```
foremanctl deploy --add-feature foreman_azure_rm
foremanctl deploy --add-feature foreman_google
```

Related tasks
* https://github.com/theforeman/foreman-oci-images/commit/5c09bfbc3dc5522e63991798b8484e04ee6d23c1

### DIFF
--- a/tests/foreman_compute_resources_test.py
+++ b/tests/foreman_compute_resources_test.py
@@ -1,0 +1,10 @@
+import pytest
+
+FOREMAN_HOST = 'localhost'
+FOREMAN_PORT = 3000
+
+@pytest.mark.parametrize("compute_resource", ['AzureRm', 'EC2', 'GCE', 'Libvirt', 'Openstack', 'Vmware'])
+def test_foreman_compute_resources(server, compute_resource):
+    hammer = server.run("hammer compute-resource create --help | grep provider")
+    assert hammer.succeeded
+    assert compute_resource in hammer.stdout
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -105,6 +105,9 @@ jobs:
       - name: Add optional feature - foreman-proxy
         run: |
           ./foremanctl deploy --add-feature foreman-proxy
+      - name: Add optional feature - foreman_azure_rm and foreman_google
+        run: |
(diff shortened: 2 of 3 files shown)


tests/foreman_compute_resources_test.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+FOREMAN_HOST = 'localhost'
+FOREMAN_PORT = 3000
+
+@pytest.mark.parametrize("compute_resource", ['AzureRm', 'EC2', 'GCE', 'Libvirt', 'Openstack', 'Vmware'])
+def test_foreman_compute_resources(server, compute_resource):
+    hammer = server.run("hammer compute-resource create --help | grep provider")
+    assert hammer.succeeded
+    assert compute_resource in hammer.stdout
```
evgeni: I was about to suggest to use the API directly (thus saving the requirement to have hammer set up), but the code gets ugly:

```
>>> next(pp for pp in next(p for p in foremanapi.resource('compute_resources').action('create').params if p.name == 'compute_resource').params if pp.name == 'provider').description
'\nProviders include Libvirt, Ovirt, EC2, Vmware, Openstack, AzureRm, GCE\n'
```

evgeni: For the non-core ones you could call `foremanapi.list('plugins')`, something like:
```python
plugins = [plugin['name'] for plugin in foremanapi.list('plugins')]
assert 'foreman_google' in plugins
```

stejskalleos: added [tests/foreman_plugins.py](https://github.com/theforeman/foremanctl/pull/344/commits/3ac369e1a7f833270618edbc7ab1970049edcdbc)

tests/foreman_plugins.py
evgeni: this file needs to be called `foreman_plugins_test.py` to be picked up by pytest

evgeni (approved):

PR #345, opened 2025-12-02
pass in the configured DB ca (if any) to the container

evgeni:


ehelms:
This looks related to https://github.com/theforeman/foremanctl/pull/141 but is not an aspect I have incorporated. I was wanting to get in a version of remote database testing and then harden it with more of the options.

evgeni:
It is, @Gauravtalreja1 ran into this when testing ext db stuff

src/roles/foreman/tasks/main.yaml
```diff
@@ -8,7 +8,7 @@
   containers.podman.podman_secret:
     state: present
     name: foreman-database-url
-    data: "postgresql://{{ foreman_database_user }}:{{ foreman_database_password }}@{{ foreman_database_host }}:{{ foreman_database_port }}/{{ foreman_database_name }}?pool={{ foreman_database_pool }}&sslmode={{ foreman_database_sslmode }}{% if foreman_database_ssl_ca is defined %}&sslrootcert={{ foreman_database_ssl_ca }}{% endif %}" # yamllint disable-line rule:line-length
+    data: "postgresql://{{ foreman_database_user }}:{{ foreman_database_password }}@{{ foreman_database_host }}:{{ foreman_database_port }}/{{ foreman_database_name }}?pool={{ foreman_database_pool }}&sslmode={{ foreman_database_sslmode }}{% if foreman_database_sslrootcert is defined %}&sslrootcert=/etc/foreman/db-ca.crt{% endif %}" # yamllint disable-line rule:line-length
```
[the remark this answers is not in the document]
evgeni: Updated.

.github/workflows/test.yml
```diff
@@ -53,6 +53,9 @@ jobs:
           - certificate_source: default
             security: none
             database: external
+          - certificate_source: default
+            security: none
+            database: externalssl
```
evgeni: I personally think that users should never run external db without SSL, but I know that today we document and support that, so I added this as another matrix entry instead of repurposing the `external` one. But do we really need to test both?

evgeni:
@ehelms look, no draft!

development/playbooks/remote-database/metadata.obsah.yaml
```diff
@@ -0,0 +1,4 @@
+variables:
+  database_ssl:
+    type: Boolean
```
Gauravtalreja1: Do we really need a new variable here? Could we use the existing `--database-ssl-ca`, given that it would be only defined for the external SSL scenario, right?

.github/workflows/test.yml
```diff
@@ -69,11 +72,11 @@ jobs:
...
       - name: Start VMs
         run: |
-          ./forge vms start --vms "quadlet client ${{ matrix.database == 'external' && 'database' || '' }}"
+          ./forge vms start --vms "quadlet client ${{ matrix.database != 'internal' && 'database' || '' }}"
       - name: Configure remote-database
-        if: matrix.database == 'external'
+        if: matrix.database != 'internal'
         run: |
-          ./forge remote-database
+          ./forge remote-database ${{ matrix.database == 'externalssl' && '--database-ssl true' || ''}}
```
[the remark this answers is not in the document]
evgeni: no, this just instructs the `forge` tool (which we don't ship to users, it's CI/dev only) to deploy a DB with SSL

development/playbooks/remote-database/metadata.obsah.yaml
evgeni: for `forge` we don't have `--database-ssl-ca` as it will generate an *own* CA for the DB

.github/workflows/test.yml
evgeni: I see little chance that non-SSL ext-DB breaks while SSL-enabled continues to work.

and you can still verify that in robotello for formal support, I just don't think it needs to happen here

src/roles/check_database_connection/tasks/check.yaml
```diff
@@ -1,14 +1,41 @@
+- name: Store CA cert to a temporary file
```
evgeni: This is the only place that is not tested for the non-ssl case right now, as check_database_connection only runs for external DB.

I think that's OK

ehelms: I read your original statement backwards. Carry on.

development/playbooks/remote-database/metadata.obsah.yaml
evgeni: I did that now, to keep the diff smaller

PR #346, opened 2025-12-02
Refactor SSH config generation from vagrant to ansible inventory

ehelms:
This is intended to allow running the test suite against more targets than just Vagrant environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ build/
 foremanctl-*.tar.gz
 
 .var
+.tmp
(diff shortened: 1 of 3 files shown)


evgeni (approved):

ehelms:
> The idea is what? As long as my ansible inventory declares one host with the name `quadlet` and one with `client` my tests still work fine, even if the source of those hosts is not Vagrant (even tho it by default is)?

Yes. I, for example, wanted to be able to the run tests against localhost in one case and another remote host target in another.

PR #348, opened 2025-12-03
refactor ext db test into normal test matrix

evgeni:


evgeni:
@ehelms I didn't want to refactor it directly in #141, but I think this way it's nicer as it automatically benefits from any changes to our normal test workflow and the chances we forget something is smaller

evgeni:
Like this?
image

PR #352, opened 2025-12-05
Run `pulpcore-manager check --deploy` as part of the tests

evgeni:
Django has a checks framework that can detect problems in a deployment. It's also extensible and allows Pulp and plugin developers to add their own checks. This allows detection of misconfigurations.

Some checks run implicitly before running certain commands but others don't for performance reasons.

`--deploy` signals that a production setup is used, which enables more checks.

Ideally, we'd also run with `--fail-level WARNING`, but there are too many warnings right now to fix them all.

https://docs.djangoproject.com/en/5.2/topics/checks/

### DIFF
--- a/tests/pulp_test.py
+++ b/tests/pulp_test.py
@@ -69,3 +69,7 @@ def test_pulp_worker_target(server):
     pulp_worker_target = server.service("pulp-worker.target")
     assert pulp_worker_target.is_running
     assert pulp_worker_target.is_enabled
+
+def test_pulp_manager_check(server):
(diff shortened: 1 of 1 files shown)


evgeni:
This ports @ekohl's https://github.com/theforeman/puppet-pulpcore/pull/155 over to the container world.
Enjoy

evgeni:
Let's continue that discussion in #353

ekohl:
I'm surprised this passed, but my Puppet version failed with many warnings. Any idea why?

PR #362, opened 2025-12-17
exclude collection tests from the tarball

evgeni:


adamruzicka (approved):
Did the trick in #312 , let's get this in

PR #375, opened 2026-02-15
Set explicit volume mount permissions

ehelms:
Adds an Ansible lint rule to help ensure any future volume mounts that are added follow this rule.

### DIFF
--- a/.ansible-lint-rules/explicit_volume_mode.py
+++ b/.ansible-lint-rules/explicit_volume_mode.py
@@ -1,0 +1,53 @@
+"""Implementation of explicit-volume-mode rule."""
+
+from __future__ import annotations
+
+import re
+from typing import TYPE_CHECKING, Any
+
+from ansiblelint.rules import AnsibleLintRule
+
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import Lintable
+
+_VOLUME_PATTERN = re.compile(r'^[^:]+:[^:]+:(rw|ro)(?:,[zZ])?$')
+
+
+class ExplicitVolumeModeRule(AnsibleLintRule):
+    """Volume mounts must specify explicit :rw or :ro mode."""
+
+    id = "explicit-volume-mode"
+    description = "Volume mounts must explicitly specify :rw or :ro mode, to ease auditing of access control."
+    severity = "HIGH"
+    tags = ["idiom"]
+    version_added = "custom"
+
+    def matchtask(self, task: dict[str, Any], file: Lintable | None = None) -> bool | str:
+        """Return a match if any volume mount lacks explicit mode."""
+        action = task.get("action", {})
+        if action.get("__ansible_module__") != "containers.podman.podman_container":
+            return False
+
+        for volume_key in ("volume", "volumes"):
+            volume_data = action.get(volume_key)
+            if not volume_data:
+                continue
+
+            if isinstance(volume_data, str):
+                specs = [volume_data]
+            elif isinstance(volume_data, list):
+                specs = [s for s in volume_data if isinstance(s, str)]
+            else:
+                continue
+
+            for spec in specs:
+                if not _is_valid(spec):
+                    return f"Volume mount '{spec}' missing explicit mode (:rw or :ro)"
+
+        return False
+
+
+def _is_valid(spec: str) -> bool:
+    if "{{" in spec:
+        return True
+    return bool(_VOLUME_PATTERN.match(spec.strip().strip("'\"")))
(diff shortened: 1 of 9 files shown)


.ansible-lint
```diff
@@ -4,3 +4,7 @@ enable_list:
 exclude_paths:
   - .ansible/
   - .github/
+
+# Enable custom rules for volume permission validation
+rulesdir:
```
evgeni: Don't do that without `use_default_rules: true`, as otherwise this (silently) disables all default rules.
Yes, even the ones in the `enable_list`.

ehelms: > Don't do that without `use_default_rules: true`, as otherwise this (silently) disables all default rules.
> Yes, even the ones in the `enable_list`.

:facepalm:

evgeni: Well, not disable, they are just not loaded as they are not in the "rulesdir" anymore. And enablement silently ignores that, as those rules are not disabled/opt-in, they are just "not there"

evgeni:
This needs a rebase

ehelms:
Given https://github.com/theforeman/foremanctl/pull/526 adds quite a bit more structure around custom lint rules, I'm inclined to get yours in and then I'll rebase / update mine based on what is established there.

ehelms:
I've put this in draft for now, and rebased it on https://github.com/theforeman/foremanctl/pull/526 as I would like to pursue it once that PR establishes the baseline.

evgeni:
Why is the default, implicit, rw bad?

development/requirements.txt
```diff
@@ -8,3 +8,4 @@ pytest-testinfra
 pytest-durations
 python-dateutil
 pyyaml
+ansible-lint
```
evgeni: line 1 already contains it

.ansible-lint-rules/no_volume_permissions.py
```diff
@@ -0,0 +1,54 @@
...
+
+from ansiblelint.rules import AnsibleLintRule
+
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import Lintable
+
+_VOLUME_PATTERN = re.compile(r'^[^:]+:[^:]+:(rw|ro)(?:,[zZ])?$')
+
+
+class NoVolumePermissionsRule(AnsibleLintRule):
```
evgeni: This should have some tests :)

ehelms:
> Why is the default, implicit, rw bad?

The origin of this is from https://redhat.atlassian.net/browse/SAT-42077 which references the following standard: BSI SYS 1.6 A19

Quoting that section of the standard (translated by Google translate):

> SYS.1.6.A19 Mounting Data Storage in Containers (S)
Containers SHOULD only be able to access the mass storage and directories necessary for operation. Permissions SHOULD be granted explicitly only when required. If the container runtime mounts local storage for a container, file system access rights SHOULD be restricted to the container's service account. If network storage is used, permissions SHOULD be set on the network storage itself.

evgeni:
So the idea is that we explicitly opt in to rw and use that to think whether ro would have been sufficient?

ehelms:
> So the idea is that we explicitly opt in to rw and use that to think whether ro would have been sufficient?

That's an element of it. My other take away is that users will need to audit this, and it helps if the permissions are explicitly stated whenever a mount is present.

.ansible-lint-rules/no_volume_permissions.py
```diff
@@ -0,0 +1,53 @@
...
+    """Volume mounts must specify explicit :rw or :ro permissions."""
+
+    id = "no-volume-permissions"
+    description = "Volume mounts must specify explicit :rw or :ro permissions."
```
evgeni: Technically, this is not "permission" but "mode", see `podman-run(1)`:
> ## Write Protected Volume Mounts
> Add :ro or :rw option to mount a volume in read-only or read-write mode, respectively. By default, the volumes are mounted read-write.

I'd also factor in the why, so how about:

```suggestion
    description = "Volume mounts must explicitly specify :rw or :ro mode, to ease auditing of access control."
```

.ansible-lint-rules/no_volume_permissions.py
```diff
@@ -0,0 +1,53 @@
...
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import Lintable
+
+_VOLUME_PATTERN = re.compile(r'^[^:]+:[^:]+:(rw|ro)(?:,[zZ])?$')
+
+
+class NoVolumePermissionsRule(AnsibleLintRule):
+    """Volume mounts must specify explicit :rw or :ro permissions."""
+
+    id = "no-volume-permissions"
```
evgeni: I think this should be `explicit-volume-mode` (also the filename etc)

evgeni (approved):
Wording nitpicks, but ACK on the concept

PR #403, opened 2026-03-09
Switch container networking from host to named bridge networks

pablomh:
Currently all containers run with `network: host`, which means they
share the host network namespace and communicate via localhost. This
works but provides no network isolation between services.

This PR replaces host networking with named Podman bridge networks:

- `foreman-db` (internal, isolated) — PostgreSQL; accessible only to
  services that need database access
- `foreman-cache` (internal, isolated) — Valkey; accessible only to
  services that need cache access
- `foreman-app` — Foreman, Pulp, Candlepin, httpd; shared application
  network
- `foreman-proxy` — Foreman and the smart proxy; used when the
  `foreman-proxy` feature is enabled
- `iop-core-network` (`10.130.0.0/24`) — the IOP services, when the
  `iop` feature is enabled; the IOP gateway is dual-homed onto this
  network and `foreman-app` so Foreman can reach it directly

Services communicate by container DNS name (e.g. `postgresql`,
`candlepin`, `iop-core-gateway`) rather than `localhost`. A dedicated
TLS certificate with `SAN=candlepin` is issued for the Candlepin
container, since Foreman now connects to it via the bridge DNS name
and validates the hostname. The IOP gateway's certificate covers both
`localhost` and `iop-core-gateway` so the same endpoint keeps working
from the host and from the bridge network.

PostgreSQL is accessed over its Unix socket rather than TCP/localhost
wherever a client and the database container share a host (backup,
health checks, and the internal-database roles); this avoids
publishing a database port at all when nothing outside the host needs
TCP access to it.

Candlepin does not publish any ports to the host — Foreman reaches
it directly over the bridge. Port publishing is minimised throughout:
PostgreSQL and Valkey never publish ports; Pulp API and content
endpoints publish to `127.0.0.1` for the httpd proxy; the smart
proxy publishes to `0.0.0.0` for external clients.

The `foreman-proxy` deployment always registers the smart proxy with
Foreman under its own real FQDN (independent of the bridge-network
split). `--registration-url` remains available as a separate,
optional override for load-balancer or multi-proxy setups: when set,
it's written into the proxy's own settings as the endpoint newly
registering hosts should use, without changing how Foreman itself
manages or validates the proxy.

Containers inherit host `/etc/hosts` entries via Podman's
`base_hosts_file` setting, so host-only name mappings remain usable
from inside containers even though they're no longer on the host
network.

The old-installer-answers migration path (`foremanctl migrate`) now
also rewrites an internal database host of `localhost`/`127.0.0.1` to
the `postgresql` container DNS name, since a bridge-networked
deployment can no longer reach the database container via loopback.

The development environment keeps `network: host` for
`postgresql`/`valkey`/`candlepin`/`pulp` to avoid reconfiguring every
service URL for a Rails development workflow where the Foreman
process runs directly on the host.

### DIFF
--- a/development/playbooks/deploy-dev/deploy-dev.yaml
+++ b/development/playbooks/deploy-dev/deploy-dev.yaml
@@ -46,10 +46,25 @@
     - role: systemd_target
     - role: certificates
     - role: postgresql
+      vars:
+        postgresql_network: host
     - role: valkey
+      vars:
+        valkey_network: host
     - role: candlepin
+      vars:
+        candlepin_networks: host
+        candlepin_database_host: localhost
+        candlepin_healthcheck_host: localhost
+        candlepin_tomcat_key: "{{ localhost_key }}"
+        candlepin_tomcat_certificate: "{{ localhost_certificate }}"
     - role: httpd
     - role: pulp
+      vars:
+        pulp_networks: host
+        pulp_migration_networks: host
+        pulp_database_host: localhost
+        pulp_redis_url: "redis://localhost:6379/8"
     - role: foreman_development
       vars:
         foreman_development_oauth_consumer_key: "{{ foreman_oauth_consumer_key }}"
--- a/development/playbooks/remote-database/remote-database.yaml
+++ b/development/playbooks/remote-database/remote-database.yaml
@@ -18,6 +18,8 @@
     - role: pre_install
     - role: certificates
     - role: postgresql
+      vars:
+        postgresql_network: host
 
   tasks:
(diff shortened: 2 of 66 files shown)


ehelms:
@pablomh A lot has changed since you opened this, with some things being removed and others being simplified. I moved to draft. If you get time, consider rebasing and updating the design based on the current state.

pablomh:
Hi @ekohl, @ehelms! Based on the comments in https://github.com/theforeman/foremanctl/pull/728, I feel like this approach is not the desired one so I'm closing it.

PR #421, opened 2026-03-25
Auto-detect and normalize installer certificates

ehelms:

Migrating from a foreman-installer deployment to foremanctl requires converting
both the answer file parameters and the certificate layout. Previously this had
to be done manually. The `migrate` subcommand now handles this in a single step,
but the migrate playbook was running without elevated privileges, causing a
permission error when reading `/etc/foreman-installer/` files.

* Add `migrate_foreman_installer` role that runs during `foremanctl migrate` to
  copy certificates from `/root/ssl-build/` into `/root/certificates/`, persist
  the CA passphrase into `parameters.yaml`, and back up the original directory
* Auto-detect custom server certificates by comparing the internal CA against
  the server CA; write `certificates_source: custom_server` to `parameters.yaml`
  when they differ so subsequent deploys do not overwrite the custom cert
* Remove the installer certificate source — migrated certs use standard default
  source paths after normalization; mark certificate path parameters as `IGNORE`
  in answer file migration since the role handles cert files directly
* Consolidate `default_certificates.yml` and `custom_server_certificates.yml`
  into a single `certificates.yml` (paths were identical after normalization)
* Add a dedicated CI job that exercises the full migrate workflow: install
  foreman-installer, run `foremanctl migrate`, deploy, and test
* Fix permission error in migrate playbook by adding `become: true` to the
  answer file migration play so the module can read files under
  `/etc/foreman-installer/`

Steps to reproduce:

* Run the migration CI job and verify it passes end-to-end
* On a host with an existing foreman-installer install, run `foremanctl migrate`
  and confirm a valid `parameters.yaml` is produced and certs are copied into place
* Verify a host with a custom server cert gets `certificates_source: custom_server`
  written to `parameters.yaml`
* Run `foremanctl deploy` after migration and confirm it succeeds without
  re-running certificate migration

#### Checklist


evgeni:
@ehelms could you please rebase this one

src/roles/certificates/tasks/main.yml
```diff
@@ -1,9 +1,24 @@
...
+    - certificates_ca
+    - not certificates_installer_ca.stat.exists
+
+- name: Apply custom server certificates
+  ansible.builtin.include_tasks: custom.yml
+  when: certificates_custom_server_certificate is defined
 
-- name: 'Issue other certificates'
+- name: Issue host certificates
   ansible.builtin.include_tasks: issue.yml
```
evgeni, line 13: This fails for "installer" certs, as the CA password is not properly loaded.

OTOH, I don't see why it would re-generate the installer issued certs to begin with.

src/roles/certificates/tasks/setup.yml
```diff
@@ -0,0 +1,24 @@
...
+  ansible.builtin.file:
+    path: "{{ certificates_ca_directory_certs }}"
+    state: directory
+    mode: '0755'
+
+- name: 'Create keys directory'
+  ansible.builtin.file:
+    path: "{{ certificates_ca_directory_keys }}"
+    state: directory
+    mode: '0755'
```
evgeni: the keys really should not be world readable

evgeni:
certs-check is right, `/root/certificates/certs/quadlet.example.com.crt` is signed by `Foreman Self-signed CA`, not by `Custom Test CA` - but why?!

"TASK [certificates : Sign server certificate] **********************************" did re-sign the cert, it really should not for "custom" certs, those can't be managed by us.

src/roles/certificates/tasks/main.yml
```diff
@@ -1,9 +1,24 @@
...
+
+- name: Normalize installer certificates
+  ansible.builtin.include_tasks: normalize.yml
+  when: certificates_installer_ca.stat.exists
+
+- name: Generate CA certificate
   ansible.builtin.include_tasks: ca.yml
-  when: certificates_ca
+  when:
+    - certificates_ca
```
evgeni: At this point, will that ever be false?

src/roles/certificates/tasks/main.yml
```diff
@@ -1,9 +1,24 @@
...
+    - not certificates_installer_ca.stat.exists
```
evgeni: normalize.yml deletes `/root/ssl-build/katello-default-ca.crt` after execution, so on the next run of the role `not certificates_installer_ca.stat.exists` will be `true` and `ca.yml` will run -- is that intentional?

src/playbooks/deploy/deploy.yaml
```diff
@@ -6,7 +6,7 @@
   vars_files:
     - "../../vars/defaults.yml"
     - "../../vars/flavors/{{ flavor }}.yml"
-    - "../../vars/{{ certificate_source }}_certificates.yml"
+    - "../../vars/default_certificates.yml"
```
evgeni: arguably, this file now should be just `certificates.yml`, right?

src/roles/certificates/tasks/main.yml
evgeni: yepp, ran into this in https://github.com/theforeman/foremanctl/actions/runs/24706818357/job/72261872791?pr=421

evgeni:
The more I think about it, the more I think `normalize` should be part of the "one time migration from old installer" (as started in #446), not a thing that "continuously" runs as I think copying files is not sufficient for that step. We need to at least also make foremanctl aware of the passphrase of the CA (I hacked that up with `slurp` now, but I hate it) and ideally also import the old CA properties so that it does not get regenerated on follow up runs.

and the `custom` cert setup is a nice addition, but is also valuable outside the `normalize` flow, so IMHO this PR should be two?

ehelms:
I agree I can now pull out the custom certificates to their own PR. And that, installer certificates via migration handling makes more sense as a one time operation.

Do you generally agree on the direction of installer certs support? That is should be "converted" so `foremanctl` manages it like our default ones?

ehelms:
Custom certificate handling now exists at: https://github.com/theforeman/foremanctl/pull/462
This will need to be incorporated into the `migrate` command once that's available. Setting this to draft for now as it will require re-work once those two conditions are met.

evgeni:
> Do you generally agree on the direction of installer certs support? That is should be "converted" so `foremanctl` manages it like our default ones?

Yes, as the setup (well, at the very least the CA) need to be fully usable in foremanctl to deploy new proxies etc.

src/roles/migrate_foreman_installer/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+migrate_foreman_installer_ca_directory: /root/certificates
```
evgeni: This is not correct anymore

src/roles/migrate_foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,152 @@
...
+      ansible.builtin.copy:
+        src: /root/ssl-build/
+        dest: /root/ssl-build.bak/
+        remote_src: true
+        mode: preserve
+
+    - name: Remove original installer certificate directory
+      ansible.builtin.file:
+        path: /root/ssl-build
+        state: absent
```
evgeni, line 153: Can/should this be a `command: mv`? Neither seems too Ansible-style to me, but I wanted to ask.

src/playbooks/migrate/migrate.yaml
```diff
@@ -27,3 +28,10 @@
       vars:
         _unmappable_warning: "Warning: {{ migration_result.unmappable | length }} parameter(s) could not be mapped - see warnings above"
         _output_file_msg: "Output written to: {{ migration_result.output_file }}"
+
+- name: Migrate foreman-installer certificates
```
evgeni: Given this alters the system, should it be behind a flag? Users might want to call `foremanctl migrate` many times before they are satisfied with the answers-migration.

docs/user/certificates.md
```diff
@@ -174,10 +171,9 @@ For `certificate_source: custom_server`:
 2. **Custom Server Certificates**: Copy the custom server cert, key, and CA bundle from user-provided paths to `/root/certificates/` (only when certificate paths are provided)
 3. **Host Certificate Issuance**: Generate client certificate and localhost certificate signed by the internal CA (server cert for FQDN is skipped)
 
-For `certificate_source: installer`:
+#### Migration from foreman-installer
 
-- Uses existing certificates from `/root/ssl-build/` generated by foreman-installer
-- No certificate generation performed; files must already exist
+The `foremanctl migrate` command includes a `migrate_foreman_installer` role that normalizes `foreman-installer` certificates into the canonical `/root/certificates/` structure. It also reads the CA passphrase from the installer's password file and persists it into foremanctl's configuration so that subsequent deploys can issue new certificates using the original CA.
```
evgeni: `/var/lib/…`, not `/root`

evgeni:
> certs-check is right, `/root/certificates/certs/quadlet.example.com.crt` is signed by `Foreman Self-signed CA`, not by `Custom Test CA` - but why?!
> 
> "TASK [certificates : Sign server certificate] **********************************" did re-sign the cert, it really should not for "custom" certs, those can't be managed by us.

This is still happening. The CA private key remains stable, but the self-signed cert is updated (from `C=US, ST=North Carolina, L=Raleigh, O=Katello, OU=SomeOrgUnit, CN=quadlet.example.com` to `CN=Foreman Self-signed CA`), and then everything falls apart.

The problem is that the CSR doesn't match anymore (different CN etc), and this triggers a regeneration of the cert. Logical, but painful.

src/roles/migrate_foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,153 @@
...
+
+    - name: Read CA password from installer
+      ansible.builtin.slurp:
+        src: /root/ssl-build/katello-default-ca.pwd
+      register: migrate_foreman_installer_installer_ca_password
+      no_log: true
+
+    - name: Persist CA password to foremanctl configuration
+      ansible.builtin.copy:
+        dest: "{{ obsah_state_path }}/certificates-ca-password"
```
evgeni, line 94: Can't we use `certificates_ca_password_file` here?

src/roles/certificates/tasks/ca.yml
```diff
@@ -58,6 +60,7 @@
       - digitalSignature
     key_usage_critical: true
     create_subject_key_identifier: true
+    force: "{{ certificates_renew_ca }}"
```
evgeni: This is not sufficient to guard against re-generation when input parameters change :/

src/roles/migrate_foreman_installer/tasks/main.yml
ehelms: Should we then do both? I thought this file was the source of truth?

evgeni:
```
src/vars/base.yaml:certificates_ca_password_file: "{{ obsah_state_path }}/certificates-ca-password"
```

The variable has exactly the same value, so it ends up in the same file.
I was merely pointing at the fact that we have a var for that, and using it will ensure we don't forget to update it if we ever have to ;)

ehelms: Fair, just means I gotta go make another role variable yadda yadda. Maybe we need some canonical variables that are acceptable across roles...

evgeni (approved):

PR #391, opened 2026-02-27
Remove artemis

jturel:
Katello no longer consumes events from Candlepin, and does not connect to the embedded Artemis broker. This is part of an effort to clean the remaining pieces which are:

- https://github.com/theforeman/puppet-candlepin/pull/278
- https://github.com/theforeman/katello-selinux/pull/65
- https://github.com/theforeman/puppet-katello/pull/525
- https://github.com/theforeman/foreman-packaging/pull/13091
- https://github.com/theforeman/foremanctl/pull/391
- https://github.com/theforeman/puppet-katello_devel/pull/321

The test failure seems expected - hammer ping will pass once new hammer and katello RPMs reach nightly  repos. I was able to spin up a dev environment and see that Candlepin was healthy.

PR #424, opened 2026-03-26
Switch to using nightly images

ehelms:
The mainline development branch for the installer should use the most up to date images as we iterate. This would change to use the nightly image tags going foward.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,17 +1,53 @@
 # Release
 
-To create a release, create a tag.
-It must follow the x.y.z pattern without any prefix.
+## Version Scheme
+
+- `master` always contains a tag with the current development version: `X.Y-develop`
+- Stable branches are named `X.Y-stable`
+- Release candidate tags on stable branches follow `X.Y.Z-rcN` (e.g. `3.18.0-rc1`)
+- Release tags on stable branches follow `X.Y.Z` (e.g., `3.18.0`, `3.18.1`)
+
+## Branching for a Release
+
+When ready to stabilize a version (e.g., 3.18):
+
+1. Create the stable branch from master:
+
+   ```
+   git fetch origin
+   git checkout -b 3.18-stable origin/master
+   git push --set-upstream origin HEAD
+   ```
+
+2. Update `src/vars/images.yml` on the stable branch to reference the release images (e.g., `3.18`).
+
+   ```
+   sed -i '/container_tag_stream:/ s/:.*/: "3.18"/' src/vars/images.yml
+   ```
+
+3. Bump master to the next development version:
+
+   ```
+   git checkout master
+   git tag -s "3.19-develop" -m "Start 3.19 development"
+   git push --follow-tags
+   ```
+
+## Cutting a Release
+
+To release from a stable branch (e.g., 3.18.0):
 
 ```
-VERSION=x.y.z
-git tag -s "$VERSION" -m "Release $VERSION"
+git checkout 3.18-stable
+git tag -s "3.18.0" -m "Release 3.18.0"
 git push --follow-tags
 ```
 
 This will create a GitHub release and attach the created tarball to it.
 
-Once that is done, the packaging is handled in the `foreman-packaging` repository where the spec file is maintained:
+## Packaging
+
+Once that is done, the packaging is handled in the [foreman-packaging](https://github.com/theforeman/foreman-packaging) repository where the spec file is maintained:
 
     obal update foremanctl --version $VERSION
     gh pr create --base rpm/develop
--- a/src/vars/images.yml
+++ b/src/vars/images.yml
@@ -6,7 +6,7 @@ postgresql_registry_auth_file: "{{ registry_auth_file }}"
 pulp_registry_auth_file: "{{ registry_auth_file }}"
 redis_registry_auth_file: "{{ registry_auth_file }}"
 
-container_tag_stream: "3.18"
+container_tag_stream: "nightly"
 candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "foreman-{{ container_tag_stream }}"
 foreman_container_image: quay.io/foreman/foreman


ehelms:
@evgeni @stejskalleos @arvind4501 -- https://github.com/theforeman/foremanctl/pull/391 presents a solid reason to switch to using nightly images where the code base has an architectural deviation from release-based images. As more developers start contributing here, I think we need to be able to pull in the nightly code bases.

src/vars/images.yml
```diff
@@ -6,7 +6,7 @@ postgresql_registry_auth_file: "{{ registry_auth_file }}"
 pulp_registry_auth_file: "{{ registry_auth_file }}"
 redis_registry_auth_file: "{{ registry_auth_file }}"
 
-container_tag_stream: "3.18"
+container_tag_stream: "nightly"
```
stejskalleos, line 9: Seems reasonable to me, but checking the [quay](https://quay.io/repository/foreman/foreman?tab=tags&tag=latest), the latest nightly is 15 days old.

IMO, we should make sure we have regular, real nightly builds first.

arvind4501:
I think having nightly pinned is good for development.

Thinking out loud, at some point we need to think how can we ship foremanctl with stable images(3.18, 3.19) or maybe align foremanctl releases with foreman, but not necessarily today.

src/vars/images.yml
ehelms: While I agree that would be nice,  we can build nightly updates when we need to so I do not think that should block this.

evgeni:
Please keep in mind that this change means:
- We are at risk of becoming the CI for nightly, when changes to the `nightly` container tags are pushed out without proper testing of the containers (which could totally be implemented by running `foremanctl` pointing at some staged containers, but that's not how things are done today)
- We need to create stable branches for setting the image stream to a stable Foreman release/branch
- The above breaks `git describe` for `master` as the tags will be only inside the stable branches (you can see that effect in `theforeman/foreman` today) and thus breaks the [nifty VERSION handling Ewoud built](https://github.com/theforeman/foremanctl/commit/92bb1cd485e52da8289f11a2e081215ed4f6da40) and Packit RPM versioning

That shouldn't mean I am against that change, quite the opposite. But it's more than one line :/

ehelms:
> Please keep in mind that this change means:
> 
> * We are at risk of becoming the CI for nightly, when changes to the `nightly` container tags are pushed out without proper testing of the containers (which could totally be implemented by running `foremanctl` pointing at some staged containers, but that's not how things are done today)

Arguably, `foremanctl` and it's test should be the CI. This would just mean right now that PRs to this repository would be that CI rather than PRs to the image repositories themselves. I agree we should close that gap. I suppose this is an order of operations question.

> * We need to create stable branches for setting the image stream to a stable Foreman release/branch

Do I translate this correctly to "foremanctl needs to be in the branching procedure for Foreman releases" ?

> * The above breaks `git describe` for `master` as the tags will be only inside the stable branches (you can see that effect in `theforeman/foreman` today) and thus breaks the [nifty VERSION handling Ewoud built](https://github.com/theforeman/foremanctl/commit/92bb1cd485e52da8289f11a2e081215ed4f6da40) and Packit RPM versioning

This translates to, we gotta make this nifty version thing nightly aware?

> 
> That shouldn't mean I am against that change, quite the opposite. But it's more than one line :/

evgeni:
> > Please keep in mind that this change means:
> > 
> > * We are at risk of becoming the CI for nightly, when changes to the `nightly` container tags are pushed out without proper testing of the containers (which could totally be implemented by running `foremanctl` pointing at some staged containers, but that's not how things are done today)
> 
> Arguably, `foremanctl` and it's test should be the CI. This would just mean right now that PRs to this repository would be that CI rather than PRs to the image repositories themselves. I agree we should close that gap. I suppose this is an order of operations question.

The code should be used in CI, yes.
What I want to avoid is that a PR is opened against `foremanctl` and *that* PR is red because a container image changed.

> > * We need to create stable branches for setting the image stream to a stable Foreman release/branch
> 
> Do I translate this correctly to "foremanctl needs to be in the branching procedure for Foreman releases" ?

Correct

> > * The above breaks `git describe` for `master` as the tags will be only inside the stable branches (you can see that effect in `theforeman/foreman` today) and thus breaks the [nifty VERSION handling Ewoud built](https://github.com/theforeman/foremanctl/commit/92bb1cd485e52da8289f11a2e081215ed4f6da40) and Packit RPM versioning
> 
> This translates to, we gotta make this nifty version thing nightly aware?

Yes, albeit I am not sure this is actually possible.

ehelms:
> The code should be used in CI, yes.
> What I want to avoid is that a PR is opened against `foremanctl` and _that_ PR is red because a container image changed.

I have an initial attempt at that: https://github.com/theforeman/foreman-oci-images/pull/27
This is failing at the moment because it's a chicken and egg problem.

Nightly has changed behavior, so a test is failing. However, that change can't be made here until the repository relies on the nightly images. However, we can't build the nightly images because the tests fail....

ehelms:
> > > * The above breaks `git describe` for `master` as the tags will be only inside the stable branches (you can see that effect in `theforeman/foreman` today) and thus breaks the [nifty VERSION handling Ewoud built](https://github.com/theforeman/foremanctl/commit/92bb1cd485e52da8289f11a2e081215ed4f6da40) and Packit RPM versioning
> > 
> > 
> > This translates to, we gotta make this nifty version thing nightly aware?
> 
> Yes, albeit I am not sure this is actually possible.

I am not following how this is affected by using images built based on the "nightly" code bases.

evgeni:
> > > > * The above breaks `git describe` for `master` as the tags will be only inside the stable branches (you can see that effect in `theforeman/foreman` today) and thus breaks the [nifty VERSION handling Ewoud built](https://github.com/theforeman/foremanctl/commit/92bb1cd485e52da8289f11a2e081215ed4f6da40) and Packit RPM versioning
> > > 
> > > 
> > > This translates to, we gotta make this nifty version thing nightly aware?
> > 
> > 
> > Yes, albeit I am not sure this is actually possible.
> 
> I am not following how this is affected by using images built based on the "nightly" code bases.

Not the images, but the fact we use stable branches. Probably easiest is to go back to a normal `VERSION` file.

evgeni:
> > The code should be used in CI, yes.
> > What I want to avoid is that a PR is opened against `foremanctl` and _that_ PR is red because a container image changed.
> 
> I have an initial attempt at that: [theforeman/foreman-oci-images#27](https://github.com/theforeman/foreman-oci-images/pull/27) This is failing at the moment because it's a chicken and egg problem.
> 
> Nightly has changed behavior, so a test is failing. However, that change can't be made here until the repository relies on the nightly images. However, we can't build the nightly images because the tests fail....

In forklift, we worked around this by annotating the tests with data which component versions they apply to. Should we do something similar here? Or just drop the `candlepin_events` tests knowing it's on the way out anyway

ehelms:
> In forklift, we worked around this by annotating the tests with data which component versions they apply to. Should we do something similar here? Or just drop the `candlepin_events` tests knowing it's on the way out anyway

I think you are asking a big question here: do we want `foremanctl` main branch to track Foreman develop or do we want foremanctl main branch to track support across versions?

For tests, I feel like they should be smarter and only execute if the feature is present. This is what I tried to do with `iop`.

evgeni:
Yes and no. "Run IOP tests only if IOP feature is enabled" makes sense, but what if the set of tests for said feature vary by version?

ehelms:
> Yes and no. "Run IOP tests only if IOP feature is enabled" makes sense, but what if the set of tests for said feature vary by version?

Good question. I think that goes back to my other question: do we want `foremanctl` to be multi-version or for it to track the version of Foreman via branching structure?

ehelms:
> > This translates to, we gotta make this nifty version thing nightly aware?
> 
> Yes, albeit I am not sure this is actually possible.

If I run make, right now I get `foremanctl-1.2.0-44-g7aeaf71.tar.gz` which seems reasonable as it has the git commit in it and the last tagged version. I don't think we need `nightly` or such in the name, but this does always use the base of the last release.

I get that means we cannot pre-emptively update the version to indicate what the next target version is. However, if we always work from the viewpoint that if there is a git hash it's saying it's version X + some amount of nightly git bits. I think that works pretty good for a mental model.

evgeni:
> > Yes and no. "Run IOP tests only if IOP feature is enabled" makes sense, but what if the set of tests for said feature vary by version?
> 
> Good question. I think that goes back to my other question: do we want `foremanctl` to be multi-version or for it to track the version of Foreman via branching structure?

I think we don't need it to be multi version if we branch accordingly.



> > > This translates to, we gotta make this nifty version thing nightly aware?
> > 
> > 
> > Yes, albeit I am not sure this is actually possible.
> 
> If I run make, right now I get `foremanctl-1.2.0-44-g7aeaf71.tar.gz` which seems reasonable as it has the git commit in it and the last tagged version. I don't think we need `nightly` or such in the name, but this does always use the base of the last release.

It should be 2.0.0-… - you probably are missing a rebase, but that's not important for the argument.

> I get that means we cannot pre-emptively update the version to indicate what the next target version is. However, if we always work from the viewpoint that if there is a git hash it's saying it's version X + some amount of nightly git bits. I think that works pretty good for a mental model.

And exactly this won't work (automatically) once we move to stable branches as `git describe` uses the newest tag it can find in the ancestry, but as the new release tags are on stable branches, it will pick up something old. Try it out in `foreman.git` or `foreman-installer.git` to see what I mean.

ehelms:
Thinking and playing around with this more I see two strategies:

#### Direct Version Management

This would be reverting to the classical approach of keeping the version number in `VERSION` file. The version in master would track the "next" version and branches would contain and be updated with the current version for said branch.

#### Keeping `export-subst` 

Keep the current method of using `export-subst` and implement the following:

 * In master, set `VERSION` set it to `%H$` so that the VERSION file has the commit hash of the point in time the tarball was created
 * In branches, set `VERSION` to `%(describe)$` so that VERSION file has the tag commit of tarball

evgeni:
Given that both strategies involve touching `VERSION`, *plus* `%H$` not sorting correctly when used as an RPM version (so we'd have to use `NEXT_VESION-%H` or similar to achieve that), I'd vote for "Direct Version Management"

ekohl:
> The above breaks `git describe` for `master` as the tags will be only inside the stable branches (you can see that effect in `theforeman/foreman` today) and thus breaks the [nifty VERSION handling Ewoud built](https://github.com/theforeman/foremanctl/commit/92bb1cd485e52da8289f11a2e081215ed4f6da40) and Packit RPM versioning

For Foreman I already thought about this and we have the same problem. There we have commits like https://github.com/theforeman/foreman/commit/ba8f2688e160b1cc225b02b5830e540d43cf84fa and I'd suggest to re-implement those as git tags when using `export-subst`.

Big question: what is that develop version? Do we want to start tying the foremanctl version numbers to Foreman or stick to foremanctl's own versions? It would be easier for users to see what Foreman version they will get when they install it.

ehelms:
> For Foreman I already thought about this and we have the same problem. There we have commits like [theforeman/foreman@ba8f268](https://github.com/theforeman/foreman/commit/ba8f2688e160b1cc225b02b5830e540d43cf84fa) and I'd suggest to re-implement those as git tags when using `export-subst`.

Seeing if I follow. You are proposing to tag the repo when we "bump" the version to a development version. So we might have:

1. master tagged with `3.18-develop`
2. branch to `3.18-stable`
3. tag master `3.19-develop`
4. `3.18-stable` branch gets a `3.18.0` tag on release, etc.

ekohl:
> Seeing if I follow. You are proposing to tag the repo when we "bump" the version to a development version. So we might have:

Kind of yes. Perhaps this helps
```mermaid
---
config:
    gitGraph:
        mainBranchName: develop
---
gitGraph
    commit tag: "3.18-develop"
    commit
    commit
    branch 3.18-stable
    checkout 3.18-stable
    commit
    commit tag: "3.18.0-rc1"
    checkout develop
    commit tag: "3.19-develop"
    commit
```

So at the end I'd expect `git describe` to show `3.19-develop-1-6f07a2d` because it's 1 commit after `3.19-develop` with git sha `6f07a2d`.

RELEASE.md
```diff
@@ -1,17 +1,48 @@
 # Release
 
-To create a release, create a tag.
-It must follow the x.y.z pattern without any prefix.
+## Version Scheme
+
+- `master` is always tagged with the current development version: `X.Y-develop`
```
ekohl: I think we don't keep the tag updated. The tag is created when we branch a stable release
```suggestion
- `master` always contains a tag with the current development version: `X.Y-develop`
```

RELEASE.md
```diff
@@ -1,17 +1,48 @@
...
+- Stable branches are named `X.Y-stable`
+- Release tags on stable branches follow `X.Y.Z` (e.g., `3.18.0`, `3.18.1`)
```
ekohl, line 8: Perhaps also mention the RC tags? Not sure if it should be be separated with a dash or dot and also uppercase vs lowercase. So `3.18.0-RC1`, `3.18.0.RC1`, `3.18.0-rc1`, or `3.18.0.rc1`.

RELEASE.md
```diff
@@ -1,17 +1,48 @@
...
+## Branching for a Release
+
+When ready to stabilize a version (e.g., 3.18):
+
+1. Create the stable branch from master:
+
+   ```
+   git checkout master
+   git checkout -b 3.18-stable
+   git push origin 3.18-stable
```
ekohl: You don't need to be on a branch to branch it there and you can push the current branch with HEAD, avoiding copy-paste errors. Then if you assume the remote is `origin` you can also ensure the latest version is used, even if your local branch is older. I also like setting up branch tracking by using `--set-upstream`
```suggestion
   git fetch origin
   git checkout -b 3.18-stable origin/master
   git push --set-upstream origin HEAD
```

Or even without touching your local checkout at all:
```suggestion
   git fetch origin
   git push --set-upstream origin origin/master:3.18-stable
```

But that may end up being a bit too magical.

ehelms:
Updated with a release process. I'd like to propose the order of operations:

1. Merge this change here with release documentation
2. Follow up with branch / release updates to https://github.com/theforeman/theforeman-rel-eng

RELEASE.md
```diff
@@ -1,17 +1,49 @@
...
+
+1. Create the stable branch from master:
+
+   ```
+   git fetch origin
+   git checkout -b 3.18-stable origin/master
+   git push --set-upstream origin HEAD
+   ```
+
+2. Update `src/vars/images.yml` on the stable branch to reference the release images (e.g., `foreman-3.18`).
```
ekohl: Perhaps suggest a command?
```
sed -i '/container_tag_stream/ s/nightly/foreman-3.18/' src/vars/images.yml
```

ekohl (approved):
That plan makes sense to me.

RELEASE.md
```diff
@@ -1,17 +1,49 @@
 # Release
 
-To create a release, create a tag.
-It must follow the x.y.z pattern without any prefix.
+## Version Scheme
+
+- `master` always contains a tag with the current development version: `X.Y-develop`
+- Stable branches are named `X.Y-stable`
+- Release candidate tags on stable branches follow `X.Y.Z-rc` (e.g. `3.18.0-rc1`)
```
evgeni:
```suggestion
- Release candidate tags on stable branches follow `X.Y.Z-rcN` (e.g. `3.18.0-rc1`)
```

RELEASE.md
evgeni:
```
sed -i '/container_tag_stream:/ s/:.*/: "${{ matrix.upgrade_to }}"/' src/vars/images.yml
```
is what we use in our CI

evgeni: the reference is `3.18`, not `foreman-3.18`
```suggestion
2. Update `src/vars/images.yml` on the stable branch to reference the release images (e.g., `3.18`).
```

evgeni (approved):
nitpicks :)

ehelms:
Addressed nitpicks. I am going to opt to merge this in now, so that https://github.com/theforeman/foremanctl/pull/391 can go in and `foremanctl` can reflect the state of nighty before branching.

PR #429, opened 2026-03-28
Fixes #72 - Replace sidekiq-selinux wrapper with direct sidekiq call

### LINKED ISSUE #72: Review dynflow-sidekiq exec command
Based on https://github.com/evgeni/talks/pull/4#discussion_r1932083989 we should review the command:

https://github.com/theforeman/foreman-quadlet/blob/177348fc149e74a4adb6680cb239181f16cf8476/roles/foreman/tasks/main.yaml#L90

This copies the `DYNFLOW_SIDEKIQ_SCRIPT` (that's also set as an env var) and hardcodes the Rails environment, but do we even need to expose this in the first place? It's based on https://github.com/theforeman/foreman/blob/74ddd8851cebd21c268002e0c82a3619d2a9db66/extras/systemd/dynflow-sidekiq%40.service#L12-L20 but is it better to provide a better `/usr/libexec/foreman/sidekiq-selinux` wrapper script that has these defaults built in?

jakduch:
The sidekiq-selinux wrapper at /usr/libexec/foreman/sidekiq-selinux is a packaging artifact that only exists for SELinux domain transition on bare-metal RPM installations. Its entire content is:

    exec sidekiq "\'$@"\'

In containers where SELinux is disabled (label=disable), this wrapper serves no purpose. This replaces the wrapper call with a direct sidekiq invocation and uses an absolute path for the dynflow-sidekiq.rb script.

Based on the discussion in https://github.com/evgeni/talks/pull/4#discussion_r1932083989

cc @ekohl @evgeni

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -154,7 +154,7 @@
       DYNFLOW_REDIS_URL: "redis://localhost:6379/6"
       REDIS_PROVIDER: "DYNFLOW_REDIS_URL"
       FOREMAN_ENABLED_PLUGINS: "{{ foreman_plugins | join(' ') }}"
-    command: "/usr/libexec/foreman/sidekiq-selinux -e production -r ./extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/%i.yml"
+    command: "sidekiq -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/%i.yml"
     quadlet_options:
       - |
         [Install]


ekohl:
I must have missed the ping.

IMHO there's still value in keeping RPM and non-RPM installations equivalent. That's why I'm tempted to go in a slightly different direction: have a more complete wrapper that is something like `/usr/libexec/foreman/foreman-worker` that's something like:
```sh
#!/bin/sh
exec sidekiq -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C "$1"
```

Or even include the path already:

```sh
#!/bin/sh
exec sidekiq -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C "/etc/foreman/dynflow/$1.yml"
```

If that's included in Foreman itself and thus the RPM, all tooling can use this. If SELinux is present then it can apply the correct label.

A big benefit of this is that the orchestration layer (that foremanctl really is) becomes less aware of container internals.

ehelms:
@jakduch Would you still like to pursue this based on @ekohl feedback?

jakduch:
Thanks @ekohl, that makes sense — I agree the orchestration layer shouldn't need to know the exact sidekiq invocation, and keeping RPM and container installs equivalent is worth it. @ehelms yes, I'd like to pursue it.

I'll go with your second variant — the wrapper builds the config path so callers only pass the instance name — shipped where `sidekiq-selinux` lives today (Foreman / the RPM) and keeping the SELinux domain transition so bare-metal behaves as before:

```sh
#!/bin/sh
exec sidekiq -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C "/etc/foreman/dynflow/$1.yml"
```

This PR's quadlet command then becomes just `foreman-worker %i`, which drops all the sidekiq/dynflow specifics from foremanctl.

I'll open the wrapper PR first and rebase this one on top to call it.

ehelms:
@jakduch Checking back in with you as you said you'd like to pursue it.

jakduch:
Thanks for the reminder. I added the generic worker entry point in theforeman/foreman#11275, its RPM installation in theforeman/foreman-packaging#14150, and the SELinux label in theforeman/foreman-selinux#189. This PR is now rebased and only calls `foreman-worker %i`, so foremanctl no longer carries the Sidekiq details.

jakduch:
The deployment matrix is currently failing because the nightly Foreman image does not yet contain `/usr/libexec/foreman/foreman-worker` (the SOS report shows exit 127 for all three Dynflow services). This PR depends on theforeman/foreman#11275 and theforeman/foreman-packaging#14150; once those are merged and the nightly image is rebuilt, the command will be available. Lint and devel tests are green.

PR #431, opened 2026-03-29
Fixes #353 - Add health check command

### LINKED ISSUE #353: provide a healthcheck option
In https://github.com/theforeman/foremanctl/pull/352 it came up that it'd be cool if foremanctl would be able to perform health checks
- as the last step of the deployment
- as a standalone command

jakduch:
Adds a new `foremanctl health` command that performs post-install health checks on all Foreman services.

**Checks performed:**
- Core services running (foreman, httpd, redis, postgresql)
- Dynflow workers running (orchestrator, worker, worker-hosts-queue)
- Pulp services running (pulp-api, pulp-content)
- Candlepin service running
- Foreman API responding (GET /api/v2/ping)
- Foreman tasks status (via Katello ping)

The command reports individual results for each check and provides a summary. It exits non-zero when any check fails, making it suitable for scripting and CI use.

As discussed in the issue, this can also be integrated as a final step of deployment in the future.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -160,6 +160,9 @@ jobs:
             --add-feature remote-execution \
             --add-feature bmc \
             ${{ matrix.iop == 'enabled' && '--add-feature iop' || '' }}
+      - name: Run health check
+        run: |
+          ./foremanctl health
       - name: Run tests
         run: |
           ./forge test --pytest-args="--certificate-source=${{ matrix.certificate_source }} --database-mode=${{ matrix.database }}"
@@ -284,6 +287,9 @@ jobs:
       - name: Run deployment
         run: |
           ./foremanctl deploy
+      - name: Run health check
+        run: |
+          ./foremanctl health
       - name: Run tests
         run: |
           ./forge test
--- a/src/filter_plugins/foremanctl.py
+++ b/src/filter_plugins/foremanctl.py
@@ -101,6 +101,11 @@ def available_foreman_proxy_plugins(_value):
     return compact_list(plugins)
 
 
+def has_feature(features, feature):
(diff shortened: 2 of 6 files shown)


pablomh:
Could you please show the output? Thx!

jakduch:
Good questions.

**Service vs API checks - do we need both?**

They operate at different layers. Service checks verify that systemd units are running - this works offline and does not depend on a functional API. The API check (`/api/v2/ping`) verifies end-to-end functionality including database connectivity and task runner status. A service can be running but not responding (misconfiguration, certificate issues), and conversely the API might not report on every systemd unit. That said, if the ping endpoint already covers everything we need, we could simplify to just the API check. Happy to go either way.

**Conditional checks based on features - yes, agreed.**

Updated Pulp and Candlepin checks to only run when `content/` features are enabled (`enabled_features | select("contains", "content/")`). Same condition on the Katello tasks status check in `check_foreman_api`. Core services and dynflow are always checked. Force-pushed.

**Validation - could you clarify what you have in mind?**

An `ansible-playbook --syntax-check` for the new playbook, a test scenario in CI, or something else?

ehelms:
@jakduch Would you be able to rebase this?

jakduch:
> @jakduch Would you be able to rebase this?

Rebased, the Ansible Lint failure looks like Galaxy timeout, not related to the changes.

qcjames53 (approved):
Thanks for the PR, Jakub. It looks really solid to me. Service checking passes when valid and fails when needed. The feature matching to determine status checks works and is appropriate. The tasks status check looks good as well. I think running the health playbook both on tests and in upgrade makes sense to me. I see no syntax issues so this looks good to merge for me!

ehelms (approved):

PR #439, opened 2026-03-31
Add custom Pulp containerfile and use information

ianballou:
When upgrading Pulp, Katello developers need a way to install new, untested versions of Pulpcore and plugins into development environments. Previously this meant install new versions with `pip`. Now, we can simply use a new Pulp container. However, we cannot use the container shipped by Pulp because it does not include pulp_smart_proxy.

I've added a containerfile that builds the plugins Katello requires into Pulp. It also includes some code to hack the pulp_smart_proxy Pulpcore requirement so it can run with newer Pulp without updating the bound before the plugin is ready.

The documentation has some usage information with instructions on how to run `deploy-dev` using new containerfiles. I also included a part about trying newer versions of postgres, since it was necessary when running Pulp with newer Django. I don't imagine we'll be using this that frequently, but it's good to have on record.

Question to maintainers - the containerfile will be used somewhat infrequently which is why I figured the somewhat ugly comments would be fine. If it's preferred, I could make a second containerfile just for building the hacked pulp_smart_proxy.


docs/development-environment.md
```diff
@@ -65,6 +65,62 @@ By default `hammer` feature will set up `hammer-cli` and `hammer-cli-foreman`, `
...
+
+```bash
+./forge deploy-dev \
+    --target-host=my-dev-box \
+    --extra-vars pulp_container_image="quay.io/yourusername/pulp" \
+    --extra-vars pulp_container_tag="custom" \
+    --add-feature=foreman-proxy
+```
+
+### Customizing PostgreSQL Version
```
ehelms: Do you feel this section is still needed given the code base has upgraded to PG 16?

ianballou: I don't, this whole thing will get a look over once https://github.com/theforeman/pulp-oci-images/pull/13 is in. I'd like to keep it simple.

ehelms (approved):
As this is only docs changes, the failing test is unrelated.

PR #443, opened 2026-04-02
Add features support for dev env

ehelms:
#### What is the purpose of this pull request?

 The development environment had its own hardcoded plugin registry and  `--foreman-development-enabled-plugin` CLI parameter, completely independent of the feature  system used by production deployments. This duplicated plugin metadata between `src/` and  `development/` with different data structures.

  - Add `development/features.yaml` that extends `src/features.yaml` with `development:`  blocks (git repos, settings templates, module configs) and defines dev-only features  (foreman-tasks, discovery, openscap, etc.)
  - Add `development/filter_plugins/foremanctl_dev.py` with a `features_to_dev_plugins` filter  that deep-merges dev features with src/ and resolves transitive dependencies, importing  shared logic from `src/filter_plugins/foremanctl.py`
  - Update task loops in main.yaml, hammer/main.yml, and smart-proxy/main.yml to use  `enabled_features |  features_to_dev_plugins` instead of the old plugin registry
  - Remove `foreman_development_plugin_registry`, `foreman_development_enabled_plugins`, and  the `--foreman-development-enabled-plugin` parameter
  - Add `ansible`, `rh-cloud`, and `iop` feature definitions to `src/features.yaml`

  #### What are the testing steps for this pull request?

  - Run `./forge deploy-dev` and verify default plugins (katello, remote-execution + dynflow)  are deployed
  - Run `./forge deploy-dev --add-feature=ansible` and verify ansible and its dependencies are  set up
  - Enable `hammer` and `foreman-proxy` features and confirm their plugin loops pick up the correct dev configs


ekohl:
Conceptually I like the blending of dev and env and going to throw some possibly wild idea out there. We already have features for `foreman` and `foreman-proxy`:
https://github.com/theforeman/foremanctl/blob/fbe19fc1631d4c1bded938a6b29a17d4baaefd3b/src/features.yaml#L2-L5
Should we take this a step further and implement a `foreman-dev` and `foreman-proxy-dev` to get the development server? Perhaps the specific feature can add an override. Something like:
```yaml
tasks:
  internal: true
  foreman-dev:
    gem_name: foreman-tasks
```

Then the flow can be:
```
foremanctl deploy --flavor development # gives you the foreman-dev feature
foremanctl deploy --flavor development --add-feature foreman-proxy # to use a production Foreman Proxy
foremanctl deploy --flavor development --add-feature foreman-proxy-dev # to use a development Foreman Proxy
```

PR #444, opened 2026-04-02
Add foreman_ansible

ehelms:


src/features.yaml
```diff
@@ -29,6 +29,15 @@ remote-execution:
...
     - dynflow
+foreman-ansible:
+  description: Ansible plugin for Foreman
+  foreman:
+    plugin_name: foreman_ansible
+  foreman_proxy:
+    plugin_name: ansible
+  hammer: foreman_ansible
+  dependencies:
+    - dynflow
```
arvind4501: I see remote_execution as dependecy of smart proxy ansible plugin too. https://github.com/theforeman/smart_proxy_ansible/blob/master/smart_proxy_ansible.gemspec#L32C31-L32C63, so by the metadata defination, should we add

```
remote-execution:
  internal: true
  foreman_proxy:
    plugin_name: remote_execution_ssh
```
here?

ehelms:
I started adding this to help ensure the IoP use-case could work out of the box by having ansible support for remediation. There is more nuance and testing for this feature that needs to happen to be complete.

When I test this locally with the updated containers built locally, I run into the hammer sync failing with no clear reason why. That will need investigation, i.e.:

```
hammer ansible roles sync
```

.github/workflows/test.yml
```diff
@@ -229,6 +232,9 @@ jobs:
       - name: Add optional features - azure-rm, google and remote-execution
         run: |
           ./foremanctl deploy --add-feature azure-rm --add-feature google --add-feature remote-execution
+      - name: Add optional feature - foreman-ansible
+        run: |
+          ./foremanctl deploy --add-feature foreman-ansible
```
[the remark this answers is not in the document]
ehelms: Yes - I can do that for efficiency.

tests/foreman_ansible_test.py
```diff
@@ -0,0 +1,83 @@
...
+def test_run_command_via_ansible(foremanapi, server_fqdn):
+    templates = foremanapi.list('job_templates', search='name = "Run Command - Ansible Default"')
+    job = foremanapi.create('job_invocations', {
+        'job_template_id': templates[0]['id'],
+        'inputs': {'command': 'uptime'},
+        'search_query': f'name = {server_fqdn}',
+        'targeting_type': 'static_query',
+    })
+    task = foremanapi.wait_for_task(job['task'])
+    assert task['result'] == 'success'
```
evgeni: these need to go to `tests/client_test.py` (well, it doesn't *have* to, but so far all client-related tests are there) and executed on the client, not the server. the server does not exist in Foreman as an entity and even if it would it would not have the ssh keys set up.

see https://github.com/theforeman/foremanctl/blob/488b150346b6c4435d59587d6f28beb034311fba/tests/client_test.py#L12-L19

tests/feature/ansible/base_test.py
```diff
@@ -0,0 +1,83 @@
...
+def ansible_proxy_id(foremanapi):
+    proxies = foremanapi.list('smart_proxies')
+    for proxy in proxies:
+        if any(f['name'] == 'Ansible' for f in proxy.get('features', [])):
+            return proxy['id']
+    pytest.skip("No smart proxy with Ansible feature found")
+
+
+@pytest.fixture(scope="module")
+def ansible_role(server, foremanapi, ansible_proxy_id):
```
evgeni, line 20: this needs to run in a way that the result ends up *inside* the proxy container, otherwise ansible will never see the roles

tests/foreman_ansible_test.py
pablomh: This is what I had:

```
def test_foreman_ansible_rex(client_environment, activation_key, organization, foremanapi, client, client_fqdn):
    client.run('dnf install -y subscription-manager')
    rcmd = foremanapi.create('registration_commands', {'organization_id': organization['id'], 'insecure': True, 'activation_keys': [activation_key['name']], 'force': True})
    client.run_test(rcmd['registration_command'])
    job = foremanapi.create('job_invocations', {'job_template': 'Run Command - Ansible Default', 'inputs': {'command': 'uptime'}, 'search_query': f'name = {client_fqdn}', 'targeting_type': 'static_query'})
    task = foremanapi.wait_for_task(job['task'])
    assert task['result'] == 'success'
    foremanapi.delete('hosts', {'id': client_fqdn})
```

src/roles/foreman_proxy/templates/settings.d/ansible.yml.j2
```diff
@@ -0,0 +1,5 @@
+---
+:enabled: {{ feature_enabled }}
+:ansible_dir: /usr/share/foreman-proxy
+:working_dir: /tmp
```
adamruzicka: Where is this coming from? The example setting file doesn't have it https://github.com/theforeman/smart_proxy_ansible/blob/master/settings.d/ansible.yml.example

adamruzicka (commented):
@ehelms what are your plans with this?

src/roles/foreman_proxy/templates/settings.d/ansible.yml.j2
ehelms: This comes from the default value in `puppet-foreman_proxy`

adamruzicka: Huh, it is actually a valid setting in sp-ansible, but it doesn't seem to be used anywhere in the codebase.

tests/feature/ansible/base_test.py
```diff
@@ -0,0 +1,118 @@
...
+    setup = server.run(f"podman exec foreman-proxy mkdir -p /etc/ansible/roles/{ROLE_NAME}/tasks")
```
adamruzicka: If we're bind-mounting `/var/lib/foreman-proxy/ansible/roles` to `/etc/ansible/roles` in the container, wouldn't it be easier to just write everything to the host-side of the mount without having to take the round trip through `podman exec`?

src/roles/foreman_proxy/templates/settings.d/ansible.yml.j2
ehelms: :facepalm:

tests/feature/ansible/base_test.py
```diff
@@ -0,0 +1,88 @@
...
+    try:
+        foremanapi.delete('hosts', {'id': client_fqdn})
+    except Exception:
+        pass
+
+
+def test_run_ansible_role(ansible_role, organization, registered_client, foremanapi, server):
+    org_id = organization['id']
+
+    assign = server.run(f"hammer host ansible-roles assign --organization-id {org_id} --name {registered_client} --ansible-roles {ansible_role}")
```
evgeni: this shouldn't be using hammer

tests/feature/ansible/base_test.py
```diff
@@ -0,0 +1,88 @@
...
+    assert sync.succeeded
+
+    for task in foremanapi.list('foreman_tasks', search='label ~ SyncRolesAndVariables and state != stopped'):
+        foremanapi.wait_for_task(task)
+
+    yield ROLE_NAME
+
+
+def test_import_ansible_role(ansible_role, server):
+    role_list = server.run(f"hammer --output csv --no-headers ansible roles list --search='name={ansible_role}'")
```
evgeni: this shouldn't be running hammer

src/features.yaml
```diff
@@ -47,6 +47,15 @@ remote-execution:
...
     - tasks
+ansible:
+  description: Ansible plugin for Foreman
+  foreman:
+    plugin_name: foreman_ansible
+  foreman_proxy:
+    plugin_name: ansible
+  hammer: foreman_ansible
+  dependencies:
+    - tasks
```
adamruzicka:
```suggestion
    - remote_execution
```

Ansible also needs rex. And since rex needs tasks...

tests/feature/ansible/base_test.py
```diff
@@ -0,0 +1,88 @@
...
+            return proxy['id']
+    pytest.skip("No smart proxy with Ansible feature found")
+
+
+@pytest.fixture(scope="module")
+def ansible_role(server, foremanapi, ansible_proxy_id):
+    setup = server.run(f"podman exec foreman-proxy mkdir -p /etc/ansible/roles/{ROLE_NAME}/tasks")
+    assert setup.succeeded
+
+    write = server.run(f"podman exec foreman-proxy bash -c \"echo '- command: uptime' > /etc/ansible/roles/{ROLE_NAME}/tasks/main.yml\"")
```
evgeni: I don't think that's how a user would use it, right?

tests/feature/ansible/base_test.py
evgeni: I've updated this now

evgeni:
@adamruzicka I cleaned up what I disliked, but would now require someone with Ansible knowledge to look at this :)

src/roles/foreman_proxy/tasks/feature/ansible.yaml
```diff
@@ -0,0 +1,48 @@
...
+    mode: '0644'
+    owner: root
+    group: root
+  notify:
+    - Restart Foreman Proxy
+    - Refresh Foreman Proxy
+
+- name: Create ansible roles directory
+  ansible.builtin.file:
+    path: /var/lib/foreman-proxy/ansible/roles
```
adamruzicka, line 26: Previously we relied on `/etc/ansible/roles` for "user roles" (and `s/role/collection/g`). Having everthing in `/var/lib` would be cleaner, but also make the migration harder. Personally I'd be willing to make that tradeoff, but still worth calling out.

src/roles/foreman_proxy/tasks/feature/ansible.yaml
```diff
@@ -0,0 +1,48 @@
...
+    state: directory
+    mode: '0755'
+
+- name: Mount ansible directories
+  ansible.builtin.copy:
+    dest: /etc/containers/systemd/foreman-proxy.container.d/ansible-volumes.conf
+    content: |
+      [Container]
+      Volume=/var/lib/foreman-proxy/ansible/roles:/etc/ansible/roles:rw,Z,U
+      Volume=/var/lib/foreman-proxy/ansible/collections:/etc/ansible/collections:rw,Z,U
```
adamruzicka: Why `rw`? I'd expect the proxy to only consume the content that's there.

Also til about `U`, nice.

src/roles/foreman_proxy/templates/ansible.env.j2
```diff
@@ -0,0 +1,13 @@
+export ANSIBLE_CALLBACK_WHITELIST="theforeman.foreman.foreman"
+export ANSIBLE_CALLBACKS_ENABLED="theforeman.foreman.foreman"
+export ANSIBLE_LOCAL_TEMP="/tmp"
+export ANSIBLE_HOST_KEY_CHECKING="False"
+export ANSIBLE_ROLES_PATH="/etc/ansible/roles:/usr/share/ansible/roles"
+export ANSIBLE_COLLECTIONS_PATHS="/etc/ansible/collections:/usr/share/ansible/collections"
```
adamruzicka: > [DEPRECATION WARNING]:
ANSIBLE_COLLECTIONS_PATHS option, does not fit var naming standard, use the singular form ANSIBLE_COLLECTIONS_PATH instead. This feature will be removed from ansible-core in version 2.19. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```suggestion
export ANSIBLE_COLLECTIONS_PATH="/etc/ansible/collections:/usr/share/ansible/collections"
```

src/roles/foreman_proxy/templates/ansible.env.j2
```diff
@@ -0,0 +1,13 @@
+export ANSIBLE_CALLBACK_WHITELIST="theforeman.foreman.foreman"
```
adamruzicka: Iirc whitelist was replaced with the one below in ansible 2.15. It should be safe to drop this one.
```suggestion
```

adamruzicka (commented):
Seems to work well (ad-hoc rex-like jobs, role management, role application, reports from roles). Some minor suggestions inline

tests/feature/ansible/base_test.py
```diff
@@ -0,0 +1,80 @@
+import pytest
+
+ROLE_NAME = "theforeman.foremanctltest"
+
+
+def test_foreman_ansible_plugin_installed(foremanapi):
```
adamruzicka: https://github.com/theforeman/foremanctl/pull/713#discussion_r3683573940 :)

evgeni: that's a shiny mirror you have there. where did you get it?

evgeni:
@adamruzicka addressed all your code-comments

src/roles/foreman_proxy/tasks/feature/ansible.yaml
evgeni: I do like that it's separated from the Ansible of the host, which might (intentionally or not) contain other content too

adamruzicka (approved):
This is good. There will be a followup or two to close the feature gap, but this is a good start.

PR #446, opened 2026-04-08
Add migration subcommand

### LINKED ISSUE #52: How will RPM-based installations migrate to a container-based installation?

archanaserver:
Fixes #52 

Create foremanctl migration subcommand to convert foreman-installer answer files to foremanctl YAML configuration.

Mappings include:
- Database configuration (host, port, credentials, mode)
- Foreman admin credentials
- Certificate paths

Implements [SAT-43246](https://redhat.atlassian.net/browse/SAT-43246)


ehelms:
Good to reference the repository issue -- https://github.com/theforeman/foremanctl/issues/52

docs/migrate-command.md
```diff
@@ -0,0 +1,128 @@
+# Migrate Command
```
ehelms: I'd consider making this guide about migrating from foreman-installer to foremanctl rather than the command itself.

docs/migrate-command.md
```diff
@@ -0,0 +1,128 @@
...
+# Migrate from backup
+foremanctl migrate --root /backup --output /tmp/config.yaml
+
+# Output to stdout
+foremanctl migrate --answer-file /path/to/answers.yaml
+```
+
+### Command Options
+
+- `--answer-file PATH` - Path to the foreman-installer answer file (default: `/etc/foreman-installer/scenarios.d/satellite.yaml`)
```
ehelms:
```suggestion
- `--answer-file PATH` - Path to the foreman-installer answer file (default: `/etc/foreman-installer/scenarios.d/foreman-answers.yaml`)
```

docs/migrate-command.md
```diff
@@ -0,0 +1,128 @@
...
+- `--output PATH` - Path for the migrated configuration (default: stdout)
+- `--root PATH` - Root directory for finding the answer file, useful for migrations from backups (default: /)
```
ehelms: Is the expectation when using `--root` that there is a directory that has this `/backup/foreman-answers.yaml` ?  If so, would it be better to make the user be explicit here and just have the `--answer-file` parameter?

docs/migrate-command.md
```diff
@@ -0,0 +1,128 @@
...
+## Usage
+
+### Basic Usage
+
+```bash
+# Migrate from default location
+foremanctl migrate --output /tmp/new-config.yaml
+
+# Migrate from custom location
+foremanctl migrate --answer-file /path/to/answers.yaml --output /tmp/config.yaml
```
ehelms: My first thought is what do I as a user do with this file once I generate it?

docs/migrate-command.md
```diff
@@ -0,0 +1,128 @@
...
+- `--root PATH` - Root directory for finding the answer file, useful for migrations from backups (default: /)
+
+## How It Works
+
+1. **Reads** the old YAML answer file
+2. **Parses** the nested parameter structure (e.g., `foreman::db_host`)
+3. **Maps** old parameter names to new names using a mapping table
+4. **Transforms** values where needed (e.g., boolean to string)
+5. **Writes** the new configuration file
+6. **Reports** any unmappable parameters as warnings (does not fail)
```
ehelms: The bolding in this list feels unnecessary

docs/migrate-command.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Parameter Mapping
+
+The migration uses a hardcoded mapping table in `src/playbooks/migrate/library/migrate_answers.py`:
+
+```python
+PARAMETER_MAP = {
```
ehelms: Is it possible for this to be a separate file that is loaded?

docs/migrate-command.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+### Adding New Mappings
+
+To add new parameter mappings:
+
+1. Open `src/playbooks/migrate/library/migrate_answers.py`
+2. Add entries to the `PARAMETER_MAP` dictionary
+3. Use format: `(old_module, old_param): new_param`
+4. For transformations: `(old_module, old_param): (new_param, transform_function)`
```
ehelms: Ahh, I guess this might be why it is storing it all in python?

src/playbooks/migrate/metadata.obsah.yaml
```diff
@@ -0,0 +1,15 @@
+---
+help: |
+  Migrate foreman-installer answer file to foremanctl configuration format.
+
+  This command reads the old Satellite 6.18 answer file and converts it to the
```
ehelms: This is overly Satellite specific.

src/playbooks/migrate/migrate.yaml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Migrate foreman-installer answer file to foremanctl format
+  hosts:
+    - localhost
+  connection: local
+  gather_facts: false
+  vars:
+    root_dir: "{{ root | default('/') }}"
+    default_answer_file: "{{ root_dir }}/etc/foreman-installer/scenarios.d/satellite.yaml"
```
ehelms: This is not the answer file, this is the configuration file.
This should not default to satellite.yaml.
I would expect this to default to `last_scenario.yaml` as it's a symlink to the current scenario that is enabled.

docs/migrate-command.md
archanaserver: Hmm I see the `--root` parameter adds little complexity but my intent was to simplify backup/restore scenarios where you have a full system backup mounted (e.g., at /backup) - using `--root /backup` is shorter than typing the full path. However, I agree the explicit `--answer-file` approach is clearer. Does foreman-installer have a similar` --root` parameter that users might expect? If not, I'm happy to remove it and keep things simple - just using `--answer-file` makes the code easier to understand. What's your preference?

  ex:                                                                                                                                                                                                                                                                      

  - With ` --root: foremanctl migrate --root /backup --output config.yaml     `                                                                                                                                                                                                    
  - Without `--root: foremanctl migrate --answer-file /backup/etc/foreman-installer/scenarios.d/last_scenario.yaml --output config.yaml `

docs/migrate-command.md
archanaserver: Good catch! Updated doc now makes this clear upfront.

docs/migrate-command.md
archanaserver: Yeah some parameters need special conversion logic, so we needed Python code anyway, I just put the whole mapping table in the same file to keep it simple and easier. Do you prefer other way doing it?

src/playbooks/migrate/migrate.yaml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Migrate foreman-installer answer file to foremanctl format
+  hosts:
+    - localhost
+  connection: local
+  gather_facts: false
+  vars:
+    root_dir: "{{ root | default('/') }}"
+    default_answer_file: "{{ root_dir }}/etc/foreman-installer/scenarios.d/last_scenario.yaml"
```
evgeni: The `last_scenario` symlink does not point to an answer file, it points to `<scenario>.yaml`.

But you can find the answer file path in that file under the `:answer_file:` key

src/playbooks/migrate/library/migrate_answers.py
```diff
@@ -0,0 +1,193 @@
...
+
+def write_output(data, output_path=None):
+    """Write migrated configuration to file or stdout."""
+    yaml_content = yaml.dump(data, default_flow_style=False, sort_keys=True)
+
+    if output_path:
+        with open(output_path, 'w') as f:
+            f.write(yaml_content)
+    else:
+        print(yaml_content)
```
evgeni: this `print` is not seen when ran from Ansible:

```
# foremanctl migrate --answer-file /etc/foreman-installer/scenarios.d/katello-answers.yaml

PLAY [Migrate foreman-installer answer file to foremanctl format] 

TASK [Run migration] 
ok: [localhost]

TASK [Display migration results] 
ok: [localhost] => 
    msg: |-
        Migration completed successfully!
…
        Output written to stdout

PLAY RECAP 
localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

docs/migrate-command.md
evgeni: I think using `--root` here and pointing it at a backup gives a false idea of the migration concept/flow. A backup (as it's taken today) does not contain the answer file directly -- you'd need to unpack `config_files.tar.gz` and fetch the relevant yamls from there. And while at it, also grab the certs, as you'll need them for the migration too.

Ideally `foremanctl` will be able to do all that: point it at a valid backup and it will restore it including the migration from packages/puppet to containers/ansible. But I'd not put it here in this PR and focus solely on providing an entrypoint to start the "simple" "in-place" migration.

src/plugins/modules/migrate_answers.py
```diff
@@ -0,0 +1,193 @@
...
+    )
+
+    try:
+        old_config = load_answer_file(module.params['answer_file'])
+
+        migration_result = apply_mappings(old_config)
+
+        result['mapped_count'] = len(migration_result['mapped'])
+        result['unmappable_count'] = len(migration_result['unmappable'])
+        result['unmappable'] = migration_result['unmappable']
```
evgeni, line 209: The linked Jira says we should raise an error when there are unmappable parameters, which contradicts the help text ("Unmappable parameters are reported as warnings but do not cause the command to fail."), but I also don't see any warnings issued here.

I do think that issuing warnings is the correct path here, but that should be reflected everywhere.

src/playbooks/migrate/library/migrate_answers.py
```diff
@@ -0,0 +1,193 @@
...
+IGNORE_PARAMS = {'IGNORE'}
+
+
+def load_answer_file(file_path):
+    """Load and parse YAML answer file."""
+    if not os.path.exists(file_path):
+        raise FileNotFoundError(f"Answer file not found: {file_path}")
+
+    if not os.access(file_path, os.R_OK):
+        raise PermissionError(f"Cannot read answer file: {file_path}")
```
evgeni: `open` will already raise `PermissionError` and `FileNotFoundError` in those cases, so I don't think we need to special-handle them

src/playbooks/migrate/library/migrate_answers.py
```diff
@@ -0,0 +1,193 @@
...
+                result['changed'] = True  # File was written
+
+        module.exit_json(**result)
+
+    except FileNotFoundError as e:
+        module.fail_json(msg=str(e), **result)
+    except PermissionError as e:
+        module.fail_json(msg=str(e), **result)
+    except ValueError as e:
+        module.fail_json(msg=str(e), **result)
```
evgeni:
```suggestion
    except (FileNotFoundError, PermissionError, ValueError) as e:
        module.fail_json(msg=str(e), **result)
```

docs/migration-guide.md
```diff
@@ -0,0 +1,115 @@
...
+
+When the migration completes, you may see warnings like:
+
+```
+Warning: The following parameters could not be mapped:
+  - katello::enable_ostree
+  - foreman::some_other_param
+```
+
+These parameters need to be manually reviewed and added to the new configuration if needed. Check the [parameters documentation](parameters.md) for equivalent foremanctl parameters.
```
evgeni, line 116: Translating this in my head: once satisfied with the result, the user would put that as `/var/lib/foremanctl/parameters.yaml` and run `foremanctl deploy`?

archanaserver: yeah, added a section to the docs that explains this workflow now

docs/migration-guide.md
```diff
@@ -0,0 +1,129 @@
+# Migrating from foreman-installer to foremanctl
+
+## Overview
+
+When upgrading from foreman-installer to foremanctl, the `foremanctl migrate` command helps convert your existing configuration to the new format.
```
Gauravtalreja1, line 5: Do you think it would be good to have steps here to install foremanctl on foreman-installer based setups?

src/playbooks/migrate/metadata.obsah.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+  This command reads foreman-installer answer files and converts them to the
+  new foremanctl configuration format. Unmappable parameters are reported
+  as warnings but do not cause the command to fail.
+
+variables:
+  answer_file:
+    help: Path to the foreman-installer answer file to migrate. If not specified, attempts to read from the default location.
+  output:
+    help: Path where the migrated configuration should be written. If not specified, outputs to stdout.
```
Gauravtalreja1, line 14: If not specified, it outputs to stdout currently, so just a thought should we also put those into default config /var/lib/foremanctl/parameters.yaml as well?

src/playbooks/migrate/migrate.yaml
```diff
@@ -0,0 +1,61 @@
...
+            resolved_answer_file: "{{ scenario_data[':answer_file'] | default('') }}"
+
+        - name: Fail if no answer file found in scenario
+          ansible.builtin.fail:
+            msg: "Scenario file does not contain :answer_file: key"
+          when: resolved_answer_file == ''
+
+    - name: Run migration
+      migrate_answers:
+        answer_file: "{{ answer_file | default(resolved_answer_file) }}"
```
evgeni: I would have done `answer_file | default(omit)` and do the "load last_scenario and find answer file there" logic in the `migrate_answers` module instead.

not blocking, just an opinion :)

src/playbooks/migrate/migrate.yaml
```diff
@@ -0,0 +1,61 @@
...
+    default_scenario_file: "/etc/foreman-installer/scenarios.d/last_scenario.yaml"
+  tasks:
+    - name: Determine answer file to migrate
+      when: answer_file is not defined
+      block:
+        - name: Read scenario file
+          ansible.builtin.slurp:
+            src: "{{ default_scenario_file }}"
+          register: scenario_content
+
```
Gauravtalreja1: Currently, if I touch any random file and run the migration by passing that file with `--answer-file`, it passes with `Migration completed successfully!`, so I suggest adding a YAML validation check before reading it, to check if its valid yaml and valid answers file

src/playbooks/migrate/metadata.obsah.yaml
```diff
@@ -0,0 +1,13 @@
+---
+help: |
+  Migrate foreman-installer answer file to foremanctl configuration format.
+
+  This command reads foreman-installer answer files and converts them to the
+  new foremanctl configuration format. Unmappable parameters are reported
+  as warnings but do not cause the command to fail.
+
+variables:
+  answer_file:
```
Gauravtalreja1, line 10: when using `--answer-file` once, it gets stored in `/var/lib/foremanctl/parameters.yaml`, so it is run always with same answers file unless we override it again, so just wanted to check if this is expected design?

src/playbooks/migrate/migrate.yaml
```diff
@@ -0,0 +1,61 @@
...
+        - name: Parse scenario file
+          ansible.builtin.set_fact:
+            scenario_data: "{{ scenario_content['content'] | b64decode | from_yaml }}"
+
+        - name: Extract answer file path from scenario
+          ansible.builtin.set_fact:
+            resolved_answer_file: "{{ scenario_data[':answer_file'] | default('') }}"
+
```
Gauravtalreja1: Can we combine this both set_fact tasks, as scenario_data seems redundant, can you try something like 
`(scenario_content.content | b64decode | from_yaml).get(':answer_file', '')`?

not blocking, just an opinion to optimize?

src/playbooks/migrate/migrate.yaml
```diff
@@ -0,0 +1,61 @@
...
+        - name: Fail if no answer file found in scenario
+          ansible.builtin.fail:
+            msg: "Scenario file does not contain :answer_file: key"
+          when: resolved_answer_file == ''
+
+    - name: Run migration
+      migrate_answers:
+        answer_file: "{{ answer_file | default(resolved_answer_file) }}"
+        output: "{{ output | default(omit) }}"
```
Gauravtalreja1, line 10: if I pass `--output diff.yaml` , where file doesn't exist previously, and the file gets created under `/usr/share/foremanctl/src/playbooks/migrate/diff.yaml` instead of present working directory, is this expected?

docs/migration-guide.md
```diff
@@ -0,0 +1,129 @@
+# Migrating from foreman-installer to foremanctl
```
ehelms, line 1: Given the title and focus of the doc, I think the filename could use an update to reflect this.

docs/migration-guide.md
```diff
@@ -0,0 +1,129 @@
...
+```bash
+foremanctl migrate
+```
+
+## Command Options
+
+- `--answer-file PATH` - Path to the foreman-installer answer file. If not specified, reads the currently active scenario and extracts the answer file path from it.
+- `--output PATH` - Path for the migrated configuration (default: stdout)
+
+## Parameter Mappings
```
ehelms, line 72: We have an overlap now with `parameters.md`. I think either combine these two, or link out and capture this information in there.

src/playbooks/migrate/migrate.yaml
archanaserver: yeah, sounds clean moving all the logic in the module. on it

src/playbooks/migrate/migrate.yaml
archanaserver: no it was not expected, i see a bug now, let me fix that and address other comments as well

src/playbooks/migrate/metadata.obsah.yaml
archanaserver: So if no `--answer-file` is specified, it first checks the scenario file (our defaultt one) and it only takes effect if you explicitly providing `--answer-file` in a previous run but users can always override it by specifying a new` --answer-file` that was my thinking behind it  and this persistance happens because of https://github.com/theforeman/foremanctl/blob/master/foremanctl#L8 in the foremanctl wrapper script.

docs/migration-guide.md
archanaserver: i've linked it, looks better?

src/playbooks/migrate/migrate.yaml
archanaserver: already addressed! I moved all scenario file reading logic to the module

docs/migration-guide.md
archanaserver: not sure about it, but my thinking is installation steps are probably better suited for the main README or deployment docs rather than the migration guide docs or can add a simple "prerequisites" section with a link to installation docs. WDYT?

@ehelms @evgeni thoughts?

src/playbooks/migrate/metadata.obsah.yaml
archanaserver: I kept stdout as the default for the safer side, so that users can review output before it overwrites existing config. However, I believe we could document the recommended command: `foremanctl migrate --output /var/lib/foremanctl/parameters.yaml`. What's your preference?

docs/migration-guide.md
ehelms: I think since this guide is about how to upgrade from foreman-installer to foremanctl, that it should cover those steps for now. Like, prerequisites (e.g. foreman-installer installed, foremanctl present on the system) then run the steps. Or point to docs and say what steps should be run.

src/playbooks/migrate/metadata.obsah.yaml
evgeni: Given they will be only used for a limited amount of time, and the user might be "experimenting" (= putting different data into them) while doing so, I'd lean to not persisting answers-file and output.

Since https://github.com/theforeman/obsah/pull/111/ you can mark them as not persisting by adding `persist: false` to the metadata of each variable.

src/playbooks/migrate/library/migrate_answers.py
```diff
@@ -0,0 +1,240 @@
...
+                continue
+
+            if isinstance(mapping, tuple):
+                new_key, transform_func = mapping
+                new_value = transform_func(old_value)
+            else:
+                new_key = mapping
+                new_value = old_value
+
+            result[new_key] = new_value
```
evgeni:
```suggestion
            if new_value is not None:
                result[new_key] = new_value
```

If the `new_value` is `None`, we should avoid returning it, as it basically means "use the default", but setting something to `null` (which is what `None` will be translated to in YAML) will override the defaults we have in our vars files.

src/plugins/modules/migrate_answers.py
```diff
@@ -0,0 +1,240 @@
...
+
+    Example:
+        {'foreman': {'db_host': 'localhost'}}
+    becomes:
+        {('foreman', 'db_host'): 'localhost'}
+    """
+    items = []
+    for key, value in nested_dict.items():
+        if isinstance(value, dict):
+            items.extend(flatten_nested_dict(value, key).items())
```
evgeni, line 107: This removes the "root" `key` from the result, which makes it harder to detect enabled plugins (you can detect them by being "non False", so an empty dict is an enabled plugin.

```suggestion
        if isinstance(value, dict):
            items.append(key, True)
            items.extend(flatten_nested_dict(value, key).items())
```

This would add `key` with a `True` value to the list and we could then later process it.

At the same time, this is nothing that we need in the initial "here is the command", so feel free to say we add it later.

archanaserver: Hmm I was checking that earlier, it's small change but gonna differ here, and needs testing and more. but let's add this enhancement in follow-up to keep this PR focused on the migration functionality for now.

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+
+When upgrading from foreman-installer to foremanctl, the `foremanctl migrate` command helps convert your existing configuration to the new format.
+
+This guide explains how to migrate your foreman-installer answer files to foremanctl configuration files.
+
+## Prerequisites
+
+Before migrating, ensure the following:
+
+1. **foreman-installer is installed** - You should have an existing foreman-installer setup with answer files to migrate from.
```
evgeni: I'd phrase that a bit differently, as the main point is not that foreman-installler is installed, but you have a Foreman installed by foreman-installer, so how about:

```suggestion
1. **Foreman deployment using foreman-installer** - You should have an existing Foreman deployment has been installed using foreman-installer and has an answers file to migrate from.
```

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+  initial_admin_username: admin
+  initial_admin_password: changeme
+```
+
+### Output (foremanctl format)
+
+```yaml
+database_host: database.example.com
+database_port: 5432
+database_mode: internal
```
evgeni:
```suggestion
database_mode: external
```

not that it technically matters, but your example uses a db host, so this is external db :)

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Handling Unmapped Parameters
+
+When the migration completes, you may see warnings like:
+
+```
+Warning: The following parameters could not be mapped:
+  - katello::enable_ostree
+  - foreman::some_other_param
+```
```
evgeni: the warning format changed in one of the recent pushes, so this probably deserves an update to match code

evgeni:
There is a failing test, and I posted a few nitpicks, but I think this reflects what we wanted to expose as the first version of the migrate command (read: ACK once my comments are addressed ;) )

src/plugins/modules/migrate_answers.py
evgeni: I've added this to my list of things that we need to implement for the final version.

archanaserver: You mean in that prepare work epic related to this one, right?

src/playbooks/migrate/metadata.obsah.yaml
ehelms: That logic makes sense to me.

src/plugins/modules/migrate_answers.py
evgeni: Yes

src/playbooks/migrate/library/migrate_answers.py
```diff
@@ -0,0 +1,241 @@
...
+                new_value = old_value
+
+            if new_value is not None:
+                result[new_key] = new_value
+        else:
+            if isinstance(old_key, tuple):
+                param_name = '::'.join(old_key)
+            else:
+                param_name = str(old_key)
+            unmappable.append(param_name)
```
evgeni:
```suggestion
            if old_value is not None:
                unmappable.append(param_name)
```

as with the mapped version, if the value is `None`, we can safely drop the entry.

src/plugins/modules/migrate_answers.py
```diff
@@ -0,0 +1,241 @@
...
+    ('foreman', 'db_manage_rake'): 'IGNORE',  # Not needed in new format
+
+    # Foreman configuration
+    ('foreman', 'initial_admin_username'): 'foreman_initial_admin_username',
+    ('foreman', 'initial_admin_password'): 'foreman_initial_admin_password',
+
+    # Certificate configuration
+    ('foreman', 'server_ssl_cert'): 'server_certificate',
+    ('foreman', 'server_ssl_key'): 'server_key',
+    ('foreman', 'server_ssl_ca'): 'ca_certificate',
```
evgeni, line 33: I think this mapping is wrong, we should rather map the data from `certs` (so `('certs', 'server_cert')`, `(…, 'server_key')` and `(…, 'server_ca_cert')` to the custom cert variables introduced in #421, but we don't need to wait on that and can update the mapping later, the current setup does not break anything

src/plugins/modules/migrate_answers.py
```diff
@@ -0,0 +1,241 @@
+#!/usr/bin/python3
```
ehelms, line 1: As this file is a module, I am thinking it should live in:

```
src/plugins/modules/migrate_answers.py
```

src/playbooks/migrate/migrate.yaml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Migrate foreman-installer answer file to foremanctl format
+  hosts:
+    - localhost
```
ehelms: I notice we are locking this command into running locally only. As in, I cannot point this at an installed host and say "go migrate the answer file over there". That would be nice to have for testing.

archanaserver: So i've tried dropping `connection: local` limit with added `migrate_host` inventory group, i've also tried few different approaches but no help. But though this approach is solving the remote execution issue, we can now create a custom inventory file and run migration on remote host,, I'm not sure this approach is good enough here, I can't think of else but WDYT?

src/plugins/modules/migrate_answers.py
archanaserver: ACK, I'll update the certificate mappings in a follow-up PR.

src/playbooks/migrate/migrate.yaml
evgeni: I think to make ehelms' request work, just using `quadlet` here would have been sufficient? Then everyone can adjust the meaning of `quadlet` like they desire? (But I could also miss-reading it, so…)

archanaserver: well yes, that would be sufficient, thanks, updated :)

docs/migration-guide.md
```diff
@@ -0,0 +1,138 @@
...
+foreman_database_user: foreman_user
+foreman_initial_admin_password: changeme
+foreman_initial_admin_username: admin
+```
+
+## Handling Unmapped Parameters
+
+When the migration completes, you may see warnings like:
+
+> [!WARNING]
```
ehelms, line 111: This is going to be tricky in practice. Users will need guidance on parameters that could not be mapped. Perhaps for now we can include as part of the warning output the high-level possible reasons they could not be mapped? 

For example:

1. There was an error trying to map it
2. The value has no equivalent parameter / is unsupported
3. other reasons?

evgeni: I think that's something that we'll have to learn (and document) as we go.

evgeni (approved):

docs/migration-guide.md
archanaserver: Yeah, that is something we can focus on follow up work as well. Thanks for pointing it out.

archanaserver:
All green now, can we get this in, or is there anything else this needs to be covered? @evgeni @ehelms

ehelms (approved):

PR #450, opened 2026-04-09
Add initial organization and initial location parameters

arvind4501:
Allow setting the initial organization and location during deployment

### DIFF
--- a/src/playbooks/deploy/metadata.obsah.yaml
+++ b/src/playbooks/deploy/metadata.obsah.yaml
@@ -7,6 +7,12 @@ variables:
     help: Initial username for the admin user.
   foreman_initial_admin_password:
     help: Initial password for the admin user.
+  foreman_initial_organization:
+    help: Name of an initial organization.
+    parameter: --initial-organization
+  foreman_initial_location:
+    help: Name of an initial location.
+    parameter: --initial-location
   foreman_puma_workers:
     help: Number of workers for Puma.
   pulp_worker_count:
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -116,6 +116,8 @@
       - 'foreman-client-key,type=mount,target=/etc/foreman/client_key.pem'
       - 'foreman-db-ca,type=mount,target={{ foreman_database_ssl_ca_path }}'
     env:
+      SEED_ORGANIZATION: "{{ foreman_initial_organization }}"
+      SEED_LOCATION: "{{ foreman_initial_location }}"
       FOREMAN_PUMA_THREADS_MIN: "{{ foreman_puma_threads_min }}"
       FOREMAN_PUMA_THREADS_MAX: "{{ foreman_puma_threads_max }}"
       FOREMAN_PUMA_WORKERS: "{{ foreman_puma_workers }}"
@@ -203,6 +205,9 @@
       - 'foreman-ca-cert,type=mount,target=/etc/foreman/katello-default-ca.crt'
       - 'foreman-client-cert,type=mount,target=/etc/foreman/client_cert.pem'
       - 'foreman-client-key,type=mount,target=/etc/foreman/client_key.pem'
+    env:
+      SEED_ORGANIZATION: "{{ foreman_initial_organization }}"
+      SEED_LOCATION: "{{ foreman_initial_location }}"
     quadlet_options:
       - |
         [Unit]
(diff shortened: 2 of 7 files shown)


src/roles/foreman/tasks/main.yaml
```diff
@@ -33,6 +33,24 @@
     - Restart foreman
     - Restart dynflow-sidekiq@
 
+- name: Create secret for SEED_ORGANIZATION
```
ehelms: These one time creation secrets provide a lot of "waste". They are created, persisted, and mounted into the container, used the first time you ever do something and then hang around providing no value. 

We can stick with this paradigm for now if that's easiest.

I think we should examine other methods for handling these, e.g.:

 1. Running an action in the container that sets these as environment variables, and runs `db:seed`
 2. Stop using `db:seed` and rely on the API to create and maintain these idempotently through our Ansible modules

arvind4501: Yes agree, we don't use most of these secret after first deployment

Here are my thoughts on alternative approches:-

> Running an action in the container that sets these as environment variables, and runs db:seed

You mean mount ENVs directly to container instead creating podman secret? yes i think we can do that for organization/location but i think we should not remove that for config files as change in config should trigger handlers.

> Stop using db:seed and rely on the API to create and maintain these idempotently through our Ansible modules

do you mean removing seed entirely? i seed does far more than just user/org/location  data so not sure if we want to remove that.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -7,6 +7,10 @@ variables:
     help: Initial username for the admin user.
   foreman_initial_admin_password:
     help: Initial password for the admin user.
+  initial_organization:
```
evgeni: I think I'd call the variables `foreman_*`, as that where they logically belong to, but override the parameter to be `--initial-*` only.

evgeni (commented):
Can we set the org/loc for CI (in `.github/workflows`, and probably manually in the testing doc) to something like "Foreman CI" and "Internet" and add a pytest that asserts that these exist via the API?

src/roles/foreman/tasks/main.yaml
evgeni: >     2. Stop using `db:seed` and rely on the API to create and maintain these idempotently through our Ansible modules

I've been called an "advanced user" for proposing this solution yesterday ;-)

The resulting discussion resulted in "we probably want to have a simple wizard on first launch, but that's a lot of change for the current effort, so let's go with seeds"

ehelms: I'll create an issue to track this since I am still not a fan of having one-and-done environment variables on the runtime container persistently.

ehelms: https://github.com/theforeman/foremanctl/issues/452

arvind4501:
> Can we set the org/loc for CI (in `.github/workflows`, and probably manually in the testing doc) to something like "Foreman CI" and "Internet" and add a pytest that asserts that these exist via the API?

Sorry, i don't follow, did you mean passing initial-organization/location in CI and having a pytest which validates that with foremanapi?

evgeni:
> > Can we set the org/loc for CI (in `.github/workflows`, and probably manually in the testing doc) to something like "Foreman CI" and "Internet" and add a pytest that asserts that these exist via the API?
> 
> Sorry, i don't follow, did you mean passing initial-organization/location in CI and having a pytest which validates that with foremanapi?

Yes, exactly that's what I meant

arvind4501:
With these change now, `./forge test` fails if we don't pass the hardcoded values of oorg and location in first deployment https://github.com/theforeman/foremanctl/blob/master/DEVELOPMENT.md#deployment

evgeni:
> With these change now, `./forge test` fails if we don't pass the hardcoded values of oorg and location in first deployment https://github.com/theforeman/foremanctl/blob/master/DEVELOPMENT.md#deployment

The same is true for the `changeme` password and tuning profile, so probably just extend the example `deploy` call in the linked doc?

evgeni (approved):

nacoool:
QE Ack : 
Tested on build : `foremanctl-2.0.0-1.20260417070511952327.pr450.6.g0d9aa1e.el9.noarch`
Result : Success
Scenarios 

1.  Deploy fresh formenactl with `foremanctl deploy --initial-organization first --initial-location first`
2. Deploy fresh foremanctl wtih `foremanctl deploy --foreman-initial-admin-username "admin" --foreman-initial-admin-password "changeme" --add-feature hammer --add-feature foreman-proxy --initial-organization "second" --initial-location "second"`
3. Deploy foremanctl on already installed foremanctl `foremanctl deploy --foreman-initial-admin-username "admin" --foreman-initial-admin-password "changeme" --add-feature hammer --add-feature foreman-proxy --initial-organization "second" --initial-location "second"`
4. Deploy fresh formenactl without `--initial-organization <value> --initial-location <value>`

Observation : 

1. foremanctl deployed with org and loc with the value `first` successfully.
2. foremanctl deployed with org and loc with the value `second` successfully.
3. foremanctl did not deployed with org and loc with the value `second` correctly. (as expected)
4. foremanctl deployed with org and loc with `default` values.

PR #455, opened 2026-04-17
Pulp import export support

aidenfine:
[Jira](https://redhat.atlassian.net/browse/SAT-44059?atlOrigin=eyJpIjoiNTJjYjFmMDhlYzUzNGI4ZDg3NWEyZTllMDk4MDhiMGQiLCJwIjoiaiJ9)

Add pulp content import/export arguments.

You can pass in multiple args of either `--content-import-path` or `--content-export-path` as many times as you want, since these will be appended as a list. These folders will be automatically created.

`pytest tests/pulp_test.py`

Can also manually test these steps.
```bash
# create env
./foremanctl deploy --foreman-initial-admin-password=changeme --tuning development --content-import-path /mycontentpath/import --content-export-path /mycontentpath/export --content-export-path /mycontentpath/export2

# ssh 
vagrant ssh quadlet

# verify imports (should see a list with exported/imported paths)
sudo podman exec pulp-api pulpcore-manager shell -c "from django.conf import settings; print(settings.ALLOWED_IMPORT_PATHS)"

# exports
sudo podman exec pulp-api pulpcore-manager shell -c "from django.conf import settings; print(settings.ALLOWED_EXPORT_PATHS)"
```

### DIFF
--- a/development/playbooks/deploy-dev/metadata.obsah.yaml
+++ b/development/playbooks/deploy-dev/metadata.obsah.yaml
@@ -32,3 +32,4 @@ variables:
 
 include:
   - _flavor_features
+  - _pulp
--- a/src/playbooks/deploy/metadata.obsah.yaml
+++ b/src/playbooks/deploy/metadata.obsah.yaml
@@ -53,6 +53,7 @@ constraints:
   forbidden_if:
     - [certificates_source, installer, [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]]
 
+
 include:
   - _certificate_source
   - _certificate_validity
@@ -60,3 +61,4 @@ include:
   - _database_connection
   - _tuning
   - _flavor_features
+  - _pulp
--- a/src/roles/pulp/README.md
+++ b/src/roles/pulp/README.md
@@ -1,0 +1,32 @@
+Pulp Role
+=========
+
+Deploys and manages [Pulp](https://pulpproject.org/) as Podman quadlets.
+
+Variables
+---------
+- `pulp_container_image`: Container image to use for Pulp (default: `quay.io/foreman/pulp`)
+- `pulp_container_tag`: Container image tag (default: `"3.73"`)
+- `pulp_registry_auth_file`: Path to the registry authentication file (default: `/etc/foreman/registry-auth.json`)
+- `pulp_worker_count`: Number of Pulp workers (default: min of 8 or CPU count)
+- `pulp_content_origin`: URL for the Pulp content service (default: `http://{{ fqdn }}:24816`)
+- `pulp_pulp_url`: URL for the Pulp API service (default: `http://{{ fqdn }}:24817`)
+- `pulp_volumes`: Volume mounts for Pulp containers (default: `/var/lib/pulp:/var/lib/pulp`)
+- `pulp_enable_analytics`: Enable Pulp analytics (default: `false`)
+- `pulp_import_paths`: Paths Pulp can use for content imports (default: `[/var/lib/pulp/sync_imports, /var/lib/pulp/imports]`)
+- `pulp_export_paths`: Paths Pulp can use for content exports (default: `[/var/lib/pulp/exports]`)
+- `pulp_plugins`: Additional Pulp plugins to enable (default: `[pulp_container, pulp_rpm]`)
+- `pulp_database_name`: Name of the Pulp database (default: `pulp`)
+- `pulp_database_user`: Database user (default: `pulp`)
+- `pulp_database_host`: Database host (default: `localhost`)
+- `pulp_database_port`: Database port (default: `5432`)
+- `pulp_database_password`: Database password (required, no default)
+- `pulp_database_ssl_mode`: Database SSL mode (default: `disabled`)
+- `pulp_database_ssl_ca`: Path to the database SSL CA certificate on the control node (default: empty)
+
+Usage Inside foremanctl
+-----------------------
+When used as part of `foremanctl`, the variables are setup as the following
+- `pulp_worker_count`: `--pulp-worker-count`
+- `pulp_import_paths`: `--content-import-path` (may be specified multiple times)
+- `pulp_export_paths`: `--content-export-path` (may be specified multiple times)
--- a/src/roles/pulp/defaults/main.yaml
+++ b/src/roles/pulp/defaults/main.yaml
@@ -10,8 +10,12 @@ pulp_worker_count: "{{ [8, ansible_facts['processor_nproc']] | min }}"
 pulp_content_service_worker_count: "{{ (2 * ([8, ansible_facts['processor_nproc']] | min)) + 1 }}"
 pulp_api_service_worker_count: "{{ ([4, ansible_facts['processor_nproc']] | min) + 1 }}"
 
-pulp_volumes:
-  - /var/lib/pulp:/var/lib/pulp
+pulp_volumes: >-
+  {{
+    ['/var/lib/pulp:/var/lib/pulp'] +
+    (pulp_import_paths | map('regex_replace', '^(.+)$', '\1:\1') | list) +
+    (pulp_export_paths | map('regex_replace', '^(.+)$', '\1:\1') | list)
+  }}
 
 pulp_api_container_name: pulp-api
 pulp_content_container_name: pulp-content
@@ -21,6 +25,11 @@ pulp_content_origin: "http://{{ ansible_facts['fqdn'] }}:24816"
 pulp_pulp_url: "http://{{ ansible_facts['fqdn'] }}:24817"
 
 pulp_enable_analytics: false
+
+pulp_default_import_paths: ["/var/lib/pulp/sync_imports", "/var/lib/pulp/imports"]
+pulp_default_export_paths: ["/var/lib/pulp/exports"]
+pulp_import_paths: []
+pulp_export_paths: []
 
 pulp_default_plugins:
   - pulp_certguard
@@ -66,6 +75,10 @@ pulp_settings_other_env:
   PULP_CONTENT_WORKERS: "{{ pulp_content_service_worker_count }}"
   PULP_TOKEN_AUTH_DISABLED: "true"
   PULP_FLATPAK_INDEX: "true"
+  PULP_ALLOWED_IMPORT_PATHS: >-
+    {{ pulp_default_import_paths + pulp_import_paths }}
+  PULP_ALLOWED_EXPORT_PATHS: >-
+    {{ pulp_default_export_paths + pulp_export_paths }}
   PULP_ALLOWED_CONTENT_CHECKSUMS: >-
     ['sha1', 'sha224', 'sha256', 'sha384', 'sha512']
 
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -37,6 +37,20 @@
     - assets
     - media
 
+- name: Create import directories
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: directory
+    mode: "0755"
+  loop: "{{ pulp_default_import_paths + pulp_import_paths }}"
+
+- name: Create export directories
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: directory
+    mode: "0755"
+  loop: "{{ pulp_default_export_paths + pulp_export_paths }}"
+
 - name: Create DB password secret
   containers.podman.podman_secret:
     state: present
--- a/tests/pulp_test.py
+++ b/tests/pulp_test.py
@@ -1,10 +1,26 @@
 import json
+import os
 
 import pytest
+import yaml
 
 PULP_HOST = 'localhost'
 PULP_API_PORT = 24817
 PULP_CONTENT_PORT = 24816
+
+
+def load_pulp_paths_from_parameters():
+    test_dir = os.path.dirname(os.path.abspath(__file__))
+    foremanctl_dir = os.path.dirname(test_dir)
+    params_file = os.path.join(foremanctl_dir, '.var', 'lib', 'foremanctl', 'parameters.yaml')
+
+    if os.path.exists(params_file):
+        with open(params_file, 'r') as f:
+            params = yaml.safe_load(f)
+            import_paths = params.get('pulp_import_paths', [])
+            export_paths = params.get('pulp_export_paths', [])
+
+    return import_paths, export_paths
 
 
 @pytest.fixture(scope="module")
@@ -15,6 +31,11 @@ def pulp_status_curl(server):
 @pytest.fixture(scope="module")
 def pulp_status(pulp_status_curl):
     return json.loads(pulp_status_curl.stdout)
+
+
+@pytest.fixture(scope="module")
+def pulp_import_export_paths():
+    return load_pulp_paths_from_parameters()
 
 
 def test_pulp_api_service(server):
@@ -85,3 +106,40 @@ def test_pulp_worker_target(server):
 def test_pulp_manager_check(server):
     result = server.run("podman exec -ti pulp-api pulpcore-manager check --deploy")
     assert result.succeeded
+
+
+def test_pulp_import_export_settings(server, pulp_import_export_paths):
+    expected_import_paths, expected_export_paths = pulp_import_export_paths
+    py = 'from django.conf import settings; import json; print(json.dumps({"import": list(settings.ALLOWED_IMPORT_PATHS), "export": list(settings.ALLOWED_EXPORT_PATHS)}))'
+    result = server.run(f"podman exec pulp-api pulpcore-manager shell -c '{py}'")
+    assert result.succeeded
+    data = json.loads(result.stdout)
+    for path in expected_import_paths:
+        assert path in data['import'], f"expected {path} in Pulp ALLOWED_IMPORT_PATHS"
+    for path in expected_export_paths:
+        assert path in data['export'], f"expected {path} in Pulp ALLOWED_EXPORT_PATHS"
+
+
+def test_pulp_import_directories(server, pulp_import_export_paths):
+    import_paths, _ = pulp_import_export_paths
+    for path in import_paths:
+        assert server.file(path).is_directory
+
+
+def test_pulp_export_directories(server, pulp_import_export_paths):
+    _, export_paths = pulp_import_export_paths
+    for path in export_paths:
+        assert server.file(path).is_directory
+
+
+@pytest.mark.parametrize("container", ["pulp-api", "pulp-content", "pulp-worker-1"])
+def test_pulp_import_export_volume_mounts(server, container, pulp_import_export_paths):
+    import_paths, export_paths = pulp_import_export_paths
+    result = server.run(f"podman inspect {container} --format '{{{{json .Mounts}}}}'")
+    assert result.succeeded
+    mounts = json.loads(result.stdout)
+    destinations = [mount['Destination'] for mount in mounts]
+
+    for path in import_paths + export_paths:
+        mounted = path in destinations or any(path.startswith(d + '/') for d in destinations)
+        assert mounted, f"expected {path} to be mounted as a volume in {container}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -145,7 +145,9 @@ jobs:
             --initial-admin-password=changeme \
             --initial-organization "Foreman CI" \
             --initial-location "Internet" \
-            --tuning development
+            --tuning development \
+            --content-import-path /custom/import \
+            --content-export-path /custom/export
       - name: Deploy features
         run: |
           ./foremanctl deploy \
--- a/development/playbooks/_pulp/metadata.obsah.yaml
+++ b/development/playbooks/_pulp/metadata.obsah.yaml
@@ -1,0 +1,12 @@
+---
+variables:
+  pulp_import_paths:
+    help: Extra file path that Pulp can use for content imports. Argument may be used more than once.
+    action: append_unique
+    type: AbsolutePath
+    parameter: --content-import-path
+  pulp_export_paths:
+    help: Extra file path that Pulp can use for content exports. Argument may be used more than once.
+    action: append_unique
+    type: AbsolutePath
+    parameter: --content-export-path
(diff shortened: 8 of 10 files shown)


src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -23,6 +23,17 @@ variables:
     action: append_unique
     type: FQDN
     parameter: --certificate-cname
+  pulp_extra_import_paths:
```
ehelms: The user doesn't care that it's Pulp right? How about:


```suggestion
  content_extra_import_paths:
```

What makes them "extra" ?

src/roles/pulp/README.md
```diff
@@ -0,0 +1,34 @@
...
+- `pulp_plugins`: Additional Pulp plugins to enable (default: `[pulp_container, pulp_rpm]`)
+- `pulp_database_name`: Name of the Pulp database (default: `pulp`)
+- `pulp_database_user`: Database user (default: `pulp`)
+- `pulp_database_host`: Database host (default: `localhost`)
+- `pulp_database_port`: Database port (default: `5432`)
+- `pulp_database_password`: Database password (required, no default)
+- `pulp_database_ssl_mode`: Database SSL mode (default: `disabled`)
+- `pulp_database_ssl_ca`: Path to the database SSL CA certificate on the control node (default: empty)
+
+Usage Inside foremanctl
```
ehelms, line 27: I am not sure this detail makes sense at this level. I'd rather see a section in the deployment document around content or add a content document.

src/roles/pulp/README.md
```diff
@@ -0,0 +1,34 @@
+Pulp Role
+=========
+
+Deploys and manages [Pulp](https://pulpproject.org/) as Podman containers.
+
+Variables
+---------
+- `pulp_container_image`: Container image to use for Pulp (default: `quay.io/foreman/pulp`)
```
ehelms, line 8: I am wary of storing defaults in the documentation here. The code can show those and reduce our overhead.

src/roles/pulp/README.md
```diff
@@ -0,0 +1,34 @@
+Pulp Role
+=========
+
+Deploys and manages [Pulp](https://pulpproject.org/) as Podman containers.
```
ehelms:
```suggestion
Deploys and manages [Pulp](https://pulpproject.org/) as Podman quadlets.
```

src/roles/pulp/tasks/main.yaml
```diff
@@ -36,6 +36,22 @@
     - tmp
     - assets
     - media
+    - imports
+    - exports
+
+- name: Create extra import directories
```
ehelms: I assume these would have to be mounted into the container?

src/roles/pulp/defaults/main.yaml
```diff
@@ -66,6 +73,10 @@ pulp_settings_other_env:
   PULP_CONTENT_WORKERS: "{{ pulp_content_service_worker_count }}"
   PULP_TOKEN_AUTH_DISABLED: "true"
   PULP_FLATPAK_INDEX: "true"
+  PULP_ALLOWED_IMPORT_PATHS: >-
+    {{ pulp_default_import_path + pulp_extra_import_paths }}
+  PULP_ALLOWED_EXPORT_PATHS: >-
+    {{ pulp_default_export_path + pulp_extra_export_paths }}
```
sjha4: Locally I have this:
 PULP_ALLOWED_IMPORT_PATHS: >-
   ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
PULP_ALLOWED_EXPORT_PATHS: >-
 ['/var/lib/pulp/exports']

These would be the default paths..

src/roles/pulp/defaults/main.yaml
```diff
@@ -22,6 +22,13 @@ pulp_pulp_url: "http://{{ ansible_facts['fqdn'] }}:24817"
 
 pulp_enable_analytics: false
 
+pulp_default_import_path:
```
ehelms: This says `import_path` but it's an array.
I wonder can we have a single variable for import and one for export, and they are just lists that appended to?

src/playbooks/deploy/metadata.obsah.yaml
aidenfine: Good point. I used extra because users can add extra or additional import/export paths. Would you prefer something other than extra?

src/roles/pulp/defaults/main.yaml
aidenfine: I think this is a good idea. So instead of 4 variables (2 default, 2 lists to hold user passed values). We could just have 2 lists both by default containing the default path and just append user passed values to these lists.

Am I understanding this correctly?

src/roles/pulp/tasks/main.yaml
aidenfine: Yeah I didn't think about that, I will add this

src/roles/pulp/README.md
aidenfine: I was hoping someone would comment on the docs. I tried to follow similar patterns from the `roles/hammer/README.md`. I also feel this is kinda an info dump but wanted to get opinions on it. I don't mind removing this doc from the PR if it makes more sense not to have this

src/playbooks/deploy/metadata.obsah.yaml
ianballou: I'm trying to remember, I think the installer uses this "extra" terminology.

ianballou:
Can we make these import/export arguments also available via `forge deploy-dev` ?

src/playbooks/deploy/metadata.obsah.yaml
ehelms: Ahh, it uses `additional`. So there are the "you can use these out of the box paths" and then there are "user has some setup, they want to use these paths and accept the requirements to make this work".

src/roles/pulp/README.md
ehelms: I like the having of a README. I can see now that this is following roughly the outline from the hammer role. I think as long as we are consistent in the structure then keep it.

ehelms:
> Can we make these import/export arguments also available via `forge deploy-dev` ?

Consider this method for defining them -- https://github.com/theforeman/foremanctl/blob/master/docs/developer/playbooks-and-roles.md#shared-metadata-fragments

src/roles/pulp/README.md
ehelms: And we should define that roughly in https://github.com/theforeman/foremanctl/blob/master/docs/developer/playbooks-and-roles.md#shared-metadata-fragments

You do not have to do that here in this PR. I am noting this as a reminder.

ianballou:
I just tried this out, seeing the following in my pulp-api service definition:

```
Environment='PULP_ALLOWED_IMPORT_PATHS=['"'"'/var/lib/pulp/imports'"'"', '"'"'/var/lib/pulp/my_imports'"'"', '"'"'/var/lib/pulp/my_exports'"'"', '"'"'/var/lib/pulp/exports'"'"']'
Environment='PULP_ALLOWED_EXPORT_PATHS=['"'"'/var/lib/pulp/exports'"'"', '"'"'/var/lib/pulp/my_exports'"'"']'
```
So far so good, assuming the escaping is correct. And yes, I like to be able to import from my export directory.

ianballou:
Export and import worked! I was able to generate  `/var/lib/pulp/exports/Default_Organization/Export-Zoo-1/1.0/2026-04-17T17-37-04-00-00/` and import it back into another repository. Perfect.

aidenfine:
@ehelms I am not too sure on how to make sure the created import and export directories are mounted into the container. Can you point me into the right direction. Maybe somewhere in the code we handle a similar workflow

ehelms:
> @ehelms I am not too sure on how to make sure the created import and export directories are mounted into the container. Can you point me into the right direction. Maybe somewhere in the code we handle a similar workflow

You'll need to do something like this -- https://github.com/theforeman/foremanctl/blob/master/src/roles/pulp/tasks/main.yaml#L137

src/roles/pulp/defaults/main.yaml
```diff
@@ -22,6 +22,10 @@ pulp_pulp_url: "http://{{ ansible_facts['fqdn'] }}:24817"
 
 pulp_enable_analytics: false
 
+# pulp import/export paths
+pulp_import_paths: ["/var/lib/pulp/sync_imports", "/var/lib/pulp/imports"]
+pulp_export_paths: ["/var/lib/pulp/exports"]
```
evgeni: Should these be *always* present, or only until the user overrides them via `--content-*-path`?
(`action: append(_unique)` does not consider role defaults)

aidenfine: I am not sure I assumed that these defaults would always be present, and users can add extra paths as needed. From some docs I read on how this works it does sound like they should always be present. Can someone else confirm this?

ehelms: I read https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L67-L70 the same way. There are default "always" working paths. And users can add additional paths.

ianballou: Let's keep these as the defaults and let users further modify the paths. I'm indifferent though as to whether or not we append or let users overwrite the entire paths.

src/roles/pulp/defaults/main.yaml
```diff
@@ -22,6 +26,10 @@ pulp_pulp_url: "http://{{ ansible_facts['fqdn'] }}:24817"
 
 pulp_enable_analytics: false
 
+# pulp import/export paths
```
ehelms: This is redundant.

src/roles/pulp/defaults/main.yaml
```diff
@@ -10,8 +10,12 @@ pulp_worker_count: "{{ [8, ansible_facts['processor_nproc']] | min }}"
 pulp_content_service_worker_count: "{{ (2 * ([8, ansible_facts['processor_nproc']] | min)) + 1 }}"
 pulp_api_service_worker_count: "{{ ([4, ansible_facts['processor_nproc']] | min) + 1 }}"
 
-pulp_volumes:
-  - /var/lib/pulp:/var/lib/pulp
+pulp_volumes: >-
+  {{
+    ['/var/lib/pulp:/var/lib/pulp'] +
+    (pulp_import_paths | map('regex_replace', '^(.+)$', '\1:\1') | list) +
```
ehelms, line 16: What are these regexs doing?

development/playbooks/deploy-dev/metadata.obsah.yaml
```diff
@@ -29,6 +29,16 @@ variables:
   foreman_development_github_username:
     help: GitHub username to add as additional remote for git checkouts
     action: store
+  pulp_import_paths:
```
ehelms: You could define these in a file like https://github.com/theforeman/foremanctl/blob/master/src/playbooks/_database_connection/metadata.obsah.yaml and then include that file in both deploy and deploy-dev.

src/roles/pulp/defaults/main.yaml
aidenfine: lets take one item from `pulp_import_paths` and call it *string1*, so `string1="/foo/bar"`
this regex turns `string1` into the same format as the line above. (`'/var/lib/pulp:/var/lib/pulp'`). 

So after `string1` runs through regex, we get `"/foo/bar:/foo:bar"`

ehelms: As `/var/lib/pulp` is already mounted as a volume, do we want / need to mount the default paths? I realize this can make the logic a bit more complicated but I worry about performance of mounting unnecessary volumes, especially those that are sub-directories of an existing mount.

src/roles/pulp/defaults/main.yaml
```diff
@@ -10,8 +10,12 @@ pulp_worker_count: "{{ [8, ansible_facts['processor_nproc']] | min }}"
 pulp_content_service_worker_count: "{{ (2 * ([8, ansible_facts['processor_nproc']] | min)) + 1 }}"
 pulp_api_service_worker_count: "{{ ([4, ansible_facts['processor_nproc']] | min) + 1 }}"
 
-pulp_volumes:
-  - /var/lib/pulp:/var/lib/pulp
+pulp_volumes: >-
+  {{
+    ['/var/lib/pulp:/var/lib/pulp'] +
+    ((pulp_default_import_paths + pulp_import_paths) | map('regex_replace', '^(.+)$', '\1:\1') | list) +
```
ehelms:
```suggestion
    ((pulp_import_paths) | map('regex_replace', '^(.+)$', '\1:\1') | list) +
```

I think this would be the way to prevent known double-mounting? Same for line below.

aidenfine: Okay I will make this change today

aidenfine: Okay pushed that change with some test changes as well

tests/pulp_test.py
```diff
@@ -73,3 +83,25 @@ def test_pulp_worker_target(server):
...
+    for path in import_paths:
+        assert server.file(path).is_directory
+
+def test_pulp_export_directories(server, pulp_import_export_paths):
+    _, export_paths = pulp_import_export_paths
+    for path in export_paths:
+        assert server.file(path).is_directory
+
+@pytest.mark.parametrize("container", ["pulp-api", "pulp-content", "pulp-worker-1"])
+def test_pulp_import_export_volume_mounts(server, container, pulp_import_export_paths):
```
ehelms, line 136: My read on this is that it will check only the defaults right now, since neither the tests or the test setup is specifying any additional paths. Have you thought about that and adding specification of additional paths to the test setup?

aidenfine: It will also read custom paths as well the fixture will pull items from the environment and verify that any path in the `PULP_ALLOWED_IMPORT_PATHS` or `PULP_ALLOWED_EXPORT_PATHS` is present AND a volume mount.

ehelms:
@aidenfine this will need a rebase now

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -29,6 +29,16 @@ variables:
     action: append_unique
     type: FQDN
     parameter: --certificate-cname
+  pulp_import_paths:
```
ehelms: Use the fragment you created instead of repeating these.

src/roles/pulp/defaults/main.yaml
```diff
@@ -10,8 +10,12 @@ pulp_worker_count: "{{ [8, ansible_facts['processor_nproc']] | min }}"
 pulp_content_service_worker_count: "{{ (2 * ([8, ansible_facts['processor_nproc']] | min)) + 1 }}"
 pulp_api_service_worker_count: "{{ ([4, ansible_facts['processor_nproc']] | min) + 1 }}"
 
-pulp_volumes:
-  - /var/lib/pulp:/var/lib/pulp
+pulp_volumes: >-
+  {{
+    ['/var/lib/pulp:/var/lib/pulp'] +
+    ((pulp_import_paths) | map('regex_replace', '^(.+)$', '\1:\1') | list) +
```
ehelms:
```suggestion
    (pulp_import_paths | map('regex_replace', '^(.+)$', '\1:\1') | list) +
```

src/roles/pulp/defaults/main.yaml
```diff
@@ -10,8 +10,12 @@ pulp_worker_count: "{{ [8, ansible_facts['processor_nproc']] | min }}"
...
+    ((pulp_export_paths) | map('regex_replace', '^(.+)$', '\1:\1') | list)
```
ehelms:
```suggestion
    (pulp_export_paths | map('regex_replace', '^(.+)$', '\1:\1') | list)
```

tests/pulp_test.py
ehelms: My point is that this isn't testing the user input anywhere and therefore not checking whether the custom parts are working. Consider adding these to CI to be able to assert that they work.

ehelms:
Consider also:

  - Major: docs/user/parameters.md — Add `--content-import-path` and `--content-export-path` to the Mapped section with their foreman-installer equivalents. Remove                               
  `--foreman-proxy-content-pulpcore-additional-import-paths` from the Undetermined list on line 158.
  - Minor: Consider adding a section to docs/user/ (either a new file or a section in an existing doc) explaining the content import/export path configuration — what the paths are used for, 
  when users need to set them, and example usage with --content-import-path.

ianballou:
Just wanted to note that this is the upstream issue for it: https://github.com/theforeman/foremanctl/issues/445

tests/pulp_test.py
aidenfine: I think I fixed this in my latest commit going to watch the CI run

src/roles/pulp/defaults/main.yaml
evgeni: While it shouldn't have a perf-impact, I'd avoid mounting anything under /var/lib/pulp extra, just to make the resulting container less confusing to look at.

But, if you look closely, this uses `pulp_import_paths` which is the user-supplied value, not the `*_default_*` values we have, so it "should do the right thing" already

tests/pulp_test.py
```diff
@@ -16,6 +16,14 @@ def pulp_status_curl(server):
...
     return json.loads(pulp_status_curl.stdout)
 
+@pytest.fixture(scope="module")
+def pulp_import_export_paths(server):
+    result = server.run("podman inspect pulp-api --format '{{json .Config.Env}}'")
+    assert result.succeeded
+    env = {v.split('=', 1)[0]: v.split('=', 1)[1] for v in json.loads(result.stdout)}
+    import_paths = json.loads(env['PULP_ALLOWED_IMPORT_PATHS'].replace("'", '"'))
+    export_paths = json.loads(env['PULP_ALLOWED_EXPORT_PATHS'].replace("'", '"'))
+    return import_paths, export_paths
```
evgeni: This implies setting of the env vars was correct, which is part of the flow we're supposed to test.

We had a similar issue in #484 and I think the correct path here would to load `parameters.yaml` and taking the user-input from there assert that the values have been properly propagated to the various consumers.

src/roles/pulp/defaults/main.yaml
evgeni, line 16: As an alternative, you could write: `pulp_import_paths | zip(pulp_import_paths) | map('join', ':') | list`

I am not sure it's actually nicer than the regex, but wanted to still mention it :)

evgeni (changes requested):
the (newly added) ruff check is complaining, and I think this needs some polishing on the tests side, but conceptually this looks solid!

aidenfine:
@evgeni I updated the PR to use the `parameters.yaml` file in the import/export tests and fixed ruff errors. Tests are passing

tests/pulp_test.py
```diff
@@ -85,3 +106,42 @@ def test_pulp_worker_target(server):
 def test_pulp_manager_check(server):
     result = server.run("podman exec -ti pulp-api pulpcore-manager check --deploy")
     assert result.succeeded
+
+
+@pytest.mark.parametrize("container", ["pulp-api", "pulp-content", "pulp-worker-1"])
+def test_pulp_import_export_env_vars(server, container, pulp_import_export_paths):
```
evgeni: instead of testing the env vars, you could test whether the setting is actually passed on to pulp as you write in your PR description via `pulpcore-manager`, seems cleaner to me?

evgeni (approved):

aidenfine:
Will fix these failing tests today.

On another note @evgeni is this test change what you expected? I had Claude help create this test for me.

ehelms:
Testing out a skill that analyzes failures, so bear with me a bit here :)

```
Universal Failure

  tests/pulp_test.py::test_pulp_import_export_settings

  Error: SyntaxError: invalid syntax inside pulpcore-manager shell -c

  Root cause: Shell quoting collision. The test constructs a Python one-liner with single-quoted dict keys and wraps it in single quotes for the -c argument:

  # tests/pulp_test.py:114
  py = "from django.conf import settings; import json; print(json.dumps({'import': list(settings.ALLOWED_IMPORT_PATHS), 'export': list(settings.ALLOWED_EXPORT_PATHS)}))"
  result = server.run(f"podman exec pulp-api pulpcore-manager shell -c '{py}'")

  The f-string wraps py in single quotes. The shell then sees '...{' as the first token and import as a bare (unquoted) word — which Python parses as the import keyword, producing the
  SyntaxError. The actual command that reaches pulpcore-manager is:

  ...shell -c 'from django.conf import settings; import json; print(json.dumps({import: ..., export: ...}))'

  Suggested fix — use double quotes for the dict keys so they survive the single-quote shell wrapping (tests/pulp_test.py:114):

  py = 'from django.conf import settings; import json; print(json.dumps({"import": list(settings.ALLOWED_IMPORT_PATHS), "export": list(settings.ALLOWED_EXPORT_PATHS)}))'
  result = server.run(f"podman exec pulp-api pulpcore-manager shell -c '{py}'")
```

aidenfine:
> Testing out a skill that analyzes failures, so bear with me a bit here :)
> 
> ```
> Universal Failure
> 
>   tests/pulp_test.py::test_pulp_import_export_settings
> 
>   Error: SyntaxError: invalid syntax inside pulpcore-manager shell -c
> 
>   Root cause: Shell quoting collision. The test constructs a Python one-liner with single-quoted dict keys and wraps it in single quotes for the -c argument:
> 
>   # tests/pulp_test.py:114
>   py = "from django.conf import settings; import json; print(json.dumps({'import': list(settings.ALLOWED_IMPORT_PATHS), 'export': list(settings.ALLOWED_EXPORT_PATHS)}))"
>   result = server.run(f"podman exec pulp-api pulpcore-manager shell -c '{py}'")
> 
>   The f-string wraps py in single quotes. The shell then sees '...{' as the first token and import as a bare (unquoted) word — which Python parses as the import keyword, producing the
>   SyntaxError. The actual command that reaches pulpcore-manager is:
> 
>   ...shell -c 'from django.conf import settings; import json; print(json.dumps({import: ..., export: ...}))'
> 
>   Suggested fix — use double quotes for the dict keys so they survive the single-quote shell wrapping (tests/pulp_test.py:114):
> 
>   py = 'from django.conf import settings; import json; print(json.dumps({"import": list(settings.ALLOWED_IMPORT_PATHS), "export": list(settings.ALLOWED_EXPORT_PATHS)}))'
>   result = server.run(f"podman exec pulp-api pulpcore-manager shell -c '{py}'")
> ```

Lets see if it passes 🤞

evgeni:
Yes, the test looks like what I was thinking of, thanks!

While the existing quoting seems to be fixed, for future reference: [shlex.quote](https://docs.python.org/3/library/shlex.html#shlex.quote) can be useful in such cases, similar to Ruby's `Shellwords`.

aidenfine:
> Yes, the test looks like what I was thinking of, thanks!
> 
> While the existing quoting seems to be fixed, for future reference: [shlex.quote](https://docs.python.org/3/library/shlex.html#shlex.quote) can be useful in such cases, similar to Ruby's `Shellwords`.

Great, thank you I wasn't aware of shlex. Are we good to merge once all tests pass?

PR #456, opened 2026-04-19
Add certs expiry params

ShimShtein:

https://redhat.atlassian.net/browse/SAT-43479

We need to expose the certificate validity to the users, so it would be possible to align it with the local security policies.

* Created new ansible variables to parameterize the expiration date
* Exposed the variables to the user through `metadata.obsah.yaml`

Steps to reproduce:

* Run an installation procedure
* Make sure to set custom expiration for each of the parameters
* Validate the expiration of certificates

#### Checklist


ShimShtein:
@stejskalleos One thing I see already as a problem: the certs are not regenerated if I change the expiry date.

evgeni:
That's quite intentional (see https://docs.ansible.com/projects/ansible/latest/collections/community/crypto/x509_certificate_module.html#parameter-selfsigned_not_after and https://docs.ansible.com/projects/ansible/latest/collections/community/crypto/x509_certificate_module.html#parameter-ownca_not_after) as otherwise the use of relative timestamps would mean that the cert is regenerated on every run.

ShimShtein:
@evgeni should we switch to absolute timestamps then? Maybe it will make more sense to us, so we will be able to prolong the validity without too much hassle?

evgeni:
My gut feeling is that it wouldn't make the UX any better, but I might have not thought is through completely.

ShimShtein:
I would say that the client and server certs are safe to regenerate, which is not really possible without the option (unless we want to split it into a different feature).
For the CA it's a bit more problematic, since we will need to create a specific merged CA cert, distribute it and switch to a new one. I would vote not to allow the regeneration altogether (only by deleting the current cert).

evgeni:
You don't want to regenerate certificates continuously. This will trigger service restarts, which users don't like. And also could have negative effects if people use certificate pinning or similar (based on the actual cert, not the CA, this is clearly "don't do it like this" territory, but often people don't listen to such advice). Also, this is kinda against the whole idea of certificate validity, as it just pushes it out "forever".

So we could either use absolute timestamps (which means we'd have to store them), or dynamically set `ignore_timestamps=false` when the user indicates they *actually* want to perform a certificate renewal?

ShimShtein:
I was thinking about using absolute timestamps, as I said previously. 
So what I thought was:
1. Use absolute timestamps for all three fields (ca, server, client)
2. Set the `ignore_timestamps` statically to `false` for server and client
3. Add a `regenerate_ca` flag that will control the `ignore_timestamps` for the CA field.

What do you think, would it make more sense? This way we enable easy change to the more common certs, and a confirmation flag to CA, that requires more attention from the user.

evgeni:
In your setup, how would a user renew certificates that are bound to expire in (say) two weeks?

ShimShtein:
If we are talking about server or client certs it would be running 
```
foremanctl deploy --cert-server-validity 20-04-2027 --cert-client-validity 20-04-2027
```
But if the user wants to change the CA validity, then it would be something like
```
foremanctl deploy --cert-ca-validity 20-04-2027 --regenerate-ca
```

evgeni:
Okay. And my alternative (for server/client) would be:
Define the validity once via `foremanctl --certificate-days 7300`
Renew at any time with `foremanctl --certificate-renew` 
(Or whatever the names end up with, thats not important)

Does anyone else have an opinion/preference?

ekohl:
I think a user wants minimal downtime and secure communication. For that we chose to use certificates and they need to be valid.

@evgeni already mentioned to avoid service restarts due to certificate changes, but downtime can also be caused by expired certificates.

How do we translate that into a process and parameters? You need to create a certificate and determine a validity period (measured in days). Then you also need to determine when to renew. That probably shouldn't be the validity period, otherwise you'll renew it all the time.

Let's Encrypt has [proposed](https://letsencrypt.org/2025/09/16/ari-rfc) a process via [RFC 9773](https://www.rfc-editor.org/rfc/rfc9773.html). While I don't expect us to implement this right now, I like the idea and we can learn from the concepts. `suggestedWindow` jumps out to me: can we also determine this for our certificates?

A basic implementation to have a minimum validity period.

> Define the validity once via `foremanctl --certificate-days 7300`

I like this.

> Renew at any time with `foremanctl --certificate-renew`

I'm thinking of automating it and avoid introducing new parameters by using a formula. For example, take half the `certificate_days` but at most renew every 30 days: `min(certificate_days / 2, 30)`.

I don't know if this also works for the CA or it needs a more nuanced implementation.

ShimShtein:
If we will introduce a single `--certificate-renew` flag, we won't be able to manage separate validity periods. I would suggest having a parameter per certificate type (ca, client, server).
Specifically for the CA, the problem is that we need to distribute the new CA in a secure way (which means installing a dual CA for a while on the server, distributing the new CA to all the known clients and reinstalling the new CA on the server). I would argue that the CA renewal process should be handled by a separate command.
The problem with automation is not about regeneating the certs, it's about distributing those certs to the clients (smart proxies). This means keeping a list of all known hosts (we can derive it from the list of smart-proxies in Foreman) and orchestrating the upload of the certificates to those hosts.

evgeni:
while I can see different validity for CA and non-CA, I see no reason to differentiate between server and client certs.

ShimShtein:
I think I have addressed everything. How does it look now?

ekohl:
Should we track the renewal in the same PR? I think there's a lot to discuss on that topic and I don't think we should hold up adding the expiration params on that.

src/roles/certificates/tasks/issue.yml
```diff
@@ -1,4 +1,17 @@
 ---
+- name: Remove existing server and client certificate material when renewing
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: absent
+  loop:
+    - "{{ certificates_ca_directory_keys }}/{{ certificates_hostname }}.key"
```
evgeni: keys don't need to be deleted, they don't contain any dates

src/roles/certificates/tasks/issue.yml
```diff
@@ -1,4 +1,17 @@
...
+    - "{{ certificates_ca_directory_requests }}/{{ certificates_hostname }}.csr"
```
evgeni: csrs don't need to be deleted, they don't contain any dates

src/roles/certificates/tasks/issue.yml
```diff
@@ -1,4 +1,17 @@
 ---
+- name: Remove existing server and client certificate material when renewing
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: absent
```
evgeni: instead of deleting the certs, we could set `ignore_timestamps: "{{ not certificates_renew }}"` (or whatever the right syntax is for inverting a boolean in jinja)

not sure what's cleaner tho

evgeni:
> Should we track the renewal in the same PR? I think there's a lot to discuss on that topic and I don't think we should hold up adding the expiration params on that.

I think the current state is "good enough" to be used as a starting ground for any follow up discussions if needed (or to be ripped out, if we decide to go a different route).

src/roles/certificates/tasks/issue.yml
ekohl: Or use [force: true](https://docs.ansible.com/projects/ansible/latest/collections/community/crypto/x509_certificate_module.html#parameter-force).

evgeni: Or that

src/playbooks/_certificate_validity/metadata.obsah.yaml
```diff
@@ -0,0 +1,13 @@
+---
+variables:
+  certificates_ca_validity_days:
+    help: Lifetime of the generated CA certificate, in days.
+    parameter: --cert-ca-validity-days
```
ehelms: We should be consistent: `--cert` or `--certificate`

src/roles/certificates/tasks/issue.yml
```diff
@@ -1,4 +1,17 @@
 ---
+- name: Remove existing server and client certificate material when renewing
```
ehelms:
```suggestion
- name: Remove existing server and client certificates for renewal
```

src/vars/default_certificates.yml
```diff
@@ -1,4 +1,6 @@
 ---
+certificates_ca_validity_days: 7300
+certificates_validity_days: 7300
```
ehelms: Do we need to restate these here?

src/vars/installer_certificates.yml
```diff
@@ -1,4 +1,6 @@
 ---
+certificates_ca_validity_days: 7300
+certificates_validity_days: 7300
```
ehelms: Right now, these have no affect on installer certificates, as those are not managed. That proposed handling change is in https://github.com/theforeman/foremanctl/pull/421

ehelms:
Configuration without deployment is a heavy operation for a simple change. `foremanctl --certificate-days 7300` will update the configuration, change nothing but force a full run of the installer. That makes me wonder if we should have a sub-command for this situation generically (e.g. `foremanctl config`) or have a dedicated certificates management command. Here's how I'd describe or see those two commands if we went either route:

### `foremanctl config`

This would allow users to set or modify input configuration, and thus change what's in parameters.yaml without initiating a change. 

`foremanctl config create --certificate-days=7300`
`foremanctl config update --certificate-days=2000`

### `foremanctl certs`

This command would allow users to manage the certificates. Update configuration, regenerate, switch to custom or full stack CA, or configure / connect to an external CA service potentially. This command would not deploy. Deployment and thus rotation or updating of certificates would happen via an explicit deploy command. Deploy would still generate a default set of certificates if none exist.

`foremanctl certificates update --certificate-days-7300`
`foremanctl certificates renew`
`foremanctl certificates renew-ca`
`foremanctl certificates generate-bundle`

ShimShtein:
Now the PR should be in a workable state.

As for the separate commands: I think the certificates renew command that will do a (partial) deploy does not seem to fit properly here.
I would argue that there is a place for the `foremanctl config` command. Where the user can "prepare" the environment with multiple calls to the `config` command and then execute a full deployment with `foremanctl deploy`.
Of course bundle generation command should be separate from the deploy command.

src/playbooks/_certificate_validity/metadata.obsah.yaml
```diff
@@ -0,0 +1,13 @@
...
+    help: Lifetime of the generated CA certificate, in days.
+    parameter: --certificate-ca-validity-days
+  certificates_validity_days:
+    help: Lifetime of the generated server and client certificates, in days.
+    parameter: --certificate-validity-days
+  certificates_renew:
+    help: Remove and regenerate server and client certificates for each issued hostname (does not regenerate the CA).
+    parameter: --certificate-renew
+    action: store_true
+    persist: false
```
evgeni, line 13: Flashbacks to https://github.com/theforeman/foreman-installer/pull/764 and this nice feeling that today it's just a `persist: false`.

src/vars/default_certificates.yml
evgeni: nope, we can rely on role default or user-provided-custom-data

src/vars/installer_certificates.yml
evgeni: and even if they had, the argument from https://github.com/theforeman/foremanctl/pull/456/changes#r3117445074 applies here too

evgeni (changes requested):
ACK once the vars files are cleaned up :)

src/playbooks/_certificate_validity/metadata.obsah.yaml
```diff
@@ -0,0 +1,13 @@
+---
+variables:
+  certificates_ca_validity_days:
+    help: Lifetime of the generated CA certificate, in days.
+    parameter: --certificate-ca-validity-days
+  certificates_validity_days:
+    help: Lifetime of the generated server and client certificates, in days.
+    parameter: --certificate-validity-days
+  certificates_renew:
+    help: Remove and regenerate server and client certificates for each issued hostname (does not regenerate the CA).
```
evgeni:
```suggestion
    help: Regenerate server and client certificates for each issued hostname (does not regenerate the CA).
```

src/roles/certificates/tasks/issue.yml
```diff
@@ -28,7 +28,8 @@
     ownca_path: "{{ certificates_ca_directory_certs }}/ca.crt"
     ownca_privatekey_path: "{{ certificates_ca_directory_keys }}/ca.key"
     ownca_privatekey_passphrase: "{{ certificates_ca_password }}"
-    ownca_not_after: "+7300d"
+    ownca_not_after: "+{{ certificates_validity_days }}d"
+    force: "{{ certificates_renew | bool }}"
```
ekohl, line 32: IMHO here we see the real value of using `community.crypto`. With our hand crafted openssl statements this would have been a pain.

src/playbooks/_certificate_validity/metadata.obsah.yaml
ekohl: I'd start the last bit as a new sentence. Also questioning if the "for each issued hostname" is really relevant. How about:
```suggestion
    help: Force renewal of the server and client certificates. Does not regenerate the CA.
```

evgeni: As we're adding custom certs in #462 and those would not be renewed by this, I wonder if there is a word we could use to differentiate the certs here. Something like "*managed* server and client certificates"? Doesn't have to be solved in this PR, but having clear wording which movable parts of the stack are "bring your own" and which are "we do it for you" would be cool.

ekohl: The old `default CA` and `server CA` naming was really poor, but I struggle to come up with something better. Internal vs External perhaps?

evgeni: "Force renewal of the server and client certificates issued by the internal CA."? Sounds cludgy.

ShimShtein: What about `Regenrate server and client certificates previously generated by foremanctl. Does not regenerate the CA.`

evgeni: Technically they are not regenerated (the private and public keys remain the same, only the signature changes), but I don't really mind.

src/roles/certificates/defaults/main.yml
```diff
@@ -7,3 +7,7 @@ certificates_ca_directory_requests: "{{ certificates_ca_directory }}/requests"
 certificates_cnames: []
 certificates_algorithm_type: RSA
 certificates_algorithm_size: 4096
+# Validity periods in days (used as +Nd for community.crypto not_after).
```
ehelms: A bit nitpicky here -- the variable name tells me this information already. And I do not understand the part in parenthesis to feel like it adds value.

ehelms (changes requested):
This is missing documentation updates and tests at least for the renew workflow.

ShimShtein:
Updated the documentation, but couldn't understand how to properly add a test that will rerun the installer.

docs/user/certificates.md
```diff
@@ -71,11 +71,31 @@ To remove all CNAMEs from an existing certificate, use `--reset-certificate-cnam
 
 On each run the role compares the Subject Alternative Names in the existing server certificate against the desired list (hostname + any `--certificate-cname` values). If they differ, the CSR and certificate are regenerated automatically. This means both adding and removing CNAMEs are handled transparently without manual cleanup.
 
+### Validity period
```
ehelms, line 74: Can you reduce the bolding in this section?

docs/user/certificates.md
```diff
@@ -71,11 +71,31 @@ To remove all CNAMEs from an existing certificate, use `--reset-certificate-cnam
...
+foremanctl deploy --certificate-renew
+```
+
+The `--certificate-renew` flag is **not persisted** in foremanctl’s answers file (one-shot).
+
 ### Current Limitations
 
 - Cannot provide custom certificate files during deployment
-- Fixed 20-year certificate validity period
-- Limited certificate customization options
```
evgeni, line 78: This limitation is still there.

docs/user/certificates.md
```diff
@@ -149,10 +166,12 @@ The `certificate_checks` role uses `foreman-certificate-check` binary to validat
...
 **Certificate Properties:**
-- Key Size: 4096-bit RSA
-- Hash Algorithm: SHA256
-- Validity Period: 7300 days (20 years)
-- Extensions: serverAuth, clientAuth, nsSGC, msSGC
+- Key algorithm / size: RSA 4096 by default (`certificates_algorithm_type`, `certificates_algorithm_size`)
+- Hash Algorithm: SHA256 (via `community.crypto` defaults)
+- Validity:
+  - CA: `certificates_ca_validity_days` (default 7300 days)
+  - Server and client leaf certs: `certificates_validity_days` (default 7300 days)
```
evgeni: what are "leaf" certs?

src/playbooks/_certificate_validity/metadata.obsah.yaml
```diff
@@ -0,0 +1,13 @@
...
+variables:
+  certificates_ca_validity_days:
+    help: Lifetime of the generated CA certificate, in days.
+    parameter: --certificate-ca-validity-days
+  certificates_validity_days:
+    help: Lifetime of the generated server and client certificates, in days.
+    parameter: --certificate-validity-days
+  certificates_renew:
+    help: Regenrate server and client certificates previously generated by foremanctl. Does not regenerate the CA.
+    parameter: --certificate-renew
```
amolpati30, line 11: If I set the expiration period for both the `client` and `server` certificates to 5 days each, my question is whether the customer will receive a notification before the certificates expire, as I do not see that happening here.

My second point is that `--certificate-renew` does not renew the CA certificate. If I set the CA certificate validity to 5 days, how will the CA certificate be renewed?

Also, you mentioned that the certificate is considered to have a lifetime. What exactly does that mean if I explicitly set its validity period to 5 days?

docs/user/certificates.md
ShimShtein: Done

docs/user/certificates.md
ShimShtein: done

src/playbooks/_certificate_validity/metadata.obsah.yaml
ShimShtein: > whether the customer will receive a notification before the certificates expire, as I do not see that happening here.

Is it something that we previously set up? 

> My second point is that --certificate-renew does not renew the CA certificate. If I set the CA certificate validity to 5 days, how will the CA certificate be renewed?

We have discussed it previously: CA renewal is a process and I want to treat it in a separate story.

> Also, you mentioned that the certificate is considered to have a lifetime. What exactly does that mean if I explicitly set its validity period to 5 days?

I may be missing something, but if you set validity period to 5 days, it means that the lifetime of the certificate would be 5 days from the run of the module.

docs/user/certificates.md
```diff
@@ -149,10 +167,12 @@ The `certificate_checks` role uses `foreman-certificate-check` binary to validat
...
-- Key Size: 4096-bit RSA
-- Hash Algorithm: SHA256
-- Validity Period: 7300 days (20 years)
-- Extensions: serverAuth, clientAuth, nsSGC, msSGC
+- Key algorithm / size: RSA 4096 by default (`certificates_algorithm_type`, `certificates_algorithm_size`)
+- Hash Algorithm: SHA256 (via `community.crypto` defaults)
+- Validity:
+  - CA: `certificates_ca_validity_days` (default 7300 days)
+  - Server and client certs: `certificates_validity_days` (default 7300 days)
+- Extensions: serverAuth / clientAuth as appropriate for server vs client certs
```
evgeni:
```suggestion
  - Server and client certificates: `certificates_validity_days` (default 7300 days)
- Extensions: serverAuth / clientAuth as appropriate for server and client certificates
```

I think we should not say "certs" in human readable docs.

evgeni (approved):
wording nitpick, but otherwise ACK

src/playbooks/_certificate_validity/metadata.obsah.yaml
amolpati30: > > whether the customer will receive a notification before the certificates expire, as I do not see that happening here.
> 
> Is it something that we previously set up?

I guess no, i do not see in the old installer.

amolpati30: If the server and client certificates have expired and I want to renew them, will this work using this option?
`foremanctl deploy --certificate-validity-days 365 --certificate-renew` or
`foremanctl deploy --certificate-renew`

I tried renewing the expired server and client certificates, but it does not seem to be supported. Is it required to have a valid CA certificate first before renewing the server and client certificates?

evgeni: `foremanctl deploy --certificate-renew` should be sufficient, in my understanding

was the CA expired too?

ShimShtein: You can do both. In the shorter case, I guess you will be getting the default validity (7300 days). Unless @evgeni says that we preserve the validity days parameter by default.

ShimShtein: >Is it required to have a valid CA certificate first before renewing the server and client certificates?

Yes, the CA has to be valid. We are signing the server and client certs with the CA.

amolpati30: > `foremanctl deploy --certificate-renew` should be sufficient, in my understanding
> 
> was the CA expired too?

Yes, the CA certificate has also expired.

evgeni: > You can do both. In the shorter case, I guess you will be getting the default validity (7300 days). Unless @evgeni says that we preserve the validity days parameter by default.

If you haven't marked it as `persist: false`, it will of course be preserved (which I think is the correct way to do it)

ShimShtein:
All the `certs` were fixed.

ehelms (approved):

amolpati30 (approved):
I have done below testing scenarios here to validate the certificates:

Verify that the default validity period is correctly applied to the certificates.

Explicitly set the certificate validity period, validate it, and then update the server and client certificate validity again using the --certificate-renew parameter, followed by validation.

PR #458, opened 2026-04-20
Add apache config support for /pypi

pavanshekar:

Add Apache config support for /pypi in foremanctl

Fixes [SAT-44475](https://redhat.atlassian.net/browse/SAT-44475)

* Added `<Location "/pypi">` block in `foreman-ssl-vhost.conf.j2` to proxy requests to Pulp API backend
* Configured required headers: X-CLIENT-CERT, X-FORWARDED-PROTO
* Set timeout to 600 seconds (matching other Pulp API routes)

Steps to reproduce:

* Deploy the changes
* SSH into the VM
* Verify the Apache configuration was generated:
   `grep -A 6 'Location "/pypi"' /etc/httpd/conf.d/foreman-ssl.conf`
* Expected: Should show the /pypi Location block with correct headers and proxy settings

#### Checklist

### DIFF
--- a/src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
+++ b/src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
@@ -68,6 +68,10 @@
     ProxyPassReverse {{ httpd_pulp_content_backend }}/pulp/content
   </Location>
 
+{% for httpd_pulp_snippet in httpd_enabled_pulp_snippets %}
+{% include httpd_pulp_snippet+'.j2' %}
+{% endfor %}
+
   <Location "/pulp/api/v3">
     RequestHeader unset REMOTE_USER
     RequestHeader unset REMOTE-USER
--- a/src/vars/base.yaml
+++ b/src/vars/base.yaml
@@ -26,6 +26,7 @@ httpd_server_ca_certificate: "{{ server_ca_certificate }}"
 httpd_client_ca_certificate: "{{ client_ca_certificate }}"
 httpd_server_certificate: "{{ server_certificate }}"
 httpd_server_key: "{{ server_key }}"
+httpd_enabled_pulp_snippets: "{{ ['pypi'] if 'pulp_python' in pulp_plugins else [] }}"
 
 pulp_content_origin: "https://{{ ansible_facts['fqdn'] }}"
 pulp_pulp_url: "https://{{ ansible_facts['fqdn'] }}"
--- a/tests/httpd_test.py
+++ b/tests/httpd_test.py
@@ -63,6 +63,14 @@ def test_https_pulp_auth(server, certificates, server_fqdn):
     assert cmd.stdout == '200'
 
 
+def test_https_pypi_endpoint(server, certificates, server_fqdn):
+    cmd = server.run(f"curl --cacert {certificates['server_ca_certificate']} https://{server_fqdn}/pypi/test/")
+    assert cmd.succeeded
+    # Verify route proxies to Pulp's Python plugin by checking for PythonDistribution in response
+    # (Rails or unconfigured routes would return different errors)
+    assert "PythonDistribution" in cmd.stdout
+
+
 def test_pub_directory_exists(server):
     pub_dir = server.file(HTTPD_PUB_DIR)
     assert pub_dir.exists
--- a/src/roles/httpd/defaults/main.yml
+++ b/src/roles/httpd/defaults/main.yml
@@ -3,6 +3,7 @@ httpd_pulp_api_backend: http://localhost:24817
 httpd_pulp_content_backend: http://localhost:24816
 httpd_foreman_backend: http://localhost:3000
 httpd_pub_dir: /var/www/html/pub
+httpd_enabled_pulp_snippets: []
 
 # MPM event module defaults
 httpd_server_limit: 25
--- a/src/roles/httpd/templates/pypi.j2
+++ b/src/roles/httpd/templates/pypi.j2
@@ -1,0 +1,7 @@
+  <Location "/pypi">
+    RequestHeader unset X-CLIENT-CERT
+    RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT
+    RequestHeader set X-FORWARDED-PROTO expr=%{REQUEST_SCHEME}
+    ProxyPass {{ httpd_pulp_api_backend }}/pypi timeout=600
+    ProxyPassReverse {{ httpd_pulp_api_backend }}/pypi
+  </Location>
--- a/src/vars/flavors/katello.yml
+++ b/src/vars/flavors/katello.yml
@@ -5,4 +5,5 @@ flavor_features:
   - content/ansible
   - content/container
   - content/deb
+  - content/python
   - content/rpm


src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -68,6 +68,14 @@
     ProxyPassReverse {{ httpd_pulp_content_backend }}/pulp/content
   </Location>
 
+  <Location "/pypi">
```
ehelms: In `foreman-installer` this was added conditionally if python content type was enabled. I think we should do the same here so as not to expose routes that have no effect.

pavanshekar: I've updated the template to conditionally include the /pypi Location block only when pulp_python is in the enabled plugins list.

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -68,6 +68,16 @@
     ProxyPassReverse {{ httpd_pulp_content_backend }}/pulp/content
   </Location>
 
+{% if 'pulp_python' in (pulp_enabled_plugins | default([])) %}
```
ehelms: The problem is that this variable is from the pulp role, and this is the httpd role.

@evgeni -- do you think we should go with role variables `httpd_pulp_python` or we should have Pulp role require/expect httpd, and use our drop in directory to have the `pulp` role drop in chunks?

pavanshekar: Just checking in on this when you have a moment. Happy to proceed based on your suggestion.

evgeni: Jinja has an [`include`](https://jinja.palletsprojects.com/en/stable/templates/#include) tag, so we could do something like:

```
# httpd/defaults/main.yml
httpd_enabled_pulp_snippets: []

# vhost.conf.j2:
{% for httpd_pulp_snippet in httpd_enabled_pulp_snippets %}
{% include httpd_pulp_snippet+'.j2' %}
{% endfor %}
```

and then feed `httpd_enabled_pulp_snippets` with the right list of enabled pulp plugins at the `foremanctl` level (and of course provide those snippets)

pavanshekar: Thanks! Implemented using httpd_enabled_pulp_snippets with includes as suggested.

src/vars/base.yaml
```diff
@@ -25,17 +25,13 @@ httpd_server_ca_certificate: "{{ server_ca_certificate }}"
 httpd_client_ca_certificate: "{{ client_ca_certificate }}"
 httpd_server_certificate: "{{ server_certificate }}"
 httpd_server_key: "{{ server_key }}"
+httpd_enabled_pulp_snippets: "{{ ['pypi'] if 'pulp_python' in pulp_enabled_plugins else [] }}"
```
ehelms: I think here we need to tie this into our features. For example, see https://github.com/theforeman/foremanctl/blob/06add8019025c2fbb5ff44dc4aaac75943d10d79/src/vars/flavors/katello.yml#L5

And you can look down at line 32 in this file to see it being referenced.

@evgeni -- did we have an intent to define "meta" features within the `features.yaml`? e.g. `content/*` features

pavanshekar: Updated to use `enabled_features` instead of `pulp_enabled_plugins` and added `content/python` to the katello flavor, following the same pattern as `content/ansible`.

ehelms:
Design looks good -- this could use some tests and consider if documentation would benefit from an update.

pavanshekar:
I added a test in tests/httpd_test.py::test_https_pypi_endpoint that verifies the Apache route proxies correctly to Pulp.

For documentation, since this follows the existing content/* pattern and katello.yml already shows what's included, I didn't add any. Happy to document something specific if you think it would be helpful!

tests/httpd_test.py
```diff
@@ -52,6 +52,13 @@ def test_https_pulp_auth(server, certificates, server_fqdn):
     assert cmd.succeeded
     assert cmd.stdout == '200'
 
+def test_https_pypi_endpoint(server, certificates, server_fqdn):
+    cmd = server.run(f"{CURL_CMD} --cacert {certificates['ca_certificate']} --write-out '%{{http_code}}' https://{server_fqdn}/pypi/")
```
ianballou:
```suggestion
    cmd = server.run(f"{CURL_CMD} --cacert {certificates['server_ca_certificate']} --write-out '%{{http_code}}' https://{server_fqdn}/pypi/")
```

I'm seeing server_ca_certificate used elsewhere in the  httpd_test.

pavanshekar: Updated to use server_ca_certificate.

tests/httpd_test.py
```diff
@@ -52,6 +52,13 @@ def test_https_pulp_auth(server, certificates, server_fqdn):
     assert cmd.succeeded
     assert cmd.stdout == '200'
 
+def test_https_pypi_endpoint(server, certificates, server_fqdn):
+    cmd = server.run(f"{CURL_CMD} --cacert {certificates['server_ca_certificate']} --write-out '%{{http_code}}' https://{server_fqdn}/pypi/")
+    assert cmd.succeeded
+    # Verify route proxies to Pulp: 404 from Pulp confirms correct configuration
+    # (unconfigured routes return 503)
+    assert cmd.stdout == '404'
```
evgeni: Rails also does return 404 here?
```
$ curl -k --silent --output /dev/null --write-out '%{http_code}' https://localhost/pypi/
404
$ curl -k --silent --output /dev/null --write-out '%{http_code}' https://localhost/pypifail/
404
```

src/vars/base.yaml
```diff
@@ -26,6 +26,7 @@ httpd_server_ca_certificate: "{{ server_ca_certificate }}"
 httpd_client_ca_certificate: "{{ client_ca_certificate }}"
 httpd_server_certificate: "{{ server_certificate }}"
 httpd_server_key: "{{ server_key }}"
+httpd_enabled_pulp_snippets: "{{ ['pypi'] if 'content/python' in enabled_features else [] }}"
```
evgeni:
```suggestion
httpd_enabled_pulp_snippets: "{{ ['pypi'] if 'pulp_python' in pulp_plugins else [] }}"
```

I wonder if it's easier to create an (empty) `pulp_NAME.j2` for every plugin and just make it:

```suggestion
httpd_enabled_pulp_snippets: "{{ pulp_plugins }}"
```

evgeni: (not blocking)

tests/httpd_test.py
evgeni: as an alternative, you could assert the content of the 404 message.

apache:
```
$ curl -k --silent https://localhost/pub/lol
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL was not found on this server.</p>
</body></html>
```

rails:
```
$ curl -k --silent https://localhost/nope
<!DOCTYPE html>
<html>
<head>
  <title>The page you were looking for doesn't exist (404)</title>
  <meta name="viewport" content="width=device-width,initial-scale=1">
  <style>
  body {
    background-color: #EFEFEF;
    color: #2E2F30;
    text-align: center;
    font-family: arial, sans-serif;
    margin: 0;
  }

  div.dialog {
    width: 95%;
    max-width: 33em;
    margin: 4em auto 0;
  }

  div.dialog > div {
    border: 1px solid #CCC;
    border-right-color: #999;
    border-left-color: #999;
    border-bottom-color: #BBB;
    border-top: #B00100 solid 4px;
    border-top-left-radius: 9px;
    border-top-right-radius: 9px;
    background-color: white;
    padding: 7px 12% 0;
    box-shadow: 0 3px 8px rgba(50, 50, 50, 0.17);
  }

  h1 {
    font-size: 100%;
    color: #730E15;
    line-height: 1.5em;
  }

  div.dialog > p {
    margin: 0 0 1em;
    padding: 1em;
    background-color: #F7F7F7;
    border: 1px solid #CCC;
    border-right-color: #999;
    border-left-color: #999;
    border-bottom-color: #999;
    border-bottom-left-radius: 4px;
    border-bottom-right-radius: 4px;
    border-top-color: #DADADA;
    color: #666;
    box-shadow: 0 3px 8px rgba(50, 50, 50, 0.17);
  }
  </style>
</head>

<body>
  <!-- This file lives in public/404.html -->
  <div class="dialog">
    <div>
      <h1>The page you were looking for doesn't exist.</h1>
      <p>You may have mistyped the address or the page may have moved.</p>
    </div>
    <p>If you are the application owner check the logs for more information.</p>
  </div>
</body>
</html>
```

pulp (django):
```
$ curl -k --silent https://localhost/pypi

<!doctype html>
<html lang="en">
<head>
  <title>Not Found</title>
</head>
<body>
  <h1>Not Found</h1><p>The requested resource was not found on this server.</p>
</body>
</html>
```

src/vars/base.yaml
evgeni: so far we have not yet

evgeni (approved):
none of the comments are blocking, so ack

tests/httpd_test.py
pavanshekar: Yes, this is better, thanks! I will update the test to check for Pulp's 404 message "The requested resource was not found on this server." to verify the route is proxying to Pulp instead of Rails.

evgeni: You could also ask for `/pypi/test/` which will return a 404 with the JSON:
```
{"detail":"No PythonDistribution found for base_path test"}
```
so you can assert `PythonDistribution` in there

src/vars/base.yaml
pavanshekar: Thanks for the suggestion! I've implemented the first approach.

tests/httpd_test.py
pavanshekar: I've implemented this.

evgeni (approved):

pavanshekar:
Rebased to resolve merge conflicts with master.

evgeni:
Ah yes, and now the newly added `ruff` check complains. Sorry.

pavanshekar:
> Ah yes, and now the newly added `ruff` check complains. Sorry.

Added required blank lines for ruff check.

ehelms (approved):

PR #460, opened 2026-04-21
Remove candlepin events service and ssl config

amolpati30:
Since the `candlepin_events` service has been removed [here](https://github.com/theforeman/puppet-katello/pull/525), I am removing it from foremanctl as well, as it is no longer required.


ehelms:
Thanks @amolpati30 -- this is a duplicate of https://github.com/theforeman/foremanctl/pull/391

PR #462, opened 2026-04-21
Add custom server certificate support

ehelms:

Allow users to provide their own server certificates via `--certificates-server-certificate`, `--certificates-server-key`, and `--certificates-server-ca-certificate` flags on foremanctl deploy. The user must then specify the use of custom server certificates with `--certificate-source=custom_server`. Custom certificates are copied into the canonical /root/certificates/ structure, while client certificates and localhost certificates continue to be managed by the internal CA. The custom certificates do not need to persist on disk after the first time they are deployed.

* Adds new user parameters for providing server certificate, key and CA
* Includes custom certificates in the test matrix

Steps to reproduce:

* Perform a standard setup for testing and stop before `foremanctl deploy`
* Run `./forge custom-certs` to generate custom certs on the host
* Deploy with custom certs: `./foremanctl --certificate-source=custom_server --certificate-server-certificate /root/custom-certificates/certs/quadlet.example.com.crt --certificate-server-key /root/custom-certificates/private/quadlet.example.com.key --certificate-server-ca-certificate /root/custom-certificates/certs/server-ca.crt`

#### Checklist


tests/conftest.py
```diff
@@ -16,6 +16,7 @@
 def pytest_addoption(parser):
     parser.addoption("--certificate-source", action="store", default="default", choices=('default', 'installer'), help="Where to obtain certificates from")
     parser.addoption("--database-mode", action="store", default="internal", choices=('internal', 'external'), help="Whether the database is internal or external")
+    parser.addoption("--certificate-source", action="store", default="default", choices=('default', 'installer', 'custom'), help="Certificate source used during deployment")
```
evgeni: you can't add the same option twice ;)


```suggestion
    parser.addoption("--certificate-source", action="store", default="default", choices=('default', 'installer', 'custom'), help="Certificate source used during deployment")
    parser.addoption("--database-mode", action="store", default="internal", choices=('internal', 'external'), help="Whether the database is internal or external")
```

.github/workflows/test.yml
```diff
@@ -61,6 +61,10 @@ jobs:
             security: none
             database: external
             box: centos/stream9
+          - certificate_source: custom
```
evgeni: `certificate_source: custom` is unknown in https://github.com/theforeman/foremanctl/blob/72b80524e8226affaa1f2a8c9772cc2a81b41bb5/src/playbooks/_certificate_source/metadata.obsah.yaml#L3-L7 and also there is no `custom_certificates.yaml` in vars (this was not needed in #421 as there you folded all vars into one file again, but here it's a required intermediate step)

docs/user/certificates.md
```diff
@@ -51,6 +51,21 @@ After deployment, certificates are available at:
...
+```bash
+foremanctl deploy \
+  --tls-server-certificate /path/to/server.crt \
+  --tls-server-key /path/to/server.key \
+  --tls-server-ca-certificate /path/to/ca-bundle.crt
+```
+
+All three flags must be provided together. The custom server certificate, key, and CA are copied into `/root/certificates/` and used for all server-facing TLS. An internal CA is still generated (or preserved from a previous deploy) to manage client certificates and the localhost certificate.
+
+On subsequent deploys, the custom certificates persist — you only need to pass the flags again if you want to update them (e.g., for certificate rotation).
```
evgeni: Which meaning of persist is used here?

The parameter values are stored (persisted) in `params.yaml`, so the files need to persist on disk as otherwise things will go bad (the same as in the old installer).
And this also effectively means you don't *need* to pass them on update if the paths don't change.

(Yes, I am picking words, but I want to make sure the implementation matches the intent)

docs/user/certificates.md
```diff
@@ -73,9 +88,8 @@ On each run the role compares the Subject Alternative Names in the existing serv
 
 ### Current Limitations
 
-- Cannot provide custom certificate files during deployment
 - Fixed 20-year certificate validity period
-- Limited certificate customization options
+- Custom certificates only supported with `certificate_source: default`
```
evgeni: But you pass `custom` in CI

evgeni (changes requested):

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -29,6 +29,22 @@ variables:
     action: append_unique
     type: FQDN
     parameter: --certificate-cname
+  certificates_custom_server_certificate:
+    help: Path to a custom server certificate to use instead of the auto-generated one.
+    type: AbsolutePath
+    parameter: --tls-server-certificate
```
evgeni: why "tls"? i'd drop that part from the param name

evgeni:
@ehelms I tried to make minimal changes to the tests so they match the docs (no "custom" source, just "default" with additional params) and picked the one change I had in #421 for "don't re-gen the server cert", please have a look if you agree

docs/user/certificates.md
```diff
@@ -51,6 +51,21 @@ After deployment, certificates are available at:
 - Server Certificate: `/root/ssl-build/<hostname>/<hostname>-apache.crt`
 - Client Certificate: `/root/ssl-build/<hostname>/<hostname>-foreman-client.crt`
 
+#### Custom Server Certificates
+
+To use certificates signed by your own CA instead of foremanctl's self-signed certificates:
```
ekohl: I'd like to avoid "own CA" (since it can be some third party company) but struggle to come up with good wording.

ekohl (commented):
Zooming out: what is the server CA certificate? I know we're in a messy place, but can we at least better define this?

Today we really mix a few concerns so I'm going to start with what certbot creates because I think it's a really great setup. For every certificate it creates files `/etc/letsencrypt/live/$domain`:
* `cert.pem` only the issued certificate
* `chain.pem` the intermediate certificates between the issued certificate and the root
* `fullchain.pem` the issued certificate AND the intermediate certificates (the same as `cert.pem` and `chain.pem` combined)
* `privkey.pem` the private key

Note it doesn't write the actual root CA anywhere and it assumes that's already in the trust stores somewhere.

You can configure Apache in 2 ways correctly serve to clients. What we do today:
```
SSLCertificateFile /etc/letsencrypt/live/$domain/cert.pem
SSLCertificateChainFile /etc/letsencrypt/live/$domain/chain.pem
SSLCertificateKeyFile /etc/letsencrypt/live/$domain/privkey.pem
```

But quoting https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcertificatechainfile
> SSLCertificateChainFile became obsolete with version 2.4.8, when [SSLCertificateFile](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcertificatefile) was extended to also load intermediate CA certificates from the server certificate file.

These days it's recommended to do:

```
SSLCertificateFile /etc/letsencrypt/live/$domain/fullchain.pem
SSLCertificateKeyFile /etc/letsencrypt/live/$domain/privkey.pem
```

Note that if you use multiple algorithms (typically RSA, DSA, ECC, but with PQC also ML-DSA) you need to repeat the `SSLCertificateFile` statement. How it exactly matches certs and keys is unclear to me. This also implies the foremanctl `--tls-server-certificate` parameters needs to be repeatable.

In Foreman we also need the actual root CA because we want to distribute that to clients. It's also used to only allow certificates signed by a specific CA to do client authentication. If you use a third party CA that is already well known (like Let's Encrypt) then is there really a point to specifying it? Right now our tooling expects it so I can accept we'll do that for now.

As we're moving to a whole new installer, I think this is the moment to properly define the inputs and what we expect. As we're also looking at PQC then I think the `fullchain.pem` approach is really the only sane approach but not 100% sure all of our tooling supports that today.

docs/user/certificates.md
evgeni: Just ditch the "own"? "To use certificates signed by a CA instead of self-signed certificates"

evgeni:
We only use the server_ca for Apache as a "server" thing, so that part should work.
We also pass it to Foreman Proxy, and Hammer to validate the connection to the server, but there it's already used as a "CA file" which can contain multiple certs, and one more cert in that file won't hurt.

src/roles/certificates/tasks/ca.yml
```diff
@@ -67,3 +67,19 @@
     privatekey_passphrase: "{{ certificates_ca_password }}"
     provider: selfsigned
     selfsigned_not_after: "+7300d"
+
+- name: 'Copy CA as server CA certificate'
+  ansible.builtin.copy:
+    src: "{{ certificates_ca_directory_certs }}/ca.crt"
+    dest: "{{ certificates_ca_directory_certs }}/server-ca.crt"
+    remote_src: true
+    force: false
```
shubhamsg199, line 76: I wonder if,  with `force: false` here, if a user previously deployed with custom certs and then re-deploys without the custom cert flags (intending to revert to default), this task won't overwrite the old custom server-ca.crt. The server would end up with a server cert signed by the internal CA but server-ca.crt still pointing to the old custom CA. Is this scenario something we need to handle, or is reverting from custom to default an unsupported path?

evgeni: It is a supported path, as far as I know, but not a common one I think

evgeni: We probably could do something like `force: "{{ certificates_custom_server_certificate is not defined }}"` instead

shubhamsg199: Yes, this should work

src/playbooks/deploy/metadata.obsah.yaml
ehelms: I wanted to group certificate params through a common prefix. I will switch to `certificate`.

src/roles/certificates/defaults/main.yml
```diff
@@ -1,9 +1,11 @@
 ---
+certificate_source: default
```
evgeni: ansible lint wants this var to be `certificates_source`

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -33,18 +33,25 @@ variables:
...
     help: Path to the CA certificate that signed the custom server certificate.
     type: AbsolutePath
     parameter: --certificate-server-ca-certificate
+    persist: false
 
 constraints:
   required_together:
     - [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]
+  required_if:
+    - [certificate_source, custom_server, [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]]
```
evgeni: Isn't `certificate_source` persisted, so you'll end up having to always pass the custom certs for things not to break?

evgeni: (CI says "no", but I am confused)

ehelms: I didn't get a chance to write an update after I pushed. Here is the design approach I was going for:

After the first time custom certificates are supplied, they do not need to be supplied again and are not required to remain on disk. I am trying to solve the problem where users have to keep around and backup their certificates.

The design is that the first time `deploy` is ran with custom certificates, those are copied in place, and the `certificate_source` is persisted but not the paths to the input certificates.

evgeni: I like that design :)

ehelms:
> As we're moving to a whole new installer, I think this is the moment to properly define the inputs and what we expect. As we're also looking at PQC then I think the `fullchain.pem` approach is really the only sane approach but not 100% sure all of our tooling supports that today.

Yea, we'd have to audit and potentially update any component to work this same way that uses public facing certificates. And we'd be putting the onus on the user to concatenate their CA and server certificate together. They are capable of this, but we'd have to do error checking and what not to ensure the user does it correctly. I'm hesitant to go that route, at least right now. Especially with not having resolved our installer certificate strategy. I am hopeful https://github.com/theforeman/foremanctl/pull/421 is the right approach and that would allow us to then consider this route.

ekohl:
I'm wondering what we should do. I'd like to have a GA quality with Foreman 3.19, but we can "migrate" in Foreman 3.20. But at that point I'd really want to have finalized the meaning of parameters for at least some time.

Looking at https://community.theforeman.org/t/foreman-3-19-schedule-and-planning/45747 we have prep week next week so need to decide soon. Not saying we can't merge this in 3.19, but then we also need to plan follow up issues.

ehelms:
> I'm wondering what we should do. I'd like to have a GA quality with Foreman 3.19, but we can "migrate" in Foreman 3.20. But at that point I'd really want to have finalized the meaning of parameters for at least some time.

Does GA quality meaning supporting upgrades from release to release?

ekohl:
IMHO yes

docs/user/certificates.md
```diff
@@ -152,18 +197,19 @@ The `certificate_checks` role uses `foreman-certificate-check` binary to validat
...
 ```
 
-**OpenSSL Configuration:**
-- Custom configuration template supports SAN extensions
-- Multiple DNS entries supported: `subjectAltName = DNS:{{ certificates_hostname }}{% for cname in certificate_cname %},DNS:{{ cname }}{% endfor %}`
-- Uses OpenSSL's `req` and `ca` commands for generation and signing
+**Certificate Generation:**
+- Uses Ansible's `community.crypto` collection for all certificate operations
+- Subject Alternative Names (SANs) supported for multi-domain certificates
 - CNAMEs configured via `certificate_cname` variable (list of additional DNS names)
```
evgeni: SANs and CNAMEs are the same feature, no need to list them twice

tests/foreman_proxy_test.py
```diff
@@ -17,3 +17,15 @@ def test_foreman_proxy_service(server):
...
     foreman_proxy = server.addr('localhost')
     assert foreman_proxy.port(FOREMAN_PROXY_PORT).is_reachable
+
+def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn):
+    cmd = server.run(
+        f"curl --cacert {certificates['server_ca_certificate']} "
+        f"--cert {certificates['client_certificate']} "
+        f"--key {certificates['client_key']} "
+        f"--silent --output /dev/null --write-out '%{{http_code}}' "
+        f"https://{server_fqdn}/api/v2/ping"
```
evgeni: `api/v2/ping` doesn't require auth, so that doesn't seem like a good endpoint to test it.

src/playbooks/_certificate_source/metadata.obsah.yaml
```diff
@@ -1,7 +1,9 @@
...
 variables:
-  certificate_source:
-    help: Where certificates are coming from. Currently default Ansible role or the foreman-installer.
+  certificates_source:
+    help: Where certificates are coming from. Currently default Ansible role, the foreman-installer, or custom server certificates.
+    parameter: --certificate-source
     choices:
       - default
       - installer
+      - custom_server
```
ShimShtein, line 9: Do we really need this option explicitly? Can't we use the other two options, and if certificates_custom_server* options are specified, they will take precedence over the default paths specified by certificate-source?
This will also address my concern about combining installer client certs with custom server ones.

ShimShtein (commented):
I would prefer generating the bundle internally, and not exposing it to the user.
Something along the lines of this pseudocode:
```
ca_sources = ([client_ca_path] + [server_ca_path]).unique

ca_bundle = File.combine(ca_sources)
```
Then you will just need to set the `server_ca` parameter to whatever value you want - it can be internally generated, installer generated or custom.

docs/user/certificates.md
ehelms: Fair... which one should we lean into? I would guess SAN is more of an implementation detail and users will think in CNAMEs. I'll stick with that.

src/playbooks/_certificate_source/metadata.obsah.yaml
ehelms: I do, for this design to work where the user does not have to persist the certificates used as input on disk. We need some indicator of the mode they are operating in. I'll update the PR description as this design has deviated since I first opened it.

docs/user/certificates.md
evgeni: For now I'd go with CNAME as that's what we use in docs today, but we have a story to unify the naming to something that is disconnected from (partial) implementation details.

See my host header injection PR for another place where the naming gets muddy.

tests/foreman_proxy_test.py
```diff
@@ -17,3 +17,15 @@ def test_foreman_proxy_service(server):
...
     foreman_proxy = server.addr('localhost')
     assert foreman_proxy.port(FOREMAN_PROXY_PORT).is_reachable
+
+def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn):
+    cmd = server.run(
+        f"curl --cacert {certificates['client_ca_certificate']} "
+        f"--cert {certificates['client_certificate']} "
+        f"--key {certificates['client_key']} "
+        f"--silent --output /dev/null --write-out '%{{http_code}}' "
+        f"https://{server_fqdn}/api/v2/smart_proxies"
```
evgeni: This endpoint doesn't support proxy auth.

```suggestion
        "--data '{\"config_report\": {\"host\": \"test.example.com\", \"reported_at\": \"2010-11-19 02:39:04 UTC\"}}' --header 'Content-Type: application/json' "
        f"https://{server_fqdn}/api/v2/config_reports"
```

tests/foreman_proxy_test.py
```diff
@@ -17,3 +17,15 @@ def test_foreman_proxy_service(server):
...
+    )
+    assert cmd.succeeded
+    assert cmd.stdout == '200'
```
evgeni: if you use the command above, the code will be 201

evgeni:
Welp, this fails with `No smart proxy server found on ["quadlet.example.com"] and is not in trusted_hosts` as the proxy has no Report-enabled features :see_no_evil:

evgeni:
needs a rebase after I just merged #456

src/roles/certificates/tasks/ca.yml
```diff
@@ -67,3 +67,20 @@
...
+    mode: '0444'
+
+- name: 'Create CA bundle'
+  ansible.builtin.copy:
+    src: "{{ certificates_ca_directory_certs }}/ca.crt"
+    dest: "{{ certificates_ca_directory_certs }}/ca-bundle.crt"
+    remote_src: true
+    force: false
+    mode: '0444'
+>>>>>>> 5008544 (Add custom server certificate support)
```
evgeni: choo choo, the conflict train!

ehelms: :facepalm:

src/roles/certificates/tasks/issue.yml
```diff
@@ -1,35 +1,39 @@
 ---
-- name: 'Create server private key'
-  community.crypto.openssl_privatekey:
-    path: "{{ certificates_ca_directory_keys }}/{{ certificates_hostname }}.key"
-    type: "{{ certificates_algorithm_type }}"
-    size: "{{ certificates_algorithm_size }}"
-    mode: '0600'
+- name: Issue server certificate
+  when:
+    - (certificates_source != 'custom_server') or (certificates_hostname == 'localhost')
```
ShimShtein, line 4: It means we will issue a new key and server cert if the hostname is localhost regardless of the certificate source. Why would you like to do it in case the user had specified `certificate_source = custom_server and hostname=localhost`

ehelms: `localhost` is for internal communications: Candlepin and IoP. The user cannot supply these certificates. They can only provide custom server certificates for public facing interfaces.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -29,6 +29,29 @@ variables:
...
+    help: Path to the CA certificate that signed the custom server certificate.
+    type: AbsolutePath
+    parameter: --certificate-server-ca-certificate
+    persist: false
+
+constraints:
+  required_together:
+    - [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]
+  required_if:
+    - [certificates_source, custom_server, [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]]
```
evgeni: this seems not to work, as running `./foremanctl deploy --certificate-source custom_server` starts a deployment.

(also, it would explode on subsequent calls to deploy, where you don't pass the certs anymore)

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -29,6 +29,29 @@ variables:
...
+  mutually_exclusive_if:
+    - [certificates_source, installer, [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]]
```
evgeni: this too seems not to work, as `./foremanctl deploy --certificate-source installer --certificate-server-ca-certificate /dev/null` starts a deployment

evgeni: well, that explains, there is no `mutually_exclusive_if` my friend

src/playbooks/deploy/metadata.obsah.yaml
evgeni: https://github.com/theforeman/obsah/pull/116

ehelms: Good find, thanks!

src/playbooks/deploy/metadata.obsah.yaml
ehelms: Well... yea. I failed hard there. Fixed.

src/playbooks/deploy/metadata.obsah.yaml
evgeni: This PR got released as 1.8.1, mind bumping the minimum obsah version in this PR, so we ensure we run with the right constraints

evgeni: finally, its failing CI as predicted :D

ehelms: Except that it failed in the *wrong* way. I dropped this line for now. I think this could be cleaner for the user, but this feels like a fundamental problem.

I want to guard against **user** input. This is failing because it's reading from `parameters.yaml` and treating that as part of the constraints. The problem is: with this design `parameters.yaml` *should* have only this value. But if the user supplies `--certificate_source=custom_server` then they should be providing those other three values.

evgeni: Yeah, right now there is no way to differentiate between "this comes from a stored answer" vs "the user entered it".

docs/user/certificates.md
```diff
@@ -94,9 +129,10 @@ The `--certificate-renew` flag is **not persisted** in foremanctl’s answers fi
 
 ### Current Limitations
 
-- Cannot provide custom certificate files during deployment
 - Uses the same lifetime for both client and server certificates
 - Limited certificate customization options
+- Custom server certificates cannot be combined with `certificate_source: installer`
+- CNAMEs are only applied to certificates generated by the default CA
```
evgeni: In the "Custom Server Source" section you call it "internal CA", should we use the same term here?

src/roles/certificates/tasks/issue.yml
```diff
@@ -1,35 +1,39 @@
...
+        privatekey_path: "{{ certificates_ca_directory_keys }}/{{ certificates_hostname }}.key"
+        common_name: "{{ certificates_hostname }}"
+        subject_alt_name: "{{ _certificates_desired_server_sans }}"
+        key_usage:
+          - digitalSignature
+          - keyEncipherment
+        extended_key_usage:
+          - serverAuth
+      vars:
+        _certificates_desired_server_sans: "{{ (([certificates_hostname] + certificates_cnames) | map('regex_replace', '^', 'DNS:') | list) }}"
```
evgeni, line 25: unrelated to this PR, but this means we're adding the CNAMEs to *all* server certs we generate -- including `localhost`, which is wrong.

evgeni (approved):

src/roles/certificates/tasks/issue.yml
ehelms: Tracked: https://github.com/theforeman/foremanctl/issues/487

PR #463, opened 2026-04-22
validate certs when using the API

evgeni:

we should not skip certificate validation

pass our ca cert to ansible

Steps to reproduce:

* CI will validate it

#### Checklist

### DIFF
--- a/development/roles/foreman_development/tasks/smart-proxy/main.yml
+++ b/development/roles/foreman_development/tasks/smart-proxy/main.yml
@@ -114,8 +114,8 @@
 
 - name: Wait for smart-proxy development server to be accessible
   ansible.builtin.uri:
-    url: "http://{{ ansible_facts['fqdn'] }}:8000/features"
-    validate_certs: false
+    url: "https://{{ ansible_facts['fqdn'] }}:8443/features"
+    ca_path: "{{ foreman_development_ca_certificate }}"
   until: foreman_development_smart_proxy_status.status == 200
   retries: 30
   delay: 5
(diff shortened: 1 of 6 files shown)


src/roles/foreman_proxy/handlers/main.yml
```diff
@@ -10,4 +10,4 @@
     server_url: "{{ foreman_url }}"
     oauth1_consumer_key: "{{ foreman_oauth_consumer_key }}"
     oauth1_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
-    validate_certs: false
+    ca_path: "{{ foreman_ca_certificate }}"
```
ehelms: I guess we kinda broke our best practice in the `foreman_proxy` role with this variable and the oauth variables on lines 11-12.

evgeni: psssssssssssssst! ;)

evgeni: the same is also true for foreman_url ;)

want me to add `foreman_proxy_*: "{{ foreman_* }}"` vars?

evgeni: https://github.com/theforeman/foremanctl/pull/470

development/roles/foreman_development/tasks/smart-proxy/main.yml
```diff
@@ -115,7 +115,6 @@
 - name: Wait for smart-proxy development server to be accessible
   ansible.builtin.uri:
     url: "http://{{ ansible_facts['fqdn'] }}:8000/features"
-    validate_certs: false
```
ShimShtein, line 118: Shouldn't we switch to the `https` endpoint for the smart proxy? Then we will need to add the `ca_path` as in other places.
While technically correct, I think we added a couple of features to the proxy with `:enabled: https`.

evgeni: let's see if it's as easy as changing the port

ehelms (approved):

ShimShtein (approved):

PR #457, opened 2026-04-20
POC: Files for Agentic AI development

stejskalleos:
:warning: **Disclaimer**

* Proof of concept, I have not tested it (yet) for actual development
* Generated with multiple AI tools (Claude, Opus 4.6 & Gemini)
* The files & instructions should be model/tool agnostic

#### Why are you introducing these changes? (Problem description, related links)

This PR is my test & POC for Agentic AI development.
I'd like to start this effort & discussion as part of the [Foreman organization effort](https://community.theforeman.org/t/rfc-agentic-workflow-practices-for-the-community/46232)

#### What are the changes introduced in this pull request?

`CONTRIBUTING_AI.md` file with supporting files for:

- Agents
- Commands
- Rules
- Skills

#### Comments

Please take a look and share your thoughts, opinions, and comments. This PR is by no means complete; take it as a starting point for AI development efforts.

PR #469, opened 2026-04-23
Initialize files & dirs for Agentic AI development

stejskalleos:
- AGENTS.md
- .agents dir with agents, skills, rules
- settings file for Claude
- links for Claude files

Follow up to https://github.com/theforeman/foremanctl/pull/457, where I introduced quite a big change for an initial commit. This is a reduced version that should build the foundation for Agentic AI development.

- AGENTS.md
- .agents dir with agents, skills, rules
- settings file for Claude
- links for Claude files

Use your AI tool and observe behavior and functionality.

#### ⚠️ Disclaimer

* Proof of concept, I have not tested it (yet) for actual development
* Generated with multiple AI tools (Claude, Opus 4.6 & Gemini)
* The files & instructions should be model/tool agnostic


.agents/rules/podman-secrets-rule.md
```diff
@@ -0,0 +1,72 @@
+---
```
ehelms: I am wondering if we need this rule or can point to, and document this information in our existing docs. This ends up making it very AI-centric rather than human and AI centric. Whereas https://github.com/theforeman/foremanctl/blob/master/DEVELOPMENT.md#service-configuration has some information and could be enhanced with what you have here.

I don't know how well AI will treat the information in `DEVELOPMENT.md` compared to a "rule" but I think we should start with the human / context centric approach until we find it's not working.

.agents/settings.json.EXAMPLE
```diff
@@ -0,0 +1,14 @@
+{
```
ehelms, line 1: This is an area I have been unsure about. I lean towards starting with `settings.json.example` and having users copy this to `settings.local.json` with this file in gitignore. While convenient, I don't think we want to force permissions out of the box on users.

.claude
```diff
@@ -0,0 +1 @@
+.agents
```
ehelms, line 1: Should this be a symlink?

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+foremanctl is a deployment tool for Foreman and Katello using Podman quadlets and Ansible. It provides a command-line interface built on top of `obsah` (an Ansible wrapper) for managing containerized Foreman deployments.
+
+The repository contains two main wrapper scripts:
+
+- `foremanctl` - Production deployment and management
+- `forge` - Development environment management and testing
+
+Both scripts wrap `obsah`, which provides a structured way to run Ansible playbooks with parameter validation and persistence.
+
+## Supported Platforms
```
ehelms: Here we are repeating ourselves, and I'm not sure this will be hyper relevant for AI tools.

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+  agents/          # Specialist agent personas with defined scopes
+  rules/           # Coding standards and constraints (all agents must follow)
+  skills/          # Domain knowledge references
+```
+
+### Agents (`.agents/agents/`)
+
+Focused personas with defined scopes and technology boundaries:
+
+| Agent | Purpose |
```
ehelms: This list will need updating as you scoped this down.

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+|-------|---------|
+| `ansible-playbook-agent` | Playbooks and `metadata.obsah.yaml` files |
+| `ansible-role-agent` | Roles, tasks, handlers, templates |
+| `feature-agent` | End-to-end feature addition workflow |
+| `test-agent` | pytest/testinfra tests |
+| `lint-agent` | ansible-lint and Python linting |
+| `podman-agent` | Podman quadlets, containers, secrets |
+| `code-review-agent` | Read-only code review and analysis |
+
+### Commands (`.agents/commands/`)
```
ehelms: Can be dropped

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+
+Three inventory configurations:
+
+- `localhost` - Local deployment
+- `local_vagrant` - Vagrant-managed local VMs
+- `broker.py` - Dynamic inventory script
+
+## Developer Documentation
+
+Additional developer guides (from PR [#435](https://github.com/theforeman/foremanctl/pull/435)):
```
ehelms: I think the PR is irrelevant.

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+
+- [How to Add a Feature](docs/developer/how-to-add-a-feature.md) -- end-to-end feature development
```
ehelms: I like this section being concise and to the point.

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+Tests use pytest with testinfra for infrastructure testing. Key fixtures in `tests/conftest.py`:
+
+- `server_hostname`, `server_fqdn` - Default to `quadlet` / `quadlet.example.com`
+- `client_hostname`, `client_fqdn` - Default to `client` / `client.example.com`
+- `certificates` - Certificate configuration based on `--certificate-source`
+- SSH configuration generated from Ansible inventory at `.tmp/ssh-config`
+
+Test files follow the pattern `tests/<component>_test.py` and use testinfra's server fixture to execute commands and check system state.
+
+## Build System
```
ehelms: This feels like a less used detail that maybe shouldn't eat context in our entrypoint.

ehelms:
Great start, I've been looking at https://github.com/NVIDIA/OpenShell/blob/main/AGENTS.md (and the repo) as good examples of well written and structured assets.

.claude
stejskalleos: It is a symlink:
```
lrwxrwxrwx. 1 lstejska lstejska 1M Apr 24 10:30 .claude -> .agents
```

stejskalleos:
@ehelms applied changes as per your comments

stejskalleos:
@ehelms I'm thinking about adding a [Caveman skill](https://github.com/JuliusBrussee/caveman/blob/main/README.md), what do you think?

> A [Claude Code](https://docs.anthropic.com/en/docs/claude-code) skill/plugin and Codex plugin that makes agent talk like caveman — cutting ~75% of output tokens while keeping full technical accuracy. Now

ehelms:
> @ehelms I'm thinking about adding a [Caveman skill](https://github.com/JuliusBrussee/caveman/blob/main/README.md), what do you think?

In my experience, most of my time is reading the final outputs of agents. Having to do mental gymnastics to understand the output does not seem worth it at this point.

ehelms:
Is `.agents/agents/ansible-playbook-agent.md` an empty file?

CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
ehelms, line 1: Should this be a symlink?

.agents/skills/lint-skill.md
```diff
@@ -0,0 +1,91 @@
...
+- `development` -> lint only the development codebase
+
+## Workflow
+
+### 1. Ansible Lint
+
+ansible-lint must be run from within each data directory separately, matching CI behavior:
+
+```bash
+cd src; ansible-lint
```
ehelms: If the user was not in the virtualenv, or does not have `ansible-lint`, then it will attempt a convoluted process to try and find it or install it. 

I think we should instruct the skill to use it from the virtual env `.venv` if it exists, use it locally, or halt if it cannot find the executable and tell the user to make it available.

e.g. `cd src; ../.venv/bin/ansible-lint`

.agents/skills/obsah-metadata-skill.md
```diff
@@ -0,0 +1,196 @@
+---
+name: obsah-metadata
```
evgeni: having it here in such a verbosity felt wrong, so I went ahead and updated the obsah docs in https://github.com/theforeman/obsah/pull/115

ekohl: Which is published to https://obsah.readthedocs.io/en/latest/. Perhaps the development page should be split and a `using.rst` is better for people / agents that only want to consume obsah.

.agents/skills/lint/SKILL.md
```diff
@@ -0,0 +1,91 @@
+---
```
ehelms: I think the file should be:

```
.agents/skills/lint/SKILL.md
```

AGENTS.md
```diff
@@ -0,0 +1,216 @@
...
+# Start development VM
+./forge vms start
+
+# Deploy Foreman in development mode
+./foremanctl deploy --foreman-initial-admin-password=changeme --tuning development
+
+# Stop development VM
+./forge vms stop
+```
+
```
arvind4501, line 55:
```suggestion
See [development guide](DEVELOPMENT.md) for the full development setup workflow.
```

AGENTS.md
```diff
@@ -0,0 +1,216 @@
...
+### Custom Plugins
+
+- `src/callback_plugins/foremanctl.py` - Ansible callback plugin for foremanctl-specific output
+- `src/filter_plugins/foremanctl.py` - Custom Jinja2 filters
+
+### Features and Flavors
+
+Features are modular capabilities that can be enabled. See [docs/developer/how-to-add-a-feature.md](docs/developer/how-to-add-a-feature.md) for the complete feature development guide.
+
+Current features:
```
arvind4501: Features listing is something that will grow, i wonder if we should keep this updating or update it to contain `Example features:`

.agents/skills/obsah-metadata-skill.md
```diff
@@ -0,0 +1,196 @@
...
+description: >-
+  Deep reference on the metadata.obsah.yaml schema used to define CLI
+  subcommands in foremanctl and forge. Covers help, variables, constraints,
+  includes, and all variable properties.
+technologies:
+  - yaml
+  - ansible
+references:
+  - docs/developer/playbooks-and-roles.md
+  - https://github.com/theforeman/obsah/blob/master/docs/source/development.rst
```
ekohl: Should https://obsah.readthedocs.io/en/latest/ be added or replace this entry?

stejskalleos:
Updated:

- Fixed .agents/agents files
- Simplified podman-secrets-rule
- Restructured skills
- Applied the suggestions from the comments

.agents/agents/ansible-role-agent.md
```diff
@@ -0,0 +1,70 @@
...
+Follow the instructions in the `docs/developer/playbooks-and-roles.md` document.
+
+## Workflow
+
+1. **Identify the role** -- determine which existing role to modify, or whether a new role is needed.
+2. **Follow conventions** -- use snake_case naming, `.yaml` extensions, standard directory layout.
+3. **Write idempotent tasks** -- all tasks must be safe to run multiple times.
+4. **Use handlers** -- notify handlers for service restarts rather than inline restarts.
+5. **Template with Jinja2** -- use `.j2` templates for configuration files, reference variables from the vars system.
+6. **Lint** -- run `cd src; ansible-lint` or `cd development; ansible-lint`.
```
ehelms: Should this tell the agent to use the lint skill?

.agents/agents/ansible-role-agent.md
```diff
@@ -0,0 +1,70 @@
...
+    <static files>
+  vars/
+    main.yaml
+```
+
+## How to
+
+Follow the instructions in the `docs/developer/playbooks-and-roles.md` document.
+
+## Workflow
```
ehelms: 2,3,4,5 are all guidelines on our design / style. I feel like we should extract those. Kind of like these two sections (https://github.com/theforeman/foremanctl/blob/master/DEVELOPMENT.md#service-configuration) where we are laying down the conventions for our project.

.agents/rules/podman-secrets-rule.md
```diff
@@ -0,0 +1,23 @@
...
+name: podman-secrets
```
ehelms: I still struggle with this being a rule. I am wondering if we should focus on just encoding the podman secrets context in it's current location and see how well the agents do. And then if they are struggling we come back and make a rule.

AGENTS.md
```diff
@@ -0,0 +1,213 @@
...
+
+Flavors define the base feature set:
+
+- `katello` - Currently the only flavor (Foreman 3.18 + Katello 4.20 + Pulp 3.85 + Candlepin 4.6)
+
+The `enabled_features` variable is computed as: `flavor_features + features`
+
+### Deployment Variables
+
+Key deployment parameters:
```
ehelms: I'd drop this section. I am not sure what makes any of these "key" to the agent or human since they are situational.

AGENTS.md
```diff
@@ -0,0 +1,213 @@
...
+Configuration is managed through:
+
+1. **Vars files** (`src/vars/`):
+   - `defaults.yml` - Base defaults
+   - `flavors/` - Deployment flavors (currently only `katello.yml`)
+   - `tuning/` - Resource profiles (default, development, medium, large, extra-large, extra-extra-large)
+   - `images.yml` - Container image references
+   - `database.yml`, `foreman.yml` - Service-specific configuration
+
+2. **Podman secrets**: Configuration files and credentials are stored as Podman secrets following naming conventions:
```
ehelms: I think we can point at the file that contains this information.

AGENTS.md
```diff
@@ -0,0 +1,213 @@
...
+- `smoker/` - Smoker test execution
+- `deploy-dev/` - Development deployment variations
+- `security/`, `sos/`, `lock/` - Utilities
+
+### Roles
+
+Roles in `src/roles/` correspond to services and deployment stages:
+
+- Service roles: `foreman`, `pulp`, `candlepin`, `postgresql`, `redis`, `httpd`
+- Feature roles: `hammer`, `foreman_proxy`
```
ehelms, line 127: This is debateable -- since foreman-proxy is a service.

.agents/skills/obsah-metadata/SKILL.md
```diff
@@ -0,0 +1,21 @@
+---
```
ehelms: This is another one I'd vote for just being in our documentation and ensuring the other docs point to it where necessary.

ehelms:
I took this for a test drive implementing something. Here are my recommendations after that experience:

1. Split out the foundational parts to a PR -- i.e. AGENTS.md
2. The two agents are presenting more like skills rather than agents. Agents should be able to perform some autonomous action and our repository design means that splitting up how roles and playbooks interact is not a clean cut. There are often features and changes that need to touch the whole of `src/`. However, encapsulating how roles work, playbooks work into skills that can be used is helpful to the built in agents or future agents we might build ourselves. These skills would allow builtin agents (such as Claude's Explore and Plan agents) to use them.
 3. Let's also get in the lint skill and try it out.

ehelms:
Follow up, I tried implementing a feature with this PR + turning those two agents into skills and had good results. I then tried to follow up with a "testing agent" which focused on running a deploy, running tests, addressing failures. Here is the first attempt at that agent I used: https://github.com/theforeman/foremanctl/pull/481

.agents/agents/ansible-role-agent.md
stejskalleos: Good point, I'm thinking of reducing the workflow to these three steps:

1. **Identify the role** -- determine which existing role to modify, or whether a new role is needed.
2. **Write idempotent tasks** -- all tasks must be safe to run multiple times.
3. **Lint** -- Use Lint skill defined in `.agents/skills/lint/SKILL.md`

WDYT?

.agents/rules/podman-secrets-rule.md
stejskalleos: So should I remove the file?

AGENTS.md
```diff
@@ -0,0 +1,193 @@
...
+- `hammer` - CLI tool
+- `foreman-proxy` - Smart proxy functionality
+- `remote-execution` - Remote execution via SSH
+- `google` - Google Compute Engine integration
+- `azure-rm` - Azure Resource Manager integration
+- `katello` - Content and subscription management
+
+Flavors define the base feature set:
+
+- `katello` - Currently the only flavor (Foreman 3.18 + Katello 4.20 + Pulp 3.85 + Candlepin 4.6)
```
ekohl: A flavor doesn't define the versions and we've updated to always follow nightly. Perhaps leave out the versions?
```suggestion
- `katello` - Currently the only flavor (Foreman + Katello + Pulp + Candlepin)
```

Instead of listing them, should it perhaps be instructed to read https://github.com/theforeman/foremanctl/tree/master/src/vars/flavors?

If you need a description, should we add that to a flavor? I can imagine we'll have `foremanctl list-flavors` or `foremanctl flavor --list` or some variant like that which should tell users what to select.

AGENTS.md
ekohl: Can it be an instruction to read https://github.com/theforeman/foremanctl/blob/master/src/features.yaml? Because the list is already out of date.

ehelms (approved):
I think this is a good start, and the only way for this area to get better is to use it and iterate.

PR #470, opened 2026-04-23
use role-specific vars for url and oauth in foreman_proxy and hammer

evgeni:

we forgot to use role-specific vars for accessing foreman inside the hammer and foreman_proxy roles

* introduce and use those vars

CI

#### Checklist

### DIFF
--- a/src/vars/base.yaml
+++ b/src/vars/base.yaml
@@ -32,5 +32,9 @@ pulp_plugins: "{{ enabled_features | select('contains', 'content/') | map('repla
 
 hammer_ca_certificate: "{{ server_ca_certificate }}"
 hammer_plugins: "{{ enabled_features | features_to_hammer_plugins }}"
+hammer_foreman_server_url: "{{ foreman_url }}"
 
 foreman_proxy_plugins: "{{ enabled_features | features_to_foreman_proxy_plugins }}"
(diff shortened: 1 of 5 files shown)


src/vars/base.yaml
```diff
@@ -32,5 +32,9 @@ pulp_plugins: "{{ enabled_features | select('contains', 'content/') | map('repla
 
 hammer_ca_certificate: "{{ server_ca_certificate }}"
 hammer_plugins: "{{ enabled_features | features_to_hammer_plugins }}"
+hammer_foreman_server_url: "{{ foreman_url }}"
 
 foreman_proxy_plugins: "{{ enabled_features | features_to_foreman_proxy_plugins }}"
+foreman_proxy_foreman_server_url: "{{ foreman_url }}"
```
arvind4501, line 38: I wonder why we need this defined twice, isn't `/foreman-proxy/defaults/main.yaml` works? or do we use it outside of that role too?

evgeni: the one in defaults uses the fqdn, which is a "good guess" for the role standalone, but the idea is that foremanctl is able to override that globally and that happens in the vars file here

arvind4501 (approved):

PR #475, opened 2026-04-27
Swap development to use oauth to register smart_proxy

ehelms:

The production deployment swapped to registering smart-proxies with oauth. This moves development to follow the same model.

* smart-proxy registration in development deployment will now use oauth

Steps to reproduce:

* Deploy a foremanctl development environment
* No errors

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -174,7 +174,7 @@ jobs:
           ./forge setup-repositories
       - name: Run deployment
         run: |
-          ./forge deploy-dev --foreman-development-enabled-plugin foreman_ansible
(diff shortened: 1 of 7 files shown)


jeremylenz (approved):
No idea what the test failures mean, but this seems to execute just fine in my environment

evgeni:
```
Uncaught Error: There is already an interval running and registered for: SUBSCRIPTIONS_TASK_SEARCH.
```

This is an UI error that we've seen every now and then (also in normal pipelines, without containers), no idea what happens there tho.

evgeni (approved):

PR #476, opened 2026-04-27
Move default certificate path to /var/lib/foremanctl/certs

ehelms:

Certificates that are generated and managed by `foremanctl` should live in an appropriate location on the file system that indicates to the admin. In `foreman-installer`, these certificates lived in `/root/ssl-build` which was always seen as a non-standard location. These certificates should not be touched by the user, and thus `/var` is the best location to store and indicate this to the user.

* Moves certificate generation and management from `/root/certificates` to `/var/lib/foremanctl/certs`.

Steps to reproduce:

* `./foremanctl deploy`
* Check that certificates are present in `/var/lib/foremanctl/certs`

#### Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -48,7 +48,7 @@ def certificates(pytestconfig, server_fqdn):
     source = pytestconfig.getoption("certificate_source")
     env = Environment(loader=FileSystemLoader("."), autoescape=select_autoescape())
     template = env.get_template(f"./src/vars/{source}_certificates.yml")
-    context = {'certificates_ca_directory': '/root/certificates',
+    context = {'certificates_ca_directory': '/var/lib/foremanctl/certs',
                'ansible_facts': {'fqdn': server_fqdn}}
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -48,7 +48,7 @@ def certificates(pytestconfig, server_fqdn):
     source = pytestconfig.getoption("certificate_source")
     env = Environment(loader=FileSystemLoader("."), autoescape=select_autoescape())
     template = env.get_template(f"./src/vars/{source}_certificates.yml")
-    context = {'certificates_ca_directory': '/root/certificates',
+    context = {'certificates_ca_directory': '/var/lib/foremanctl/certs',
```
evgeni, line 53: this is not needed after e64547966d17a1eff67e539ada0c0e7dfb60aacf

evgeni:
this needs a rebase

evgeni (approved):

PR #477, opened 2026-04-28
Update smart-proxy name configuration for development

pondrejk:

This change makes "internal" capsule naming consistent with downstream which in turn makes robottelo easier to use against dev environment

* proxy name from hostname-dev to hostname

### DIFF
--- a/development/roles/foreman_development/tasks/smart-proxy/main.yml
+++ b/development/roles/foreman_development/tasks/smart-proxy/main.yml
@@ -138,7 +138,7 @@
 
 - name: Configure smart-proxy for development
   theforeman.foreman.smart_proxy:
-    name: "{{ ansible_facts['fqdn'] }}-dev"
+    name: "{{ ansible_facts['fqdn'] }}"
     url: "https://{{ ansible_facts['fqdn'] }}:8443"
     server_url: "{{ foreman_development_url }}"
     username: "{{ foreman_development_admin_user }}"


ehelms:
> This change makes "internal" capsule naming consistent with downstream which in turn makes robottelo easier to use against dev environment

You got me curious. How is this used inside robottelo? What tests / workflows?

pondrejk:
> You got me curious. How is this used inside robottelo? What tests / workflows?

I only started exploring this, but the idea is to use the setup fixtures from robottelo to create reproducers on devel setup that would be long/annoying to prepare by hand (things like external auth, living hosts registered through external capsules, provisioned hosts, things like that)

ehelms:
> I only started exploring this, but the idea is to use the setup fixtures from robottelo to create reproducers on devel setup that would be long/annoying to prepare by hand (things like external auth, living hosts registered through external capsules, provisioned hosts, things like that)

Fair. I am curious then, how is the smart-proxy / foreman-proxy / capsule referenced inside robottelo? How is it being used in this fixture creation?

pondrejk:
> Fair. I am curious then, how is the smart-proxy / foreman-proxy / capsule referenced inside robottelo? How is it being used in this fixture creation?

https://github.com/SatelliteQE/robottelo/blob/6f474e655f5648feaf3a7c8b2befa8c49b6e9c8d/pytest_fixtures/component/smartproxy.py#L7 
Basically the expectation is for default proxy to be the same hostname as Satellite, current workaround for me is to add `-dev` at line 10

ehelms:
I think this identifies that, just as we are breaking our assumptions in our deployment, that the testing needs to make less assumptions and lean into the features. In this case, instead of assuming a default proxy and that it will have certain capabilities, we should be leaning into finding a proxy with the features that is needed for the test. We might can get away with ensuring the `url` has the same hostname to find the "local" foreman-proxy. Relying on the `name` however, just won't cut it anymore.

ehelms:
@pondrejk Any updates? Do you want to continue persuing this?

pondrejk:
> @pondrejk Any updates? Do you want to continue persuing this?

Hey, reading back what you wrote I'm not sure what to make of it. For the way things are now, the small change in this pr makes sense to me for a little bit of convenience. 
When it comes to how robottelo finds the default capsule, I'm sure it needs to adjust to what the fmctl default will be (if any) but that something for a design discussion with wider team, bit out of my scope in this PR

ehelms:
As there does not appear to be an immediate need to solve this in this way, and as discussed, there is a better design principle that should be followed I am choosing to close this.

PR #479, opened 2026-04-28
Restart PostgreSQL when quadlet env vars change

shubhamsg199:
The Deploy PostgreSQL container task updates the quadlet file with tuning-specific env vars (POSTGRESQL_MAX_CONNECTIONS, POSTGRESQL_SHARED_BUFFERS, POSTGRESQL_EFFECTIVE_CACHE_SIZE) but does not notify the Restart postgresql handler. As a result, switching tuning profiles (e.g. medium to default) updates the quadlet file and triggers daemon_reload, but PostgreSQL keeps running with stale env vars from the previous profile.

* Adding `notify: Restart postgresql` to the container task ensures the handler fires whenever the quadlet content changes, restarting the service with the updated env vars.

Steps to reproduce:

* Run `foremanctl deploy --tuning medium`
* Check the postgresql tuning values
* Run  `foremanctl deploy --reset-tuning`
* Check the postgresql tuning values again
* The tuning values will be for the medium profile

#### Checklist

### DIFF
--- a/src/roles/postgresql/tasks/main.yml
+++ b/src/roles/postgresql/tasks/main.yml
@@ -43,6 +43,8 @@
         WantedBy=default.target foreman.target
         [Unit]
         PartOf=foreman.target
+  notify:
+    - Restart postgresql
 
 - name: Configure SSL certificates
   when:


ehelms:
This reads like we would benefit from a test that switches tuning profiles and checks that the values for said tuning profile are updated within the running container(s) that it applies to.

evgeni:
> This reads like we would benefit from a test that switches tuning profiles and checks that the values for said tuning profile are updated within the running container(s) that it applies to.

That's what https://github.com/SatelliteQE/robottelo/pull/21357 is about where this was found, yeah.

I am not sure we want to test every permutation of parameters in our primary CI, as that will be expensive (time wise)

evgeni (approved):
Tested with `foremanctl-2.0.0-1.20260428184728543158.pr479.12.ge60ca68.el9.noarch`, PostgreSQL is correctly restarted and the new `max_connections` value is used.

PR #484, opened 2026-05-04
Add BMC feature to foreman-proxy

arvind4501:
Enable BMC feature for proxy

* BMC feature for foreman-proxy
* Default bmc provider ipmi

foremanctl deploy --add-feature bmc

Steps to reproduce:

*

#### Checklist


arvind4501:
To test this with foremanctl we need https://github.com/theforeman/foreman-packaging/pull/13417

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -29,6 +29,15 @@ variables:
     action: append_unique
     type: FQDN
     parameter: --certificate-cname
+  foreman_proxy_bmc_default_provider:
+    help: Default BMC provider.
+    choices:
+      - freeipmi
+      - ipmitool
+      - redfish
```
evgeni: While the parameter is called "default provider", the overall behavior is a bit different.

The smart proxy code takes the provider from the request, and falls back to the default if unset (honoring the "default" definition):
https://github.com/theforeman/smart-proxy/blob/a5780bc6e0132d7e91fa3074845b2a3d1cc30f01/modules/bmc/bmc_api.rb#L392-L393

But then Foreman comes (who is the only consumer of this API) and *always* sets the provider *unless* the selected provider in Foreman is "IPMI":
https://github.com/theforeman/foreman/blob/f2d892f7585c61d71f671298c8158712daace8e8/app/models/nic/bmc.rb#L27

Which basically makes the setting "used IPMI implementation" (as one can't pick one in Foreman), not "fallback provider".

To add insult to injury, everyone disagrees what a provider is. Foreman defines [IPMI, Redfish and SSH](https://github.com/theforeman/foreman/blob/f2d892f7585c61d71f671298c8158712daace8e8/app/models/nic/bmc.rb#L3), but the Proxy knows about [freeipmi, ipmitool, redfish, shell and ssh](https://github.com/theforeman/smart-proxy/blob/a5780bc6e0132d7e91fa3074845b2a3d1cc30f01/modules/bmc/bmc_api.rb#L402-L443) 🤯 

I don't think we can (or even should) fix all of that mess here and now, but maybe calling this `ipmi_implementation` instead of `default_provider` and dropping `redfish` would slightly reduce the user confusion?

👀 @stejskalleos 

PS: I couldn't find any users of the "Shell" provider, which is basically giving power control over the machine running the smart proxy. First I thought that'd be used in Discovery, but there we have an own implementation. It was added in https://projects.theforeman.org/issues/2387 but the explanation is rather… slim

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -29,6 +29,15 @@ variables:
...
+  foreman_proxy_bmc_redfish_verify_ssl:
```
evgeni, line 55: Should we expose the "foreman_proxy" part to the user? Or should this be `--bmc-…` given the feature is "bmc"?

evgeni (commented):
needs a rebase please

src/playbooks/deploy/metadata.obsah.yaml
arvind4501: i think it should be `--bmc-*`, as we have done with foreman related params such as `--initial-organization` and `--initial-location`, will update it

src/playbooks/deploy/metadata.obsah.yaml
arvind4501: > But then Foreman comes (who is the only consumer of this API) and always sets the provider unless the selected provider in Foreman is "IPMI":
https://github.com/theforeman/foreman/blob/f2d892f7585c61d71f671298c8158712daace8e8/app/models/nic/bmc.rb#L27

In this case of user selected IPMI in foreman UI, the smart proxy does not get any bmc_provider in request and falls back to `Proxy::BMC::Plugin.settings.bmc_default_provider` , which is [ipmitool](https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/init.pp#L398) if its installed via foreman-installer.

> To add insult to injury, everyone disagrees what a provider is. Foreman defines [IPMI, Redfish and SSH](https://github.com/theforeman/foreman/blob/f2d892f7585c61d71f671298c8158712daace8e8/app/models/nic/bmc.rb#L3), but the Proxy knows about [freeipmi, ipmitool, redfish, shell and ssh](https://github.com/theforeman/smart-proxy/blob/a5780bc6e0132d7e91fa3074845b2a3d1cc30f01/modules/bmc/bmc_api.rb#L402-L443) 🤯

Yes and also with ssh provider, i am not sure if its supported as provider by foreman, our [doc](https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#enabling-power-management-on-hosts_foreman) does not say that and even i can't see any ssh key deployed at https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/init.pp#L401 path

tests/bmc_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+def is_bmc_enabled():                                                                                                                                                                                                                                                            
+    if os.path.exists(params_file):
+        with open(params_file, 'r') as f:
+            params = yaml.safe_load(f)
+            features = params.get('features', [])
+            if isinstance(features, str):
+                features = features.split()
+            return 'bmc' in features
+    return False
```
evgeni: There is a similar thing for IOP in https://github.com/theforeman/foremanctl/blob/44c09ec034eb0f894b865115cad2f1e7abbf37e5/tests/conftest.py#L195-L208

Should we have an `enabled_features` helper that then can be inspected further?

tests/bmc_test.py
evgeni, line 1: any reason this is an own file and not in foreman_proxy_test.py?

arvind4501: TBH no strong reason, foreman_proxy_test.py is a general place to test foreman proxy and its feature, the features we are asserting in foreman_proxy_test are added in CI too and works fine, but for BMC i need to implement `is_bmc_enabled` and foreman_proxy_test.py does not look right place for that.
as features grow for foreman-proxy i thought of having tests per feature

evgeni: BMC is a proxy feature, so all helpers for that are IMHO welcome in the proxy test file, but as noted above I actually think these helpers should be made more generic and moved to conftest.py

arvind4501: fair, moved to proxy test

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -44,6 +44,17 @@ variables:
     type: AbsolutePath
     parameter: --certificate-server-ca-certificate
     persist: false
+  foreman_proxy_bmc_default_provider:
+    parameter: --bmc-default-provider
+    help: Default BMC provider.
+    choices:
```
stejskalleos, line 52: In the [bmc.yaml.example ](https://github.com/theforeman/smart-proxy/blob/develop/config/settings.d/bmc.yml.example#L5) are more providers:

```yaml
# Available providers:
# - freeipmi / ipmitool - requires the appropriate package installed, and the rubyipmi gem
# - redfish - requires the redfish_client gem
# - shell - for local reboot control (requires sudo access to /sbin/shutdown for the proxy user)
# - ssh - limited remote control (status, reboot, turn off)
:bmc_default_provider: freeipmi
```

arvind4501: I have opened an [RFC](https://community.theforeman.org/t/rfc-remove-support-for-ssh-and-shell-providers-for-bmc/46375) for that and looking at that i think we should only stick to supported providers. and  c22c9211ad0f952038c4246b3d27e34984350716  can be droped?

src/roles/foreman_proxy/templates/settings.d/bmc.yml.j2
```diff
@@ -0,0 +1,16 @@
+---
+:enabled: {{ feature_enabled }}
+:bmc_default_provider: {{ foreman_proxy_bmc_default_provider }}
+
+{% if foreman_proxy_bmc_default_provider == 'redfish' %}
+:redfish_verify_ssl: {{ foreman_proxy_bmc_redfish_verify_ssl }}
+{% endif %}
```
evgeni: one can  use redfish without it being the default provider, so I'd not hide this behind an `if` (yes, we do this in Puppet too, and there too it's wrong, IMHO)

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -44,6 +44,37 @@ variables:
...
+      - freeipmi
```
evgeni, line 53: freeipmi is not installed in the container right now, so switching the default provider to it would break things, right?

evgeni:
```
May 08 09:36:26 quadlet.example.com foreman-proxy[133744]: 2026-05-08T09:36:26 3ae9d6df [I] Started GET /bmc/192.168.124.1/chassis/power/status 
May 08 09:36:26 quadlet.example.com foreman-proxy[133744]: 2026-05-08T09:36:26 3ae9d6df [W] freeipmi specified but it is not installed
May 08 09:36:26 quadlet.example.com foreman-proxy[133744]: 2026-05-08T09:36:26 3ae9d6df [W] Using ipmitool as the default BMC provider
```

mhhh, unsure what I dislike more :)

tests/foreman_proxy_test.py
```diff
@@ -33,3 +51,9 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
     )
     assert cmd.succeeded
     assert cmd.stdout == '201'
+
+@pytest.mark.skipif("not is_bmc_enabled()")
+def test_bmc_default_provider(server):
```
evgeni: cmd and get_default_bmc_provider and up reading /etc/foreman-proxy/settings.d/bmc.yml in the container, so they will always match?

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -44,6 +44,37 @@ variables:
...
+    help: BMC SSH user.
+  foreman_proxy_bmc_ssh_key:
+    parameter: --bmc-ssh-key
+    help: BMC SSH key.
+  foreman_proxy_bmc_ssh_poweron:
+    parameter: --bmc-ssh-poweron
+    help: BMC SSH power on command.
+  foreman_proxy_bmc_ssh_poweroff:
+    parameter: --bmc-ssh-poweroff
+    help: BMC SSH power off command.
```
amolpati30: Don’t you think we can create a single resource for both on and off, and provide those options as choices?
Like -
```
foreman_proxy_bmc_ssh_power_on/off:
    parameter: --bmc-ssh-power-action
    choices:
       - on
       - off
    help: BMC SSH power on/off command
```

src/playbooks/deploy/metadata.obsah.yaml
arvind4501: ah.. yes, need to look into that

src/playbooks/deploy/metadata.obsah.yaml
arvind4501: the thing is it will only set value(on/off) for a single variable either  `foreman_proxy_bmc_ssh_power_on` or `foreman_proxy_bmc_ssh_poweroff`. but anyway we are removing the support of these providers.

tests/foreman_proxy_test.py
arvind4501: yeah, thanks for pointing that out.

evgeni:
tested this with ipmitool and virtualbmc, worked well

tests/foreman_proxy_test.py
```diff
@@ -1,8 +1,24 @@
 import datetime
 import json
 import pytest
+import yaml
+from conftest import enabled_features
 
 FOREMAN_PROXY_PORT = 8443
+BMC_CONFIG = "/etc/foreman-proxy/settings.d/bmc.yml"
```
evgeni: this is now unused

tests/foreman_proxy_test.py
```diff
@@ -1,8 +1,24 @@
...
+
+def get_proxy_feature_setting(server, feature, default=None):
+    cmd = server.run(f"podman exec foreman-proxy cat /etc/foreman-proxy/settings.d/{feature}.yml")
+    if cmd.succeeded:
+        return yaml.safe_load(cmd.stdout)
+    return default
```
evgeni: should the default be `{}` (or `if default is None: default = {}`)?
otherwise you return `None` and the called tries `.get` on `NoneType`.

tests/foreman_proxy_test.py
```diff
@@ -33,3 +51,7 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
     )
     assert cmd.succeeded
     assert cmd.stdout == '201'
+
+@pytest.mark.skipif("not is_bmc_enabled()")
```
evgeni, line 63: shall we add `--add-feature bmc` to the enablement in `.github/workflows/test.yml` so that it gets executed?

tests/foreman_proxy_test.py
arvind4501: {} works, updated

src/playbooks/deploy/metadata.obsah.yaml
arvind4501: Ok, i see the issue now, `ipmitool` package gets installed in proxy container as its [required](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec#L22C11-L22C19) by rubyipmi gem, but as `freeipmi` is not in that list it won't get installed out of the box. i see two ways to do this

- install `freeipmi` in [Containerfile](https://github.com/theforeman/foreman-oci-images/blob/master/images/foreman-proxy/Containerfile)
-  or add it as dependency for rubyipmi gem

which one do you prefer?

evgeni: Given it's `foreman-proxy` who is giving us the explicit choise, we could also add it as a dependency of `foreman-proxy`, but then it's also installed in (non container) setups that don't do BMC at all.

Puppet does the selection at install time: https://github.com/theforeman/puppet-foreman_proxy/blob/30e0e0f565b31e40e217aafb7ae11d7dbe1efd78/manifests/install.pp#L18-L20
Given we don't have that "luxury" in containers, the next best equivalent would be the Containerfile?

That all said, I would accept any of the solutions.

arvind4501: https://github.com/theforeman/foreman-oci-images/pull/32

evgeni: Deployed the container from that PR (thanks @Odilhao for the hint about the Konflux built PR images) via
```
diff --git src/vars/images.yml src/vars/images.yml
index 6535633..7669fa5 100644
--- src/vars/images.yml
+++ src/vars/images.yml
@@ -24,8 +24,8 @@ candlepin_container_image: quay.io/foreman/candlepin
-foreman_proxy_container_image: "quay.io/foreman/foreman-proxy"
-foreman_proxy_container_tag: "{{ container_tag_stream }}"
+foreman_proxy_container_image: "quay.io/foreman/foreman-proxy-stage"
+foreman_proxy_container_tag: "on-pr-b64b6ec3d5f0845a8631f30e5211136f9b703f0c"
 
 postgresql_container_image: quay.io/sclorg/postgresql-13-c9s
 postgresql_container_tag: "latest"
```
 and things work fine with freeipmi too.

src/playbooks/deploy/metadata.obsah.yaml
evgeni: Yeah, I think it can be dropped. If the RFC ends up with a different result (which I don't expect), we can still implement that after the fact.

arvind4501: thanks, removed it

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -46,6 +46,17 @@ variables:
     type: AbsolutePath
     parameter: --certificate-server-ca-certificate
     persist: false
+  foreman_proxy_bmc_default_provider:
+    parameter: --bmc-default-provider
+    help: Default BMC provider.
```
stejskalleos: The description is basically the parameter name, so not much of an explanation there for me.

The default implies (to me) the existence of a non-default parameter, which I don't think we have. Why not name it `--bmc-provider`?

---

I see it's named the same as in the bmc.yml, which I think should also be named just `:bmc_provider:`. (not a problem of this PR)

stejskalleos (changes requested):

src/playbooks/deploy/metadata.obsah.yaml
arvind4501: > The default implies (to me) the existence of a non-default parameter, which I don't think we have. Why not name it --bmc-provider?

 I think this goes back to how we expose default provider in smart-proxy itself, if you see https://github.com/theforeman/smart-proxy/blob/a5780bc6e0132d7e91fa3074845b2a3d1cc30f01/modules/bmc/bmc_api.rb#L393
 we look for `bmc_default_provider` in `bmc.yml`. i could change the option to `--bmc-provider` but that then does not change the fact that actually it affects `bmc_default_provider`.

I think for this change we need to make changes in `smart-proxy` as well, which IMHO can be looked sepratelyy. what do you think?


> The description is basically the parameter name, so not much of an explanation there for me.

description is inherited from puppet
https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/init.pp#L208

src/features.yaml
```diff
@@ -44,3 +44,7 @@ iop:
   description: iop services
   dependencies:
     - rh-cloud
+bmc:
+  description: Power management for bare metal hosts (IPMI, Redfish, SSH)
```
evgeni:
```suggestion
  description: Power management for bare metal hosts (IPMI, Redfish)
```

src/playbooks/deploy/metadata.obsah.yaml
evgeni: It's not even the BMC provider, it's the **IPMI** provider/implementation.

https://github.com/theforeman/foremanctl/pull/484/changes#r3186541821

stejskalleos: `bmc_default_provider`, what's the default value?
The name IMHO does not make sense. But we are not really good at naming. We use anything except `provider`, see `use_provider` in https://github.com/theforeman/smart-proxy/blob/develop/config/settings.d/dns.yml.example#L10 and https://github.com/theforeman/smart-proxy/blob/develop/config/settings.d/dhcp.yml.example

evgeni: So call it `--bmc-ipmi-implementation` and only allow `freeipmi` and `ipmitool` here?

stejskalleos: Or even just `--bmc-ipmi`

arvind4501: but we also support redfish, so including ipmi in name itself is misleading

evgeni: Right, but the only place the "default_bmc_provider" is used (effectively) is to select freeipmi or ipmitool when the user selects "IPMI".
redfish is always passed as "redfish" provider

arvind4501: ah.. yes, true in case of foreman is using smart-proxy. as smart-proxy providers APIs i think anyone can consume those(not only foreman), and that then makes me think smart-proxy implemtation of default-provider is fine in itself if foreman is not the only cosumer.

evgeni: can, sure, but I don't think there actually are any other consumers.

evgeni (approved):
This matches the current Puppet implementation, so ACK.

However, I would prefer if we could take the chance to clear things up and rename `--bmc-default-provider` to `--bmc-ipmi-implementation`, only allowing `freeipmi` and `ipmitool` as values, leaving `redfish` to be set explicitly by Foreman. Not going to block on this though.

src/playbooks/deploy/metadata.obsah.yaml
arvind4501: i could make the options here in foremanctl as `--bmc-ipmi-implementation` and save it in bmc.yml as `bmc_default_provider` . should i do the changes in smart-proxy implemenation as well

ekohl:
> However, I would prefer if we could take the chance to clear things up and rename `--bmc-default-provider` to `--bmc-ipmi-implementation`, only allowing `freeipmi` and `ipmitool` as values, leaving `redfish` to be set explicitly by Foreman. Not going to block on this though.

Agreed. IMHO the Puppet implementation was always a bit confusing.

tests/foreman_proxy_test.py
```diff
@@ -33,3 +50,7 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
...
+def test_bmc_default_provider(server):
+    assert get_default_bmc_provider(server) == 'ipmitool'
```
evgeni: I'm debating whether the test should be called like the settings entry or the cli option (ipmi implementation), and leaning to the latter?

evgeni (approved):

tests/conftest.py
```diff
@@ -194,22 +193,20 @@ def wait_for_tasks(foremanapi, search=None):
...
             if isinstance(features, str):
                 features = features.split()
-            return 'iop' in features
-
-    return False
+            return features
+    return []
 
+def is_iop_enabled():
+    return 'iop' in enabled_features()
```
ekohl, line 223: Nit: there are supposed to be 2 empty lines between top level function definitions
```suggestion

def is_iop_enabled():
    return 'iop' in enabled_features()

```

ekohl (commented):
I think it is good to stick to https://peps.python.org/pep-0008/#blank-lines

> Surround top-level function and class definitions with two blank lines.

And perhaps also enforce that via pylint.

tests/conftest.py
arvind4501: > And perhaps also enforce that via pylint.

Sure, i am thinking to implement https://docs.astral.sh/ruff/ , i have created a seprate PR for that https://github.com/theforeman/foremanctl/pull/505

evgeni:
@stejskalleos you had requested changes in the past, were your concerns addressed?

stejskalleos (approved):
LGTM, my requested changes have been implemented.
(I have not tested the feature myself tho)

tests/foreman_proxy_test.py
```diff
@@ -1,16 +1,33 @@
...
+
 def test_foreman_proxy_features(server, certificates, server_fqdn):
     cmd = server.run(f"curl --cacert {certificates['server_ca_certificate']} --silent https://{server_fqdn}:{FOREMAN_PROXY_PORT}/features")
     assert cmd.succeeded
     features = json.loads(cmd.stdout)
     assert "logs" in features
     assert "script" in features
     assert "dynflow" in features
+    if is_bmc_enabled():
+        assert "bmc" in features
```
ekohl, line 33: Perhaps also add the negative so expect it isn't enabled?
```suggestion
    if is_bmc_enabled():
        assert "bmc" in features
    else:
        assert "bmc" not in features
```

tests/foreman_proxy_test.py
```diff
@@ -1,16 +1,33 @@
 import datetime
 import json
 import pytest
+import yaml
+from conftest import enabled_features
 
 FOREMAN_PROXY_PORT = 8443
 
+def get_proxy_feature_setting(server, feature):
+    cmd = server.run(f"podman exec foreman-proxy cat /etc/foreman-proxy/settings.d/{feature}.yml")
```
ekohl: Perhaps not for this PR, but I think we should avoid reading config files where possible (treat them as internal details), but there is a way to expose settings in the Smart Proxy. Then it shows up in `/v2/features` as an exposed setting and you can query it from the outside.

tests/foreman_proxy_test.py
```diff
@@ -1,16 +1,33 @@
...
+
+def is_bmc_enabled():
+    return 'bmc' in enabled_features()
+
+def get_default_bmc_provider(server):
+    bmc_setting = get_proxy_feature_setting(server, 'bmc')
+    return bmc_setting.get(':bmc_default_provider')
+
 def test_foreman_proxy_features(server, certificates, server_fqdn):
     cmd = server.run(f"curl --cacert {certificates['server_ca_certificate']} --silent https://{server_fqdn}:{FOREMAN_PROXY_PORT}/features")
```
ekohl, line 26: Should this query `/v2/features` and also verify the capability? https://github.com/theforeman/smart-proxy/blob/4830f1b0f4ba552d6c124e615722b80aa7846eaf/modules/bmc/bmc_plugin.rb#L13 makes sure you can query that it's capable of both `ipmitool` and `freeipmi` which helps you assert that both tools are present in the container.

ekohl:
In case it wasn't clear: don't consider my review as blocking but as hints to further assert a correct implementation.

amolpati30 (approved):
```
I have tested the following:

Verified the parameter behavior during and after deployment, and no failures were observed.
Tested the BMC functionality on the iDRAC server and performed power on/off actions, which worked as expected.
```

tests/foreman_proxy_test.py
arvind4501: I tried to see `/v2/features` endpoint but could not see anything related settings, i am only able to query `/features` which gives me
```
 [
    "bmc",
    "logs"
]
``` 
but nothing in case of /v2/features

I think https://github.com/theforeman/smart-proxy/blob/develop/modules/bmc/bmc_plugin.rb does not expose any extra setting as done in https://github.com/theforeman/smart-proxy/blob/4830f1b0f4ba552d6c124e615722b80aa7846eaf/modules/tftp/tftp_plugin.rb#L14
 or did i missinterpreted you?

ekohl: > I tried to see `/v2/features` endpoint but could not see anything related settings, i am only able to query `/features` which gives me

You need to be authenticated so see that since settings can contain confidential information. See https://theforeman.org/2019/04/smart-proxy-capabilities-explained.html for more information.

> I think https://github.com/theforeman/smart-proxy/blob/develop/modules/bmc/bmc_plugin.rb does not expose any extra setting as done in https://github.com/theforeman/smart-proxy/blob/4830f1b0f4ba552d6c124e615722b80aa7846eaf/modules/tftp/tftp_plugin.rb#L14

Correct, which is why you can't test it in this PR. However, we are in control of the BMC plugin so if this is something we care about then perhaps we should expose it.

The larger point: how do we test functionality going forward? Reading the actual config files is IMHO more of a unit test level thing. However, I'd expect the end-to-end test suite to be more black box and verify behavior. So let Foreman Proxy actually parse the config and read the state of the service. This will also uncover any typos that you may have made in config files or other bugs.

Was that clearer?

arvind4501: Yes, thanks for the explanation. so we can merge this PR and tackle the tests sepreately as requires changes in smart-proxy repo

evgeni: that's `/features` what you pasted, `/v2/features` needs auth and gives you:
```
{
    "dynflow": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {},
        "state": "disabled",
        "capabilities": []
    },
    "ansible": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {},
        "state": "disabled",
        "capabilities": []
    },
    "script": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {},
        "state": "disabled",
        "capabilities": []
    },
    "facts": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {},
        "state": "disabled",
        "capabilities": []
    },
    "dns": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "use_provider": null
        },
        "state": "disabled",
        "capabilities": []
    },
    "templates": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "template_url": null
        },
        "state": "disabled",
        "capabilities": []
    },
    "tftp": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "tftp_servername": null
        },
        "state": "disabled",
        "capabilities": [
            "bootloader_universe"
        ]
    },
    "dhcp": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "use_provider": null
        },
        "state": "disabled",
        "capabilities": []
    },
    "puppetca": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "use_provider": null
        },
        "state": "disabled",
        "capabilities": []
    },
    "puppet": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "use_provider": null
        },
        "state": "disabled",
        "capabilities": []
    },
    "bmc": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {},
        "state": "running",
        "capabilities": [
            "freeipmi",
            "ipmitool",
            "power_action_v2",
            "redfish",
            "shell",
            "ssh"
        ]
    },
    "realm": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "use_provider": null
        },
        "state": "disabled",
        "capabilities": []
    },
    "logs": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {},
        "state": "running",
        "capabilities": []
    },
    "httpboot": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "http_port": null,
            "https_port": null
        },
        "state": "disabled",
        "capabilities": []
    },
    "registration": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "registration_url": null
        },
        "state": "disabled",
        "capabilities": []
    }
}
```

(via `curl --cacert certificates/certs/ca.crt --cert certificates/certs/quadlet.example.com-client.crt --key certificates/private/quadlet.example.com-client.key -k -v https://$(hostname -f):8443/v2/features | python3 -m json.tool`)

and there you can see in BMC what it can do, but sadly not what the default provider is.

arvind4501: > sadly not what the default provider is.

yes, thats need to added

evgeni: and yes, I agree, let's merge as is, and then polish up the tests a bit

arvind4501: ah ruff conflicts :disappointed:

ekohl: > and there you can see in BMC what it can do, but sadly not what the default provider is.

As I noted below (https://github.com/theforeman/foremanctl/pull/484#discussion_r3232477843) I'd still argue that asserting `freeipmi` and `ipmitool` as capabilities is a good test. It asserts that both tools are available in the container image and Foreman Proxy detects them as usable. Pretty sure it would uncover when https://github.com/theforeman/foreman-packaging/pull/13417 isn't in the container image.

>```json
>     "bmc": {
>         "http_enabled": true,
>         "https_enabled": true,
>         "settings": {},
>         "state": "running",
>         "capabilities": [
>             "freeipmi",
>             "ipmitool",
>             "power_action_v2",
>             "redfish",
>             "shell",
>             "ssh"
>         ]
>     },
>```

IMHO this also uncovers another TODO item: `http_enabled` should IMHO be `false` so that we only expose it over HTTPS.

tests/foreman_proxy_test.py
```diff
@@ -38,3 +58,8 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
     )
     assert cmd.succeeded
     assert cmd.stdout == '201'
+
+
+@pytest.mark.skipif("not is_bmc_enabled()")
```
ekohl, line 63: In https://github.com/theforeman/smoker/pull/46/changes/1f7fd3a9fc5c67df625e8d5b8b0485f1c11c2a06 I implemented a mechanism in smoker to dynamically read the plugins. Perhaps for a follow up to implement this for enabled features so you can use:
```suggestion
@pytest.mark.feature("bmc")
```

That would provide a much more generic basis for testing features.

tests/foreman_proxy_test.py
arvind4501: would you like me to set `http_enabled` false in bmc config, here or https://github.com/theforeman/smart-proxy/blob/4a468d730426ca561f9e2f088589427ae2f96843/lib/proxy/settings/plugin.rb#L4?

evgeni: given we don't expose the http port today, this is not relevant, IMHO

ekohl: Fair, but it's something to keep in mind. Within the provisioning context we will see it and need to track that work when we do expose the HTTP port.

arvind4501: https://github.com/theforeman/smart-proxy/pull/943 for exposing default provider setting

tests/foreman_proxy_test.py
arvind4501: i have updated it in b9bb58b43310f6931d321f918014c6cfacc87c8f

tests/foreman_proxy_test.py
ekohl: I couldn't help myself: https://github.com/theforeman/foremanctl/pull/508.

evgeni:
Let's continue this in #508

PR #485, opened 2026-05-04
Skip tests for features not enabled and show pytest results

ehelms:

When certain features are not enabled, their tests are marked as failed which can make debugging confusing. 

* Checks for feature enablement and skips tests if their feature is not enabled
* Uses flavor + parameters.yaml to determine enabled features
* Add get_enabled_features() and has_feature() helpers to conftest.py that replicate the Ansible enabled_features calculation by combining flavor_features with user features from parameters.yaml.
* Use pytest.mark.skipif to gate tests on their required features: foreman-proxy, hammer, azure-rm, google, remote-execution, and iop.
* Read the Foreman admin password from
`.var/lib/foremanctl/foreman-admin-init-passwd` instead of hardcoding.
* Update the test playbook to display pytest output before failing so pass/fail details are visible in Ansible output.

Steps to reproduce:

* Run a deployment with some set of enabled features (e.g. `foremanctl deploy --add-feature iop`
* Check which tests are skipped and ensure the list looks appropriate

#### Checklist


ekohl (commented):
I hadn't seen this. https://github.com/theforeman/foremanctl/pull/508 works towards a more generalized mechanism to skip tests for disabled features.

I like the idea of leveraging the directory structure more. How about:
* tests
  * feature
    * foreman
      * test_base.py
    * foreman-proxy
      * test_base.py
      * test_bmc.py
    * iop
      * test_advisor.py

Then taking your approach (and very similarly https://stackoverflow.com/questions/57031403/pytest-marks-mark-entire-directory-package) we can automate it. I've taken a stab at that in https://github.com/theforeman/foremanctl/pull/509.

tests/conftest.py
```diff
@@ -83,11 +83,20 @@ def ssh_config(server_hostname):
 
 
 @pytest.fixture(scope="module")
-def foremanapi(ssh_config, server_fqdn):
+def foreman_admin_password():
+    test_dir = os.path.dirname(os.path.abspath(__file__))
+    foremanctl_dir = os.path.dirname(test_dir)
+    passwd_file = os.path.join(foremanctl_dir, '.var', 'lib', 'foremanctl', 'foreman-admin-init-passwd')
+    with open(passwd_file) as f:
```
ekohl, line 91: I also ran into this. I'd love to see it in a dedicated commit/PR. Please then also use `rootpath`
```suggestion
def foreman_admin_password(pytestconfig):
    passwd_file = pytestconfig.rootpath / '.var' / 'lib' / 'foremanctl' / 'foreman-admin-init-passwd'
    with passwd_file.open() as f:
```

PR #488, opened 2026-05-04
Fix IoP advisor integration: Kafka listener binding and engine rule packages

jeremylenz:
## Summary

- Fix Kafka listener binding to all interfaces so connections survive container restarts
- Add `prodsec.rules` and `telemetry.rules.plugins` to the engine package list so diagnostic rules are loaded and advisor receives findings

## Details

**Kafka listener fix** (`src/roles/iop_kafka/templates/kafka/kraft.j2`): The previous config bound Kafka to the container hostname (`iop-core-kafka:9092`), which resolves to the container's IP. When the container restarts it gets a new IP, causing connection refused errors in the advisor service and other consumers. Binding to all interfaces (`PLAINTEXT://:9092`) avoids this.

**Engine packages** (`src/roles/iop_engine/defaults/main.yaml`): Without `prodsec.rules` and `telemetry.rules.plugins`, the engine only loads `insights_kafka_service.rules` which contains no diagnostic rules — only metadata. Advisor receives results with 0 reports on every upload.

## Test plan

- [ ] Deploy IoP with `deploy-dev`
- [ ] Register a host with `insights-client`
- [ ] Verify advisor service logs a non-zero report count
- [ ] Verify Kafka connections survive an engine or advisor container restart

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/docs/iop.md
+++ b/docs/iop.md
@@ -136,4 +136,14 @@ Gateway certificates are configured per certificate source:
 
 ### Container Images
 
-All IOP images default to `quay.io/iop/<service>:foreman-3.16`. Each role exposes `iop_<role>_container_image` and `iop_<role>_container_tag` variables to override.
+All IOP images default to `quay.io/iop/<service>:foreman-3.18`. Each role exposes `iop_<role>_container_image` and `iop_<role>_container_tag` variables to override.
+
+### Engine Rule Packages
+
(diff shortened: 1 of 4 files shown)


ehelms (approved):

vkrizan (approved):
Thanks!

More concerns deserve a commit each. 😁

jeremylenz:
Split into 2 commits.

PR #491, opened 2026-05-05
set the FQDN setting to our real FQDN so that derived settings work

### LINKED ISSUE #482: Container hostname change breaks unattended_url for external host registration
## Problem

Commit 4ef4632b7271 ("use shorter internal hostnames for the pulp containers")
changed the Foreman container hostname from `{{ ansible_facts['fqdn'] }}` to
`{{ ansible_facts['hostname'] }}.local` for all containers, including the
Foreman application containers.

Foreman derives the `unattended_url` setting from the container's internal
hostname. After this change, `unattended_url` becomes `http://satellite.local`
instead of `https://satellite.example.com`, causing the `Host initial
configuration` template to generate callbacks to an unreachable hostname.

## Symptoms

All hosts register successfully via subscription-manager (POST /rhsm/consumers
works), but fail at the host init phase with:

```
ERROR: Host [hostname] configured, but failed to report status to Foreman.
```

The generated `/root/registration_host_init.sh` contains:
```
curl https://satellite.local/unattended/built?token=...
```

which is unreachable from external hosts.

## Impact

- All host registrations using the global registration template fail
  at the host init callback phase
- CI does not catch this because it does not test the full registration
  flow with host init template execution
- Workaround: `hammer settings set --name unattended_url --value "https://$(hostname -f)"`

evgeni:

Foreman uses the `domain` and `fqdn` settings (which default to their respective values from Facter) to calculate several defaults for URL and E-mail related settings. As our containers don't have a correct FQDN (since 4ef4632b72711eb704c989544dc2961eb954f772), this creates wrong defaults, breaking certain workflows.

See #482 for details.

* The `fqdn` and `domain` settings are set to sane values.

Steps to reproduce:

* The updated tests should be sufficient.

#### Checklist

Fixes: #482


evgeni:
Should we add a test that asserts that all settings that contain fqdns/domains have the right value?

ehelms:
What about the scenario where the user is providing a cname?

evgeni:
What about it? That wasn't working for unattended url before, and doesn't work now ;)

evgeni:
Tests updated

tests/foreman_test.py
```diff
@@ -75,6 +75,20 @@ def test_foreman_recurring_services_exist(server, instance):
...
     assert delivery_method_setting[0]['value'] == 'smtp'
+
+
+@pytest.mark.parametrize("setting", ["foreman_url", "unattended_url"])
+def test_foreman_fqdn_in_url_settings(foremanapi, server_fqdn, setting):
+    settings = foremanapi.list('settings', search=f'name={setting}')
+    assert server_fqdn in settings[0]['value']
+
+
+@pytest.mark.parametrize("setting", ["administrator", "email_reply_address"])
```
amolpati30, line 90: If I am referring to the `administrator`, should it search for a domain such as `rhos-01.prod.psi.rdu2.redhat.com`?

evgeni: that's what the test does. yeah

amolpati30: Ah, I was testing this, but with the patch it was not showing the correct result—possibly due to some issues on my server. I tried it on a fresh server, and it works as expected.

amolpati30 (approved):

src/vars/base.yaml
```diff
@@ -19,7 +19,8 @@ foreman_ca_certificate: "{{ ca_bundle }}"
 foreman_client_key: "{{ client_key }}"
 foreman_client_certificate: "{{ client_certificate }}"
 foreman_plugins: "{{ enabled_features | features_to_foreman_plugins }}"
-foreman_url: "https://{{ ansible_facts['fqdn'] }}"
+foreman_name: "{{ ansible_facts['fqdn'] }}"
```
archanaserver, line 22: Not blocking but just a thought, i believe removing `foreman_name` from `src/vars/base.yaml` and keep it only in `defaults/main.yaml` would let users to override it when needed, because as `vars/` go with  higher precedence than `defaults/`, so defaults definition is currently being ignored.

src/roles/foreman/templates/settings.yaml.j2
```diff
@@ -1,4 +1,6 @@
 ---
+:domain: {{ ansible_facts['domain'] }}
+:fqdn: {{ foreman_name }}
```
archanaserver, line 3: Isn't this setting will rewritten every time ansible runs, so If a user manually changes these settings via hammer, will they be overwritten when `foremanctk deploy` runs again?

evgeni: once something is set in settings.yaml, it can't be overriden via hammer anymore

src/vars/base.yaml
evgeni: if the user should be able to override this, we'd have to expose it in metadata anyway, and that gets higher priority than vars.

we've been following "the role has a sensible default, but the real value comes from vars" so far, but I am not sure we ever written it down so explicitly.

archanaserver: hmm, make sense

archanaserver (approved):

PR #305, opened 2025-11-17
Set the libvirt management domain

ekohl:
This makes sure the domain is set when the management domain is created, which makes libvirt configure dnsmasq as a proper DNS server. This can be used together with systemd-resolved on the host to look up guest IPs.

It also configures Vagrant in CI with DNS set up to utilize this.

PR #494, opened 2026-05-05
Scope Vagrant libvirt domains to allow multiple checkouts

ehelms:

I sometimes have 2-3 different directories working at the same time, this change will prevent clashes in those environments. 

Use a hash of the absolute working directory path to generate a unique libvirt domain prefix per checkout, preventing collisions when multiple clones or git worktrees deploy Vagrant boxes simultaneously.

Both the libvirt domain name prefix and the VM hostnames are scoped using the prefix, ensuring full isolation between deployments.

Steps to reproduce:

* `./forge vms start`
* `vagrant ssh quadlet`
* `hostname`

Check that the hostname is not `quadlet.example.com`

#### Checklist

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,4 +1,7 @@
+require 'digest'
+
 DOMAIN = ENV.fetch('VAGRANT_DOMAIN', 'example.com'.freeze)
+prefix = ENV.key?('FOREMANCTL_VAGRANT_PREFIX') ? Digest::MD5.hexdigest(Dir.pwd)[0..7] : nil
 
 Vagrant.configure("2") do |config|
   config.vm.synced_folder ".", "/vagrant"
@@ -18,9 +21,10 @@ Vagrant.configure("2") do |config|
 
   config.vm.define "quadlet" do |override|
     override.vm.box = ENV.fetch("FOREMANCTL_BASE_BOX", "centos/stream9")
-    override.vm.hostname = "quadlet.#{DOMAIN}"
+    override.vm.hostname = prefix ? "quadlet.#{prefix}.#{DOMAIN}" : "quadlet.#{DOMAIN}"
 
     override.vm.provider "libvirt" do |libvirt, provider|
+      libvirt.default_prefix = prefix if prefix
       libvirt.memory = 10240
       libvirt.cpus = 4
       libvirt.machine_virtual_size = 30
@@ -29,18 +33,20 @@ Vagrant.configure("2") do |config|
 
   config.vm.define "client" do |override|
     override.vm.box = "centos/stream9"
-    override.vm.hostname = "client.#{DOMAIN}"
+    override.vm.hostname = prefix ? "client.#{prefix}.#{DOMAIN}" : "client.#{DOMAIN}"
 
     override.vm.provider "libvirt" do |libvirt, provider|
+      libvirt.default_prefix = prefix if prefix
       libvirt.memory = 1024
     end
   end
 
   config.vm.define "database" do |override|
     override.vm.box = "centos/stream9"
-    override.vm.hostname = "database.#{DOMAIN}"
+    override.vm.hostname = prefix ? "database.#{prefix}.#{DOMAIN}" : "database.#{DOMAIN}"
 
     override.vm.provider "libvirt" do |libvirt, provider|
+      libvirt.default_prefix = prefix if prefix
       libvirt.memory = 2048
     end
   end
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -63,7 +63,7 @@ def server_hostname():
 
 @pytest.fixture(scope="module")
 def server_fqdn(server):
-    return server.check_output('hostname -f')
+    return server.backend.hostname
 
 
 @pytest.fixture(scope="module")
@@ -73,7 +73,7 @@ def client_hostname():
 
 @pytest.fixture(scope="module")
 def client_fqdn(client):
-    return client.check_output('hostname -f')
+    return client.backend.hostname
 
 
 @pytest.fixture(scope="module")
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -39,6 +39,19 @@ source .venv/bin/activate
 ./foremanctl deploy --initial-admin-password=changeme --tuning development --initial-organization "Foreman CI" --initial-location "Internet"
 ```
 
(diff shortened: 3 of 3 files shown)


Vagrantfile
ekohl, line 1: Can we first merge https://github.com/theforeman/foremanctl/pull/305 and then rebase this?

tests/conftest.py
```diff
@@ -30,8 +30,8 @@ def server_hostname():
 
 
 @pytest.fixture(scope="module")
-def server_fqdn(server_hostname):
-    return f'{server_hostname}.example.com'
+def server_fqdn(server):
+    return server.check_output('hostname -f')
```
ekohl: Do we want to get what the client reports or just what we configured? I'd like to avoid running `hostname -f` because it does a DNS lookup and can be unpredictable.
```suggestion
    return server.backend.hostname
```

tests/conftest.py
```diff
@@ -40,8 +40,8 @@ def client_hostname():
 
 
 @pytest.fixture(scope="module")
-def client_fqdn(client_hostname):
-    return f'{client_hostname}.example.com'
+def client_fqdn(client):
+    return client.check_output('hostname -f')
```
ekohl: Same as above
```suggestion
    return client.backend.hostname
```

.github/workflows/test.yml
```diff
@@ -122,7 +128,7 @@ jobs:
...
+          ./foremanctl deploy --certificate-source=${{ matrix.certificate_source }} ${{ matrix.database == 'external' && format('--database-mode=external --database-host={0} --database-ssl-ca $(pwd)/.var/lib/foremanctl/db-ca.crt --database-ssl-mode verify-full', env.DATABASE_FQDN) || '' }} ${{ matrix.certificate_source == 'custom_server' && format('--certificate-server-certificate /root/custom-certificates/certs/{0}.crt --certificate-server-key /root/custom-certificates/private/{0}.key --certificate-server-ca-certificate /root/custom-certificates/certs/server-ca.crt', env.QUADLET_FQDN) || '' }} --foreman-initial-admin-password=changeme --initial-organization "Foreman CI" --initial-location "Internet" --tuning development
```
ekohl: At which point are we going to split up this invocation to multi-line?

.github/workflows/test.yml
```diff
@@ -90,6 +90,12 @@ jobs:
       - name: Start VMs
         run: |
           ./forge vms start --vms "quadlet client ${{ matrix.database == 'external' && 'database' || '' }}"
+      - name: Resolve VM FQDNs
```
ekohl: Rather than resolving, perhaps we should introduce a way to make the domain predictable. Using https://github.com/theforeman/foremanctl/pull/305 as a base it should be easy to use:
```ruby
DOMAIN = ENV.get('VAGRANT_DOMAIN') { "#{Digest::MD5.hexdigest(Dir.pwd)[0..7]}.example.com" }
```

tests/conftest.py
ekohl: Answering my own question: the backend "hostname" is actually the IP so this is correct.

ekohl:
Thanks for merging https://github.com/theforeman/foremanctl/pull/305. Needs a rebase now.

.github/workflows/test.yml
```diff
@@ -105,6 +105,8 @@ jobs:
         run: pip install --upgrade ansible-core
       - name: Setup environment
         run: ./setup-environment
+      - name: Set Vagrant domain
+        run: echo "VAGRANT_DOMAIN=$(echo -n "$(pwd)" | md5sum | cut -c1-8).example.com" >> "${GITHUB_ENV}"
```
ekohl: Technically in CI we don't really care about the exact domain and uniqueness. `$pwd` will be the same for every checkout anyway so it's not that you can differentiate the exact axis in the test matrix so perhaps keep it simple?
```suggestion
        run: echo "VAGRANT_DOMAIN=example.com" >> "${GITHUB_ENV}"
```

Vagrantfile
ekohl: Can the `default_prefix` be set at this level to avoid the repetition?

```suggestion
    libvirt.management_network_domain = DOMAIN
    libvirt.default_prefix = prefix
```

.github/workflows/test.yml
ehelms: This was split in a previous PR.

Vagrantfile
```diff
@@ -14,13 +17,15 @@ Vagrant.configure("2") do |config|
...
     libvirt.management_network_domain = DOMAIN
+    libvirt.default_prefix = prefix
   end
 
   config.vm.define "quadlet" do |override|
     override.vm.box = ENV.fetch("FOREMANCTL_BASE_BOX", "centos/stream9")
     override.vm.hostname = "quadlet.#{DOMAIN}"
 
     override.vm.provider "libvirt" do |libvirt, provider|
+      libvirt.default_prefix = prefix
```
ekohl: In https://github.com/theforeman/foremanctl/pull/494/changes#r3266500716 I implied that this line should no longer be needed, and the same goes for the other VM definitions.

Vagrantfile
```diff
@@ -14,6 +17,7 @@ Vagrant.configure("2") do |config|
 
   config.vm.provider "libvirt" do |libvirt|
     libvirt.management_network_domain = DOMAIN
+    libvirt.default_prefix = prefix
```
ekohl: Is it good to keep the foremanctl prefix so they're still grouped?
```suggestion
    libvirt.default_prefix = "foremanctl_#{prefix}"
```

ekohl (commented):
When we merge this, will it break existing checkouts users may have? Should the prefix be opt-in?

ekohl:
There's a merge conflict. Forgot to pull `master` before rebasing?

tests/conftest.py
```diff
@@ -63,7 +63,7 @@ def server_hostname():
 
 @pytest.fixture(scope="module")
 def server_fqdn(server):
-    return server.check_output('hostname -f')
+    return server.backend.hostname
```
ekohl, line 66: When I played with this I found that the backend "hostname" in our config really is the IP address, which is why I went for the `check_output` that you originally also went for.

ShimShtein:
Can you also output the new prefix? I don't think we show the names of the newly created machines.

PR #483, opened 2026-05-03
Add httpd MPM event module configuration

pablomh:
#### Why are you introducing these changes? (Problem description, related links)

foremanctl does not manage the Apache MPM event module configuration.
Without an explicit `event.conf`, httpd uses its compiled-in defaults:
`ServerLimit=16`, `ThreadsPerChild=25`, `MaxRequestWorkers=400`.

In foreman-installer deployments, puppet-apache sets base defaults of
`ServerLimit=25` and `ThreadsPerChild=16` (with MaxRequestWorkers
derived automatically as ServerLimit * ThreadsPerChild). The tuning
profiles then override these to `ServerLimit=64` and
`MaxRequestWorkers=1024` via custom-hiera. foremanctl has no equivalent,
leaving the system severely under-provisioned for concurrent workloads
such as bulk host registration.

With only 400 MaxRequestWorkers, httpd begins rejecting connections once
concurrent requests exceed this limit, causing registration failures that
are entirely avoidable with proper MPM configuration.

#### What are the changes introduced in this pull request?

* Add `event.conf.j2` template to the httpd role with configurable MPM
  event parameters (ServerLimit, MaxRequestWorkers, ListenBacklog, etc.)
* Add role defaults matching puppet-apache base values (ServerLimit=25,
  MaxRequestWorkers not set — Apache derives it automatically)
* Add per-profile tuning overrides for medium, large, extra-large, and
  extra-extra-large matching the foreman-installer custom-hiera values
  (ServerLimit=64, MaxRequestWorkers=1024)
* Add task to deploy `event.conf` to `/etc/httpd/conf.modules.d/`

#### How to test this pull request

Steps to reproduce:

* Deploy with `foremanctl deploy`
* Verify `/etc/httpd/conf.modules.d/event.conf` exists with
  `ServerLimit 25` and no explicit `MaxRequestWorkers`
* Deploy with `foremanctl deploy --tuning large`
* Verify `/etc/httpd/conf.modules.d/event.conf` now has
  `ServerLimit 64` and `MaxRequestWorkers 1024`
* Run `httpd -t` to confirm syntax
* Run `hammer ping` to confirm all services operational
* Register hosts concurrently — should not hit httpd connection limits

#### Checklist
* [x] Tests added/updated (if applicable)
* [ ] Documentation updated (if applicable)

PR #495, opened 2026-05-05
Add registration admission control for httpd

pablomh:
## Summary

Add registration admission control for `/rhsm` and `/register` endpoints via an Apache `mod_proxy_balancer` pool. When `httpd_registration_admission_max > 0`, requests beyond that limit are queued by Apache instead of overwhelming Puma during registration bursts.

Builds on top of #483 (MPM event module configuration).

## Problem

Under high-concurrency registration (e.g., 912+ simultaneous hosts), all requests arrive at Puma simultaneously, creating a deep backlog where tail hosts exceed subscription-manager's 180s timeout. Without admission control, pass rates drop sharply above 760 concurrent.

## Solution

A `<Proxy balancer://foreman-registration>` block with a `max=` limit on the `BalancerMember` directive limits how many registration requests reach Puma concurrently. Excess requests are held in Apache's event loop (non-blocking) until a slot opens.

The balancer block is injected conditionally in both `foreman-ssl-vhost.conf.j2` and `foreman-vhost.conf.j2` before the catch-all `ProxyPass /`, so `/rhsm` and `/register` match first while all other traffic (Web UI, API, Pulp) is unaffected.

## Configuration

Disabled by default (`httpd_registration_admission_max: 0`). Tuning profiles set values based on `puma_workers * threads * 5`:

| Profile | Value |
|---|---:|
| default | 0 (disabled) |
| medium | 300 |
| large | 600 |
| extra-large | 1200 |
| extra-extra-large | 2400 |

## Results (16-CPU large Satellite)

| Level | Without throttle | With throttle (max=600) |
|---:|---:|---:|
| 912 | 72.8% | **100%** |
| 1064 | 71.8% | **99.8%** |
| 1368 | 72.5% | **97.0%** |

## Note

A similar mechanism could be implemented for foreman-installer deployments via `puppet-foreman` using the existing `foreman::config::apache::fragment` mechanism.

## How to test

1. Deploy with a tuning profile: `foremanctl deploy --tuning large`
2. Verify throttle block: `grep foreman-registration /etc/httpd/conf.d/foreman-ssl.conf`
3. Verify config syntax: `httpd -t`
4. Register hosts concurrently and observe improved pass rates at high concurrency


ehelms:
Are there downsides or trade-offs to this approach? Does it increase memory usage temporarily for example?

pablomh:
Admission control (AC) adds essentially zero overhead. In our testing with a medium-sized Satellite, peak memory is within the baseline range (25.0-25.3G baseline vs 25.1G with AC). httpd RSS is actually lower with AC (1251M vs 1268-1329M baseline) because the balancer prevents httpd from spawning connections it can't service. Puma RSS is slightly higher (12.1G vs 10.7-11.9G) because more requests complete successfully instead of being dropped — Puma does more useful work. CPU is flat.

jakduch (commented):
I think the admission-control semantics need another pass before this leaves draft. Apache documents `max` as a per-child-process connection-pool limit; pools are not coordinated between httpd processes. With `ThreadsPerChild: 16`, values such as 150 or 600 therefore do not act as the advertised global cap.

There is no non-blocking queue here either. Without `acquire` or a balancer wait timeout, exhaustion returns `SERVER_BUSY`; the existing `timeout=900` is the backend I/O timeout.

The measured improvement may still be real, but it could come from using an explicit reusable proxy worker rather than from `max`. Could we rerun the comparison with `max=1`, `16`, and `600` while observing the actual Puma connection count?

Apache reference: https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypass

jakduch:
CI note: the CentOS Stream 9 failure is unrelated to this change. Downloading the Ansible collection failed with HTTP 522, which then left `theforeman.forklift.sos_report` unavailable. The aggregate `Test suite` job only reports that matrix failure.

This does not invalidate the measured results, but the per-httpd-child `BalancerMember max` semantics described in the review still need to be resolved.

Issue #497, opened 2026-05-06
Intermittent test failure on smoker UI test of subscriptions page


ehelms:
See https://github.com/Katello/katello/pull/11728

PR #500, opened 2026-05-07
Load boxes.yaml

ianballou:

* Adds hostmanager config, mainly so that the hypervisor's `/etc/hosts` is updated along with the VMs' hosts files.
* Increases the dev box disk size to 50 to allow more space for long-lasting logs and Pulp.
* Allows users to create a boxes.yaml file to add to the default box list in the Vagrantfile.

Steps to reproduce:

* Try to create a boxes.yaml like

```yaml
---
# User-local box definitions (this file is gitignored)
# Each entry becomes a vagrant box. Available settings:
#   box:       base box (default: centos/stream9)
#   hostname:  FQDN (default: <name>.example.com)
#   memory:    RAM in MB (default: 2048)
#   cpus:      CPU count (default: 1)
#   disk_size: disk in GB (default: unset)

katello-production:
  memory: 12288
  cpus: 4
  disk_size: 50
```

* Try `vagrant-hostmanager` and ensure the hypervisor's hosts file is updated.

#### Checklist

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -26,7 +26,7 @@ Vagrant.configure("2") do |config|
     override.vm.provider "libvirt" do |libvirt, provider|
       libvirt.memory = 10240
       libvirt.cpus = 4
-      libvirt.machine_virtual_size = 30
+      libvirt.machine_virtual_size = 50
     end
   end
 
@@ -45,6 +45,23 @@ Vagrant.configure("2") do |config|
 
     override.vm.provider "libvirt" do |libvirt, provider|
       libvirt.memory = 2048
+    end
+  end
+
+  # Load user-local box definitions from boxes.yaml (gitignored)
+  boxes_yaml = File.join(__dir__, 'boxes.yaml')
+  if File.exist?(boxes_yaml)
+    user_boxes = YAML.safe_load(File.read(boxes_yaml)) || {}
+    user_boxes.compact.each do |name, settings|
+      config.vm.define name do |override|
+        override.vm.box = settings.fetch('box') { ENV.fetch('FOREMANCTL_BASE_BOX', 'centos/stream9') }
+
+        override.vm.provider "libvirt" do |libvirt, _provider|
+          libvirt.memory = settings.fetch('memory', 3072)
+          libvirt.cpus = settings.fetch('cpus', 1)
+          libvirt.machine_virtual_size = settings.fetch('disk_size', 50)
+        end
+      end
     end
   end
 end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,4 +1,12 @@
+require 'yaml'
+
 Vagrant.configure("2") do |config|
+  if Vagrant.has_plugin?('vagrant-hostmanager')
+    config.hostmanager.enabled = true
+    config.hostmanager.manage_host = true
+    config.hostmanager.manage_guest = true
+    config.hostmanager.include_offline = true
+  end
   config.vm.synced_folder ".", "/vagrant"
 
   config.vm.provision("etc_hosts", type: 'ansible') do |ansible|
@@ -17,7 +25,7 @@ Vagrant.configure("2") do |config|
     override.vm.provider "libvirt" do |libvirt, provider|
       libvirt.memory = 10240
       libvirt.cpus = 4
-      libvirt.machine_virtual_size = 30
+      libvirt.machine_virtual_size = 50
     end
   end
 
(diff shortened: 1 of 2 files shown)


ianballou:
How do we want to advertise this boxes.yaml file? Docs?

Also this fixes https://github.com/theforeman/foremanctl/issues/437

ehelms:
The failure was due to https://github.com/theforeman/foremanctl/issues/497

Vagrantfile
```diff
@@ -38,4 +46,23 @@ Vagrant.configure("2") do |config|
       libvirt.memory = 2048
     end
   end
+
+  # Load user-local box definitions from boxes.yaml (gitignored)
```
ehelms, line 51: Reminds me we should clean up the gitignore as it has some forklift-isms.

ekohl (commented):
I'm never a fan of `vagrant-hostmanager`. Since https://github.com/theforeman/foremanctl/pull/305 we configure the management network to have a DNS domain set so you actually have DNS between guests. The `etc_hosts` hack is no longer needed then.

The benefit is that it really works as a production system. Ruby's [Resolv](https://rubyapi.org/o/resolv) module famously has a line:
> /etc/nsswitch.conf is not supported.

It may not be obvious, but glibc reads `/etc/nsswitch.conf` and if there's a `files` entry for `hosts` then it reads `/etc/hosts`. Foreman uses `Resolv` in some of its codebase, so 

However, vagrant-libvirt only applies that to newly created networks so we should get people to migrate.

To also get the host to respond I use https://github.com/voxpupuli/setup-vagrant/blob/v0/network-hook to configure resolvectl. The only problem I have is that SELinux on my Fedora doesn't allow running Python for hooks. I think https://bugzilla.redhat.com/show_bug.cgi?id=2456505 is the same problem.

ianballou:
> I'm never a fan of `vagrant-hostmanager`. Since #305 we configure the management network to have a DNS domain set so you actually have DNS between guests. The `etc_hosts` hack is no longer needed then.
> 
> The benefit is that it really works as a production system. Ruby's [Resolv](https://rubyapi.org/o/resolv) module famously has a line:
> 
> > /etc/nsswitch.conf is not supported.
> 
> It may not be obvious, but glibc reads `/etc/nsswitch.conf` and if there's a `files` entry for `hosts` then it reads `/etc/hosts`. Foreman uses `Resolv` in some of its codebase, so
> 
> However, vagrant-libvirt only applies that to newly created networks so we should get people to migrate.
> 
> To also get the host to respond I use https://github.com/voxpupuli/setup-vagrant/blob/v0/network-hook to configure resolvectl. The only problem I have is that SELinux on my Fedora doesn't allow running Python for hooks. I think https://bugzilla.redhat.com/show_bug.cgi?id=2456505 is the same problem.

It's great to hear that we can rely on proper DNS between vagrant hosts. Having the hypervisor be able to connect to vagrant hosts is the more common issue (to me), so I'd like something 'official' for developers to follow. Do you think the `network-hook` that you mentioned could suit this? Whatever we end up going with, I'd like to have it in the developer documentation. If `network-hook` is a decent path forward, I'm happy to update the PR here to include a mention of it.

ekohl:
I found a solution to the SELinux issue and wrote up my notes in https://community.theforeman.org/t/setting-up-libvirt-with-dns-resolution-for-vagrant/46567.

Vagrantfile
```diff
@@ -38,4 +46,23 @@ Vagrant.configure("2") do |config|
...
+  boxes_yaml = File.join(__dir__, 'boxes.yaml')
+  if File.exist?(boxes_yaml)
+    user_boxes = YAML.safe_load(File.read(boxes_yaml)) || {}
+    user_boxes.each do |name, settings|
+      next if settings.nil?
```
ekohl:
```suggestion
    user_boxes.compact.each do |name, settings|
```

Vagrantfile
```diff
@@ -38,4 +46,23 @@ Vagrant.configure("2") do |config|
...
+      config.vm.define name do |override|
+        override.vm.box = ENV.fetch("FOREMANCTL_BASE_BOX", settings.fetch('box', 'centos/stream9'))
```
ekohl: Should the order be reversed and have the settings take priority?
```suggestion
        override.vm.box = settings.fetch('box') { ENV.fetch('FOREMANCTL_BASE_BOX', 'centos/stream9') }
```

Vagrantfile
```diff
@@ -38,4 +46,23 @@ Vagrant.configure("2") do |config|
...
+        override.vm.hostname = settings.fetch('hostname', "#{name}.example.com")
```
ekohl: Once you rebase, please respect the Vagrant domain here.

ekohl (commented):
I'd prefer to steer away from `vagrant-hostmanager` and direct to the DNS-based solution, but can see value in a `boxes.yaml`. Perhaps split it up?

Vagrantfile
ianballou: I thought about this, and I arrived at the conclusion that it's probably unlikely for people to be using the FOREMANCTL_BASE_BOX env var when they could just define it in their boxes.yaml.  If you're already customizing the base box, it is the same effort and leaves a better paper trail to set it in boxes.yaml than the environment variable.  I even considered removing the FOREMANCTL_BASE_BOX option from this but we might as well keep it so it's generally respected.

Anyway, long-winded way of agreeing with you.

ianballou:
@ekohl to not let this linger longer I removed out the vagrant-hostmanager bits and scoped  this down to only load the boxes.yaml file. I snuck in a disk boost to 50 GB since it's such a pain when you run out of disk, and 30 is so little to be running Pulp for a long time.

ehelms (approved):

PR #501, opened 2026-05-08
Add Podman quadlet .image file support

### LINKED ISSUE #277: Define .image files for a single definition of images
Today we have multiple files that all write down the image, but podman supports defining `.image` files. That means we can have both `foreman.container` and `dynflow-sidekiq-%i` refer to a `foreman.image`. In this .image file we can define the the exact image to be used in a single place.

This can also be interesting for disconnected installations. The documentation states you can use:

```ini
[Image]
Image=docker-archive:/tmp/archive-file.tar
ImageTag=quay.io/podman/stable:latest
```

This could save writing custom setup scripts and simply deploy a drop in file.

ehelms:

Container image references were embedded directly in each service role as variable defaults, making it difficult to override images for product builds, disconnected installs, or developer testing without modifying foremanctl-managed files.

Resolves: #277

**Image unit generation**

* Introduce a shared `images` role with `deploy_image.yaml` that generates a `.image` quadlet unit file and `.image.d/` drop-in directory for each container image
* Each service role gains an `image.yaml` task that calls the `images` role with its image name and tag, removing the per-role `*_container_image` variable defaults
* Container roles now reference `Image=<name>.image` instead of full registry URLs — systemd ensures the image unit is active before any dependent container starts

**Drop-in override hierarchy**

Three tiers of override are supported without touching foremanctl-managed files:

| Layer | Path | Who writes it |
|---|---|---|
| Vendor/RPM | `/usr/share/containers/systemd/<name>.image.d/10-product.conf` | Product RPM |
| ISO/archive | `/usr/share/containers/systemd/<name>.image.d/20-archive.conf` | ISO extraction |
| User | `/etc/containers/systemd/<name>.image.d/90-user.conf` | Operator |

The last `.image.d` drop-in wins. `registries.conf` is the right tool when a private registry mirrors upstream names exactly; `.image.d` drop-ins are required when image names or tags change (e.g. `foreman:nightly` → `foreman-rhel9:stream`).

**Authenticated registry handling**

Each image unit gets a systemd service drop-in at `/etc/systemd/system/<name>-image.service.d/auth.conf` that sets `REGISTRY_AUTH_FILE=/etc/foreman/registry-auth.json`. Users with authenticated registries run `podman login <registry> --authfile=/etc/foreman/registry-auth.json` before deploying; podman silently ignores the file if it does not exist.

**pull-images**

`pull-images` deploys image units (respecting any vendor drop-ins already in place), then creates a temporary `Policy=always` drop-in (`00-pull-always.conf`) for each image, performs a `daemon_reload`, restarts all `*-image.service` units in parallel, and removes the drop-ins in an `always:` block so cleanup happens even on failure. The pull honours any `.image.d` overrides already present.

**Migration containers converted to quadlet oneshot services**

`foreman-db-migrate` and `pulpcore-manager-migrate` are now declared as quadlet `.container` units with `Type=oneshot` rather than being run as ephemeral containers via `podman_container` with `detach: false`. Ansible starts them with `state: started` and waits via `async`/`async_status`. This keeps migration runs in the systemd journal and makes failure state visible via `systemctl status`.

**Tests**

* `tests/images_test.py` — core image files, `.image.d/` drop-in directories, `*-image` service existence, auth drop-in presence and content, postgresql conditional on `database_mode`, foreman-proxy conditional on `enabled_features`
* `tests/iop/images_test.py` — same checks for all IOP image units, gated on the `iop` feature marker

1. Deploy with `./foremanctl deploy`
2. Confirm `.image` files exist under `/etc/containers/systemd/` for each service and that `<name>-image.service` units are present
3. Confirm `/etc/systemd/system/<name>-image.service.d/auth.conf` exists for each image
4. Run `./foremanctl pull-images` and verify images are pulled; re-run and confirm it pulls again (Policy=always enforced, not just missing)
5. Place a drop-in override:
   ```
   # /etc/containers/systemd/foreman.image.d/90-user.conf
   [Image]
   Image=quay.io/foreman/foreman:pr-12345
   ```
   Run `systemctl daemon-reload` and confirm `systemctl cat foreman-image.service` reflects the override
6. Run `./forge test` and confirm `images_test.py` and `iop/images_test.py` pass

#### Checklist


ekohl:
Do you intend to resolve https://github.com/theforeman/foremanctl/issues/277 with this?

ekohl:
> Add a Podman quadlet `.image` unit system with three-tier precedence: admin overrides (`/etc/foremanctl/images.d/`) > vendor/RPM overrides (`/usr/share/foremanctl/images.d/`) > generated defaults from Ansible variables

I'm wondering if we can leverage the built in `/usr/share/containers/systemd/` somehow. It's a major change, but I've been wondering if we should ship the container definitions in an RPM and deploy them to `/usr/share/containers/systemd`, letting foremanctl only add overrides in `/etc/containers/systemd`.

Also not sure if you can use `/etc/containers/systemd/foreman.image.d/override.conf` to replace parts of it. If so, perhaps an RPM can ship `/usr/share/containers/systemd/foreman.image.d/disconnected.conf` to provide disconnected installation instructions.

ehelms:
> I'm wondering if we can leverage the built in `/usr/share/containers/systemd/` somehow. It's a major change, but I've been wondering if we should ship the container definitions in an RPM and deploy them to `/usr/share/containers/systemd`, letting foremanctl only add overrides in `/etc/containers/systemd`.

As we stated before, tieing ourselves to an RPM laying down some base container will slow us down as any time it requires a change we have to wait for the RPM to cycle and it would split the base from any overrides. I think we would end up just deploying everything via overrides. Just feel like friction right now.

I did think about this and here is an alternative using this concept:


### Updated Proposal

  Place .image files alongside .container files in /etc/containers/systemd/, using quadlet's native drop-in mechanism for overrides.

  Layout:

  /etc/containers/systemd/
    foreman.image                        # base, templated by foremanctl
    foreman.image.d/
      50-disconnected.conf               # optional: disconnected registry override
      90-user.conf                       # optional: user override

  Precedence (last wins):

  1. foreman.image -- foremanctl default
  2. 50-disconnected.conf -- disconnected layer
  3. 90-user.conf -- user layer

  Numbered prefixes enforce ordering with gaps for future layers.

  What this eliminates:

  - Custom /etc/foremanctl/images.d/ and /usr/share/foremanctl/images.d/ directories
  - Manual stat/copy/conditional precedence logic in deploy_image.yaml
  - The entire images role

  What stays the same:

  - Container roles reference Image=foreman.image (no change)
  - foremanctl templates the base .image file (just changes the destination)
  - daemon-reload after deployment (already happening)

ekohl:
> Updated Proposal

This is what I had in mind.

> Custom /etc/foremanctl/images.d/ and /usr/share/foremanctl/images.d/ directories

This is IMHO the big benefit.

evgeni:
I like the idea!

There is an alternative approach from @bochi in #503, which is much simpler in the diff size, but requires custom code vs using native systemd/podman override options. Mostly mentioning it here as I wonder if this PR would also solve their usecase.

bochi:
i like this approach too and will change my PR to only have the feature addition but no changes to image locations. looking forward to this getting merged 🙂

ianballou:
Being able to override the container images easily would be really beneficial for development. If we eventually have containers built from PRs much like packit does with RPMs today, you could just drop those in. Or, when we need to test new versions of Pulp, we could simply drop that new Pulp container in (https://github.com/theforeman/pulp-oci-images/pull/13).

docs/developer/deployment.md
```diff
@@ -52,23 +52,85 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+```
+
+All containers that share a base image (e.g., foreman, dynflow-sidekiq, foreman-recurring) reference the same `.image` unit. systemd ensures the image is pulled before any dependent container starts.
+
+#### Image Precedence
+
+During `foremanctl deploy`, foremanctl resolves which `.image` files to place in `/etc/containers/systemd/` using a three-tier precedence model:
+
+1. **Admin overrides** (`/etc/foremanctl/images.d/`) — highest priority. Place custom `.image` files here to override any default.
+2. **Vendor/RPM overrides** (`/usr/share/foremanctl/images.d/`) — used by packaging or vendor layers to set image sources via RPM.
```
ekohl: Do we still want this or rather fully lean into the native systemd/podman? Did you try mixing drop-ins in `/usr` while the base definition is in `/etc/`?

src/roles/candlepin/tasks/main.yml
```diff
@@ -55,17 +58,10 @@
...
-    name: "{{ candlepin_container_image }}:{{ candlepin_container_tag }}"
-    state: present
-  environment:
-    REGISTRY_AUTH_FILE: "{{ candlepin_registry_auth_file }}"
-
 - name: Deploy Candlepin quadlet
   containers.podman.podman_container:
     name: "candlepin"
-    image: "{{ candlepin_container_image }}:{{ candlepin_container_tag }}"
+    image: candlepin.image
```
ekohl, line 64: I like how this mostly makes this the whole definition almost (except the hostname and `candlepin_database_ssl_ca_path`) static. Works towards our goal of only configuring things via secrets.

src/roles/iop_advisor_frontend/defaults/main.yaml
```diff
@@ -1,6 +1,5 @@
 ---
 iop_advisor_frontend_container_image: "quay.io/iop/advisor-frontend"
 iop_advisor_frontend_container_tag: "foreman-3.18"
```
ekohl, line 3: Unrelated, but shouldn't this follow nightly?

tests/images_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+@pytest.fixture(params=CORE_IMAGES)
+def core_image(request):
+    return request.param
+
+
+def test_image_file_exists(server, core_image):
+    image_file = server.file(f"/etc/containers/systemd/{core_image}.image")
+    assert image_file.exists
+    assert image_file.is_file
```
ekohl: IMHO a good test only has a single `assert` statement and pytest is great at introspecting variables so I'd prefer
```suggestion
    assert image_file.exists and image_file.is_file
```

docs/developer/deployment.md
```diff
@@ -52,23 +52,84 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
-1. **Login to your registry** using Podman and save credentials to the default auth file location:
+**Upstream default (no user action):** foremanctl generates `.image` files from its built-in `images.yml`:
+
+```ini
+# /etc/containers/systemd/foreman.image (generated by foremanctl)
+[Image]
+Image=quay.io/foreman/foreman:nightly
+```
+
+**User's own registry:** The user creates a `90-user.conf` drop-in to point at their own registry. This overrides the base layer:
```
ekohl: This particular use case may be easier with [containers-registries.conf](https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md). From reading the docs I think adding the following to `/etc/containers/registries.conf` should work:
```ini
[[registry]]
prefix = "quay.io/theforeman"
location = "katello.example.com/Default_Organization"
```

ekohl (approved):

docs/developer/deployment.md
```diff
@@ -52,23 +52,84 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+1. **Login to your registry** using Podman and save credentials to an auth file:
 ```bash
 podman login <registry> --authfile=/etc/foreman/registry-auth.json
 ```
 
-2. **Deploy as usual** - foremanctl will automatically detect and use the authentication file:
+2. **Set the global auth file** so all generated `.image` units include it:
 ```bash
-./foremanctl deploy
+./foremanctl deploy --registry-auth-file=/etc/foreman/registry-auth.json
```
arvind4501: I don't see `--registry-auth-file` as exposed options in https://github.com/theforeman/foremanctl/blob/master/src/playbooks/deploy/metadata.obsah.yaml , i think that should be added

src/playbooks/pull-images/pull-images.yaml
```diff
@@ -11,27 +11,49 @@
   roles:
     - role: pre_install
   post_tasks:
-    - name: Pull an image
-      containers.podman.podman_image:
+    - name: Deploy core image units
```
arvind4501, line 14: From what i understand is this  only writes .image files to /etc/containers/systemd/ and does a daemon_reload but does that actually pulls a image? i mean does it download the images that can be used without re-pulling in deploy?

arvind4501: https://docs.podman.io/en/latest/markdown/podman-image.unit.5.html#usage-summary says `enerating a systemd .service that runs podman image pull.` is confusing, does it pull by default or on start of .service

ekohl: I think it generates a `.service` (like `foreman.image.service`) file but it requires someone/something to start it. Whether that's a user running `systemctl` or another service depending on it (I'd expect `foreman.service` to).

In this case I'd expect `foremanctl pull-images` to ensure the service is started, but some questions that pop up:
* What if the service (like `foreman.image.service`) was already started? Is it a noop or does it run again?
* What happens if an updated image is pulled? Does it restart dependent services?

Looking at the docs [Policy](https://docs.podman.io/en/latest/markdown/podman-image.unit.5.html#policy-always) influences this.

ehelms: The quadlet generator makes each .image unit a Type=oneshot service. When a container starts, systemd activates its image service dependency, which runs podman image pull (or no-ops if the image already exists), then exits. The service transitions active → inactive when the oneshot finishes.

This strategy has lots of benefits and one small downside, on systemctl start foreman.target:

  - Each container Requires its image service
  - Since the image services are inactive (they completed as oneshots), systemd re-runs them
  - Each one checks whether the image exists, which is fast if it's cached — but with 20 image services being re-checked concurrently on a busy system, it adds latency

Because of this I had to increase the foreman.target retries when we test start/stop/restart of it.

src/roles/foreman/tasks/main.yaml
```diff
@@ -1,10 +1,6 @@
 ---
-- name: Pull the Foreman container image
-  containers.podman.podman_image:
-    name: "{{ foreman_container_image }}:{{ foreman_container_tag }}"
-    state: present
-  environment:
-    REGISTRY_AUTH_FILE: "{{ foreman_registry_auth_file }}"
+- name: Deploy foreman image
+  ansible.builtin.include_tasks: image.yaml
```
evgeni, line 3: what's the reason behind using `include_tasks` where the tasks file only has a single task?

docs/developer/deployment.md
ehelms: This has been cleaned up.

src/roles/foreman/tasks/main.yaml
ehelms: So that it can be re-used by the pull-images command -- https://github.com/theforeman/foremanctl/pull/501/changes#diff-378b85a60f2035996631e7a45762e3f8173ba482dbe8ccdf9e992beda3f1fb3aR15

src/roles/images/tasks/deploy_image.yaml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Generate image file for {{ images_definition.name }}
+  ansible.builtin.template:
```
evgeni: there is also `containers.podman.podman_image`, did you compare the two approaches?

src/roles/images/tasks/deploy_image.yaml
```diff
@@ -0,0 +1,12 @@
...
+    src: image.j2
+    dest: "{{ images_quadlet_dir }}/{{ images_definition.name }}.image"
+    mode: "0644"
+
+- name: Create drop-in directory for {{ images_definition.name }}
```
evgeni, line 15: we don't deploy anything here, right? it's just "so it's there so someone else can do it"?

src/roles/images/tasks/deploy_image.yaml
ehelms: Huh, I thought I had looked at that and didn't see any quadlet support, but there it is. I'll compare.

src/vars/images.yml
```diff
@@ -1,23 +1,4 @@
...
-iop_remediation_registry_auth_file: "{{ registry_auth_file }}"
-iop_advisor_registry_auth_file: "{{ registry_auth_file }}"
-iop_gateway_registry_auth_file: "{{ registry_auth_file }}"
-iop_engine_registry_auth_file: "{{ registry_auth_file }}"
-iop_yuptoo_registry_auth_file: "{{ registry_auth_file }}"
-iop_puptoo_registry_auth_file: "{{ registry_auth_file }}"
-iop_ingress_registry_auth_file: "{{ registry_auth_file }}"
-iop_vulnerability_frontend_registry_auth_file: "{{ registry_auth_file }}"
-iop_advisor_frontend_registry_auth_file: "{{ registry_auth_file }}"
+images_registry_auth_file: ""
```
evgeni: the old default was /etc/foreman/registry-auth.json, why not anymore?

src/roles/foreman/tasks/main.yaml
evgeni: sneaky!

docs/developer/deployment.md
```diff
@@ -52,23 +52,122 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+
+```ini
+# /etc/containers/systemd/foreman.image.d/90-user.conf
+[Image]
+Image=quay.io/foreman/foreman:pr-12345
+```
+
+#### Authenticated Registry Handling
+
+Podman searches standard credential locations automatically — no foremanctl configuration required. Login once before deploying:
```
evgeni: What are those locations? I thought `podman login` by default stores the token in `/run`, which is gone on the next boot.

src/roles/images/tasks/deploy_image.yaml
ehelms: Yes. So, arguably, I can drop this here and just let whatever needs to do this create them (e.g. downstreams with different images or upstream release RPMs). Thoughts?

evgeni: I thought packages would not touch it here anyway but use /usr/lib?

src/playbooks/pull-images/pull-images.yaml
ekohl: The docs mention `Policy=always` and that implies it always connects over the network. Perhaps we should use `Policy=missing` so it's only a local operation after the first installation?

Then we do need an explicit process to update images, but we already have an update and upgrade guide. Perhaps that's good anyway? Otherwise you can unexpectedly pull in z-streams after a reboot.

tests/iop/images_test.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+pytestmark = pytest.mark.iop
```
ekohl: In b6d8f08c488d5810146631ff281d4627f84e086f we introduced feature markers
```suggestion
pytestmark = pytest.mark.feature("iop")
```

src/roles/images/tasks/deploy_image.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Generate image file for {{ images_definition.name }}
+  containers.podman.podman_image:
+    name: "{{ images_definition.image }}"
+    state: quadlet
+    quadlet_dir: "{{ images_quadlet_dir }}"
+    quadlet_filename: "{{ images_definition.name }}"
+    quadlet_file_mode: "0644"
+    quadlet_options: "{{ ['AuthFile=' + images_registry_auth_file] if images_registry_auth_file else [] }}"
```
ekohl: There's a dedicated `auth_file` option: https://docs.ansible.com/projects/ansible/latest/collections/containers/podman/podman_image_module.html#parameter-auth_file
```suggestion
    auth_file: "{{ images_registry_auth_file }}"
```

src/roles/images/tasks/deploy_image.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Generate image file for {{ images_definition.name }}
+  containers.podman.podman_image:
```
ekohl, line 3: Following up on https://github.com/theforeman/foremanctl/pull/501#discussion_r3267447972: I think we should set `Policy=missing` so it doesn't automatically update on every reboot/restart.

Then we can use [podman auto-update](https://docs.podman.io/en/latest/markdown/podman-auto-update.1.html) to force updates.

ekohl (changes requested):

src/roles/images/tasks/deploy_image.yaml
ehelms: `/usr/share` but yes you are correct. This override directory would be used for these two scenarios:

- development / testing image overrides (e.g. developers being able to do https://github.com/theforeman/pulp-oci-images/pull/13 and then drop it in and test it)
- "hotfix" or modified containers with a different name:tag

One argument is these are specialized and thus the directory should be created at the time of use. This adds a step the user or developer has to remember. Whereas, having the directory allows quick and easy drop in with less overhead remembering two steps.

What do you think?

src/roles/images/tasks/deploy_image.yaml
ehelms: That's only consumed by Ansible when performing image operations. We need this to persist and exist in the quadlet definition file.

src/roles/images/tasks/deploy_image.yaml
ekohl: > One argument is these are specialized and thus the directory should be created at the time of use. This adds a step the user or developer has to remember. Whereas, having the directory allows quick and easy drop in with less overhead remembering two steps.

IMHO you can't count on things existing just because Ansible created it so I'd prefer the tools to be robust by having a `mkdir -p` in there. So my suggestion is to remove the drop-in directory from Ansible.

src/roles/images/tasks/deploy_image.yaml
ekohl: I think that's a bug so I filed https://github.com/containers/ansible-podman-collections/issues/1034.

src/roles/images/tasks/deploy_image.yaml
```diff
@@ -0,0 +1,17 @@
...
+    name: "{{ images_definition.image }}"
+    state: quadlet
+    quadlet_dir: "{{ images_quadlet_dir }}"
+    quadlet_filename: "{{ images_definition.name }}"
+    quadlet_file_mode: "0644"
+    quadlet_options:
+      - "Policy=missing"
+      - "Environment=REGISTRY_AUTH_FILE={{ images_registry_auth_file }}"
```
ekohl: Per https://github.com/containers/ansible-podman-collections/issues/1034#issuecomment-4491635048 you can use `authfile`:
```suggestion
    authfile: "{{ images_registry_auth_file }}"
```

ehelms: Nope. This will set `AuthFile`  and this will enforce the existence of this file. What we want is to set the environment variable like we do in puppet-iop.

ehelms:
I'm not yet sure what change in the design is not leading to the failure. I will need to follow up with deeper investigation.

src/playbooks/pull-images/pull-images.yaml
arvind4501: > When a container starts, systemd activates its image service dependency, which runs podman image pull

so technically it happens when containers start not when we run `foremanctl pull-images`? thats seems odd with what we expect from pull-images, as a user i can pull-images and then deploy which reduces the deployment time as i already have latest images.
or i am interpreting wrong?

src/roles/images/tasks/deploy_image.yaml
arvind4501:
```suggestion
      - "AUTH_FILE={{ images_registry_auth_file }}"
```
we can't use Environment or REGISTRY_AUTH_FILE directly in `[Image]`  section as thats not one of the valid options(https://docs.podman.io/en/latest/markdown/podman-image.unit.5.html#options)
we could use REGISTRY_AUTH_FILE but not directly in .image file  https://docs.podman.io/en/latest/markdown/podman-image.unit.5.html#authfile-path

arvind4501: and mostly CI is also failing because of same reason

```
systemd[1]: Reloading.
May 20 09:47:23 quadlet.example.com quadlet-generator[34868]: converting "postgresql.image": unsupported key 'Environment' in group 'Image' in /etc/containers/systemd/postgresql.image
May 20 09:47:23 quadlet.example.com quadlet-generator[34868]: processing encountered some errors
May 20 09:47:23 quadlet.example.com systemd-rc-local-generator[34879]: /etc/rc.d/rc.local is not marked executable, skipping.
May 20 09:47:23 quadlet.example.com systemd[34865]: /usr/lib/systemd/system-generators/podman-system-generator failed with exit status 1.
```

ehelms: Here are the options I've come up with on how we could proceed and handle this. To recap, the previous method used the environment variable approach is this will not fail if the auth file is missing but will activate if it's present making this fairly straight forward for the end user.


  1. [Service] section with Environment=REGISTRY_AUTH_FILE= via quadlet_options
  
  The quadlet generator supports injecting a [Service] section into .image files, which passes through to the generated <name>-image.service. Setting REGISTRY_AUTH_FILE there means Podman
  has the env var set when the service runs — if the file exists, credentials are used; if it doesn't, Podman falls back gracefully to unauthenticated pulls. No hard failure. This is exactly
   the puppet-iop design you described, just expressed via quadlet rather than a container env var. The default path `/etc/foreman/registry-auth.json` is preserved as a convention.

```
  quadlet_options:
    - "Policy=missing"
    - |
      [Service]
      Environment=REGISTRY_AUTH_FILE={{ images_registry_auth_file }}
```
---
  2. --registry-auth-file as a foremanctl CLI parameter
  
  Add the parameter to the deploy playbook's metadata.obsah.yaml with no default (or empty default). AuthFile= is only written to image units when the user explicitly passes the parameter —
  which means fresh systems with no auth configured get no `AuthFile=` and no failure. The value persists across runs via obsah's state mechanism. Users with authenticated registries run
  `./foremanctl deploy --registry-auth-file=/etc/foreman/registry-auth.json` once.

  Clean and explicit, but requires user action to activate auth rather than just placing the file.

  ---
  3. Create `/etc/foreman/registry-auth.json` in pre_install
  
  The reverted approach. `{}` is valid empty Podman auth JSON, force: false never overwrites existing credentials, and `AuthFile=` always resolves. Simple, zero conditional logic.

  ---
  4. Stat-based conditional AuthFile=
  
  Check whether the auth file exists on the managed node before deploying each image unit, include `AuthFile=` only when it does. Adds a stat task per image invocation and timing sensitivity —
   the file must exist at deploy time to take effect.

evgeni: I think I prefer **1** as that's what we use today and it "just works" with no additional user interaction etc.

**2** has the downside that the user needs to pass the same filename to different tools (once to `podman login`, once to `foremanctl`) at IMHO no real benefit (I don't think anyone wants to customize the path to that file)
**3** seems like a hack
**4** would result in ugly Jinja conditions for when the file is absent, at no real benefit over the `Service.Environment` option 1

We could explore making `foremanctl` the one doing the login to the registry (or re-using the sub-man certs), but I think that's a separate issue, outside of the "let's use .image files" PR.

ehelms:
I have reworked this PR based on testing and learning between the use of `registries.conf` and `.image` quadlets. The PR description template has been updated based on the current design. Generally, this will use of the combination of the two depending on situation and where each is applicable.

docs/developer/deployment.md
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+
+```ini
+# /etc/containers/systemd/foreman.container (excerpt)
+[Container]
+Image=foreman.image
+```
+
+All containers that share a base image (e.g., foreman, dynflow-sidekiq, foreman-recurring) reference the same `.image` unit. systemd ensures the image is pulled before any dependent container starts.
+
+#### Image Overrides via Drop-ins
```
ekohl, line 75: Perhaps add a link to https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html because those are the official docs? I also just learned something new by reading that:
> If the unit contains dashes (“-“) in the name (foo-bar-baz.container), then the drop-in directories generated by truncating the name after the dash are searched as well (foo-.container.d and foo-bar-.container.d).

Not that I see an immediate use for it, but still good to keep in mind.

docs/developer/deployment.md
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+1. `foreman.image` — foremanctl default from `images.yml`
+2. `10-product.conf` — vendor/RPM provided
+3. `20-archive.conf` — ISO or archive extraction provided
+4. `90-user.conf` — user provided (highest priority)
+
+#### registries.conf vs .image.d drop-ins
+
+Both `registries.conf` and `.image.d` drop-ins can redirect where an image is pulled from, but they behave differently and suit different use cases.
+
+`registries.conf` applies a transparent redirect at pull time — the image is fetched from the `location` registry but stored in local storage under the original `prefix` name. This means `podman images` shows the upstream name (e.g., `quay.io/foreman/foreman:nightly`), and the `.image` quadlet continues to reference that same name. This works well when the private registry mirrors upstream image names and tags exactly.
```
ekohl, line 104: Link to https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md as the official docs?

docs/developer/deployment.md
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+Precedence (last wins):
+
+1. `foreman.image` — foremanctl default from `images.yml`
+2. `10-product.conf` — vendor/RPM provided
+3. `20-archive.conf` — ISO or archive extraction provided
+4. `90-user.conf` — user provided (highest priority)
+
+#### registries.conf vs .image.d drop-ins
+
+Both `registries.conf` and `.image.d` drop-ins can redirect where an image is pulled from, but they behave differently and suit different use cases.
```
ekohl, line 102: Documenting both use cases is excellent.

docs/developer/deployment.md
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+
+`registries.conf` applies a transparent redirect at pull time — the image is fetched from the `location` registry but stored in local storage under the original `prefix` name. This means `podman images` shows the upstream name (e.g., `quay.io/foreman/foreman:nightly`), and the `.image` quadlet continues to reference that same name. This works well when the private registry mirrors upstream image names and tags exactly.
+
+`.image.d` drop-ins directly replace the `Image=` value in the quadlet unit. The image is pulled from and stored under the new reference. This is required when the image name or tag changes completely (e.g., `quay.io/foreman/foreman:nightly` → `registry.example.com/org/foreman-rhel9:stream`), since `registries.conf` cannot remap image names — only registry/namespace locations.
```
ekohl, line 106: Should this point to https://docs.podman.io/en/latest/markdown/podman-image.unit.5.html?

docs/developer/deployment.md
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+
+No user action required beyond installing the RPM and logging into the product registry:
 
-1. **Login to your registry** using Podman and save credentials to the default auth file location:
 ```bash
-podman login <registry> --authfile=/etc/foreman/registry-auth.json
+podman login registry.example.com --authfile=/etc/foreman/registry-auth.json
+```
+
+##### Disconnected install from ISO
```
ekohl: ISO is a very specific distribution format.
```suggestion
##### Disconnected install from local media
```

docs/developer/deployment.md
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+
+`registries.conf` cannot express non-registry sources. ISO extraction places `.image.d` drop-ins that redirect pulls to local archive files:
```
ekohl: Perhaps explain the use case a bit more?

```suggestion
In air gapped environments the container images need to be brought in.`registries.conf` cannot express non-registry sources, but Podman can read the image from a local archive file. Users can bring these in via local media such as USB sticks. See [podman export](https://docs.podman.io/en/latest/markdown/podman-export.1.html) for more information.
```

docs/developer/deployment.md
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+# /usr/share/containers/systemd/foreman.image.d/20-archive.conf (from ISO)
+[Image]
+Image=docker-archive:/opt/foreman/images/foreman-6.17.tar
+```
+
+##### User's own registry
+
+When the private registry mirrors upstream image names and tags exactly, `registries.conf.d` handles namespace-level remapping. foremanctl images span two upstream namespaces, so two entries are needed at minimum:
+
+```ini
```
ekohl: I think this is actual toml, not ini
````suggestion
```toml
````

File: docs/developer/deployment.md [hunk 24]
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+```
+
+##### Developer testing a container build
+
+An `.image.d` drop-in overrides a single image without affecting others:
+
+```ini
+# /etc/containers/systemd/foreman.image.d/90-user.conf
+[Image]
+Image=quay.io/foreman/foreman:pr-12345
```
ekohl:
```suggestion
# To test https://github.com/theforeman/foreman-oci-images/pull/12345
Image=quay.io/foreman/stage/foreman:pr-12345
```

docs/developer/deployment.md
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
 ```
 
-2. **Deploy as usual** - foremanctl will automatically detect and use the authentication file:
+#### Authenticated Registry Handling
+
+foremanctl uses `/etc/foreman/registry-auth.json` as the default credential store. When pulling images from an authenticated registry, log in using that file:
+
 ```bash
-./foremanctl deploy
+podman login <registry> --authfile=/etc/foreman/registry-auth.json
```
ekohl: Does the order matter or can you write it like this? Removing something from the end is easier than the middle.
```suggestion
podman login --authfile=/etc/foreman/registry-auth.json <registry>
```

docs/developer/deployment.md
```diff
@@ -52,23 +52,168 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
 ```
 
-This approach integrates seamlessly with both the happy path and advanced deployment paths described above. The authentication is handled transparently during image pulling operations.
+If the auth file does not exist (unauthenticated registry), this environment variable is set but podman ignores it gracefully — no error is raised.
+
+#### Image Pulling (pull-images)
+
+The `foremanctl pull-images` command is an optional pre-deployment step that pulls all container images before running `foremanctl deploy`. This reduces deploy time and allows pre-staging images separately from deployment.
+
+`pull-images` deploys the `.image` unit files (making them available for quadlet to merge with any existing drop-ins from installed RPMs), then starts each `*-image.service` to perform the actual pull. To ensure mutable tags (such as `nightly`, `latest`, or `stream`) are always refreshed, `pull-images` temporarily creates a `Policy=always` drop-in before starting each service and removes it afterward, restoring `Policy=missing` for normal operation:
```
ekohl: Temporary files feel a bit nasty because it it's interrupted then they may not be cleaned up. I don't think we need to solve it right now, but I wonder if we could somehow leverage `podman auto-update` to pull images.

Shall we track that in a follow up issue?

src/roles/foreman/tasks/main.yaml
```diff
@@ -232,21 +228,14 @@
   loop_control:
     label: "{{ item.instance }}"
 
-- name: Run daemon reload to make Quadlet create the service files
-  ansible.builtin.systemd:
-    daemon_reload: true
-
-- name: Migrate and seed the Foreman database
+- name: Deploy foreman database migration container
```
ekohl, line 235: Now that it's a proper service (which I like), should we configure `foreman` and `dynflow-sidekiq` to require it? That way you can't start the foreman service without having migrated. Perhaps track as a separate issue?

tests/images_test.py
```diff
@@ -0,0 +1,63 @@
...
+    else:
+        assert image_file.exists and image_file.is_file
+
+
+def test_foreman_proxy_image_file(server, enabled_features):
+    image_file = server.file("/etc/containers/systemd/foreman-proxy.image")
+    if 'foreman-proxy' in enabled_features:
+        assert image_file.exists and image_file.is_file
+    else:
+        assert not (image_file.exists and image_file.is_file)
```
ekohl: Isn't this sufficient?
```suggestion
        assert not image_file.exists
```

tests/images_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_auth_dropin_content(server, core_image):
+    f = server.file(f"/etc/systemd/system/{core_image}-image.service.d/auth.conf")
+    assert "REGISTRY_AUTH_FILE" in f.content_string
+
+
+def test_postgresql_image_file(server, database_mode):
+    image_file = server.file("/etc/containers/systemd/postgresql.image")
+    if database_mode == 'external':
+        assert not (image_file.exists and image_file.is_file)
```
ekohl: Isn't this sufficient?
```suggestion
        assert not image_file.exists
```

src/roles/iop_advisor_frontend/defaults/main.yaml
ehelms: While it should, it does not today for a variety of reasons.

docs/developer/deployment.md
ehelms: No. I previously researched it and it only updates based on the policy and there is no way to "temporarily" set the policy. While I agree temporary files are not the best, there really isn't another way to do this.

Spit-balling, you could theoretically have two separate services but then you gotta manage the drop-in for both. Another idea possibly, would be to use `podman pull` by printing the service and parsing the conf file for the image name, or using `podman quadlet print`.

ekohl: I wonder if we should file a podman issue to see if there's tooling to support this.

docs/developer/deployment.md
```diff
@@ -52,23 +52,173 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
 ```bash
-podman login <registry> --authfile=/etc/foreman/registry-auth.json
+podman login --authfile=/etc/foreman/registry-auth.json registry.example.com
+```
+
+##### Disconnected install from local media
+
+In air-gapped environments, container images must be brought in without network access. `registries.conf` cannot express non-registry sources, but Podman can read images from local archive files. Images can be transported via USB or other local media using `podman save` / `podman load`, then referenced via drop-ins:
+
+See [podman-export(1)](https://docs.podman.io/en/latest/markdown/podman-export.1.html) for producing archive files.
```
ekohl: If you use the example below you don't need `podman load` because it's implicit. I also wonder what the difference between `podman save` and `podman export` is. Perhaps drop the last part?
```suggestion
In air-gapped environments, container images must be brought in without network access. `registries.conf` cannot express non-registry sources, but Podman can read images from local archive files. These images can be transported via USB or other local media, then referenced via drop-ins.
See [podman-export(1)](https://docs.podman.io/en/latest/markdown/podman-export.1.html) for producing archive files.
```

src/roles/foreman/tasks/main.yaml
```diff
@@ -257,6 +250,23 @@
...
+- name: Run daemon reload to make Quadlet create the service files
+  ansible.builtin.systemd:
+    daemon_reload: true
+
+- name: Migrate and seed the Foreman database
+  ansible.builtin.systemd:
+    name: foreman-db-migrate.service
+    state: restarted
+  async: 1800
+  poll: 10
```
evgeni: Why async/poll here? Who is waiting on this?

src/roles/pulp/tasks/main.yaml
```diff
@@ -221,39 +189,65 @@
...
+    state: restarted
+  async: 1800
+  poll: 10
+
+- name: Ensure Pulp admin user exists
+  ansible.builtin.systemd:
+    name: pulpcore-manager-admin-password.service
+    state: restarted
+  async: 600
+  poll: 10
```
evgeni: Won't admin user and db migration race here?

docs/developer/deployment.md
```diff
@@ -52,23 +52,173 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
 
-1. **Login to your registry** using Podman and save credentials to the default auth file location:
 ```bash
-podman login <registry> --authfile=/etc/foreman/registry-auth.json
+podman login --authfile=/etc/foreman/registry-auth.json registry.example.com
+```
+
+##### Disconnected install from local media
+
+In air-gapped environments, container images must be brought in without network access. `registries.conf` cannot express non-registry sources, but Podman can read images from local archive files. Images can be transported via USB or other local media using `podman save` / `podman load`, then referenced via drop-ins:
```
evgeni:
```suggestion
In air-gapped environments, container images must be brought in without network access. `registries.conf` cannot express non-registry sources, but Podman can read images from local archive files. Images can be transported via local media or other local transports using `podman save` / `podman load`, then referenced via drop-ins:
```

USB makes imho no sense (did you mean "flash drive"?), but people also can just `scp` things around

evgeni:
> **pull-images**
> 
> `pull-images` deploys image units (respecting any vendor drop-ins already in place), then creates a temporary `Policy=always` drop-in (`00-pull-always.conf`) for each image, performs a `daemon_reload`, restarts all `*-image.service` units in parallel, and removes the drop-ins in an `always:` block so cleanup happens even on failure. The pull honours any `.image.d` overrides already present.

Does this mean *every* run of `foremanctl deploy` will update the images to their latest version if we're using a `X.Y` tag?

ehelms:
> Does this mean _every_ run of `foremanctl deploy` will update the images to their latest version if we're using a `X.Y` tag?

No. Only `pull-images` will do that. The base image units are deployed with `Policy=missing` which means it only pulls if it's not found in local storage.

docs/developer/deployment.md
```diff
@@ -52,23 +52,173 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+
+See [podman-export(1)](https://docs.podman.io/en/latest/markdown/podman-export.1.html) for producing archive files.
+
+```ini
+# /usr/share/containers/systemd/foreman.image.d/20-archive.conf (from local media)
+[Image]
+Image=docker-archive:/opt/foreman/images/foreman-6.17.tar
+```
```
ehelms, line 152: There is a nuance of using this method that we need to make a design decision on from my research:

  - ISO always mounted (permanently air-gapped): Cleanup may genuinely be unnecessary. The archive is always accessible, Policy=missing finds the file, works every time.
  - ISO mounted temporarily (load and unmount): The drop-in must be removed, otherwise image service operations fail once the ISO is gone.


The question becomes do we want to tie ourselves to the local media (e.g. ISO) always being mounted / available?


For background, the way we've handled local media in previous instances is that the user had to use skopeo to copy the archive into podman's storage so that it showed up within `podman images`.

ekohl: > The question becomes do we want to tie ourselves to the local media (e.g. ISO) always being mounted / available?

IMHO no

> For background, the way we've handled local media in previous instances is that the user had to use skopeo to copy the archive into podman's storage so that it showed up within `podman images`.

I was wondering if copying the tar files to some static location is a good compromise. It will eat more disk space. I've also been thinking if they should be shipped using an RPM: both the container and the drop in file. But that's probably not a nice experience either.

Thinking out loud: should the containers be part of a backup?

evgeni: I'd prefer not being tied to anything.

When you do `Image=docker-archive:…` is the container executed from the archive directly, or is it imported like `skopeo copy …` would do it? I assume the former, but not 100% sure that's what you wrote.

Can we still use `scopeo copy` manually with those image files?

ehelms: Kinda. If you run podman images you see:

```
REPOSITORY  TAG         IMAGE ID      CREATED      SIZE
<none>      <none>      33386eb25182  5 weeks ago  369 MB
```

Which might be confusing for folks debugging.

ekohl: Should we be setting `ImageTag` as well to resolve that? https://docs.podman.io/en/latest/markdown/podman-image.unit.5.html#imagetag

evgeni: So, given this has no repo/tag, once you remove the override, the old `Image=quay.io…` won't resolve to `33386eb25182` and things won't start anymore, right?
Does this `33386eb25182` consume storage in `/var/lib/containers`, or is that still fully in `/opt/foreman/images/foreman-6.17.tar`?

ehelms: > So, given this has no repo/tag, once you remove the override, the old Image=quay.io… won't resolve to 33386eb25182 and things won't start anymore, right?

Correct.

> Does this 33386eb25182 consume storage in /var/lib/containers, or is that still fully in /opt/foreman/images/foreman-6.17.tar?

The image is loaded into containers storage.

> Should we be setting ImageTag as well to resolve that? https://docs.podman.io/en/latest/markdown/podman-image.unit.5.html#imagetag

`ImageTag` has no effect from my testing, and based on `podman pull` help that seems to make sense.

For reference, a method that is in use right now with `skopeo copy` does the following:

```python
        skopeo_process = subprocess.run((
            "skopeo",
            "copy",
            f"oci-archive:{archive_path}",
            f"containers-storage:{pullspec}",
        ))
```

evgeni: I think a `scopeo copy` (maybe with some helper on the transport medium, as the one you pasted above) makes most sense to me without being too cumbersome to handle?

docs/developer/deployment.md
```diff
@@ -52,23 +52,179 @@ IOP (Insights Operating Platform) deploys on-premise Insights services for advis
...
+
+No user action required beyond installing the RPM and logging into the product registry:
+
+```bash
+podman login --authfile=/etc/foreman/registry-auth.json registry.example.com
+```
+
+##### Disconnected install from local media
+
+In air-gapped environments, container images must be brought in without network access. Images are loaded directly into Podman's container storage using `skopeo copy` before running `foremanctl deploy`. The explicit `containers-storage:` destination ensures each image is stored under its correct pullspec — without this, images load as `<none>/<none>` and Quadlet container units cannot resolve them.
```
evgeni:
```suggestion
In air-gapped environments, container images must be brought in without network access. Images are loaded directly into Podman's container storage using `skopeo copy` before running `foremanctl deploy`.
```

This is explained later, and only confuses the user in the introduction, as they haven't seen the full `scopeo copy` invocation yet.

evgeni:
My questions about the async/poll behaviour remain unanswered, but other than that this looks good to me and we should get it in soon.

src/roles/pulp/tasks/main.yaml
ehelms: With Type=oneshot:
  - systemd blocks on restart until the container process exits
  - The Ansible systemd module waits for systemd to report completion
  - Task 2 cannot start until Task 1's migration process has fully exited

ehelms:
> My questions about the async/poll behaviour remain unanswered, but other than that this looks good to me and we should get it in soon.

Addressed or replied to.

src/roles/pulp/tasks/main.yaml
evgeni: Isn't that the same behavior as without any async/poll parameters?

ehelms: Correct. My understanding of async / poll was that it allows Ansible to be robust against disconnection for potentially long-lived SSH connections.

evgeni (approved):

docs/developer/deployment.md
ekohl: > > Should we be setting ImageTag as well to resolve that? https://docs.podman.io/en/latest/markdown/podman-image.unit.5.html#imagetag
> 
> `ImageTag` has no effect from my testing, and based on `podman pull` help that seems to make sense.

https://github.com/podman-container-tools/podman/issues/23781 reads related. Should we should follow up on this in a separate foremanctl issue and where needed create a podman issue?

PR #502, opened 2026-05-08
Enable ansible-lint var-naming to enforce scoped role variables

ehelms:

To help enforce role names being scoped per role.

* Enables the var-naming ansible-lint rule explciitly

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,4 +1,6 @@
 ---
+enable_list:
+  - var-naming[no-role-prefix]
 exclude_paths:
   - .ansible/
   - .github/


stejskalleos (approved):

evgeni:
Why was this needed? `var-naming[no-role-prefix]` should be enabled by default.

evgeni:
@ehelms @stejskalleos I'd still like to know why this needed explicit enablement, given it's a default rule.

stejskalleos:
I'll admit I was unaware that it's enabled by default.

ehelms:
> should be enabled by default.

The keyword here is `should`. This was not working, I don't know why, but at the time without this it was not being enforced.

PR #503, opened 2026-05-11
allow downstream product overrides for features

bochi:
This allows downstream products to extend foremanctl without touching the actual files (features.yaml,  foremanctl) As of now adding features  requires patching the checkout

* `foremanctl.py`: load `src/features.d/*.yaml` if the directory exists

Steps to test:

* Without creating overlays, test foremanctl, there is no change to the current behavior
* Write `features.d/whatever.yaml` with `feature: { description: my_great_feature }` and try `foremanctl features list`
* use `OBSAH_ALLOW_EXTRA_VARS=true` and set extra vars (-e var=value) and see if they are accepted

#### Checklist

### DIFF
--- a/src/filter_plugins/foremanctl.py
+++ b/src/filter_plugins/foremanctl.py
@@ -10,9 +10,18 @@ import yaml
 
 BASE_FEATURES = ['hammer', 'foreman-proxy', 'foreman']
 
-features_yaml = pathlib.Path(__file__).parent.parent / 'features.yaml'
+_SRC_ROOT = pathlib.Path(__file__).parent.parent
+features_yaml = _SRC_ROOT / 'features.yaml'
 with features_yaml.open() as features_file:
     FEATURE_MAP = yaml.safe_load(features_file)
+
+# load additional feature files under features.d  
+_features_d = _SRC_ROOT / 'features.d'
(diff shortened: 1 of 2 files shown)


src/filter_plugins/foremanctl.py
```diff
@@ -6,10 +6,19 @@
 
 BASE_FEATURES = ['hammer', 'foreman-proxy', 'foreman']
 
-features_yaml = pathlib.Path(__file__).parent.parent / 'features.yaml'
+_SRC_ROOT = pathlib.Path(__file__).parent.parent
+features_yaml = _SRC_ROOT / 'features.yaml'
 with features_yaml.open() as features_file:
     FEATURE_MAP = yaml.safe_load(features_file)
 
+# load additional feature files under features.d
```
evgeni, line 18: this goes beyond the "override images" that is also (differently) solved in #501, which I think can be a very valuable addition outside the image context

evgeni (approved):

evgeni:
Tested this to create a Satellite-specific feature using `foremanctl-3.0.0.develop-1.20260529112315925378.pr503.4.g0e50f0a.el9.noarch` and it worked like a charm.

Thanks!

PR #504, opened 2026-05-11
Make it more clear that setup-repositories is required

jeremylenz:

I didn't realize that you have to run `setup-repositories` or hammer enablement won't work.

just a note in DEVELOPMENT.md

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -41,6 +41,9 @@ source .venv/bin/activate
 
 ### Deploy hammer (optional)
 
+> [!NOTE]
+> You MUST run `setup-repositories` before enabling hammer.
+
 ```
 ./forge setup-repositories
 ./foremanctl deploy --add-feature hammer
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -41,13 +41,13 @@ source .venv/bin/activate
 
 ### Deploy hammer (optional)
 
-> [!NOTE]
-> You MUST run `setup-repositories` before enabling hammer.
-
 ```
 ./forge setup-repositories
 ./foremanctl deploy --add-feature hammer
 ```
+
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


arvind4501:
here are some thoughts

> Updated package name prefix from hammer-cli-plugin- to rubygem-hammer_cli_ in defaults

Hammer installation has been working fine with the old naming (which is alias to what you have added), as the goal is to install the respective hammer package which eventually.


> Added repository availability check that fails early with helpful error message

for production, we alredy [document](https://docs.theforeman.org/nightly/Quickstart/index-foremanctl-katello.html#configuring-repositories_quick-start) that these repos must be configured for smooth deployment.
and for development we also mention to setup repos first before deploying hammer in [DEVELOPMENT.md](https://github.com/theforeman/foremanctl/blob/master/DEVELOPMENT.md#deploy-hammer-optional)

so i wonder if we really need the check for plugins repo(or rather for any repo),  IMHO we can skip that.

jeremylenz:
@arvind4501 Okay, fair enough. I switched back to `master` and ran it again, and it indeed succeeds since the repositories are there.

I would at least like to make this fact more prominent for first-timers. I have updated the PR accordingly

DEVELOPMENT.md
```diff
@@ -41,6 +41,9 @@ source .venv/bin/activate
 
 ### Deploy hammer (optional)
 
+> [!NOTE]
+> You MUST run `setup-repositories` before enabling hammer.
+
```
arvind4501:
```suggestion
```

DEVELOPMENT.md
arvind4501, line 48:
```suggestion
> [!NOTE]
> You MUST run `setup-repositories` before enabling hammer.
```

arvind4501 (approved):
Thanks @jeremylenz , small nitpick to move note below, otherwise LGTM

jeremylenz:
I don't have merge access, feel free to merge :)

PR #508, opened 2026-05-14
Expose user enabled features as pytest markers

ekohl:

I suggested it in https://github.com/theforeman/foremanctl/pull/484#discussion_r3234542442 because I think it sets the base for future extensibility.

It reads `foremanctl features` to know which features exist and which are enabled. It then configures pytest to expose a feature marker (in `pytest_configure`) that's then read in `pytest_runtest_setup`.

With all of that set up, the existing iop feature guard is rewritten.

Steps to reproduce:

* Let CI run (or locally)
* Verify that it skips IOP when the feature is disabled
* Verify that it runs IOP tests when the feature is enabled

#### Checklist

### DIFF
--- a/docs/developer/testing.md
+++ b/docs/developer/testing.md
@@ -146,6 +146,32 @@ def test_service_port(server):
     assert server.addr("localhost").port(6379).is_reachable
 ```
 
+### Feature guarding
+
+Some functionality can only be tested when a feature is enabled.
+
+You can mark an individual test to be skipped if needed:
+
+```python
+@pytest.mark.feature("iop")
+def test_ingress_service(server):
+    service = server.service("iop-core-ingress")
+    assert service.is_running and service.is_enabled
+```
+
+Often it's better to have an entire file dedicated to a feature and mark the entire file as guarded.
+
+```python
+pytestmark = pytest.mark.feature("iop")
+
+def test_ingress_service(server):
+    service = server.service("iop-core-ingress")
+    assert service.is_running and service.is_enabled
+
+def test_ingress_http_endpoint(server):
+    # ...
+```
+
 ### API test
 
 The `foremanapi` fixture is an [apypie](https://github.com/Apipie/apypie) `ForemanApi` client that connects to the deployed Foreman instance(authenticated as `admin`/`changeme`). It maps directly to the Foreman REST API — each method takes a resource name that corresponds to an API endpoint:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
-import os
+import subprocess
 import uuid
+from functools import cached_property
 
 import apypie
 import paramiko
@@ -16,9 +17,38 @@ from requests.adapters import HTTPAdapter
 SSH_CONFIG = './.tmp/ssh-config'
 
 
+class UserParameters:
+    def __init__(self, config):
+        self._config = config
+
+    @cached_property
+    def features(self):
+        # foremanctl outputs
+        # FEATURE                   STATE               DESCRIPTION
+        # $feature                  enabled/available   $description
+        output = subprocess.check_output(['./foremanctl', 'features'], cwd=self._config.rootdir,
+                                         universal_newlines=True)
+        lines = output.splitlines(keepends=False)
+        # feature, status, description
+        return [line.split(None, 2) for line in lines[1:]]
+
+    @cached_property
+    def available_features(self):
+        return set(feature for feature, _status, _desc in self.features)
+
+    @cached_property
+    def enabled_features(self):
+        return set(feature for feature, status, _desc in self.features if status == 'enabled')
+
+
 def pytest_addoption(parser):
     parser.addoption("--certificate-source", action="store", default="default", choices=('default', 'installer', 'custom_server'), help="Certificate source used during deployment")
     parser.addoption("--database-mode", action="store", default="internal", choices=('internal', 'external'), help="Whether the database is internal or external")
+
+
+@pytest.fixture(scope="module")
+def enabled_features(pytestconfig):
+    return pytestconfig.user_parameters.enabled_features
 
 
 @pytest.fixture(scope="module")
@@ -206,35 +236,21 @@ def wait_for_metadata_generate(foremanapi):
     wait_for_tasks(foremanapi, 'label = Actions::Katello::Repository::MetadataGenerate')
 
 
-def enabled_features():
-    test_dir = os.path.dirname(os.path.abspath(__file__))
-    foremanctl_dir = os.path.dirname(test_dir)
-    params_file = os.path.join(foremanctl_dir, '.var', 'lib', 'foremanctl', 'parameters.yaml')
-    if os.path.exists(params_file):
-        with open(params_file, 'r') as f:
-            features = yaml.safe_load(f).get('features', [])
-            if isinstance(features, str):
-                features = features.split()
-            return features
-    return []
-
-
-def is_iop_enabled():
-    return 'iop' in enabled_features()
-
-
 def pytest_configure(config):
-    config.addinivalue_line("markers", "iop: tests requiring IOP to be enabled")
+    config.addinivalue_line("markers", "feature(name): mark a test as requiring a feature")
+
+    config.user_parameters = UserParameters(config)
 
 
-def pytest_collection_modifyitems(config, items):
-    if is_iop_enabled():
-        return
-
-    skip_iop = pytest.mark.skip(reason="IOP not enabled - skipping IOP tests ('iop' not in enabled_features)")
-    for item in items:
-        if "iop" in item.keywords:
-            item.add_marker(skip_iop)
+def pytest_runtest_setup(item):
+    feature_markers = set(mark.args[0] for mark in item.iter_markers(name="feature"))
+    if feature_markers:
+        invalid_features = feature_markers - item.config.user_parameters.available_features
+        if invalid_features:
+            raise pytest.PytestConfigWarning(f"Invalid feature(s) {invalid_features!r} on {item}")
+        missing = feature_markers - item.config.user_parameters.enabled_features
+        if missing:
+            pytest.skip(f"test requires feature(s) {missing!r}")
 
 
 class ResolveAdapter(HTTPAdapter):
--- a/tests/foreman_proxy_test.py
+++ b/tests/foreman_proxy_test.py
@@ -2,13 +2,8 @@ import datetime
 import json
 
 import pytest
-from conftest import enabled_features
 
 FOREMAN_PROXY_PORT = 8443
-
-
-def is_bmc_enabled():
-    return 'bmc' in enabled_features()
 
 
 def get_proxy_v2_features(server, certificates, server_fqdn):
@@ -22,14 +17,14 @@ def get_proxy_v2_features(server, certificates, server_fqdn):
     return json.loads(cmd.stdout)
 
 
-def test_foreman_proxy_features(server, certificates, server_fqdn):
+def test_foreman_proxy_features(server, certificates, server_fqdn, enabled_features):
     cmd = server.run(f"curl --cacert {certificates['server_ca_certificate']} --silent https://{server_fqdn}:{FOREMAN_PROXY_PORT}/features")
     assert cmd.succeeded
     features = json.loads(cmd.stdout)
     assert "logs" in features
     assert "script" in features
     assert "dynflow" in features
-    if is_bmc_enabled():
+    if 'bmc' in enabled_features:
         assert "bmc" in features
     else:
         assert "bmc" not in features
@@ -60,7 +55,7 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
     assert cmd.stdout == '201'
 
 
-@pytest.mark.skipif("not is_bmc_enabled()")
+@pytest.mark.feature('bmc')
 def test_bmc_capabilities(server, certificates, server_fqdn):
     features = get_proxy_v2_features(server, certificates, server_fqdn)
     assert 'bmc' in features
--- a/tests/iop/test_advisor.py
+++ b/tests/iop/test_advisor.py
@@ -1,6 +1,6 @@
 import pytest
 
-pytestmark = pytest.mark.iop
+pytestmark = pytest.mark.feature("iop")
 
 
 def test_advisor_backend_api_service(server):
--- a/tests/iop/test_advisor_frontend.py
+++ b/tests/iop/test_advisor_frontend.py
@@ -1,6 +1,6 @@
 import pytest
 
-pytestmark = pytest.mark.iop
+pytestmark = pytest.mark.feature("iop")
 
 
 def test_advisor_frontend_assets_directory(server):
--- a/tests/iop/test_cvemap_downloader.py
+++ b/tests/iop/test_cvemap_downloader.py
@@ -1,6 +1,6 @@
 import pytest
 
-pytestmark = pytest.mark.iop
+pytestmark = pytest.mark.feature("iop")
 
 
 def test_cvemap_download_script(server):
--- a/tests/iop/test_engine.py
+++ b/tests/iop/test_engine.py
@@ -1,6 +1,6 @@
 import pytest
 
-pytestmark = pytest.mark.iop
+pytestmark = pytest.mark.feature("iop")
 
 
 def test_engine_service(server):
--- a/tests/iop/test_gateway.py
+++ b/tests/iop/test_gateway.py
@@ -1,6 +1,6 @@
 import pytest
 
-pytestmark = pytest.mark.iop
+pytestmark = pytest.mark.feature("iop")
 
 
 def test_gateway_service(server):
--- a/tests/iop/test_ingress.py
+++ b/tests/iop/test_ingress.py
@@ -1,6 +1,6 @@
 import pytest
 
-pytestmark = pytest.mark.iop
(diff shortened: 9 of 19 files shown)


ekohl:
Is this a more common failure?
```
:stderr: ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device: '/home/runner/smoker/venv/lib/python3.12/site-packages/certifi-2026.4.22.dist-info'
```

tests/conftest.py
```diff
@@ -16,6 +18,38 @@
...
 
 
+class UserParameters:
+    def __init__(self, config):
+        self._config = config
+
+    @property
+    def _obsah_state(self) -> Path:
+        # mirror what foremanctl / obsah does
+        if state := os.environ.get('OBSAH_STATE'):
```
arvind4501: I am wondering if `OBSAH_STATE` and `OBSAH_BASE` will exist until we define them, in case of foremanctl we define them to be used in execution in https://github.com/theforeman/foremanctl/blob/master/foremanctl

tests/conftest.py
```diff
@@ -16,6 +18,38 @@
...
+            return Path(state)
+
+        if base := os.environ.get('OBSAH_BASE'):
+            root = Path(base)
+        else:
+            root = self._config.rootpath
+        return root / '.var' / 'lib' / 'foremanctl'
+
+    def _read_parameters(self):
+        params_file = self._obsah_state / 'parameters.yaml'
```
arvind4501: Not blocking: currently we are reading from `parameters.yaml`, and today if we add feature it first gets added to `parameters.yaml` and then the subsequent steps for setup runs in playbook and roles. 
The only case i see this becoming a issue is when --add-feature fails in between the run but feature persists in `parameters.yaml` and then that leaves that feature specific test  broken.

I also a relevent PR which persits params after playbook run https://github.com/theforeman/obsah/pull/118

tests/conftest.py
ekohl: I had assumed that you could still override them, but now that I read `foremanctl` again I see I was wrong.

My idea was to build towards running things in parallel:
```sh
export OBSAH_STATE=$(mktemp -d)
foremanctl deploy --flavor foreman
pytest
```

Then in another shell:
```sh
export OBSAH_STATE=$(mktemp -d)
foremanctl deploy --flavor katello
pytest
```

Perhaps it's better to reduce this method to the hardcoded option until we get the other pieces in place.

tests/conftest.py
ekohl: I'd argue that's a corrupted state and running a test suite on that isn't useful. If `foremanctl deploy` exits with a non-zero code then IMHO the whole execution should halt.

tests/conftest.py
arvind4501: So these envs are no use as of today, but when we have support for multiple flavours(which certanily will have diffrent set of features ) we could use envs at runtime to invoke flavour specific ci workflows?

Thinking out loud:- In foreman-installer we had answers file per scenarios, will we have similar kind of parameters files for each flavour, also there was no way of changing scenarios after you deployed with one, will that remain same or could be improved,
This is not related to the PR, but just wanted to write my mind dump out

ekohl: > So these envs are no use as of today, but when we have support for multiple flavours(which certanily will have diffrent set of features ) we could use envs at runtime to invoke flavour specific ci workflows?

Yes. I was a bit inspired by https://github.com/theforeman/foremanctl/pull/494. That uses multiple git checkouts (which has a value in its own) while the approach I had in mind here allowed testing multiple deployments from the same directory. But thinking more about it, perhaps we shouldn't build that partially and either do it completely or not at all.

> Thinking out loud:- In foreman-installer we had answers file per scenarios, will we have similar kind of parameters files for each flavour, also there was no way of changing scenarios after you deployed with one, will that remain same or could be improved,

Technically there was a way to switch scenarios, but that switching was never really well implemented / supported. Partially because the underlying system didn't really support it. I think on the end user system we should aim for a single parameters file and not allow switching between flavors.

> This is not related to the PR, but just wanted to write my mind dump out

Since we're dumping thoughts: I've been toying with the thought of remote deployments. Ansible supports running on remote hosts so can we imagine some sort of control node where you manage the actual deployment? Then by changing `OBSAH_STATE` you could perhaps manage your various environments (like test, production, etc).

arvind4501: > can we imagine some sort of control node where you manage the actual deployment? 

Yes, that should be possible with ansible, rather than hardcoding quadlet as deployment location we could give a target group(multiple hosts).

tests/conftest.py
ekohl: After playing with the tests myself I now understand what you mean. The test suite runs `--add-feature invalid-feature` which invalidates reruns.

I've also found another limitation: it doesn't read the features from the flavor which for now isn't a big deal but will be when you really want to gate all features.

ekohl: Should it use `foremanctl features --list-enabled` to get the list instead of digging in internals?

arvind4501: > Should it use foremanctl features --list-enabled to get the list instead of digging in internals?

Yes that would be a better option as `foremanctl features ` will retrieve features from [features.yml](https://github.com/theforeman/foremanctl/blob/master/src/features.yaml).

> I've also found another limitation: it doesn't read the features from the flavor which for now isn't a big deal but will be when you really want to gate all features.

This will also be solved if we use `foremanctl features` to get features list. one thing to note here is that `foremanctl features`  is not aware about flavor while pulling features, so it pulls all features from feeatures.yml, which is fine as of today because we only have one flavor.

But in future we need to make `foremanctl features` aware of flavor

arvind4501: I think once we have multiple flavors we need to extend [features.yml](https://github.com/theforeman/foremanctl/blob/master/src/features.yaml) metadata to also include flavor attribute in defination, something like:

```
foreman:
  description: Base Foreman Server
  flavor: [foreman, katello]

katello:
  description: Content and Subscription Management plugin for Foreman
  foreman:
    plugin_name: katello
  hammer: katello
  flavor: [katello]
```
It just came to mind after looking at https://github.com/theforeman/foremanctl/pull/511

ekohl: > But in future we need to make `foremanctl features` aware of flavor

While experimenting further I also realized it can be useful to expose the internal features somehow. Those do not show up in the tool, but can be useful in testing.

> I think once we have multiple flavors we need to extend [features.yml](https://github.com/theforeman/foremanctl/blob/master/src/features.yaml) metadata to also include flavor attribute in defination, something like:

It's an interesting question if we do want that. In an ideal world you can actually add Katello to an existing Foreman installation. There are technical limitations today that mean it can't always work (Foreman can nest organizations, Katello can't). However, I'd love to support that so as a new user you can start with essentially an "empty" Foreman and add functionality as you grow your need.

As an example: at first the user just wants to have an inventory of all hosts to understand what they have. Then they start provisioning new servers in a repeatable way so they add provisioning services. They still pull the content from the internet, but at some point they want to control that too. At that point Katello is added.

Now how we exactly achieve that user story is a good question. Perhaps the UI should be smarter to guide you there, perhaps adding plugins as you grow.

But to your original point: once we add the standalone proxy support, do we really want to support adding Foreman to that?

arvind4501: > once we add the standalone proxy support, do we really want to support adding Foreman to that?

I think no, proxy serves foreman so ideally we should not add foreman to proxy installation, if we try to add foreman feature to existing standalone proxy(which is registered to a foreman instance), than that would create conflict between which foreman instance the proxy serves to.

And this is the reason i raised the concern of making `foremanctl features` flavor aware, in case of stanalone proxy if we just`foremanctl features` (as of today) it will list foreman, katello, hammer etc as available features but do we ever want to add katello to proxy? i think no.

ekohl: > And this is the reason i raised the concern of making `foremanctl features` flavor aware, in case of stanalone proxy if we just`foremanctl features` (as of today) it will list foreman, katello, hammer etc as available features but do we ever want to add katello to proxy? i think no.

I think it is a good point. Thinking of user workflows.

* As an admin I have a fresh installation (plain Linux, just installed `foremanctl`)
* I run `foremanctl features --listen-enabled` to see what I'll get
* I run `foremanctl deploy --flavor $non-default`

That's probably misleading because the real features may be different.

In the testing scenario it's not going to really matter because there we only care about what has already been deployed, but when I played around I also wanted to know the internal features (`dynflow`) and it can't list those either. Building that is likely easy, but are we then intentionally leaking internal details? Perhaps if we use an env var rather than a command line parameter, but that may also be hacky.

I've also thought about using obsah's API but the flavor part is implemented in foremanctl so it's probably the wrong tool.

arvind4501: > but when I played around I also wanted to know the internal features (dynflow) and it can't list those either. Building that is likely easy, but are we then intentionally leaking internal details?

In https://github.com/theforeman/foremanctl/blob/master/docs/developer/feature-metadata.md#feature-metadata we define internal features as non user visible, honestly i would love to keep that. ideally our focus should be to test user exposed features/functionalities rather than caring about what it internally depends on. 
i know we could have possible scenarios where we can test internal feature, but IMHO if we want to test internal features then also we should make that changes in pytest to fetch internal features rather than changing `foremanctl features` behaviour.

ekohl: I'll share what triggered me to think about this:
https://github.com/theforeman/foremanctl/blob/6822461349bbf710250ad2ff686a7afdc46ed4d6/tests/foreman_proxy_test.py#L31
That really should be dependent on whether the `dynflow` feature is enabled, which is internal.

> i know we could have possible scenarios where we can test internal feature, but IMHO if we want to test internal features then also we should make that changes in pytest to fetch internal features rather than changing `foremanctl features` behaviour.

Exactly this. Question is: how? Do we introduce some reusable helper that we can load in both `foremanctl features` and pytest? Is there some other way I didn't think of?

arvind4501: > Do we introduce some reusable helper that we can load in both foremanctl features and pytest? 

i would still argue we should not need to expose internal features via `foremanctl features`. but having said that i agree that we need to test some internal feature,
What i am thinking is test user exposed features and conditionalily test internal features based on if the feature (that needs them as dependency, so in case of dynflow, it is remote-execution), is enabled.

So if `remote-execution` is enabled then only test the `dynflow` in above case.

But that also brings next challenge which is what if mutiple features needs `dynflow` as dependency.

ekohl: > But that also brings next challenge which is what if mutiple features needs `dynflow` as dependency.

You read my mind. I was thinking about Apache & PostgreSQL and if that could be tracked as an internal feature. It can be pulled in by multiple services (Foreman, Candlepin, Pulp).

arvind4501: > I was thinking about Apache & PostgreSQL and if that could be tracked as an internal feature.

not sure if we would expose postgresql as feature, but we have `remote-execution` which has `dynflow` as dependency but we might also have `dynflow` as dependency for other feature, i think ansible and that way we have two features having `dynflow` as dependency, 

and then with what i said earlier 
>  i am thinking is test user exposed features and conditionalily test internal features based on if the feature (that needs them as dependency, so in case of dynflow, it is remote-execution), is enabled.

becomes a issue as we need to test dynflow if any of these feature is enabled.

I have also asked somewhat similar https://github.com/theforeman/foremanctl/pull/444#discussion_r3038912411

tests/conftest.py
```diff
@@ -206,35 +240,18 @@ def wait_for_metadata_generate(foremanapi):
...
-    skip_iop = pytest.mark.skip(reason="IOP not enabled - skipping IOP tests ('iop' not in enabled_features)")
-    for item in items:
-        if "iop" in item.keywords:
-            item.add_marker(skip_iop)
+def pytest_runtest_setup(item):
+    feature_markers = set(mark.args[0] for mark in item.iter_markers(name="feature"))
+    if feature_markers:
+        missing = feature_markers - item.config.user_parameters.features
+        if missing:
+            pytest.skip("test requires feature(s) {!r}".format(missing))
```
shubhamsg199: Non blocking: We might never hit this but what If anyone passes an invalid feature name `pytest.mark.feature("invalid")` the test in this case will be skipped right?, where I believe it should raise an error.

shubhamsg199: The rest of the codebase appears to use f-strings. For consistency:

```suggestion
def pytest_runtest_setup(item):
    feature_markers = set(mark.args[0] for mark in item.iter_markers(name="feature"))
    if feature_markers:
        missing = feature_markers - item.config.user_parameters.features
        if missing:
            pytest.skip(f"test requires feature(s) {missing!r}")
```

docs/developer/testing.md
```diff
@@ -146,6 +146,32 @@ def test_service_port(server):
     assert server.addr("localhost").port(6379).is_reachable
 ```
 
+### Feature guarding
+
+Some functionality can only be tested when a feature is enabled.
+
+You can mark an invidivual test to be skipped if needed:
```
shubhamsg199:
```suggestion
You can mark an individual test to be skipped if needed:
```

tests/conftest.py
```diff
@@ -16,6 +18,38 @@
...
+            root = Path(base)
+        else:
+            root = self._config.rootpath
+        return root / '.var' / 'lib' / 'foremanctl'
+
+    def _read_parameters(self):
+        params_file = self._obsah_state / 'parameters.yaml'
+        if params_file.exists():
+            with params_file.open('r') as f:
+                return yaml.safe_load(f)
```
shubhamsg199: Non blocking: if parameters.yaml is ever corrupted during deployment (which is extremely rare), yaml.safe_load() returns a non-dict and .get('features', []) would raise AttributeError. Good to add a guard against it?

tests/conftest.py
ekohl: > not sure if we would expose postgresql as feature, but we have `remote-execution` which has `dynflow` as dependency but we might also have `dynflow` as dependency for other feature, i think ansible and that way we have two features having `dynflow` as dependency,

The reason I got to it is introducing a plain foreman-proxy (without content). That doesn't run Apache and PostgreSQL. I could derive it in Ansible code from which other features are enabled but then it's harder in testing. An internal feature could help.

> I have also asked somewhat similar [#444 (comment)](https://github.com/theforeman/foremanctl/pull/444#discussion_r3038912411)

Good point, I hadn't thought that far ahead yet.

arvind4501:
@coderabbitai

arvind4501:
@coderabbitai full review

tests/conftest.py
arvind4501: yes, looking at https://github.com/theforeman/foremanctl/blob/6822461349bbf710250ad2ff686a7afdc46ed4d6/tests/foreman_proxy_test.py#L31

we will mark it as  pytest.mark.feature("foreman-proxy") probabily and  i think we can stick to use 
`foremanctl feartures --list-enabled` to get features (which will  also bring default flavor feature) for now, and once we see more test required for internal features, we can extend that. what do you think?

arvind4501:
@coderabbitai review

arvind4501:
2 suggestions from coderabbit

> - Fix spelling error.
> 
>    "invidivual" should be "individual".

Thats indeed a good finding and actionable

> Add error handling for YAML parsing and validation.

> https://github.com/theforeman/foremanctl/pull/508/changes#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R39-R41

I don't think we need to do error handling here, `parameters.yaml` is obsah's space, with foremanctl we just use what obsah gives and if there are possibilities of errors we handle in obsah rather than here

Sad part is it did not even considered `foremanctl features --list-enabled` over current logic in UserParameters class

shubhamsg199:
> I don't think we need to do error handling here, `parameters.yaml` is obsah's space, with foremanctl we just use what obsah gives and if there are possibilities of errors we handle in obsah rather than here

This is something I also had in mind https://github.com/theforeman/foremanctl/pull/508#discussion_r3259964211 but the scenario which i pointed out might not ever happen.

tests/conftest.py
ekohl: Yes, let's start with `foremanctl feartures --list-enabled` and refactor as needed. IMHO even without the internal features this is already a useful framework.

tests/conftest.py
ekohl: Yes, I thought about it when I started to build on this in https://github.com/theforeman/foremanctl/pull/509. Perhaps we should also have `valid_features` to be able verify this?

I'd like to track it in an issue to follow up later.

tests/conftest.py
ekohl: https://github.com/theforeman/foremanctl/pull/508#discussion_r3246705810 is a long discussion, but with @arvind4501 I discussed how we even want to read the parameters. Short term we're narrowing in on calling `foremanctl features --list-enabled` so we don't duplicate any logic. There are limitations (like not showing internal features), but we can always refactor it later.

ekohl:
I've updated it to call `foremanctl features` and parse the output. I chose not to pass `--enabled-features` and list them all so I could determine `available_features`. That allows me to raise a config warning when an invalid feature mark is used. I'm not sure if there is a better way to do this, but I verified it works by using `@pytest.mark.feature('invalid')`.

ekohl:
And now with a few force pushes I think I pleased the linter too.

arvind4501 (approved):

tests/conftest.py
ekohl: And in the end I did implement this.

shubhamsg199: Great!

shubhamsg199 (approved):

PR #509, opened 2026-05-15
Introduce feature-based test directory structure

ekohl:

Inspired by https://github.com/theforeman/foremanctl/pull/485, but building on top of https://github.com/theforeman/foremanctl/pull/508.

It introduces a directory structure that automatically marks anything under `tests/feature/$feature` with `$feature`. This prepares for additional flavors that disable various features, such as a plain foreman (without katello) and a plain foreman-proxy.

Steps to reproduce:

* Create a deployment that includes all features and verify all tests are executed
* Create a deployment that disables some features and verify it skips the right tests

#### Checklist


ekohl:
Test failures due to Galaxy:
```
Error: : Failed to download collection tar from 'default' due to the following unforeseen error: The read operation timed out
```
```
Error: : Error when getting available collection versions for community.crypto from default (https://galaxy.ansible.com/api/) (HTTP Code: 504, Message: Gateway Timeout Code: Unknown)
```

tests/conftest.py
```diff
@@ -242,6 +243,19 @@ def pytest_configure(config):
...
+    feature_dir = config.rootdir / 'tests' / 'feature'
+    for item in items:
+        try:
+            rel_path = Path(item.fspath).relative_to(feature_dir)
+        except ValueError:
+            # Not in the features directory
+            pass
+        else:
+            feature = rel_path.parts[0]
+            item.add_marker(pytest.mark.feature(feature))
```
ehelms, line 299: This automatic marking is probably the only part that is semi-magical. However, as a convention I think this is fine. I think we should document this convention in our `testing.md` file.

Then a question. What about "meta" features? e.g. `content/rpm`, `content/pypi`

ekohl: > This automatic marking is probably the only part that is semi-magical. However, as a convention I think this is fine. I think we should document this convention in our `testing.md` file.

I also considered the magic. However, there is validation on invalid features and if we have the proper testing matrix we should exercise all code both with and without the feature present.

Good call on documenting it.

> Then a question. What about "meta" features? e.g. `content/rpm`, `content/pypi`

I've tried to keep it simple to avoid introducing too much magic. In https://github.com/theforeman/foremanctl/pull/511 I've toyed with hiding Pulp behind a feature and the testing is IMHO the biggest task on introducing additional flavors.

My preference is to have very light testing in foremanctl and then the actual end-to-end tests in [smoker](https://github.com/theforeman/smoker). So I'd avoid introducing more magic and keep it simple. Assuming we hide Pulp behind the `katello` feature and then for tests:
* `tests/feature/katello/pulp_test.py`
* `tests/feature/katello/content_rpm_test.py`
* `tests/feature/katello/content_pypi_test.py`

They will all be guarded behind the `katello` feature because of the directory structure. The content tests can use `pytestmark` at the top of the file to guard on both the katello feature and specific content types.

arvind4501:
With this we have foreman-proxy marker applied to all tests in `tests/feature/foreman-proxy/base_test.py`  but that would also have sub proxy feature(like bmc), do we need a way to create a per sub-feature file and mark that with the relevent marker?

tests/conftest.py
```diff
@@ -242,6 +243,19 @@ def pytest_configure(config):
     config.user_parameters = UserParameters(config)
 
 
+def pytest_collection_modifyitems(config, items):
+    feature_dir = config.rootdir / 'tests' / 'feature'
+    for item in items:
+        try:
+            rel_path = Path(item.fspath).relative_to(feature_dir)
```
arvind4501: i see fspath is deprecated,so good to use path instead https://docs.pytest.org/en/stable/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path

ekohl:
> With this we have foreman-proxy marker applied to all tests in `tests/feature/foreman-proxy/base_test.py` but that would also have sub proxy feature(like bmc), do we need a way to create a per sub-feature file and mark that with the relevent marker?

Technically `feature('foreman-proxy')` is applied to `tests/feature/foreman-proxy/**/*_test.py`.

You can certainly create a standalone file and if you have a bunch of tests it makes sense to create `tests/feature/foreman-proxy/bmc_test.py`:

```python
import pytest

pytestmark = pytest.mark.feature('bmc')


def test_x():
    pass


def test_y():
    pass
```

This will be equivalent to (today) creating `tests/foreman_proxy_bmc_test.py`:
```python
import pytest

pytestmark = [pytest.mark.feature('foreman-proxy'), pytest.mark.feature('bmc')]


def test_x():
    pass


def test_y():
    pass
```

But you can also create `tests/feature/foreman-proxy/base_test.py`:
```python
import pytest


def test_base():
    pass


@pytest.mark.feature('bmc')
def test_bmc():
    pass
```

All of these are functionally equivalent and it depends on how we want to organize it.

tests/conftest.py
ekohl: Neat, didn't know that was already supported. This will save me an import too.
```suggestion
            rel_path = item.path.relative_to(feature_dir)
```

Keeping it open for now to allow for more comments

arvind4501:
Yes, I was thinking about marking the bmc feature dynamically, so when foreman-proxy is enabled, base_test.py runs the base proxy tests, and if bmc is also enabled, bmc_test.py runs the BMC-specific tests. 
That way we could avoid manual `@pytest.mark.feature('bmc')` ,  the hook could derive the sub-feature marker from the filename (e.g., bmc_test.py → bmc)

Not blocking, just a thought to make things more dynamic

ekohl:
As noted in https://github.com/theforeman/foremanctl/pull/509#discussion_r3272657232, I've tried to avoid making things too dynamic and magical. I still remember https://code.djangoproject.com/wiki/RemovingTheMagic and think that's a good thing: I like it when things are easily predictable.

arvind4501:
> I've tried to avoid making things too dynamic and magical.

Fair, thanks

arvind4501:
Lets give coderabbit a try @coderabbitai review

ekohl:
> New VMaaS integration tests validate service state, systemd dependencies, Podman secrets, and data volumes.

@coderabbitai appears to be unable to track file moves so it sees this as a new tests.

arvind4501:
> You're correct — apologies for the confusion

Good to see AI accepting its mistakes gracefully :smile:

ehelms:
@ekohl I applied the suggestion but broke `ruff`:

```
Run ruff check --output-format=github tests/ src/ development/scripts/ inventories/
Error: tests/conftest.py:4:21: F401 `pathlib.Path` imported but unused
```

Mind addressing that?

ekohl:
> Mind addressing that?

Fixed up the first commit to apply the suggestion and still keep ruff happy. I have a memory of doing so already, but I guess I never pushed that change.

ehelms (approved):

ekohl:
Rebased to resolve a merge conflict: 50276b15a3ef0cd6adf48ac48bf2db4ac3203ba3 introduced `tests/iop/images_test.py` that is now also adhering to the new structure.

ehelms:
@ekohl Can I ask you for another rebase?

ekohl:
Apologies, I missed your comment. Rebased now.

ehelms (approved):
I'm generally happy with this change, needs a rebase and will require rebases by most PRs we have open. I'm not sure there is ever going to be great timing so I say we just do it. @arvind4501 thoughts given your RFC?

ekohl:
Rebased to resolve the minor merge conflict.

arvind4501 (approved):
It looks neat, and matches with my proposal in RFC about having feature markers + flavor directories

PR #511, opened 2026-05-15
Introduce a plain foreman flavor

ekohl:

Today Foreman can be deployed without Katello on both RPM and Debian installations and this allows users to mirror that.

The flavor itself is rather trivial, but there are some changes made to only deploy certain roles depending on the features enabled. That is likely incomplete. In particular, Apache is not yet modified to only deploy the required services. In the future when a plain foreman-proxy role is introduced then that will need adjustment.

There are also heavy testing changes, heavily based on https://github.com/theforeman/foremanctl/pull/509 which in turn is based on https://github.com/theforeman/foremanctl/pull/508. I'm wondering how much of those commits already belong there.

Right now the test suite fails because the `foreman` feature is only set in the flavor and not in the parameters.

It's also important to look at how this will play into https://github.com/theforeman/foremanctl/pull/431.

Steps to reproduce:

* Run `foremanctl deploy --flavor foreman`
* Observe only Foreman it deployed without Pulp and Candlepin
* Test suite passes

#### Checklist


tests/postgresql_test.py
```diff
@@ -13,18 +13,38 @@ def test_postgresql_port(database):
...
+        if "foreman" in features:
+            assert "candlepin" in result.stdout
+        assert "pulp" in result.stdout
+    else:
+        assert "candlepin" not in result.stdout
+        assert "pulp" not in result.stdout
 
 
-def test_postgresql_users(database):
+def test_postgresql_users(features, database):
```
arvind4501: Here why are we using features instead of enabled_features? features will bring all the features(available + enabled)

arvind4501:
Our CI is currently not designed to run flavor specific tests https://github.com/theforeman/foremanctl/blob/master/.github/workflows/test.yml#L54, it adds all features but surely that won't make sense, as we want to only add relevent features for a flavor(ex: add bmc to plain foreman does not make any sense), 

Thinking out loud:-
Should we make test workflow per flavor, I am still trying to get all data points which can help in multi-flavor testing

ekohl:
> Should we make test workflow per flavor, I am still trying to get all data points which can help in multi-flavor testing

If you're talking about a GitHub Actions workflow: I think that does make sense. We probably can't test all combinations, but finding a good set is exactly what I think we should do.

tests/postgresql_test.py
ekohl: Good catch. Left over from an older version where `features` was what is now `enabled_features`.

arvind4501:
> If you're talking about a GitHub Actions workflow


yes,and i think without flavor specific workflows we might end up with a single tests workflow which adds all available features(as of today) and will keep CI red

ekohl:
I'd still like to see this. Some PRs that are opened because of this, but can be merged already:
* https://github.com/theforeman/foremanctl/pull/509
* https://github.com/theforeman/foremanctl/pull/655
* https://github.com/theforeman/foremanctl/pull/656
* https://github.com/theforeman/foremanctl/pull/657

After that I'll rebase.

ekohl:
The exception is that `import_paths` and `export_paths` are unbound in this code:
https://github.com/theforeman/foremanctl/blob/4d2d7566da66b1ccdf5e9b5509d11b7b4308e4bf/tests/pulp_test.py#L11-L22
That means the `parameters.yaml` file doesn't exist.

ekohl:
> That means the `parameters.yaml` file doesn't exist.

Ah, I moved the file around and it tries to derive the root from the file location.

tests/feature/katello/pulp_test.py
```diff
@@ -4,17 +4,17 @@
 import pytest
 import yaml
 
+from tests.conftest import PARAMETERS_FILE
```
ekohl, line 7: I wasn't sure about this, but I see `tests/feature/foreman-proxy/base_test.py` also imports `FOREMAN_PROXY_PORT`.

Then there's also this:
https://github.com/theforeman/foremanctl/blob/9ffe3604f01662fcd67edc77b322556fdb1881fd/tests/migration_test.py#L7-L9
Should it be a global fixture?

ekohl:
I think this is largely ready, but we need to rethink CI a bit here. Adding IOP to plain Foreman implies adding Katello to it. While that's nice and all, it means we don't really test just Foreman. The additional proxy tests also all just fail.

src/playbooks/deploy-proxy/deploy-proxy.yaml
```diff
@@ -31,7 +31,9 @@
         - database_mode == 'internal'
     - valkey
     - httpd
-    - pulp
+    - role: pulp
+      when:
+        - "enabled_features | has_feature('katello')"
```
arvind4501: in case of `foreman-proxy-content` flavor, we don't have katello feature enabled, so that would skip pulp, but we need this https://github.com/theforeman/foremanctl/blob/9ffe3604f01662fcd67edc77b322556fdb1881fd/src/vars/flavors/foreman-proxy-content.yml#L1

ekohl: Good catch. I think we need to introduce some feature to guard for Pulp (also for https://github.com/theforeman/foremanctl/pull/527). Here I chose `katello`, but I now notice there is also a `pulp` feature that you introduced in 2198602d41c83e92c6e8568fc0bea642771a62d1. I'll update it.

tests/feature/katello/pulp_test.py
arvind4501: I think using https://github.com/theforeman/foremanctl/blob/9ffe3604f01662fcd67edc77b322556fdb1881fd/tests/conftest.py#L20 from confest make sense to me, thoughts?

tests/httpd_test.py
```diff
@@ -37,36 +37,42 @@ def test_https_foreman_ping(server, certificates, server_fqdn):
...
 
 
+@pytest.mark.feature('pulp')
 def test_https_pulp_content(server, certificates, server_fqdn):
     cmd = server.run(f"curl --silent --cacert {certificates['server_ca_certificate']} https://{server_fqdn}/pulp/content/")
     assert cmd.succeeded
     assert "Index of /pulp/content/" in cmd.stdout
 
 
+@pytest.mark.feature('pulp')
```
arvind4501, line 68: I am sure that  `./foremanctl features` will not show pulp as feature directly, its a internal feature, which can be found as `has_feature` filter but not as foremanctl features output, so these tests will be skipped always

tests/feature/katello/pulp_test.py
ekohl: Are you suggesting to also update the migration test to use the constant?

arvind4501: yes, looking at its used just to read params from the file , i think having a single defination for obsah state dir makes much more sense to me

tests/httpd_test.py
ekohl: So we're finally hitting the point we discussed at the start when we talked about feature-based tsets: the test suite needs to know about internal features. Good catch.

tests/feature/katello/pulp_test.py
ekohl: So question: should be have a global fixture or a constant? I'm leaning to avoid importing constants and instead rely on fixtures, but curious what you think.

src/playbooks/deploy/deploy.yaml
```diff
@@ -30,10 +30,15 @@
       when:
         - database_mode == 'internal'
     - valkey
-    - candlepin
+    - role: candlepin
+      when:
+        - "enabled_features | has_feature('foreman')"
+        - "enabled_features | has_feature('katello')"
```
ekohl, line 36: I missed there's now a built in `candlepin` feature that should be used.
```suggestion
        - "enabled_features | has_feature('candlepin')"
```

tests/feature/katello/pulp_test.py
arvind4501: I think those both ways are valid, but personally for me i would choose constant as the state path is going to remain static throughout the test execution and does not need the overhead of using fixtures, but i might be wrong, thats my thesis

tests/httpd_test.py
arvind4501: Yes, thats what forced me to use flavor directories for such cases https://github.com/theforeman/foremanctl/blob/master/tests/flavor/foreman-proxy-content/pulp_test.py ,
but yeah we can have a forge command to list internal features too

ekohl: At least the test suite is correctly flagging that:
```
ERROR tests/httpd_test.py::test_http_pulp_api_status - pytest.PytestConfigWarning: Invalid feature(s) {'pulp'} on <Function test_http_pulp_api_status>
ERROR tests/httpd_test.py::test_https_pulp_api_status - pytest.PytestConfigWarning: Invalid feature(s) {'pulp'} on <Function test_https_pulp_api_status>
ERROR tests/httpd_test.py::test_http_pulp_content - pytest.PytestConfigWarning: Invalid feature(s) {'pulp'} on <Function test_http_pulp_content>
ERROR tests/httpd_test.py::test_https_pulp_content - pytest.PytestConfigWarning: Invalid feature(s) {'pulp'} on <Function test_https_pulp_content>
ERROR tests/httpd_test.py::test_https_pulp_auth - pytest.PytestConfigWarning: Invalid feature(s) {'pulp'} on <Function test_https_pulp_auth>
ERROR tests/httpd_test.py::test_https_pypi_endpoint - pytest.PytestConfigWarning: Invalid feature(s) {'pulp'} on <Function test_https_pypi_endpoint>
```

tests/feature/katello/pulp_test.py
ekohl: I went with reusing the existing `obsah_params` fixture in https://github.com/theforeman/foremanctl/pull/719 instead.

ekohl:
I've decided to split up some parts. https://github.com/theforeman/foremanctl/pull/719 is already merged and I just opened https://github.com/theforeman/foremanctl/pull/722. Once that's merged, this and https://github.com/theforeman/foremanctl/pull/527 should be much smaller. Then we can focus on how to actually test those flavors.

ekohl:
Right now all the prerequisites I could think of are merged. This now needs testing; both manual and CI.

ekohl:
Did some manual testing and found that we deploy Pulp's rules in Apache unconditionally. Also the tuning checks don't really match what Foreman requires. Other than that this now fundamentally works: it only deploys the services plain Foreman needs and nothing else.

PR #512, opened 2026-05-15
Rewrite compute resources test to not require Hammer

ekohl:

Hammer is an optional feature that may not be deployed. It's also faster to query the API directly. @ehelms also ran into this with https://github.com/theforeman/foremanctl/pull/485/changes#diff-b8717c3b94dcdcb21359fefd8fc5b4ae1d84b9a1806547228b6c8f7e855891ed and here it doesn't bring value to use Hammer.

This uses the API directly to extract the right information. It would be a lot easier if there was an API endpoint to query this programatically.

Steps to reproduce:

* Let CI run
* Verify it's green

#### Checklist

### DIFF
--- a/tests/foreman_compute_resources_test.py
+++ b/tests/foreman_compute_resources_test.py
@@ -1,11 +1,9 @@
 import pytest
 
-FOREMAN_HOST = 'localhost'
-FOREMAN_PORT = 3000
-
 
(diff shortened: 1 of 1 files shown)


ekohl:
Oops, I already submitted this in https://github.com/theforeman/foremanctl/pull/510.

PR #516, opened 2026-05-21
Use 3.19 containers

evgeni:

We need a release that uses 3.19

* Use 3.19

Steps to reproduce:

* CI

#### Checklist

### DIFF
--- a/src/vars/images.yml
+++ b/src/vars/images.yml
@@ -19,7 +19,7 @@ iop_ingress_registry_auth_file: "{{ registry_auth_file }}"
 iop_vulnerability_frontend_registry_auth_file: "{{ registry_auth_file }}"
 iop_advisor_frontend_registry_auth_file: "{{ registry_auth_file }}"
 
-container_tag_stream: "nightly"
+container_tag_stream: "3.19"
 candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "foreman-{{ container_tag_stream }}"
 foreman_container_image: quay.io/foreman/foreman
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -231,7 +231,7 @@ jobs:
         upgrade_from:
           - '3.18'
         upgrade_to:
-          - 'nightly'
+          - '3.19'
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v6
--- a/.packit.yml
+++ b/.packit.yml
@@ -13,7 +13,7 @@ srpm_build_deps:
 
 actions:
   post-upstream-clone:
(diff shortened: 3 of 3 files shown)


evgeni:
@ehelms @ekohl so, following up from #424 
- I created a `2.y-stable` branch (or should it have been `2.1-stable`? idk) based on 3f6912d019208ebb5ff01414163bf431bbb8881e (the later changes are not in 3.19)
- I've opened *this* PR targeting the `2.y-stable` branch, updating `nightly` to `3.19`
- Once this lands, we tag the result as `2.2.0` (or whatever) and create a `3.0-develop` tag on `master`

src/vars/images.yml
```diff
@@ -19,7 +19,7 @@ iop_ingress_registry_auth_file: "{{ registry_auth_file }}"
 iop_vulnerability_frontend_registry_auth_file: "{{ registry_auth_file }}"
 iop_advisor_frontend_registry_auth_file: "{{ registry_auth_file }}"
 
-container_tag_stream: "nightly"
+container_tag_stream: "3.19"
```
arvind4501, line 22: Out of curiosity: what if we use `3.19-stable`, does it restrict us to update to .z images?

evgeni: The `3.19-stable` tag on https://quay.io/repository/foreman/foreman?tab=tags&tag=latest is the one we build from source (via https://github.com/theforeman/foreman/blob/develop/Dockerfile).

`3.19` is the one following .z RPM releases, currently pointing at rc1. Using that will ensure that users will get always "the latest" 3.19 we released.

evgeni:
@ehelms how should we handle dev setup when branching? Ideally, it would follow, but right now it's not possible to configure compatible gems for plugins/hammer that match the foreman version.

src/vars/images.yml
Odilhao: The tags are defined here https://github.com/theforeman/theforeman-rel-eng-konflux/blob/develop/releases/foreman/3.19/settings#L7 and later applied to https://gitlab.com/fedora/infrastructure/konflux/tenants-config/-/blob/main/clusters/kflux-fedora-01/tenants/theforeman-org-tenant/foreman/releaseplans/3.19/kustomization.yaml?ref_type=heads#L29 that creates the release objects for us on Konflux.

ehelms (approved):
Let's worry about dev after, as you note, there's a few nuances to work out. This is an example workflow where I wonder if source containers would make that alignment easier.

Odilhao:
> Let's worry about dev after, as you note, there's a few nuances to work out. This is an example workflow where I wonder if source containers would make that alignment easier.

Should it use nightly directly for `dev`? Instead of using branched/stable for devs?

PR #517, opened 2026-05-23
Only pass in Candlepin secrets instead of full configs

ekohl:

This uses an updated Candlepin image that contains templates for config files and uses environment variables to provide configuration. On the Ansible side these are provided via podman secrets.

It depends on https://github.com/theforeman/candlepin-oci-images/pull/3 and has a temporary commit pointing to a build.

This reduces the coupling between how the container is built and configured. If the container is rebuilt using EL 10 and some newer Tomcat version then the server.xml may become incompatible, but the environment variables remain the same. Effectively creating an interface.

Steps to reproduce:

* Check out code
* Run `foremanctl deploy`
* Verify `candlepin.service` only mounts in environment variables, not config files

#### Checklist


ehelms:
From my previous testing, this strategy will not work because we have dynamic configuration. See lines:

https://github.com/theforeman/foremanctl/pull/517/changes#diff-3b332a0f35360d8f303b982469421de30d4ae6db5dc4f9d7470fd9f81ce7f5ecL34-L38

https://github.com/theforeman/foremanctl/pull/517/changes#diff-3b332a0f35360d8f303b982469421de30d4ae6db5dc4f9d7470fd9f81ce7f5ecL29

Line 29 defines the name of the prefix, and then 34-38 are what use that prefix.

ekohl:
I don't quite follow. Those strings are hardcoded and https://github.com/theforeman/candlepin-oci-images/pull/3 contains those exact prefixes.

Note I've taken the approach of templating the entire Katello config in the container, not pass everything in via Candlepin's own env var mechanism.

PR #519, opened 2026-05-25
Add Templates feature to foreman-proxy

shubhamsg199:

* Enable templates feature for foreman-proxy

Steps to reproduce:

* foremanctl deploy --add-feature templates

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,9 +153,6 @@ jobs:
             --add-feature remote-execution \
             --add-feature bmc \
             --add-feature webhooks \
-            --add-feature templates \
-            --templates-listen-on http \
-            --templates-url http://quadlet.example.com:8000 \
             ${{ matrix.iop == 'enabled' && '--add-feature iop' || '' }}
       - name: Run tests
         run: |
@@ -446,7 +443,9 @@ jobs:
             --certificate-bundle $(pwd)/.var/lib/foremanctl/proxy.example.com.tar.gz \
             --foreman-fqdn quadlet.example.com \
             --oauth-consumer-key $(cat .var/lib/foremanctl/foreman-oauth-consumer-key) \
-            --oauth-consumer-secret $(cat .var/lib/foremanctl/foreman-oauth-consumer-secret)
+            --oauth-consumer-secret $(cat .var/lib/foremanctl/foreman-oauth-consumer-secret) \
+            --templates-listen-on http \
+            --templates-url http://proxy.example.com:8000
       - name: Run tests
         run: |
           ./forge test --pytest-args="--server-hostname=proxy"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,6 +153,9 @@ jobs:
             --add-feature remote-execution \
             --add-feature bmc \
             --add-feature webhooks \
+            --add-feature templates \
+            --templates-listen-on http \
+            --templates-url http://quadlet.example.com:8000 \
             ${{ matrix.iop == 'enabled' && '--add-feature iop' || '' }}
       - name: Run tests
         run: |
--- a/docs/user/parameters.md
+++ b/docs/user/parameters.md
@@ -128,6 +128,9 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--add-feature content/ostree` | Enable OSTree content type | `--foreman-proxy-content-enable-ostree` |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
+| `--add-feature templates` | Enable Templates feature on Smart Proxy | `--foreman-proxy-templates` |
+| `--templates-listen-on` | Templates proxy to listen on https, http, or both | `--foreman-proxy-templates-listen-on` |
+| `--templates-url` | URL that hosts will use to contact the proxy for provisioning templates | `--foreman-proxy-templates-url` |
 
 ### Undetermined
 
@@ -200,8 +203,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-proxy-realm-provider` | | | |
 | `--foreman-proxy-registration` | | | |
 | `--foreman-proxy-registration-url` | | | |
-| `--foreman-proxy-templates` | | | |
-| `--foreman-proxy-template-url` | | | |
 | `--puppet-server` | | puppet | server |
 | `--puppet-server-ca` | | puppet | server_ca |
 | `--puppet-dns-alt-names` | | puppet | dns_alt_names |
--- a/src/playbooks/_foreman_proxy/metadata.obsah.yaml
+++ b/src/playbooks/_foreman_proxy/metadata.obsah.yaml
@@ -10,3 +10,13 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_templates_listen_on:
+    parameter: --templates-listen-on
+    help: Templates proxy to listen on https, http, or both.
+    choices:
+      - http
+      - https
+      - both
+  foreman_proxy_templates_url:
+    parameter: --templates-url
+    help: URL that hosts will use to contact the proxy for provisioning templates (e.g. http://<proxy-fqdn>:8000). Required when templates feature is enabled.
--- a/src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
+++ b/src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
@@ -1,0 +1,4 @@
+---
+:enabled: {% if foreman_proxy_templates_listen_on | default('both') == 'both' %}true{% else %}{{ foreman_proxy_templates_listen_on }}{% endif %}
+
+:template_url: {{ foreman_proxy_templates_url }}
--- a/tests/feature/foreman-proxy/base_test.py
+++ b/tests/feature/foreman-proxy/base_test.py
@@ -27,6 +27,10 @@ def test_foreman_proxy_features(curl_request, proxy_base_url, enabled_features):
         assert "bmc" in features
     else:
         assert "bmc" not in features
+    if 'templates' in enabled_features:
+        assert "templates" in features
+    else:
+        assert "templates" not in features
 
 
 def test_foreman_proxy_service(server):
@@ -65,3 +69,10 @@ def test_bmc_capabilities(proxy_v2_features):
 def test_bmc_default_provider(proxy_v2_features):
     settings = proxy_v2_features['bmc'].get('settings', {})
     assert settings.get('bmc_default_provider') == 'ipmitool'
+
+
+@pytest.mark.feature('templates')
+def test_templates_template_url(proxy_v2_features):
+    settings = proxy_v2_features['templates'].get('settings', {})
+    template_url = settings.get('template_url')
+    assert template_url == 'http://quadlet.example.com:8000'
--- a/tests/feature/foreman-proxy/base_test.py
+++ b/tests/feature/foreman-proxy/base_test.py
@@ -72,7 +72,6 @@ def test_bmc_default_provider(proxy_v2_features):
 
 
 @pytest.mark.feature('templates')
-def test_templates_template_url(proxy_v2_features):
-    settings = proxy_v2_features['templates'].get('settings', {})
-    template_url = settings.get('template_url')
-    assert template_url == 'http://quadlet.example.com:8000'
+def test_templates_template_url(obsah_params, server_fqdn):
+    template_url = obsah_params.get('foreman_proxy_templates_url')
+    assert template_url == f'http://{server_fqdn}:8000'
--- a/src/features.yaml
+++ b/src/features.yaml
@@ -71,3 +71,7 @@ webhooks:
   foreman:
     plugin_name: foreman_webhooks
   hammer: foreman_webhooks
+templates:
+  description: Templates feature for foreman-proxy
+  foreman_proxy:
+    plugin_name: templates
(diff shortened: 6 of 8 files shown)


src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,16 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_templates_listen_on:
```
ehelms: We are using `templates` and `template` which is confusing to as a user, let's pick one and be consistent. I also wonder if we should lean into using the word "provisioning" here as I have heard users be confused about the kind of templates that are used throughout Foreman.

shubhamsg199: Do you mean tighten the help text to say something like provisioning templates everywhere or adjust flag naming altogether?

.github/workflows/test.yml
```diff
@@ -159,6 +159,7 @@ jobs:
             --add-feature google \
             --add-feature remote-execution \
             --add-feature bmc \
+            --add-feature templates \
```
stejskalleos: Can we add `--templates-listen-on` & `--templates-url` to the test as well?

stejskalleos (changes requested):

.github/workflows/test.yml
shubhamsg199: Added

ehelms:
Does templates need to be in the features.yaml? I do not see how the feature in this PR ensures foreman-proxy gets added.

shubhamsg199:
> Does templates need to be in the features.yaml? I do not see how the feature in this PR ensures foreman-proxy gets added.

Yes, templates is declared in features.yaml under the foreman_proxy key (same pattern as bmc). I don't think we have a way to ensure if foreman-proxy gets added.

shubhamsg199:
@ehelms Made some updates that ensures foreman-proxy gets added on the feature install. Not sure if that should be part of this PR.

tests/foreman_proxy_test.py
```diff
@@ -69,3 +73,16 @@ def test_bmc_capabilities(proxy_v2_features):
 def test_bmc_default_provider(proxy_v2_features):
     settings = proxy_v2_features['bmc'].get('settings', {})
     assert settings.get('bmc_default_provider') == 'ipmitool'
+
+
+@pytest.mark.feature('templates')
+def test_templates_feature_enabled(proxy_v2_features):
```
arvind4501: /v2/features will always have templates, so its redundent, we alredy tested if feature is enabled https://github.com/theforeman/foremanctl/pull/519/changes#diff-887102e2cefc161bc9879bf5389d43dbe769a1bb2dd927ace5fd2ad57fc761bcR33

src/filter_plugins/foremanctl.py
evgeni, line 1: Do the changes to this file mean that if you do `--add-feature templates` on fresh, feature-less deployment, it will also enable the (base) `foreman-proxy` feature?

shubhamsg199: Yes

evgeni: I think that's wrong, this should be done by the user deliberately.

Especially, this means that enabling any other feature that has a proxy aspect will enable the proxy (like REX), if I am reading the code correctly.

shubhamsg199: I was trying to incorporate https://github.com/theforeman/foremanctl/pull/519#issuecomment-4674550421.
`this should be done by the user deliberately.` If this is flow, then yes we don't need this.

shubhamsg199: Reverted the commit

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
```diff
@@ -0,0 +1,3 @@
+---
+:enabled: {{ feature_enabled }}
```
evgeni: as you're exposing "listen on", you'd need to alter the value of `:enabled:` accordingly. (`both` maps to `true`, the others can be used directly to only enable http/https mode.

so far we have not been opening the http port at all, and all plugins are configured for both/true.

given templates is the first one that is actually useful over plain http, we need to rethink that now

.github/workflows/test.yml
```diff
@@ -152,6 +152,9 @@ jobs:
...
+            --templates-listen-on http \
+            --templates-url http://quadlet.example.com:8000 \
```
arvind4501: I see that we are defaulting to `http` url on proxy, but AFAIK proxy is running on https https://github.com/theforeman/foreman-oci-images/blob/master/images/foreman-proxy/Containerfile#L21, do we need to enable proxy running on plain http for this?

arvind4501: I had a stab earlier to enable proxy on http https://github.com/theforeman/foremanctl/pull/520/
(which is incomplete)

tests/foreman_proxy_test.py
```diff
@@ -69,3 +73,11 @@ def test_bmc_capabilities(proxy_v2_features):
...
     settings = proxy_v2_features['bmc'].get('settings', {})
     assert settings.get('bmc_default_provider') == 'ipmitool'
+
+
+@pytest.mark.feature('templates')
+def test_templates_template_url(proxy_v2_features):
+    settings = proxy_v2_features['templates'].get('settings', {})
+    template_url = settings.get('template_url')
+    assert template_url == 'http://quadlet.example.com:8000'
+
```
evgeni: can we write a test that actually uses the template feature to get data?

docs/user/parameters.md
```diff
@@ -112,6 +112,9 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--add-feature bmc` | Enable BMC feature | `--foreman-proxy-bmc` |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
+| `--add-feature templates` | Enable Templates feature on Smart Proxy | `--foreman-proxy-templates` |
```
stejskalleos, line 131: Can you please update the https://github.com/theforeman/foremanctl/blob/master/src/plugins/modules/migrate_answers.py#L14 as well?

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
stejskalleos: What's the point of setting it to `http` or `https`? Why not just `true` / `false`?
Do we enforce SSL when the value is `https` or `true`?

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
```diff
@@ -0,0 +1,3 @@
...
+:template_url: {{ foreman_proxy_templates_url | default('http://' + ansible_facts['fqdn'] + ':8000') }}
```
stejskalleos: Question: What if we did not provide a default value and instead required users to set it manually?

Plus, you need two default values: one for HTTP and one for HTTPS.

tests/foreman_proxy_test.py
stejskalleos: +1

.github/workflows/test.yml
```diff
@@ -152,6 +152,9 @@ jobs:
             --add-feature google \
             --add-feature remote-execution \
             --add-feature bmc \
+            --add-feature templates \
+            --templates-listen-on http \
```
stejskalleos: What if we drop it? Is this parameter still relevant?

stejskalleos (changes requested):

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
stejskalleos: > No newline at end of file

That should be fixed as well.

.github/workflows/test.yml
shubhamsg199: Templates is specifically useful over plain HTTP because provisioning hosts (e.g. during iPXE/kickstart) typically can't do SSL

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
evgeni: We don't enforce SSL in that case, no.

But if a module is enabled on plain HTTP, the authorization is reduced to "the reverse DNS matches an entry in `trusted_hosts`" aka "no authorization as everyone can fake this shit"¹.

So we need to only ever allow plain HTTP to modules that don't need authorization (like templates).

¹: This is not a problem today, as we don't expose the HTTP port at all

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
ekohl: No code should use `ansible_facts['fqdn']`, other than defaults. The whole of foremanctl should assume a mindset of decoupling the service name from the actual hostname.

If it helps to not have a default, then I'm all in favor of that.

pablomh:
Shouldn't the feature by added to `foreman_proxy_base_features`?

It's currently the default in foreman-installer: https://github.com/theforeman/foreman-installer/blob/0448d5d129f8fe2ea514cfb7fe90857e0e3f15bd/config/foreman-proxy-content-answers.yaml#L30

ekohl:
@pablomh that's only for the content proxies, not for the normal install. It could be part of the proxy flavor.

docs/user/parameters.md
shubhamsg199: Updated

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
shubhamsg199: Removed the default value, `template_url` will  need to be set manually

.github/workflows/test.yml
shubhamsg199: I've removed `template_url` so I think this is irrelevant now

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
pablomh: @stejskalleos, what's the advantage of requiring users to set it manually?

stejskalleos: > what's the advantage of requiring users to set it manually?

We don't have to guess the value for multiple scenarios.
If the user enables the templates feature, they should explicitly set the value they want to use.

.github/workflows/test.yml
jeremylenz: Should ` --add-feature templates`  be under `deploy-proxy` instead of `deploy`?

cc @ehelms @parthaa

jeremylenz:
@shubhamsg199 What is the status here?

.github/workflows/test.yml
ekohl: Yes, I don't think the feature really makes a lot of sense on the server running Foreman.

shubhamsg199: Added  `--add-feature templates`  to be under `deploy-proxy` instead of `deploy`

shubhamsg199:
> @shubhamsg199 What is the status here?

@jeremylenz Updated the PR, can you take a look?

tests/feature/foreman-proxy/base_test.py
```diff
@@ -65,3 +69,9 @@ def test_bmc_capabilities(proxy_v2_features):
 def test_bmc_default_provider(proxy_v2_features):
     settings = proxy_v2_features['bmc'].get('settings', {})
     assert settings.get('bmc_default_provider') == 'ipmitool'
+
+
+@pytest.mark.feature('templates')
+def test_templates_template_url(obsah_params, server_fqdn):
```
evgeni, line 75: we'll need #703 (or similar) to make this test work

shubhamsg199: This tests is currently passing

evgeni: oh, because you changed it to read the params, not talk to the API to fetch the set value.

stejskalleos (approved):
🍏 **LGTM** 🍏 

* Code looks clear
* All comments should be addressed.
* Tests are present

Tested with [Packit](https://dashboard.packit.dev/jobs/copr/3802218), shortly:

Running `foremanctl deploy --add-feature templates --templates-listen-on http --templates-url http://foremanctl9.vagrant.lan/` does not affect the `*-pulp` proxy.

Running `foremanctl deploy --add-feature foreman-proxy` will automatically enable `templates` on the newly running Foreman proxy service.

Merging as https://github.com/theforeman/foremanctl/pull/675 is waiting for it.

src/playbooks/_foreman_proxy/metadata.obsah.yaml
```diff
@@ -10,3 +10,13 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_templates_listen_on:
+    parameter: --templates-listen-on
+    help: Templates proxy to listen on https, http, or both.
+    choices:
+      - http
+      - https
+      - both
```
ekohl, line 19: Why is it needed to expose this? In the old installer we exposed it because it was the way we did, but it was an advanced option. In foremanctl I think shouldn't expose this.

src/playbooks/_foreman_proxy/metadata.obsah.yaml
```diff
@@ -10,3 +10,13 @@ variables:
...
+  foreman_proxy_templates_url:
+    parameter: --templates-url
+    help: URL that hosts will use to contact the proxy for provisioning templates (e.g. http://<proxy-fqdn>:8000). Required when templates feature is enabled.
```
ekohl, line 22: There is a default which implies it isn't required to specify.

src/playbooks/_foreman_proxy/metadata.obsah.yaml
shubhamsg199: Addressed this in https://github.com/theforeman/foremanctl/pull/706

src/playbooks/_foreman_proxy/metadata.obsah.yaml
shubhamsg199: Updated in https://github.com/theforeman/foremanctl/pull/706

PR #507, opened 2026-05-13
Add offline backup for foremanctl

sjha4:
#### Why are you introducing these changes? (Problem description, related links)
This implements offline backup with foremanctl. 
#### What are the changes introduced in this pull request?

* Offline backup

  * Adds `foremanctl backup` command for offline backups
  * Implements comprehensive backup of core deployment components:
    * All databases (foreman, candlepin, pulp, IOP databases)
    * Foremanctl state directory
    * Pulp content directory (optional via `--skip-pulp-content`)
    * Container image metadata (digests for restore compatibility)
  * Preflight checks to ensure backup safety:
    * Verifies no running Foreman/Pulp tasks (with `--wait-for-tasks` option)
    * Database integrity verification using PostgreSQL amcheck extension
  * Automatic service restoration on backup failure
  * Adds comprehensive user documentation at `docs/user/backup.md`
  * New roles: `backup`, `check_database_index`

#### How to test this pull request
I got a foremanctl box with normal deploy. On this box, clone foremanctl repo and checkout this branch.
cd /root/foremanctl
source .venv/bin/activate
export OBSAH_STATE=/var/lib/foremanctl

Then try ./foremanctl --help

Also,

```
(.venv) [root@ip-10-0-167-40 foremanctl]# ./foremanctl backup  --help
usage: foremanctl backup [-h] [-v] [--skip-pulp-content][--wait-for-tasks] backup_dir

Create offline backup of Foreman databases and configuration

positional arguments:
  backup_dir            Directory where backup files will be stored

optional arguments:
  -h, --help            show this help message and exit
  -v, --verbose         verbose output
  --skip-pulp-content   Skip Pulp content directory backup
  --wait-for-tasks      Wait for running tasks to complete instead of failing immediately

```
You can run :
**Command:**
   ```bash
   cd /root/foremanctl
   source .venv/bin/activate
   export OBSAH_STATE=/var/lib/foremanctl
   ./foremanctl backup /var/tmp/foreman-backup-test --wait-for-tasks
   ```

Steps to reproduce:

* On a foremanctl box, run foremanctl backup BACKUP_DIR

#### Checklist
* [ ] Tests added/updated (if applicable)
* [ ] Documentation updated (if applicable)

PR #521, opened 2026-05-26
Fixes #525 - Add checks to health subcommand

### LINKED ISSUE #525: Add additional features to foremanctl health subcommand
Of the default health checks from foreman-maintain, `foremanctl health` currently implements the following checks:
- services_up
- server_ping
- foreman_tasks/not_paused

Please add the `foreman/facts_names` check and a sanity-check unit test for the health checks on a working environment. Check if there are any more relevant default checks we missed (or non-default worth including).

Note: the following checks will not need to be migrated:
- foreman_proxy/check_tftp_storage - No foreman proxy support yet.
- foreman_proxy/verify_dhcp_config_syntax - Ditto.
- puppet/verify_no_empty_cacert_requests - Puppet plugins for foreman and smart proxy are not yet compatible with foremanctl; work is underway to add support in the future.
- system_registration - This self-registration concern isn't valid for an upstream centos9 container, to the best of my knowledge.

qcjames53:
#### Why are you introducing these changes? (Problem description, related links). What are the changes introduced in this pull request?
This PR expands the existing `foremanctl health` subcommand to reach equivalency with the existing `foreman-maintain health check` tool.

The following tests are run on a default health check in foreman-maintain:
- services_up
- server_ping
- foreman_tasks/not_paused
- foreman/facts_names
- foreman_proxy/check_tftp_storage
- foreman_proxy/verify_dhcp_config_syntax
- puppet/verify_no_empty_cacert_requests
- system_registration

I've implemented these five checks:
- `health_services_up` - Checks all currently implemented quadlet services are running (with different configs supported)
- `health_server_ping` - Can we ping the foreman server?
- `health_foreman_tasks_not_paused` - Are there paused foreman tasks?
- `health_facts_count` - Does a host exist with more facts than `health_max_fact_values_per_host`?
- `check_duplicate_permissions` - Are permissions duplicated.

I've skipped these checks (reasoning provided):
- foreman_proxy/check_tftp_storage - No foreman proxy support yet.
- foreman_proxy/verify_dhcp_config_syntax - Ditto.
- puppet/verify_no_empty_cacert_requests - Puppet plugins for foreman and smart proxy are not yet compatible with foremanctl; work is underway to add support in the future.
- system_registration - This self-registration concern isn't valid for an upstream centos9 container, to the best of my knowledge. **Please double check this reasoning :)**

Steps to reproduce:

1. Spin up an environment and register a host.
2. Run `foremanctl health` and verify all checks pass.
3. `vagrant ssh quadlet` and manually stop a service. Rerun the check and confirm `health_services_up` fails. Restart the service.
4. Close the server firewall and confirm `health_server_ping` fails. Change the config back afterwards.
5. Start a sync (or similar) and pause the job. confirm `health_foreman_tasks_not_paused` fails.
6. Lower the `health_max_fact_values_per_host` value in `src/vars/health.yml` to below the number of facts of your host and confirm `health_facts_count` fails.
7. Run the unit tests and confirm they work. `./forge test` first to generate needed files then run `pytest tests/health_test.py`.
8. Create duplicate permissions in the foreman db and verify the health check fails. Run with `--fix` and see that the duplicates are cleared.

#### Checklist
* [ ] Documentation updated (if applicable) <- Didn't see anything relevant to modify

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -161,9 +161,6 @@ jobs:
             --add-feature remote-execution \
             --add-feature bmc \
             ${{ matrix.iop == 'enabled' && '--add-feature iop' || '' }}
-      - name: Run health check
-        run: |
-          ./foremanctl health
       - name: Run tests
         run: |
           ./forge test --pytest-args="--certificate-source=${{ matrix.certificate_source }} --database-mode=${{ matrix.database }}"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,8 +1,4 @@
-# AI Development Guide
-
-This file provides guidance to AI tools and agents when working with code in this repository. It is intentionally tool-agnostic -- it works with Claude Code, Cursor, GitHub Copilot, and any other AI assistant.
-
-## Project Overview
+## Foremanctl
 
 foremanctl is a deployment tool for Foreman and Katello using Podman quadlets and Ansible. It provides a command-line interface built on top of `obsah` (an Ansible wrapper) for managing containerized Foreman deployments.
 
@@ -126,8 +122,8 @@ Roles in `src/roles/` correspond to services and deployment stages:
 - Service roles: `foreman`, `pulp`, `candlepin`, `postgresql`, `redis`, `httpd`
 - Feature roles: `hammer`, `foreman_proxy`
 - Infrastructure: `certificates`, `systemd_target`
-- Checks: `check_hostname`, `check_database_connection`, `check_system_requirements`, `check_subuid_subgid`
 - Lifecycle: `pre_install`, `post_install`
+- Checks: `check_*` roles; see [documentation](docs/developer/checks.md) for more info
 
 ### Configuration System
 
@@ -170,15 +166,23 @@ Tests use pytest with testinfra for infrastructure testing. Key fixtures in `tes
 
 Test files follow the pattern `tests/<component>_test.py` and use testinfra's server fixture to execute commands and check system state.
 
-## Developer Documentation
+## Additional Documentation
 
-- [How to Add a Feature](docs/developer/how-to-add-a-feature.md) - end-to-end feature development
-- [Playbooks and Roles](docs/developer/playbooks-and-roles.md) - playbook structure, naming, metadata
-- [Testing](docs/developer/testing.md) - test infrastructure, fixtures, patterns
-- [Deployment Design](docs/developer/deployment.md) -- deployment architecture
-- [Container Image Builds](docs/developer/container-image-builds.md) - image naming, registries
-- [Development Environment](docs/developer/development-environment.md) - dev setup with git Foreman
+Developer docs:
+- [Check roles](docs/developer/checks.md) - How to integrate check roles; update as checks are created/modified
+- [Container Image Builds](docs/developer/container-image-builds.md) - Info on image naming, registries
+- [Deployment Architecture](docs/developer/deployment.md)
+- [Development Environment](docs/developer/development-environment.md) - Dev environment setup with Foreman from source
+- [How to Add a Feature](docs/developer/how-to-add-a-feature.md) - End-to-end feature development
+- [Playbooks and Roles](docs/developer/playbooks-and-roles.md) - Playbook structure, naming, metadata
+- [Testing](docs/developer/testing.md) - Additional info on test infrastructure, fixtures, patterns
 
-## Git Workflow
+User docs:
+- [Certificates](docs/user/certificates.md) - Overview of certificate sources
+- [Parameters](docs/user/parameters.md) - Map of Foreman installation parameters; update as parameters are created/modified
 
-Main branch: `master`
+- [CONTRIBUTING](CONTRIBUTING.md) - How to contribute
+- [Development](DEVELOPMENT.md) - Foremanctl development overview
+- [IOP](docs/iop.md) - Overview of insights on premise
+- [Migration Guide](docs/migration-guide.md) - Migrating from foreman-installer to foremanctl
+- [Release](RELEASE.md) - Info on Foremanctl releases
--- a/docs/developer/checks.md
+++ b/docs/developer/checks.md
@@ -1,0 +1,74 @@
+# Checks
+Foremanctl contains several "check_*" roles which are used to assist playbooks in confirming the Foreman server is configured properly. Each check is implemented as an independent Ansible role, allowing playbooks to run specific checks or groups of checks as needed.
+
+Please update this file as check usage evolves.
+
+## Check Descriptions
+### check_database_connection
+- **Description**: Validates PostgreSQL connectivity for external Foreman and Candlepin databases using configured credentials and SSL settings.
+- **Fail state**: Fails if database connection cannot be established.
+- **Rationale**: Database access is required for many foremanctl operations; external databases must be reachable using the provided credentials.
+
+### check_duplicate_permissions
+- **Description**: Queries the Foreman database for duplicate entries in the permissions table.
+- **Fail state**: Fails if duplicate permissions are detected.
+- **Rationale**: A validation was incorrectly removed which prevented users from creating duplicate Foreman permissions, causing upgrade failure. This check will need to be included until https://projects.theforeman.org/issues/38465 is addressed.
+
+### check_features
+- **Description**: Ensures that all foremanctl features requested (via the `--feature` flag) are valid.
+- **Fail state**: Fails when requested features are not recognized by foremanctl.
+- **Rationale**: This check handles input sanitization for `--feature`.
+
+### check_foreman_api
+- **Description**: Pings Foreman API (/api/v2/ping) to verify it responds. Verifies foreman_tasks service status is 'ok' when Katello/content feature is enabled.
+- **Fail state**: Fails if ping fails or service status is not 'ok'.
+- **Rationale**: An API connection is necessary for most Foreman operations. This check ensures there are no firewall or address issues.
+
+### check_foreman_tasks
+- **Description**: Queries Foreman tasks for paused, errored tasks.
+- **Fail state**: Fails if any errored tasks are found.
+- **Rationale**: Errored Foreman tasks indicate issues which need to be addressed by the user. This can be anything from a typo to systemic issues. Consulting https://community.theforeman.org may provide insight.
+- **Skipping**: This role can be skipped in `foremanctl health` by passing the `--skip-check-foreman-tasks` flag. Use only for operations where a failed Foreman task is expected or unavoidable.
+
+### check_host_facts_count
+- **Description**: Ensures all hosts' facts counts are below a maximum threshold.
+- **Fail state**: Fails if facts count exceeds threshold for any host.
+- **Rationale**: Very high host facts count causes slow facts processing. See: https://access.redhat.com/solutions/4163891 for more information.
+
+### check_hostname
+- **Description**: Validates FQDN is not 'localhost' or 'localhost.localdomain', contains at least one dot, and has no underscores.
+- **Fail state**: Fails if FQDN conditions are not met.
+- **Rationale**: Foreman requires a properly-configured server FQDN. This check ensures server hostname was modified from the default and approximates a valid FQDN.
+
+### check_services
+- **Description**: Reports the status of all non-recurring services in the foreman.target systemd dependency tree.
+- **Fail state**: Fails if any services are inactive.
+- **Rationale**: Failed or stalled foreman services indicate an issue with the server. The user should check Foreman logs to find and address the problem. Inactive foreman services may indicate a systemctl configuration problem.
+
+### check_subuid_subgid
+- **Description**: Verifies `/etc/subuid` and `/etc/subgid` have entries for the current user, which is required for rootless Podman on the Foreman server.
+- **Fail state**: Fails if these conditions are not met.
+- **Rationale**: Any Podman container operation run from a non-privileged user (pull, run, secret management) requires configuring subordinate user and group IDs. This check is unused at the moment but supports future work setting up Podman under a non-root user.
+
+### check_system_requirements
+- **Description**: This check validates the Foreman system meets minimum hardware requirements determined by the user's tuning profile.
+- **Fail state**: Fails if these conditions are not met or if the tuning profile cannot be loaded.
+- **Rationale**: A system which does not meet minimum hardware specs may stall or fail due to Foreman resource usage.
+
+## Skipping checks
+
+Users may wish to skip certain checks on some playbooks, usually to work around a known issue on particularly sensitive checks. For example `health.yaml` (`foremanctl health`) includes a pattern for skipping `check_foreman_tasks`, as errored Foreman tasks which have not yet been cleared will fail the playbook.
+
+When allowing the user to skip a check, ensure it provides tangible value greater than the consequences of user misuse. For example, allowing the user to skip `check_system_requirements` might seem harmless, but could likely lead to situations in which usage of the parameter becomes a silent default; changes in Foreman minimum system requirements would not be enforced. Be sure to document why a skip is required in the check descriptions section above.
+
+Add skip parameters following this template:
+```
+<playbook_name>_skip_<check_role_name>_param:
+  help: |
+    Identify the purpose of the skipped check.
+    Inform the user how to repair Foreman to pass this check (or point them in the right direction).
+    Indicate why the user would need this check and warn against overuse.
+  parameter: --skip-<check-role-name>
+  action: store_true
+  persist: false
+```
--- a/src/playbooks/health/health.yaml
+++ b/src/playbooks/health/health.yaml
@@ -7,14 +7,25 @@
     - "../../vars/defaults.yml"
     - "../../vars/flavors/{{ flavor }}.yml"
     - "../../vars/base.yaml"
+    - "../../vars/database.yml"
+    - "../../vars/health.yml"
   tasks:
     - name: Execute health checks
       ansible.builtin.include_role:
         name: checks
         tasks_from: execute_check
       loop:
+        - check_hostname
         - check_services
         - check_foreman_api
+        - check_database_connection
+        - check_foreman_tasks
+        - check_host_facts_count
+        - check_duplicate_permissions
+      when: >-
+        (item != 'check_foreman_tasks' or not (health_skip_check_foreman_tasks_param | default(false) | bool))
+      # Add additional skips here
+
     - name: Report status of health checks
       ansible.builtin.fail:
         msg: "{{ checks_results }}"
--- a/src/playbooks/health/metadata.obsah.yaml
+++ b/src/playbooks/health/metadata.obsah.yaml
@@ -1,5 +1,14 @@
 ---
 help: |
-  Run health checks on Foreman services
+  Check the health of your running Foreman server
+variables:
+  health_skip_check_foreman_tasks_param:
+    help: |
+      Skip checking Foreman tasks for paused/errored tasks.
+      Errored tasks can be manually resumed or cancelled using the Web UI or hammer.
+      Use only for operations where a failed Foreman task is expected or unavoidable.
+    parameter: --skip-check-foreman-tasks
+    action: store_true
+    persist: false
 include:
   - _database_mode
--- a/src/roles/check_duplicate_permissions/tasks/main.yaml
+++ b/src/roles/check_duplicate_permissions/tasks/main.yaml
@@ -1,0 +1,27 @@
+---
+# TODO: Remove this role one release after https://projects.theforeman.org/issues/38465 is addressed.
+- name: Query for duplicate permissions
+  community.postgresql.postgresql_query:
+    db: "{{ foreman_database_name }}"
+    login_user: "{{ foreman_database_user }}"
+    login_password: "{{ foreman_database_password }}"
+    login_host: "{{ foreman_database_host }}"
+    query: |
+      SELECT id, name
+      FROM permissions p
+      WHERE (SELECT count(name) FROM permissions pr WHERE p.name = pr.name) > 1
+      ORDER BY name, id
+  register: check_duplicate_permissions_result
+  changed_when: false
+
+- name: Check for duplicate permissions
+  when: not (health_fix | default(false) | bool)
+  ansible.builtin.assert:
+    that:
+      - check_duplicate_permissions_result.rowcount == 0
+    fail_msg: >-
+      Found {{ check_duplicate_permissions_result.rowcount }} duplicate permission(s) in database:
+      {{ check_duplicate_permissions_result.query_result | map(attribute='name') | unique | join(', ') }}
+
+      Duplicate permissions can cause issues with role-based access control and Foreman upgrades.
+      Please resolve the above duplicate permissions.
--- a/src/roles/check_foreman_tasks/tasks/main.yaml
+++ b/src/roles/check_foreman_tasks/tasks/main.yaml
@@ -1,0 +1,22 @@
+---
+- name: Query foreman tasks for errors
+  community.postgresql.postgresql_query:
+    db: "{{ foreman_database_name }}"
+    login_user: "{{ foreman_database_user }}"
+    login_password: "{{ foreman_database_password }}"
+    login_host: "{{ foreman_database_host }}"
+    query: |
+      SELECT count(*) AS count
+        FROM foreman_tasks_tasks
+        WHERE state IN ('paused') AND result IN ('error')
+  register: check_foreman_tasks_error_query_result
+  changed_when: false
+  when: enabled_features | has_feature('tasks')
+
+- name: Check for errored foreman tasks
+  ansible.builtin.assert:
+    that:
+      - check_foreman_tasks_error_query_result.query_result[0].count | int == 0
+    fail_msg: "{{ check_foreman_tasks_error_query_result.query_result[0].count }} foreman tasks with errors"
+    success_msg: "No foreman tasks with errors found"
+  when: enabled_features | has_feature('tasks')
--- a/src/roles/check_host_facts_count/tasks/main.yaml
+++ b/src/roles/check_host_facts_count/tasks/main.yaml
@@ -1,0 +1,27 @@
+---
+- name: Query hosts for facts count
+  community.postgresql.postgresql_query:
+    db: "{{ foreman_database_name }}"
+    login_user: "{{ foreman_database_user }}"
+    login_password: "{{ foreman_database_password }}"
+    login_host: "{{ foreman_database_host }}"
+    query: |
+      SELECT fact_values.host_id, count(fact_values.id) as count
+      FROM fact_values
+      GROUP BY fact_values.host_id
+      ORDER BY count DESC
+  register: check_host_facts_count_result
+  changed_when: false
+
+- name: Check facts count is below threshold
+  when: check_host_facts_count_result.rowcount > 0
+  vars:
+    hosts_over_limit: "{{ check_host_facts_count_result.query_result | selectattr('count', 'ge', check_host_facts_count_max_per_host) | list }}"
+  ansible.builtin.assert:
+    that:
+      - hosts_over_limit | length == 0
+    fail_msg: >-
+      {{ hosts_over_limit | length }} host(s) exceed the facts count limit of {{ check_host_facts_count_max_per_host }}:
+      {{ hosts_over_limit | map(attribute='host_id') | join(', ') }}
+
+      This can cause slow fact processing. See: https://access.redhat.com/solutions/4163891
--- a/src/roles/check_services/tasks/main.yaml
+++ b/src/roles/check_services/tasks/main.yaml
@@ -1,66 +1,44 @@
 ---
-- name: Gather service facts
-  ansible.builtin.service_facts:
+- name: Get all Foreman services as JSON
+  ansible.builtin.shell:
+    cmd: systemctl list-units --output=json --all $(systemctl list-dependencies --plain foreman.target)
+  register: check_services_all_json
+  changed_when: false
+  tags:
+    - skip_ansible_lint  # Skipping linting; systemctl JSON output not available via systemd module
 
-- name: Check core services are running
+- name: Filter out foreman-recurring services
+  ansible.builtin.set_fact:
+    check_services_all_list: >-
+      {{
+        (check_services_all_json.stdout | from_json) |
+        selectattr('unit', 'search', '\.service$') |
+        rejectattr('unit', 'search', 'foreman-recurring@') |
+        list
+      }}
+
+- name: Determine Foreman services which are inactive
+  ansible.builtin.set_fact:
+    check_services_not_running_list: >-
+      {{
+        check_services_all_list |
+        rejectattr('active', 'equalto', 'active') |
+        list
+      }}
+
+- name: Verify all Foreman services are active
   ansible.builtin.assert:
     that:
-      - "ansible_facts.services[item + '.service'] is defined"
-      - "ansible_facts.services[item + '.service']['state'] == 'running'"
-    fail_msg: "Service {{ item }} is not running"
-    success_msg: "Service {{ item }} is running"
-  loop:
-    - foreman
-    - httpd
-    - redis
+      - check_services_not_running_list | length == 0
+    fail_msg: |-
+      Some services are not running:
 
-- name: Check PostgreSQL service is running
-  ansible.builtin.assert:
-    that:
-      - "ansible_facts.services['postgresql.service'] is defined"
-      - "ansible_facts.services['postgresql.service']['state'] == 'running'"
-    fail_msg: "Service postgresql is not running"
-    success_msg: "Service postgresql is running"
-  when: database_mode | default('internal') == 'internal'
+      {% for service in check_services_not_running_list %}
+      - {{ service.unit.split('.')[0] }}: {{ service.active }} ({{ service.sub }})
+      {% endfor %}
+    success_msg: |-
+      All services are running:
 
-- name: Check dynflow services are running
-  ansible.builtin.assert:
-    that:
-      - "ansible_facts.services[item + '.service'] is defined"
-      - "ansible_facts.services[item + '.service']['state'] == 'running'"
-    fail_msg: "Service {{ item }} is not running"
-    success_msg: "Service {{ item }} is running"
-  loop:
-    - dynflow-sidekiq@orchestrator
-    - dynflow-sidekiq@worker
-    - dynflow-sidekiq@worker-hosts-queue
-
-- name: Check Pulp services are running
-  ansible.builtin.assert:
-    that:
-      - "ansible_facts.services[item + '.service'] is defined"
-      - "ansible_facts.services[item + '.service']['state'] == 'running'"
-    fail_msg: "Service {{ item }} is not running"
-    success_msg: "Service {{ item }} is running"
-  loop:
-    - pulp-api
-    - pulp-content
-  when: enabled_features | has_feature('content')
-
-- name: Check Candlepin service is running
-  ansible.builtin.assert:
-    that:
-      - "ansible_facts.services['candlepin.service'] is defined"
-      - "ansible_facts.services['candlepin.service']['state'] == 'running'"
-    fail_msg: "Service candlepin is not running"
-    success_msg: "Service candlepin is running"
-  when: enabled_features | has_feature('content')
-
-- name: Check Foreman Proxy service is running
-  ansible.builtin.assert:
-    that:
-      - "ansible_facts.services['foreman-proxy.service'] is defined"
-      - "ansible_facts.services['foreman-proxy.service']['state'] == 'running'"
-    fail_msg: "Service foreman-proxy is not running"
-    success_msg: "Service foreman-proxy is running"
-  when: enabled_features | has_feature('foreman-proxy')
+      {% for service in check_services_all_list %}
+      - {{ service.unit.split('.')[0] }}: {{ service.active }}
+      {% endfor %}
--- a/tests/health_test.py
+++ b/tests/health_test.py
@@ -1,0 +1,55 @@
+import subprocess
+
+import pytest
+
+
+@pytest.fixture
+def errored_foreman_task(database):
+    """Create an errored foreman task for testing, cleanup after test"""
+
+    database.run(
+        "podman exec postgresql psql -U foreman -d foreman -c "
+        "\"DELETE FROM foreman_tasks_tasks WHERE label = 'test_errored_task'\""
+    )
+    database.run(
+        "podman exec postgresql psql -U foreman -d foreman -c "
+        "\"INSERT INTO foreman_tasks_tasks (id, type, label, state, result, started_at, ended_at, state_updated_at) "
+        "VALUES (gen_random_uuid(), 'Actions::Test::Task', 'test_errored_task', 'paused', 'error', NOW(), NOW(), NOW())\""
+    )
+
+    yield
+
+    # cleanup
+    database.run(
+        "podman exec postgresql psql -U foreman -d foreman -c "
+        "\"DELETE FROM foreman_tasks_tasks WHERE label = 'test_errored_task'\""
+    )
+
+
+def test_health_checks_pass():
+    """Verify health checks pass on a working deployment"""
+
+    subprocess.check_call(
+        ['./foremanctl', 'health', '--skip-check-foreman-tasks']
+    )
+
+
+def test_foreman_tasks_check_detects_errors(errored_foreman_task):
+    """Verify foreman tasks check detects errored tasks and fails"""
+
+    result = subprocess.run(
+        ['./foremanctl', 'health'],
+        capture_output=True,
+        text=True
+    )
+    output = result.stdout + result.stderr
+    assert result.returncode != 0
+    assert 'foreman task' in output.lower() and 'error' in output.lower()
+
+
+def test_foreman_tasks_skipped(errored_foreman_task):
+    """Verify foreman tasks check is skipped when passed the `--skip-check-foreman-tasks` param"""
+
+    subprocess.check_call(
+        ['./foremanctl', 'health', '--skip-check-foreman-tasks']
+    )
--- a/src/filter_plugins/foremanctl.py
+++ b/src/filter_plugins/foremanctl.py
@@ -111,8 +111,10 @@ def available_foreman_proxy_plugins(_value):
 
 
 def has_feature(features, feature):
-    """Check if a feature is enabled - exact match or prefix (feature/)."""
-    return feature in features or any(f.startswith(feature + '/') for f in features)
+    """Check if a feature is enabled - exact match, prefix (feature/), or as a transitive dependency."""
+    return (feature in features
+            or any(f.startswith(feature + '/') for f in features)
+            or feature in get_dependencies(list(features)))
 
 
 class FilterModule(object):
--- a/src/roles/check_foreman_api/tasks/main.yaml
+++ b/src/roles/check_foreman_api/tasks/main.yaml
@@ -14,5 +14,6 @@
     fail_msg: "Foreman tasks status: {{ check_foreman_api_ping.json.results.katello.services.foreman_tasks.status | default('unknown') }}"
     success_msg: "Foreman tasks status: ok"
   when:
-    - enabled_features | has_feature('content')
+    - enabled_features | has_feature('tasks')
+    - enabled_features | has_feature('katello')
     - check_foreman_api_ping.json.results.katello is defined
--- a/src/roles/check_host_facts_count/defaults/main.yml
+++ b/src/roles/check_host_facts_count/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+check_host_facts_count_max_per_host: 10000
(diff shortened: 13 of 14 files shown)


qcjames53:
None of these test failures seem related. FWIW I cannot get the full unit test suite to run locally without issue either. My unit test passes, though :)

ehelms:
@qcjames53 Before I review in-depth, can you review https://github.com/theforeman/foremanctl/pull/431? This is a previous contribution that has made overlapping progress and addresses some of what would be questions I have for this PR. Ideally, we would build on #431

qcjames53:
Of course, Eric. #431 wasn't on our team's radar at all so I'll need some time to rectify. I'll reach out to make sure we're not duplicating more effort than this.

qcjames53:
This PR should now be up to par. I rebased on master and added a few additional checks to the existing `foremanctl health` subcommand (as well as a unit test). Let me know if you think we need negative-case testing for this.

qcjames53:
The newest push adds a flag to ignore errored out foreman tasks (for automation mainly). I've also removed health checks from the GH automation since the unit test checks it instead.

qcjames53:
The latest force push handles db auth correctly for external dbs.

src/playbooks/health/metadata.obsah.yaml
```diff
@@ -1,5 +1,10 @@
 ---
 help: |
   Run health checks on Foreman services
+variables:
+  ignore_foreman_tasks_check:
+    help: Ignore errored Foreman tasks
+    parameter: --ignore-foreman-tasks-check
```
ehelms: Is this running the check and ignoring the results?
Is this running the check and reporting if there are errors but not throwing an error?
Is this skipping the check all together?

src/vars/health.yml
```diff
@@ -0,0 +1,22 @@
+---
+# Health check thresholds and configuration
+
+# Maximum number of fact values per host before triggering a warning
+check_host_facts_count_max_per_host: 10000
```
ehelms: I'd move this to `defaults` in the role.

src/roles/check_services/tasks/main.yaml
```diff
@@ -2,65 +2,57 @@
...
-    that:
-      - "ansible_facts.services[item + '.service'] is defined"
-      - "ansible_facts.services[item + '.service']['state'] == 'running'"
-    fail_msg: "Service {{ item }} is not running"
-    success_msg: "Service {{ item }} is running"
-  loop:
-    - foreman
-    - httpd
-    - redis
+- name: Build list of services from role definitions
```
ehelms: This is where the `foreman.target` can come into play, as each service is a part of the target and thus you can list and control all services by invoking the target the same way you would any other process.

src/roles/check_foreman_tasks/tasks/main.yaml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Query foreman tasks for errors
+  community.postgresql.postgresql_query:
+    db: "{{ foreman_database_name }}"
+    login_user: "{{ foreman_database_user }}"
+    login_password: "{{ foreman_database_password }}"
+    login_host: "{{ foreman_database_host }}"
+    query: |
```
ehelms, line 8: I have questioned this for a long time, but foreman-maintain was so ingrained I just let it go. As we move into a new tool and thinking, I am going to raise the question again. Should we be doing direct database reads that work around our API and ORM and thus become tied to the database schema or should we be relying on our APIs / Ansible module collection to fetch data when needed?

@evgeni your thoughts on this would be welcome

qcjames53:
I just pushed some changes to add `--fix` for duplicate permissions to help out [Aneta's docs PR](https://github.com/theforeman/foreman-documentation/pull/4843). We'd have to drop the sections on fixing duplicate permissions otherwise and it was an easy add.

New unit tests added to cover this and `--ignore-foreman-tasks-check`.

src/playbooks/health/metadata.obsah.yaml
qcjames53: This flag runs the check for errored Foreman tasks but just prints the results instead of failing the playbook. I added this so I could include a full `foremanctl health` unit test in CI without all the failing foreman tasks causing issues. I could delete all the foreman tasks before running the unit test but that felt messier to me than silencing failures for this one check.

src/vars/health.yml
qcjames53: Done! If I can get away with removing the other vars with foreman.target, I'll delete this file entirely.

src/roles/check_services/tasks/main.yaml
qcjames53: Thanks for pointing this out, Eric. Fixed!

qcjames53:
Newest push should take care of linting.

src/playbooks/health/metadata.obsah.yaml
ehelms: Did you add this option because of CI only or because you expect users to use this option? If the latter, I do not think that's a good reason since users will find and use this option. I think we should ask ourselves and investigate -- why do we have so many tasks in the error state in CI?

src/roles/check_duplicate_permissions/tasks/main.yaml
```diff
@@ -0,0 +1,43 @@
+---
+- name: Query for duplicate permissions
```
ehelms, line 3: @qcjames53 do you know why this check needs to exist? I'm sure it's been around for a long time, but tihs is a good opportunity to go spelunking into our history and understand if this is something we fixed or should be fixed fundamentally at the application layer.

src/playbooks/health/metadata.obsah.yaml
qcjames53: I expect users would find the option useful since task failures aren't a real environment issue; they're more of a symptom. These task failures show up in CI due to other unit tests asserting task failure. They *could* be cleaned up as a part of the fixture cleanup for each failing test (5 as of now) but that seems like a burden on future test authors.

Here are a few options I can think of to move forward:

- Keep the `--ignore-foreman-tasks-check` as-is and add other options to ignore future tests where warranted.
- Add a method for ignoring / filtering individual checks. We explicitly excluded that from this issue's acceptance criteria but it would be "cleaner" than the above option.
- Modify `--fix` to delete failed tasks and run this unit test with `--fix` enabled. Add additional unit tests to handle failure without `--fix`.

qcjames53: What do you think, Eric?

src/roles/check_duplicate_permissions/tasks/main.yaml
qcjames53: I will look into it.

src/playbooks/health/metadata.obsah.yaml
sjha4: Is this kind of the --whitelist option in foreman-maintain: 
` -w, --whitelist whitelist    Comma-separated list of labels of steps to be skipped`
If so, can we make it more generic than skipping foreman tasks alone?

qcjames53: Yeah that would be option 2 above. Happy to add it (and think it probably should be included).

ehelms: If we are going to give the user the option to skip a particular check, I'd like to embrace `skip` as the language based on past foreman-maintain design discussions and to be targeted. So in this instance it would be `--skip-check-foreman-tasks`

src/roles/check_services/tasks/main.yaml
```diff
@@ -2,65 +2,51 @@
...
-      - "ansible_facts.services[item + '.service']['state'] == 'running'"
-    fail_msg: "Service {{ item }} is not running"
-    success_msg: "Service {{ item }} is running"
-  loop:
-    - foreman
-    - httpd
-    - redis
+- name: Get services from foreman.target
+  ansible.builtin.command:
+    cmd: systemctl list-dependencies foreman.target --plain --no-pager
```
ehelms: systemd has a lot of power and commands you can use and combine together along with getting machine readable output. For example, the following will output only failing services as JSON:

```
systemctl list-units --output=json --state=failed $(systemctl list-dependencies --plain foreman.target)
```

src/roles/check_duplicate_permissions/tasks/main.yaml
qcjames53: Here's what I've found:
- We added this check in 2020 as a part of [30527](https://projects.theforeman.org/issues/30527), fixing a duplicate permissions issue in the Sat 6.5 -> 6.6 upgrade (do not know exact Foreman versions since NOTHING mentions it; I think it's 1.21 -> 1.22?).
- Modern Foreman does not allow duplicate permissions to be created and we're well beyond supporting 1.21.

I'll drop this check from the PR and document.

src/playbooks/health/metadata.obsah.yaml
qcjames53: I've implemented a generic `--skip` param which accepts a comma seperated list, allowing the user to skip any tests they'd like. Imho this is the cleanest solution.

src/roles/check_duplicate_permissions/tasks/main.yaml
ehelms: Great finds, thanks @qcjames53

src/roles/check_services/tasks/main.yaml
qcjames53: I've implemented this suggestion; it's very snazzy and quite a bit quicker. Thanks, Eric.

ehelms: I should add, I found this by doing some research with AI. I asked it "How can I listed failed services that are dependencies of a systemd target in a machine readable format?"

qcjames53:
This most recent push should address all feedback mentioned above. I've removed the `--ignore-*` param and replaced with a `--skip` param which takes a comma seperated list of roles to skip. I've removed the duplicate permissions check and cleanup (reasoning above) and I've used your suggested systemd shell piping with json output to skip a lot of the fragile steps from before. The PR is looking good :)

ehelms:
> This most recent push should address all feedback mentioned above. I've removed the `--ignore-*` param and replaced with a `--skip` param which takes a comma seperated list of roles to skip. I've removed the duplicate permissions check and cleanup (reasoning above) and I've used your suggested systemd shell piping with json output to skip a lot of the fragile steps from before. The PR is looking good :)

As I previously tried to state, we had identified we do not want to keep this "give the user the ability to skip whatever check they want" ability. This is what foreman-maintain had, and inevitably led to having to go and create a special "tag" for unskippable actions. We want to be very prescriptive, and thoughtful, and document clearly anything we allow a user to skip with why they may need to skip it and what the consequences might be. To quote our previous foreman-maintain design changes we had planned out:

> The –whitelist option has been used without limitations to allow users to skip steps that fail, sometimes for good and sometimes with disastrous results for the upgrade, backup/store processes. There are times where whitelist options are presented that also cannot be whitelisted confusing users.
> 
> 
> Proposal
> 
> Remove –whitelist from foreman-maintain completely. The –whitelist option will be selectively replaced by options to skip known and documented actions that need to be skipped.

To that end, if we are going to allow users to skip a check it should include:

- a dedicated parameter following a naming convention, I'd suggest `--skip-<type>-<name>` (e.g. --skip-check-foreman-tasks)
- documentation of the parameter, what it does, why one might use it and the potential consequences

There is no documentation yet on this, as you are the first to be adding one. Congrats :grin:  I'd suggest start a new `doc/` that focuses on documenting checks, and any skip parameters you add. You don't have to add documentation for any that were previously added to the repository if you feel that is scope creep. Up to you. If you choose not to document all existing, please open an issue to track adding the missing parts.

qcjames53:
Ah I see what you meant and understand the rationale. I will add in a sec + document.

qcjames53:
How does the most recent push look? I think you'll be happy :)

src/playbooks/health/metadata.obsah.yaml
```diff
@@ -1,5 +1,13 @@
 ---
 help: |
-  Run health checks on Foreman services
+  Check the health of your running Foreman server
+variables:
+  health_skip_check_foreman_tasks_param:
+    help: |
+      Skip checking Foreman tasks for paused/errored tasks.
+      Errored tasks can be manually resumed or cancelled using the Web UI or hammer.
+    parameter: --skip-check-foreman-tasks
```
ehelms, line 10: Can you add a documentation section on skipping checks that includes this initial skip parameter, as well as the guidelines and naming rules that I mentioned in my previous comment?

docs/developer/checks.md
```diff
@@ -0,0 +1,65 @@
...
+check_subuid_subgid
+- Verifies /etc/subuid and /etc/subgid have entries for the current user (required for rootless Podman).
+- Fails if these conditions are not met.
+
+check_system_requirements
+- Loads tuning profile thresholds.
+- Validates system has minimum CPU cores and RAM.
+- Fails if these conditions are not met.
+
+## Check Usage
```
ehelms: I'm hesitant to include this section as this can be gleaned from the code, and is another place we have to keep up to date with the list.

docs/developer/checks.md
```diff
@@ -0,0 +1,65 @@
+# Checks
+Foremanctl contains several "check_*" roles which are used to assist playbooks in confirming the Foreman server is configured properly. Checks are implemented as independent Ansible roles, allowing any playbook to run specific checks or groups of checks. 
+
+Please update this file as check usage evolves.
+
+## Check Descriptions
```
ehelms, line 6: I like this section. A few ideas:

- Make each check it's own section (e.g. ### check_database_connection)
- Provide a description of the check, high-level description of what it does, and **why** the check exists.

docs/developer/checks.md
qcjames53: Agreed.

docs/developer/checks.md
qcjames53: Done!

src/roles/httpd/tasks/main.yml
```diff
@@ -92,6 +92,28 @@
   notify:
     - Restart httpd
 
+- name: Create systemd drop-in directory for httpd
```
ehelms, line 95: This should be it's own commit or PR as it's an independent change.

src/playbooks/health/metadata.obsah.yaml
qcjames53: Done!

docs/developer/checks.md
```diff
@@ -0,0 +1,70 @@
+# Checks
+Foremanctl contains several "check_*" roles which are used to assist playbooks in confirming the Foreman server is configured properly. Each check is implemented as an independent Ansible role, allowing playbooks to run specific checks or groups of checks as needed.
+
+Please update this file as check usage evolves.
+
+## Check Descriptions
+#### check_database_connection
```
ehelms: I'd like to further adjust the structure of each of these check sections. To recap, the elements I mentioned, which are technically captured are listed. What's hard to know is which is which, for a reader or a future developer.

- Make each check it's own section (e.g. ### check_database_connection)
- Provide a description of the check, high-level description of what it does, and why the check exists.

What I did was take those points and then prompt Claude with the following where my goal is to make sure it's obvious to a reader and to any future contributor adding a check:

> Review the structure that you landed on and see if it's obvious to identify each of the major areas that should be documented. Consider this from a reader of the documentation
  perspective. And consider this from the perspective of a future developer adding an additional check.

For this item, this is what it arrived at:

```
### check_database_connection
+
+Validates PostgreSQL connectivity for the Foreman, Candlepin, and Pulp databases using configured credentials and SSL settings.
+
+#### How it works
+
+This check runs only when `database_mode` is set to `external`. For each database, it attempts a connection using `community.postgresql.postgresql_ping` with the configured host, user, password, database name, and optional SSL CA certificate. Temporary files are created for CA certificates during the check and cleaned up afterward.
+
+#### Why it exists
+
+Direct database access is required for many foremanctl operations. This check catches misconfigured credentials, unreachable hosts, and SSL issues before they surface as cryptic failures later in a playbook run.
```

I have a whole local diff I can provide, or, if you want to give it a try yourself to go iterate on this idea and see what you come up with.

src/roles/httpd/tasks/main.yml
qcjames53: I've moved this to https://github.com/theforeman/foremanctl/pull/536. Will delete changes in this PR shortly.

docs/developer/checks.md
```diff
@@ -0,0 +1,70 @@
...
+- Pings Foreman API (/api/v2/ping) to verify it responds.
+- Verifies foreman_tasks service status is 'ok' when Katello/content is enabled.
+- Fails if ping fails or service status is not 'ok'.
+- An API connection is necessary for most Foreman operations. This check ensures there are no firewall or address issues.
+
+#### check_foreman_tasks
+- Queries Foreman tasks for paused, errored tasks.
+- Can be skipped in `foremanctl health` by passing the `--skip-check-foreman-tasks` flag.
+- Fails if any errored tasks are found.
+- Errored Foreman tasks indicate issues which need to be addressed by the user. This can be anything from a job being misconfigured up to dependency versioning issues.
```
ehelms: This is true, but I don't feel it captures why we need a check in our deployment tool. IIRC, there were cases in the past where tasks in an error state and break upgrades? @adamruzicka do you happen to recall?

docs/developer/checks.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### check_host_facts_count
+- Finds the host with the most fact values in the database.
+- Fails if facts count exceeds threshold.
+- Very high host facts counts causes slow facts processing. See: https://access.redhat.com/solutions/4163891 for more information.
+
+#### check_hostname
+- Validates FQDN is not 'localhost' or 'localhost.localdomain', contains at least one dot, and has no underscores.
+- Fails if FQDN conditions are not met.
+- The hostname remaining on 'localhost' is a common issue with foreman build failures. This is a sanity check.
```
ehelms: This should talk to the fact that Foreman and the installation require a FQDN to properly operate. @ekohl does a better job at articulating why a "real" FQDN is needed.

docs/developer/checks.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### check_services
+- Finds all units in the foreman.target systemd dependency tree, filtering foreman-recurring services.
+- Fails if any services are inactive.
+- Failed or stalled foreman services indicate an issue with the server. The user should check Foreman logs to find and address the problem. Inactive foreman services may indicate a systemctl configuration problem.
+
+#### check_subuid_subgid
+- Verifies /etc/subuid and /etc/subgid have entries for the current user (required for rootless Podman).
```
ehelms: As we are not currently running rootless podman, why is this check needed?

docs/developer/checks.md
```diff
@@ -0,0 +1,70 @@
...
+- Fails if these conditions are not met.
+- Podman requires root access without the above entries, a configuration typically avoided for security reasons.
+
+#### check_system_requirements
+- Loads tuning profile thresholds.
+- Validates system has minimum CPU cores and RAM.
+- Fails if these conditions are not met.
+- Foreman uses a lot of system resources; we attempt to proactively enforce users meeting what we consider minimum hardware specs.
```
ehelms: That's true but I would argue the point is that tuning profiles are intended for a certain set of minimum requirements, and so when a user chooses a tuning profile their system resources need to match or else they might end up tuned incorrectly and have worse performance. This check is intended to ensure those things match up.

docs/developer/checks.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Skipping checks
+
+Users may wish to skip certain checks on some playbooks, usually to work around a known issue on particularly sensitive checks. For example `health.yaml` (`foremanctl health`) includes a pattern for skipping `check_foreman_tasks`, as errored Foreman tasks which have not yet been cleared will fail the playbook.
+
+When allowing the user to skip a check, ensure it provides tangible value greater than the consequences of user misuse. For example, allowing the user to skip `check_system_requirements` might seem harmless, but could likely lead to situations in which usage of the parameter becomes a silent default; changes in Foreman minimum system requirements would not be enforced.
+
+Add skip parameters following this template:
+```
+<playbook_name>_skip_<check_role_name>_param:
```
ehelms, line 66: Food for thought. The framework allows defining parameters that are "universal" and can be included in multiple playbooks. For example, database parameters: https://github.com/theforeman/foremanctl/blob/master/src/playbooks/_database_connection/metadata.obsah.yaml

I would expect checks to be used in multiple playbooks/commands and thus the "skip" for a given one to be also re-used. I can also potentially see argument where a skip might be allowed in one command but not another. Perhaps we do not know enough yet.

docs/developer/checks.md
qcjames53: After the most recent Pulp upgrade we had a ton of erroring foreman tasks which oddly enough were the only indications inside Foreman that Postgresql needed bumped. So the failures really can be anything to be honest. Do you have a suggestion for something more specific than "look into the error and address it"?

qcjames53: Or maybe it was pulp-python? I can't remember exactly. I was installing everything from pip so it's not like that's a scenario the user is likely to ever encounter :)

docs/developer/checks.md
qcjames53: It's dead code at the moment. I didn't want to remove it without a reason and felt all checks should be documented here regardless of use.

docs/developer/checks.md
qcjames53: Updated to add tuning. What do you think about the new line?

docs/developer/checks.md
qcjames53: I've updated this with a better description. Do you have any thoughts?

docs/developer/checks.md
qcjames53: I changed this line to "Errored Foreman tasks indicate issues which need to be addressed by the user. This can be anything from a typo to systemic issues. Consulting https://community.theforeman.org may provide insight.". How's that sound?

ehelms: Sounds good, go with that.

docs/developer/checks.md
qcjames53: Sorry to keep bikeshedding on this markdown doc but I disagree with this suggestion. Ansible roles are basically self-documenting and everything in the "How it works" section is clearly apparent by just going to the role itself. My intuition says these adds don't help AI and make it too cumbersome for a person to read. Also think of the maintainability with this approach, yikes!

qcjames53: Like as a person I can read this and get the gist in 15 seconds:
```
#### check_database_connection
- Validates PostgreSQL connectivity for Foreman and Candlepin databases using configured credentials and SSL settings.
- Fails if database connection cannot be established.
- Direct database access is required for many foremanctl operations and must be configured.
```

Claude's output is way too verbose and in the weeds imho.

docs/developer/checks.md
qcjames53: I think we look into that when we start sharing skippable checks between foremanctl subcommand playbooks. Right now I'm happy with the current implementation.

docs/developer/checks.md
ehelms: I see, this comes from https://github.com/theforeman/foremanctl/pull/209 so I think the issue might be with "why this check exists" in this documentation rather than the fact it exists. The PR doesn't give us insight, but I am guessing it's to help with running "non-root user inside container" vs. rootless containers for our current setup. Let's indicate both in the explanation of why it exists.

docs/developer/checks.md
ehelms: By that argument, then this should be reduced to (since the how and failure mode is captured in the role itself):

```
#### check_database_connection
- Direct database access is required for many foremanctl operations and must be configured.
```

You could then argue that every check role should have "frontmatter" (aka a comment) explaining why the check exists instead of a single document.

I can tell you with certainty, that these checks have been added to our code bases in the past and all sense of why they exist lost. And then they persist forever, because nobody removes them because nobody understand why they were put there in the first place. So I care greatly now, and in the future, that we clearly document what the point of a given check is, what it solves, and why it was implemented in the first place.

I prefer we lean into being pedantic so it is very clear to any reader (human or otherwise) what something is representing. If I am a future developer or agent, adding a check, it is not clear in the current format that I need to document "why".

qcjames53: Thanks Eric, I understand the rationale a lot better and completely agree with preserving the reasoning for adding a check. I'm still not a fan of the subsections and whitespace (for human readability purposes, not sure makes a difference for bots) and I'm not a fan of the "How it works" section since I don't think it adds much beyond debt to maintain it.

Would you be okay with something like this?
```
#### check_name
- **Description** - Basic one or two sentence description that won't change much over time.
- **Fail State** - What causes the check to fail.
- **Why is this check needed** - Why we added this check. Can it be removed in the future?
```

ehelms: Let's give that a try.

docs/developer/checks.md
qcjames53: Done!

docs/developer/checks.md
qcjames53: All right I reformatted and reworded a bit under this new schema. Let me know how it looks!

qcjames53:
As far as `check_system_requirements` goes, I'm not a fan of a `check` role modifying the Foreman server. I think it would be prudent to make a github issue to move the tuning config to another role, and maybe just have tuning profile verification in the `check_system_requirements` role. Not in scope for this PR.

ehelms:
> As far as `check_system_requirements` goes, I'm not a fan of a `check` role modifying the Foreman server. I think it would be prudent to make a github issue to move the tuning config to another role, and maybe just have tuning profile verification in the `check_system_requirements` role. Not in scope for this PR.

I'm not either, where do you see that role modifying the server?

qcjames53:
Ah it's not modifying anything. I was mistaken. Confused loading foremanctl tuning vs modifying the OS's tuning.

docs/developer/checks.md
ehelms: Looks good.

qcjames53:
Pushed to re-run automation.

src/roles/check_foreman_tasks/tasks/main.yaml
ekohl: I'd really like to avoid that. IMHO every application should be observable from the outside and the database is an internal detail. That way users can also write up additional logic to regularly poll an endpoint. Having graphs can be extremely useful to observe trends.

Looking at https://apidocs.theforeman.org/katello/latest/apidoc/v2/foreman_tasks/index.html I'm wondering if grabbing `total` from `GET /foreman_tasks/api/tasks?search=state+%3D+paused+and+result+%3D+error&per_page=0` would be correct.

Can we leverage `apypie` in modules here?

src/roles/check_host_facts_count/tasks/main.yaml
```diff
@@ -0,0 +1,32 @@
...
+      Host {{ check_host_facts_count_result.query_result[0].host_id }} exceeds facts count limit:
+      {{ check_host_facts_count_result.query_result[0].count }} / {{ check_host_facts_count_max_per_host }}
+
+      This can cause slow fact processing. See: https://access.redhat.com/solutions/4163891
+    success_msg: "Host fact count check passed: {{ check_host_facts_count_result.query_result[0].count }} / {{ check_host_facts_count_max_per_host }}"
+
+- name: No hosts with facts found
+  when: check_host_facts_count_result.rowcount == 0
+  ansible.builtin.debug:
+    msg: "No hosts with fact values found in the database"
```
ekohl: On a fresh installation I don't think we register any host so I'd expect the facts table to be empty. Is this information helpful?

tests/health_test.py
```diff
@@ -0,0 +1,61 @@
...
+    # cleanup
+    database.run(
+        "podman exec postgresql psql -U foreman -d foreman -c "
+        "\"DELETE FROM foreman_tasks_tasks WHERE label = 'test_errored_task'\""
+    )
+
+
+def test_health_checks_pass():
+    """Verify health checks pass on a working deployment"""
+    result = subprocess.run(
```
ekohl: If you use `subprocess.check_call`, isn't the exception raised already sufficient? A simple reproducer:

```python
import subprocess


def test_call():
    subprocess.check_call('echo call 1>&2 && exit 1', shell=True)


def test_run():
    result = subprocess.run('echo run 1>&2 && exit 1', capture_output=True, text=True, shell=True)
    assert result.returncode == 0, f"Health checks failed:\n{result.stdout}\n{result.stderr}"
```

Then looking at the output:

```console
$ pytest test_subprocess.py
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.14.3, pytest-8.3.5, pluggy-1.6.0
rootdir: /home/ekohl
plugins: anyio-4.12.1
collected 2 items                                                                                                                                                                            

test_subprocess.py FF                                                                                                                                                                  [100%]

========================================================================================== FAILURES ==========================================================================================
_________________________________________________________________________________________ test_call __________________________________________________________________________________________

    def test_call():
>       subprocess.check_call('echo call 1>&2 && exit 1', shell=True)

test_subprocess.py:5: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

popenargs = ('echo call 1>&2 && exit 1',), kwargs = {'shell': True}, retcode = 1, cmd = 'echo call 1>&2 && exit 1'

    def check_call(*popenargs, **kwargs):
        """Run command with arguments.  Wait for command to complete.  If
        the exit code was zero then return, otherwise raise
        CalledProcessError.  The CalledProcessError object will have the
        return code in the returncode attribute.
    
        The arguments are the same as for the call function.  Example:
    
        check_call(["ls", "-l"])
        """
        retcode = call(*popenargs, **kwargs)
        if retcode:
            cmd = kwargs.get("args")
            if cmd is None:
                cmd = popenargs[0]
>           raise CalledProcessError(retcode, cmd)
E           subprocess.CalledProcessError: Command 'echo call 1>&2 && exit 1' returned non-zero exit status 1.

/usr/lib64/python3.14/subprocess.py:419: CalledProcessError
------------------------------------------------------------------------------------ Captured stderr call ------------------------------------------------------------------------------------
call
__________________________________________________________________________________________ test_run __________________________________________________________________________________________

    def test_run():
        result = subprocess.run('echo run 1>&2 && exit 1', capture_output=True, text=True, shell=True)
>       assert result.returncode == 0, f"Health checks failed:\n{result.stdout}\n{result.stderr}"
E       AssertionError: Health checks failed:
E         
E         run
E         
E       assert 1 == 0
E        +  where 1 = CompletedProcess(args='echo run 1>&2 && exit 1', returncode=1, stdout='', stderr='run\n').returncode

test_subprocess.py:10: AssertionError
================================================================================== short test summary info ===================================================================================
FAILED test_subprocess.py::test_call - subprocess.CalledProcessError: Command 'echo call 1>&2 && exit 1' returned non-zero exit status 1.
FAILED test_subprocess.py::test_run - AssertionError: Health checks failed:
===================================================================================== 2 failed in 0.05s ======================================================================================
```

So when I look at the output of the commands then I think it's highly subjective what is more readable, but including the `stdout` and `stderr` in the label doesn't appear to help.

docs/developer/checks.md
```diff
@@ -0,0 +1,68 @@
+# Checks
+Foremanctl contains several "check_*" roles which are used to assist playbooks in confirming the Foreman server is configured properly. Each check is implemented as an independent Ansible role, allowing playbooks to run specific checks or groups of checks as needed.
+
+Please update this file as check usage evolves.
```
ekohl, line 4: This is very likely to be forgotten. https://github.com/theforeman/foremanctl/blob/master/AGENTS.md already has mentions of checks and I fear that people will forget and they diverge.

src/playbooks/health/metadata.obsah.yaml
ekohl: Complicating matters: in upstream foreman-tasks is not guaranteed to be present. I'd like to avoid adding parameters to disable every check. Do we really need to do this?

.github/workflows/test.yml
```diff
@@ -160,9 +160,6 @@ jobs:
             --add-feature remote-execution \
             --add-feature bmc \
             ${{ matrix.iop == 'enabled' && '--add-feature iop' || '' }}
-      - name: Run health check
-        run: |
-          ./foremanctl health
```
ekohl, line 166: At first I wondered about this, but I like having the test suite really covering everything.

src/roles/check_host_facts_count/tasks/main.yaml
qcjames53: I don't think it's helpful in a zero host environment. I could make this output when host count is 1 or more but I wouldn't have the host count without a second db query. I'll drop the debug message task.

tests/health_test.py
qcjames53: The fstring on the assertion is a stupid Claude addition. I had to delete like 10 of those and just missed this one :/ 

I've switched this + the other positive unit test over to `.check_call`.

docs/developer/checks.md
```diff
@@ -0,0 +1,68 @@
...
+### check_foreman_api
+- **Description**: Pings Foreman API (/api/v2/ping) to verify it responds. Verifies foreman_tasks service status is 'ok' when Katello/content feature is enabled.
+- **Fail state**: Fails if ping fails or service status is not 'ok'.
+- **Rationale**: An API connection is necessary for most Foreman operations. This check ensures there are no firewall or address issues.
+
+### check_foreman_tasks
+- **Description**: Queries Foreman tasks for paused, errored tasks.
+- **Fail state**: Fails if any errored tasks are found.
+- **Rationale**: Errored Foreman tasks indicate issues which need to be addressed by the user. This can be anything from a typo to systemic issues. Consulting https://community.theforeman.org may provide insight.
+- **Skipping**: This role can be skipped in `foremanctl health` by passing the `--skip-check-foreman-tasks` flag.
```
ehelms: Thinking about https://github.com/theforeman/foremanctl/pull/521/changes#r3357265695 and other conversations in this PR, I think what's important for us to do is document to the best of our ability *why*. If we include a skip, we should, with what knowledge we have, indicate why a skip exists and any warnings when using it (this latter part could arguably be part of the parameter description so that users **have been warned**). The former, the *why* is to help future engineers working on the project understand the existence of the flag.

docs/developer/checks.md
qcjames53: I've updated AGENTS.md to include all markdown docs and added links to the checks.md file. Hopefully agents can pick up that this document needs to be updated when new checks are added (or in review).

src/roles/check_foreman_tasks/tasks/main.yaml
qcjames53: Authentication is outside of my area of expertise so I'm hoping someone else will know the answer.

I have a working setup using API but we would need to know the Foreman admin username and password. That can be passed in as params or stored as a secret but none of that is in foremanctl yet. Foremanctl DOES interact directly with the database in several places and has credentials already set up to log in.

I see these options:
1. Add API auth "properly" and have someone smarter than I am design a secure template. Block this PR until that is ready.
2. Use direct database access.
3. Use direct database access for now and switch to the API endpoints once auth is ready.

.github/workflows/test.yml
qcjames53: @ekohl Eric had some thoughts on this in [a comment above](https://github.com/theforeman/foremanctl/pull/521#issuecomment-4604951577) and we landed on the `--skip-check-*` format. I personally think the errored foreman tasks check is valuable for users and also would like `foremanctl health` to run as a part of automation. This option seems to be the best result of several conflicting ideas.

qcjames53: Oh the `tests/health_test.py` file covers a full run now. I just moved it.

docs/developer/checks.md
qcjames53: I've addressed this to the best of my ability in the newest push:

"This role can be skipped in `foremanctl health` by passing the `--skip-check-foreman-tasks` flag. Use only for operations where a failed Foreman task is expected or unavoidable."

"Indicate why the user would need this check and warn against overuse."

"Skip checking Foreman tasks for paused/errored tasks. Errored tasks can be manually resumed or cancelled using the Web UI or hammer. Use only for operations where a failed Foreman task is expected or unavoidable."

src/roles/check_foreman_tasks/tasks/main.yaml
ehelms: That's fair. I lean towards stick with what works, and comes from foreman-maintain with regards to this queries and file an issue to continue discussion around this topic. There will be many more to follow up on this and I don't think we can afford to block on this when we will for sure have some that come up that do not have a clear API available.

src/roles/check_host_facts_count/tasks/main.yaml
```diff
@@ -0,0 +1,27 @@
...
+    db: "{{ foreman_database_name }}"
+    login_user: "{{ foreman_database_user }}"
+    login_password: "{{ foreman_database_password }}"
+    login_host: "{{ foreman_database_host }}"
+    query: |
+      SELECT fact_values.host_id, count(fact_values.id) as count
+      FROM fact_values
+      GROUP BY fact_values.host_id
+      ORDER BY count DESC
+      LIMIT 1
```
ehelms: If I understand this correctly, this limits the results to one host. And then presents that single host to the user. If a user happens to have 25 hosts in violation, they will think it's only one host, fix that host, run this again and then get another host. Then, in theory, repeat this 23 more times until this check reports 0. That sounds rather frustrating for the end user.

src/roles/check_foreman_tasks/tasks/main.yaml
ekohl: > I have a working setup using API but we would need to know the Foreman admin username and password. That can be passed in as params or stored as a secret but none of that is in foremanctl yet.

You're right, when creating the Smart Proxy we use oauth because we can guarantee those credentials:
https://github.com/theforeman/foremanctl/blob/8044026c88c4ff359191c92df62f68c9fbf321d8/src/roles/foreman/tasks/main.yaml#L320-L323

My reasoning is that these checks probably also make sense outside of a foremanctl deployment and a hypothetical deployment in another tool (Kubernetes / OpenShift) would also benefit from this. It may be premature for that, but it's what I always keep in mind if I ask myself "does this belong in foremanctl"? So for now we can merge this, but later refactor it.

src/roles/check_duplicate_permissions/tasks/main.yaml
ekohl: > Modern Foreman does not allow duplicate permissions to be created and we're well beyond supporting 1.21.

"fun" fact: a regression was introduced. See https://projects.theforeman.org/issues/38465 for the gory details.

src/playbooks/health/metadata.obsah.yaml
ekohl: For more context: in https://github.com/theforeman/foremanctl/pull/511 there's a flavor that doesn't have foreman-tasks.

src/roles/check_foreman_tasks/tasks/main.yaml
ehelms: Agreed. @qcjames53 can you file a github issue based on the discussion? Then link it here and resolve the conversation.

qcjames53:
@ehelms @ekohl I just pushed changes that __should__ address both of your concerns. I've feature gated `check_foreman_tasks` and `check_foreman_api` properly and added unit testing. I've also changed `check_host_facts` to return the IDs of all hosts that exceed the threshold. I've updated documentation where neccessary.

src/roles/check_host_facts_count/tasks/main.yaml
qcjames53: I've addressed this in the newest push by having `check_host_facts` show IDs for all hosts exceeding the threshold.

src/playbooks/health/metadata.obsah.yaml
qcjames53: I have addressed this in the newest push. The `tasks` feature is required for certain tasks to run and is covered by unit testing.

AGENTS.md
```diff
@@ -170,15 +166,23 @@ Tests use pytest with testinfra for infrastructure testing. Key fixtures in `tes
...
+
+User docs:
+- [Certificates](docs/user/certificates.md) - Overview of certificate sources
+- [Parameters](docs/user/parameters.md) - Map of Foreman installation parameters; update as parameters are created/modified
+
+- [CONTRIBUTING](CONTRIBUTING.md) - How to contribute
+- [Development](DEVELOPMENT.md) - Foremanctl development overview
+- [IOP](docs/iop.md) - Overview of insights on premise
+- [Migration Guide](docs/migration-guide.md) - Migrating from foreman-installer to foremanctl
+- [Release](RELEASE.md) - Info on Foremanctl releases
```
ekohl, line 188: Not for this PR, but perhaps it's easier to add [mkdocs](https://www.mkdocs.org/) and instruct the agent on how to use that instead. Just like https://github.com/theforeman/foreman-infra/tree/master/docs generates https://theforeman.github.io/foreman-infra/.

qcjames53: That would be a big upgrade. I didn't want to modify the docs TOO much for a PR where sweeping changes are obviously out of scope.

ehelms:
@qcjames53 a reboot should fix the iop tests

AGENTS.md
ehelms: My opinion: building and trying to maintain a static site as the source of truth introduces overhead. Instead of reading locally, agents now have to query over the network. If you are making changes in a PR to any documentation and trying to consume those updates, you have to generate a static site. Additionally, agents then also have to parse HTML to find the relevant bits rather than just quickly parsing local markdown.

qcjames53:
Thanks Eric. I rebased and pushed to rerun CI. @ekohl would you like to review this PR again or are you okay with current changes?

src/roles/check_foreman_tasks/tasks/main.yaml
ehelms: Once this is done, I'll approve it.

sjha4: I've been working on something similar with https://github.com/theforeman/foremanctl/pull/507/changes#diff-a2aa41e3e24ae0f33fe9a19519efa3a7e637abba707fbbee4041b94aeb9c4116R7 although I have it querying the API..I can refactor this one here in my PR once I rebase on this when it is in.

src/roles/check_duplicate_permissions/tasks/main.yaml
ehelms: Lovely. Guess we gotta keep this check then.

src/roles/check_foreman_tasks/tasks/main.yaml
qcjames53: Oh no, I'd lost this thread in the collapsed comments section. I'll add an issue later this afternoon and close this out. Sorry for the ping.

qcjames53:
Ewoud uncovered https://projects.theforeman.org/issues/38465 which can cause duplicate permissions. Around 4:30 today I will re-add the duplicate permissions check for this release and include instructions in the markdown file to remove this check once the above issue is addressed.

Edit: I'm not planning on adding "--fix" since that's silly to carry over. It will be fail-only and the onus is on the user to clean up the duplicate permissions. Let me know if that's not the ideal approach.

src/roles/check_foreman_tasks/tasks/main.yaml
qcjames53: I've created https://github.com/theforeman/foremanctl/issues/551 to track this issue.

qcjames53:
I've created https://github.com/theforeman/foremanctl/issues/551 to track migrating these db queries to authenticated API calls.

I've also re-added the `check_duplicate_permissions` role (error detection only, no `--fix`) and documented that it needs to stay until https://projects.theforeman.org/issues/38465 is addressed.

qcjames53:
I force pushed to re-run automation (previous failure was unrelated to my changes).

ehelms (approved):
EL10 failure is unrelated, based on other tests I am going to merge.

src/playbooks/health/health.yaml
```diff
@@ -7,14 +7,25 @@
     - "../../vars/defaults.yml"
     - "../../vars/flavors/{{ flavor }}.yml"
     - "../../vars/base.yaml"
+    - "../../vars/database.yml"
+    - "../../vars/health.yml"
```
ianballou, line 11: This file does not exist.

src/roles/check_duplicate_permissions/tasks/main.yaml
```diff
@@ -0,0 +1,27 @@
...
+    query: |
+      SELECT id, name
+      FROM permissions p
+      WHERE (SELECT count(name) FROM permissions pr WHERE p.name = pr.name) > 1
+      ORDER BY name, id
+  register: check_duplicate_permissions_result
+  changed_when: false
+
+- name: Check for duplicate permissions
+  when: not (health_fix | default(false) | bool)
```
ianballou, line 18: what is health_fix doing here?

PR #523, opened 2026-05-27
Force FDW defaults to use undef for things that need inputs

evgeni:

Don't rely on a doc string, fail if no input was given.

* Updated defaults

Steps to reproduce:

* CI will ensure the deployment still works, the change just ensures the defaults are never used verbatim

#### Checklist

### DIFF
--- a/src/roles/iop_fdw/defaults/main.yaml
+++ b/src/roles/iop_fdw/defaults/main.yaml
@@ -2,7 +2,6 @@
 # Required parameters - must be passed by caller
 iop_fdw_database_name: "{{ undef(hint='You must specify the FDW database name') }}"
 iop_fdw_database_user: "{{ undef(hint='You must specify the FDW database user') }}"
-iop_fdw_database_password: "{{ undef(hint='You must specify the FDW database password') }}"
 iop_fdw_remote_database_name: "{{ undef(hint='You must specify the remote FDW database name') }}"
 iop_fdw_remote_user: "{{ undef(hint='You must specify the remote FDW database user') }}"
 iop_fdw_remote_password: "{{ undef(hint='You must specify the remote FDW database password') }}"
--- a/src/roles/iop_fdw/defaults/main.yaml
+++ b/src/roles/iop_fdw/defaults/main.yaml
@@ -1,11 +1,11 @@
 ---
 # Required parameters - must be passed by caller
-iop_fdw_database_name: ""
-iop_fdw_database_user: ""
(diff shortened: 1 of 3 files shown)


src/roles/iop_fdw/defaults/main.yaml
```diff
@@ -1,11 +1,11 @@
...
 # Required parameters - must be passed by caller
-iop_fdw_database_name: ""
-iop_fdw_database_user: ""
-iop_fdw_database_password: ""
-iop_fdw_remote_database_name: ""
-iop_fdw_remote_user: ""
-iop_fdw_remote_password: ""
+iop_fdw_database_name: "{{ undef(hint='You must specify the FDW database name') }}"
+iop_fdw_database_user: "{{ undef(hint='You must specify the FDW database user') }}"
+iop_fdw_database_password: "{{ undef(hint='You must specify the FDW database password') }}"
```
evgeni: I couldn't find a place where this is actually used?

evgeni:
@coderabbitai review

src/roles/iop_fdw/defaults/main.yaml
ehelms: This is unused. This was needed with how it was handled in puppet-iop, but it appears the way Ansible handles it that it's not needed. https://github.com/theforeman/puppet-iop/blob/master/manifests/postgresql_fdw.pp#L51

evgeni: So I can delete that variable (and any setting of it), and we'd be good?

ehelms: Do it

evgeni: I did it!

ehelms (approved):
This feels like a pattern we should enforce? I wonder if an Ansible lint rule is possible.

evgeni:
What would the rule be? "There are no role defaults that are empty strings or `null` (`foo_bar: `)

ehelms:
> What would the rule be? "There are no role defaults that are empty strings or `null` (`foo_bar: `)

Yea. And use that hint technique you are applying. Sounds like a good practice?

evgeni:
> > What would the rule be? "There are no role defaults that are empty strings or `null` (`foo_bar: `)
> 
> Yea. And use that hint technique you are applying. Sounds like a good practice?

https://github.com/theforeman/foremanctl/pull/526

PR #524, opened 2026-05-27
Generate passwords instead of using hard-coded ones

evgeni:

Generate passwords instead of using hard-coded ones

Generate passwords instead of using hard-coded ones

Steps to reproduce:

* Deploy
* Ensure no `CHANGME` passwords/keys are used

#### Checklist

### DIFF
--- a/docs/developer/playbooks-and-roles.md
+++ b/docs/developer/playbooks-and-roles.md
@@ -172,6 +172,45 @@ include:
   - _flavor_features
 ```
 
+## Secret Management
+
+Secrets (passwords, tokens, OAuth secrets) must never be hardcoded. Use the `ansible.builtin.password` lookup to auto-generate secrets and persist them to files under `obsah_state_path`.
+
+### Pattern
+
+Every secret needs two variables: a `_file` variable pointing to the state file, and the secret variable itself using a `lookup` against that file.
+
+```yaml
+example_database_password_file: "{{ obsah_state_path }}/example-db-password"
+example_database_password: "{{ lookup('ansible.builtin.password', example_database_password_file, chars=['ascii_letters', 'digits']) }}"
+```
+
+The `lookup` generates a random password on first run and writes it to the file. On subsequent runs, it reads the existing value, ensuring the secret is stable across deploys.
+
+### Parameters
+
+| Parameter | Usage |
+|-----------|-------|
+| `chars` | Character set for generation. Use `['ascii_letters', 'digits']` for passwords safe in URLs and connection strings. |
+| `length` | Password length. Defaults to 20 if omitted. Use `length=32` for high-entropy secrets like OAuth tokens. |
+
+### Where to define secrets
+
+Define secrets in `src/vars/` files, not in role `defaults/`. Vars files have higher Ansible precedence and are the effective source of truth at deploy time.
+
+| Secret type | Define in |
+|-------------|-----------|
+| Database passwords | `src/vars/database.yml` |
+| IOP database passwords | `src/vars/database_iop.yml` |
+| general passwords/secrets | `src/vars/base.yaml` |
+| Foreman-specific passwords/secrets | `src/vars/foreman.yml` |
+
+### Naming conventions
+
+- State file variable: `<service>_<purpose>_password_file` (or `_secret_file` for non-password secrets)
+- State file path: `{{ obsah_state_path }}/<service>-<purpose>-password` (use hyphens, no underscores)
+- Secret variable: `<service>_<purpose>_password`
+
 ## How to Add a New Command
 
 1. Create a directory under `src/playbooks/<command-name>/` (or `development/playbooks/` for dev tools).
--- a/src/vars/base.yaml
+++ b/src/vars/base.yaml
@@ -3,9 +3,12 @@ certificates_hostnames:
   - "{{ ansible_facts['fqdn'] }}"
   - localhost
 
-certificates_ca_password: "CHANGEME"
-candlepin_keystore_password: "CHANGEME"
-candlepin_oauth_secret: "CHANGEME"
+certificates_ca_password_file: "{{ obsah_state_path }}/certificates-ca-password"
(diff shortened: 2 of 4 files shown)


ekohl (commented):
Neat. Should we document this pattern in developer docs?

evgeni:
> Neat. Should we document this pattern in developer docs?

I guess that'd be a good idea

docs/developer/playbooks-and-roles.md
```diff
@@ -172,6 +172,45 @@ include:
...
+### Where to define secrets
+
+Define secrets in `src/vars/` files, not in role `defaults/`. Vars files have higher Ansible precedence and are the effective source of truth at deploy time.
+
+| Secret type | Define in |
+|-------------|-----------|
+| Database passwords | `src/vars/database.yml` |
+| IOP database passwords | `src/vars/database_iop.yml` |
+| CA/keystore passwords, OAuth secrets | `src/vars/base.yaml` |
+| Foreman admin credentials, OAuth keys | `src/vars/foreman.yml` |
```
ehelms: Why are oauth secrets and keys split?

docs/developer/playbooks-and-roles.md
```diff
@@ -172,6 +172,45 @@ include:
...
+
+Secrets (passwords, tokens, OAuth secrets) must never be hardcoded. Use the `ansible.builtin.password` lookup to auto-generate secrets and persist them to files under `obsah_state_path`.
+
+### Pattern
+
+Every secret needs two variables: a `_file` variable pointing to the state file, and the secret variable itself using a `lookup` against that file.
+
+```yaml
+example_database_password_file: "{{ obsah_state_path }}/example-db-password"
+example_database_password: "{{ lookup('ansible.builtin.password', example_database_password_file, chars=['ascii_letters', 'digits']) }}"
```
ehelms, line 185: Is it possible to turn this into a single filter that enforces this pattern? Or have an Ansible lint rule that enforces it?

docs/developer/playbooks-and-roles.md
evgeni: Because I didn't think far enough, they are both the same thing, just called differently in different places.

docs/developer/playbooks-and-roles.md
evgeni: You mean a lookup(foremanctl_password) that does the right settings for us? Not sure. 

But I'll investigate both options

evgeni: Okay, a lint rule was reasonably easy, added in https://github.com/theforeman/foremanctl/pull/526

ehelms: Nice!

nacoool (commented):
Testing in progress the following PR.

docs/developer/playbooks-and-roles.md
```diff
@@ -172,6 +172,45 @@ include:
...
+```
+
+The `lookup` generates a random password on first run and writes it to the file. On subsequent runs, it reads the existing value, ensuring the secret is stable across deploys.
```
nacoool, line 188: After 1st deploy, on subsequent runs the password is changed using `--candlepin-database-password "somepass"` should it also update the file  with the latest value?

nacoool (changes requested):

docs/developer/playbooks-and-roles.md
evgeni: No, this file is only there for the `ansible.builtin.password` lookup. Once you override it with `--…-password` it's stored in parameters.yaml instead and the lookup is not used anymore.

ekohl: If you reset the value, you will get the "cached" entry again, right?

evgeni: you mean by `--reset-candlepin-database-password`? yeah

nacoool (approved):
ACK.

ShimShtein:
Should we also remove the unused references to passwords from the tasks? Like https://github.com/theforeman/foremanctl/blob/744ba31fd0ee4e4a74eef97e2c705acca1b2961d/src/roles/iop_remediation/defaults/main.yaml#L8 or https://github.com/theforeman/foremanctl/blob/744ba31fd0ee4e4a74eef97e2c705acca1b2961d/development/playbooks/remote-database/remote-database.yaml#L11?

evgeni:
> Should we also remove the unused references to passwords from the tasks? Like
> 
> https://github.com/theforeman/foremanctl/blob/744ba31fd0ee4e4a74eef97e2c705acca1b2961d/src/roles/iop_remediation/defaults/main.yaml#L8
> or
> 
> https://github.com/theforeman/foremanctl/blob/744ba31fd0ee4e4a74eef97e2c705acca1b2961d/development/playbooks/remote-database/remote-database.yaml#L11
> ?

No, this would make the roles not work outside of `foremanctl`

#526 updates them to more sensible defaults

ShimShtein:
Well, no more comments. LGTM!

nacoool:
Tested on 

- foremanctl-3.0.0.develop-1.20260528104015158423.pr524.4.gc0733bb.el9.noarch
- Local : dev-branch : `no-changeme`

Result: **Success**

Scenarios covered:
1. foremanctl fresh deploy without any parameters
2. foremanctl rebuild using custom_password
3. foremanctl rebuild while reseting the custom_password to default password
4. foremanctl rebuild while revoking the db-password file permission
5. foremanctl rebuild while db-password file is deleted
6. Authenticate the database using valid password in the db-password file
7. Authenticate the database using invalid password
8. Authenticate the database using custom valid password provided at the time of `foremanctl deploy`
9. Authenticate the database using default password which becomes invalid because foremanctl is deployed with custom password

Observaion:

- Value provided in the db-password while authenticates when the foreman is deployed with `foremanctl deploy`
- Value provided in the db-password do not authenticates when the foreman is deployed with `foremanctl deploy`
- Value persist across rebuild
- New value is generated if the password file is deleted and is able to authenticate, while unable to authenticate with old password
- deployment gives error if sufficient file permissions are not provided to the files

PR #526, opened 2026-05-28
add ansible-lint rules to ensure no static secrets and empty defaults are present in the code

evgeni:

https://github.com/theforeman/foremanctl/pull/523#pullrequestreview-4372617563

* add a no-empty-defaults check that ensures no defaults are empty
* add a no-static-secrets check that ensure we don't use static secrets
* fix violations of those checks (partially via #524 )

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/.ansible-lint-rules/empty_defaults.py
+++ b/.ansible-lint-rules/empty_defaults.py
@@ -1,0 +1,92 @@
+"""Implementation of var-defaults rule."""
+
+from __future__ import annotations
+
+import sys
+from typing import TYPE_CHECKING
+
+from ansiblelint.rules import AnsibleLintRule
+from ansiblelint.utils import parse_yaml_from_file
+
+if TYPE_CHECKING:
+    from ansiblelint.errors import MatchError
+    from ansiblelint.file_utils import Lintable
+
+
+class EmptyDefaultsRule(AnsibleLintRule):
+    """Role default variables should not have empty values."""
+
+    id = "var-defaults"
+    severity = "HIGH"
+    tags = ["idiom"]
+    version_added = "custom"
+
+    _ids = {
+        "var-defaults[no-empty]": "Role default variables must not be null or empty strings.",
+    }
+
+    def matchyaml(self, file: Lintable) -> list[MatchError]:
+        """Return matches for empty defaults in role defaults files."""
+        results: list[MatchError] = []
+
+        if str(file.kind) != "vars" or not file.data:
+            return results
+
+        if not file.role or "defaults" not in file.path.parts:
+            return results
+
+        meta_data = parse_yaml_from_file(str(file.path))
+        if not isinstance(meta_data, dict):
+            return results
+
+        for key, value in meta_data.items():
+            if value is None or value == "":
+                results.append(
+                    self.create_matcherror(
+                        message=f"Role default variable '{key}' has an empty value. Use `undef(hint='…')` to indicate defaults that need to be overriden.",
+                        filename=file,
+                        tag="var-defaults[no-empty]",
+                        data=key,
+                    ),
+                )
+
+        return results
+
+
+if "pytest" in sys.modules:
+    from ansiblelint.config import Options
+    from ansiblelint.file_utils import Lintable
+    from ansiblelint.rules import RulesCollection
+    from ansiblelint.runner import Runner
+
+    def test_empty_defaults_are_flagged(
+        config_options: Options,
+        app: object,
+    ) -> None:
+        """Null and empty string defaults produce match errors."""
+        rules = RulesCollection(app=app, options=config_options)
+        rules.register(EmptyDefaultsRule())
+        results = Runner(
+            Lintable("tests/fixtures/ansible-lint/roles/test_empty_defaults"),
+            rules=rules,
+        ).run()
+        empty_results = [r for r in results if r.rule.id == EmptyDefaultsRule.id]
+        assert len(empty_results) == 4
+        for result in empty_results:
+            assert result.tag == "var-defaults[no-empty]"
+
+    def test_vars_file_not_checked(
+        config_options: Options,
+        app: object,
+    ) -> None:
+        """Vars files are not checked, only defaults."""
+        rules = RulesCollection(app=app, options=config_options)
+        rules.register(EmptyDefaultsRule())
+        results = Runner(
+            Lintable(
+                "tests/fixtures/ansible-lint/roles/test_empty_defaults/vars/main.yml"
+            ),
+            rules=rules,
+        ).run()
+        empty_results = [r for r in results if r.rule.id == EmptyDefaultsRule.id]
+        assert len(empty_results) == 0
--- a/.ansible-lint-rules/empty_defaults.py
+++ b/.ansible-lint-rules/empty_defaults.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import sys
 from typing import TYPE_CHECKING
 
 from ansiblelint.rules import AnsibleLintRule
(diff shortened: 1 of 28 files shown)


.ansible-lint-rules/empty_defaults.py
```diff
@@ -0,0 +1,92 @@
...
+                        filename=file,
+                        tag="var-defaults[no-empty]",
+                        data=key,
+                    ),
+                )
+
+        return results
+
+
+if "pytest" in sys.modules:
```
evgeni: this is currently not executed by our normal pytest setup

.ansible-lint-rules/empty_defaults.py
```diff
@@ -0,0 +1,92 @@
...
+        rules = RulesCollection(app=app, options=config_options)
+        rules.register(EmptyDefaultsRule())
+        results = Runner(
+            Lintable("tests/fixtures/ansible-lint/roles/test_empty_defaults"),
+            rules=rules,
+        ).run()
+        empty_results = [r for r in results if r.rule.id == EmptyDefaultsRule.id]
+        assert len(empty_results) == 4
+        for result in empty_results:
+            assert result.tag == "var-defaults[no-empty]"
```
archanaserver: but isn't this only happy path testing and this but doesn't explicitly verify that edge cases are correctly NOT flagged.

.ansible-lint-rules/no_empty_defaults.py
```diff
@@ -0,0 +1,92 @@
...
+
+        if not file.role or "defaults" not in file.path.parts:
+            return results
+
+        meta_data = parse_yaml_from_file(str(file.path))
+        if not isinstance(meta_data, dict):
+            return results
+
+        for key, value in meta_data.items():
+            if value is None or value == "":
```
archanaserver, line 39: So just to clarify, The test fixture has passed `test_empty_defaults_plugins: []` which is not flagged right. Shouldn't empty lists `([])` also be checked in that case? Or is empty list a valid default?

archanaserver (commented):
Dropped comments just to clarify some thought on this.

.ansible-lint-rules/no_empty_defaults.py
evgeni: I've kept `[]` as allowed, as my trail of thought was: "this is usually used for defining additions (e.g. plugins), so an empty list is ok"

.ansible-lint-rules/empty_defaults.py
evgeni: which edge cases? `tests/fixtures/ansible-lint/roles/test_empty_defaults/defaults/main.yml` contains 9 variables, and the assertion is that we flag 4 of them as "empty" (the ones that are either `null` or `""`), but others (e.g. `false` and `[]`) are accepted

ehelms:
This is from me testing out a review skill I wrote:

empty_defaults.py and no_static_secrets.py — redundant YAML parse

  - .ansible-lint-rules/empty_defaults.py:36 — parse_yaml_from_file(str(file.path)) re-parses a file that file.data already holds. The rule already guards against empty data with if not file.data, so meta_data = dict(file.data) is sufficient. Same pattern in no_static_secrets.py:51.

no_static_secrets rule: None/empty secret values slip through

  - .ansible-lint-rules/no_static_secrets.py:50 — The rule only matches isinstance(value, str) and not has_jinja(value). A null value for a secret variable (e.g., some_password:) is not flagged. Role defaults will catch it via var-defaults[no-empty], but non-role vars files with a null secret variable are silently ignored. Low risk given how foremanctl uses secrets, but  worth noting.

evgeni:
> empty_defaults.py and no_static_secrets.py — redundant YAML parse
> 
>   * .ansible-lint-rules/empty_defaults.py:36 — parse_yaml_from_file(str(file.path)) re-parses a file that file.data already holds. The rule already guards against empty data with if not file.data, so meta_data = dict(file.data) is sufficient. Same pattern in no_static_secrets.py:51.

`Lintable.data` uses `parse_yaml_linenumbers`, not `parse_yaml_from_file`, and while the tests do pass after the change to `file.data`, I am not versed enough in ansible-lint internals and would follow the existing core rules like [`var-naming` which use `parse_yaml_from_file` explicitly](https://github.com/ansible/ansible-lint/blob/fa50a34935b1ffa1519fd247b93d8bf320620f51/src/ansiblelint/rules/var_naming.py#L314)

src/roles/candlepin/defaults/main.yml
```diff
@@ -20,7 +20,5 @@ candlepin_database_host: localhost
 candlepin_database_port: 5432
 candlepin_database_ssl: false
 candlepin_database_ssl_mode: disable
-candlepin_database_ssl_ca:
+candlepin_database_ssl_ca: # noqa: var-defaults[no-empty]
 candlepin_database_ssl_ca_path: /etc/candlepin/certs/db-ca.crt
-candlepin_database_ssl_cert:
-candlepin_database_ssl_key:
```
evgeni, line 26: love finding unused vars :)

.ansible-lint-rules/empty_defaults.py
```diff
@@ -0,0 +1,52 @@
...
+        if not isinstance(meta_data, dict):
+            return results
+
+        for key, value in meta_data.items():
+            if value is None or value == "":
+                results.append(
+                    self.create_matcherror(
+                        message=f"Role default variable '{key}' has an empty value. Use `undef(hint='…')` to indicate defaults that need to be overriden.",
+                        filename=file,
+                        tag="var-defaults[no-empty]",
```
evgeni: this uses `id[sub-id]` style, as I've used `var-naming` rule as a base, and this has multiple sub-tags.
after using the two rules in this PR (and also flagging certain vars as `noqa` for them), I think this is wrong.

We should either:
- Have a single `var-content` rule file with multiple tags (`var-content[no-empty-defaults]`, `var-content[no-static-secrets]`)
- Have dedicated named rules like `no-empty-defaults` and `no-static-secrets`

I slightly lean towards the dedicated naming, but eager to hear what you think.

evgeni:
> no_static_secrets rule: None/empty secret values slip through
> 
>   * .ansible-lint-rules/no_static_secrets.py:50 — The rule only matches isinstance(value, str) and not has_jinja(value). A null value for a secret variable (e.g., some_password:) is not flagged. Role defaults will catch it via var-defaults[no-empty], but non-role vars files with a null secret variable are silently ignored. Low risk given how foremanctl uses secrets, but  worth noting.

I guess `null` is static by definition, but do we really need to care?
OTOH, adding it to the linter doesn't cost much, so :woman_shrugging:

.ansible-lint-rules/empty_defaults.py
archanaserver: Okay, I see your point with that current cases. :raised_hands:

.ansible-lint-rules/empty_defaults.py
ehelms: Dedicated looks good to me

evgeni: Renamed to dedicated

development/roles/foreman_development/defaults/main.yaml
```diff
@@ -12,13 +12,13 @@ foreman_development_client_certificate: "{{ foreman_client_certificate }}"
 foreman_development_client_key: "{{ foreman_client_key }}"
 
 foreman_development_admin_user: "admin"
-foreman_development_admin_password: "changeme"
+foreman_development_admin_password: "changeme" # noqa: no-static-secrets
```
ehelms, line 15: I have wondered if we should avoid setting a default in the roles as a best practice. While this can be more painful for development scenarios, preventing production deployments with default passwords is a good thing.

ehelms: I misread what role this was :) ignore me

src/roles/candlepin/defaults/main.yml
ehelms: They were planned to be used.

src/vars/installer_certificates.yml
```diff
@@ -1,5 +1,5 @@
 ---
-ca_key_password: "/root/ssl-build/katello-default-ca.pwd"
+ca_key_password: "/root/ssl-build/katello-default-ca.pwd" # noqa: no-static-secrets
```
ehelms, line 2: I suppose we should have called this `ca_key_password_file`

ehelms (approved):

src/vars/installer_certificates.yml
evgeni: Yepp, but I wasn't too sure I wanted to refactor that detail now :D

ShimShtein:
I think it misses https://github.com/theforeman/foremanctl/blob/744ba31fd0ee4e4a74eef97e2c705acca1b2961d/development/playbooks/remote-database/remote-database.yaml#L11

evgeni:
> I think it misses
> 
> https://github.com/theforeman/foremanctl/blob/744ba31fd0ee4e4a74eef97e2c705acca1b2961d/development/playbooks/remote-database/remote-database.yaml#L11

Technically, yes. This file is a playbook, and the lint rule only triggers on vars. We can extend the matching I guess, but this is not a pattern we should encounter in prod.

ehelms (approved):

evgeni:
This had been ACK'ed by Eric for a week, I want to get a release out, so I am merging this now.

PR #527, opened 2026-05-28
[INCOMPLETE] Introduce plain foreman-proxy flavor

ekohl:

This introduces a foreman-proxy flavor so it's easy for users to deploy this. It explores the problem space, but is nowhere complete.

I'm sharing this now for visibility and hope others can take lessons from it.

* It includes https://github.com/theforeman/foremanctl/pull/511 and adds commits on top of it to further isolate testing.

Steps to reproduce:

*

#### Checklist


src/vars/flavors/foreman-proxy.yml
```diff
@@ -0,0 +1,3 @@
+---
+flavor_features:
+  - foreman-proxy
```
arvind4501, line 3: Should this also include pulp(in mirrror mode) as part of flavor defaults?

src/playbooks/deploy/deploy.yaml
```diff
@@ -39,11 +64,23 @@
...
-    - httpd
-    - pulp
-    - foreman
+        - postgresql_databases | length > 0
+    - role: redis
+      when:
+        - "'redis' in enabled_features"
+    - role: candlepin
+      when:
+        - "'foreman' in enabled_features"
```
arvind4501, line 38:
```suggestion
```
Does not katello make sure foreman feature alredy exist? i can't think of a scenario where katello exist without foreman

src/vars/flavors/foreman-proxy.yml
arvind4501: Looking at https://github.com/theforeman/foremanctl/pull/527/changes#diff-2fc200de05abc2d83ea118cae7634c213de69b3e8640146a18644935bfce14f2R78-R80 , pulp does not seems to be deployed in proxy flavor

ekohl: I interpreted plain `foreman-proxy` as a vanilla Foreman without Katello.

src/playbooks/deploy/deploy.yaml
ekohl: Then how do you want to make sure that `candlepin` isn't deployed on the proxy machine?

src/vars/flavors/foreman-proxy.yml
arvind4501: ah okay, i misinterpreted it as `foreman-proxy-content`

ekohl: Even though we don't have it today, my goal here was to really isolate each sub-system and only test the relevant bits.

src/playbooks/deploy/deploy.yaml
arvind4501: I think, for plain foreman and foreman-proxy(without content), we can say `skip pulp and candlepin roles if katello is not in enabled_features`

src/vars/flavors/foreman-proxy.yml
arvind4501: Yes, I was exploring cases where our current test stategy(`test enabled features`) could fail, and one of the thing i came across is katello, katello itself is a feature which contains both pulp and candlepin, but in case of foreman-proxy-content, we need pulp not candlepin and in that scenario we need to isolate pulp and candlepin

src/playbooks/deploy/deploy.yaml
ekohl: > Does not katello make sure foreman feature alredy exist? i can't think of a scenario where katello exist without foreman

What I had in mind was that `katello` is the content feature. On a machine with `foreman` it also deploys `candlepin`. On a machine without `foreman` it just deploys pulp (in mirror mode). That's why both `foreman` and `katello` are listed. Also note that `pulp` is only gated on `katello` being enabled.

src/vars/flavors/foreman-proxy.yml
arvind4501: > I interpreted plain foreman-proxy as a vanilla Foreman without Katello.

By plain do you mean standalone? i thought it as only proxy is deployed on a machine not foreman + proxy

src/playbooks/deploy/deploy.yaml
arvind4501: > On a machine without foreman it just deploys pulp (in mirror mode)

i did not think about that, what if we deploy a standalone foreman-proxy(without content), which neither has foreman and katello? i think we did not had a particular scenario for this case in foreman-installer too, but a hack to do this https://docs.theforeman.org/nightly/Installing_Proxy/index-foreman-el.html#installing-a-smart-proxy-server

src/vars/flavors/foreman-proxy.yml
ekohl: With plain I mean without Katello (or any other plugins). So https://github.com/theforeman/foremanctl/pull/511 creates the Foreman (without Katello) flavor and this PR (https://github.com/theforeman/foremanctl/pull/527) creates the Foreman Proxy (without Katello) flavor (and happens to include https://github.com/theforeman/foremanctl/pull/511).

src/playbooks/deploy/deploy.yaml
```diff
@@ -39,11 +64,23 @@
...
         - database_mode == 'internal'
-    - redis
-    - candlepin
-    - httpd
-    - pulp
-    - foreman
+        - postgresql_databases | length > 0
+    - role: redis
+      when:
+        - "'redis' in enabled_features"
```
arvind4501: I think, This will always be false as internal features are not included in `enabled_features` , thus CI breaks at dynflow containers

arvind4501: its calculated as  `flavor_features + features` and both does not contain redis until we do `--add-feature redis`

ekohl: That's why it's marked as `INCOMPLETE`. This is very much in the idea phase.

src/playbooks/deploy/deploy.yaml
ekohl: > i think we did not had a particular scenario for this case in foreman-installer too, but a hack to do this https://docs.theforeman.org/nightly/Installing_Proxy/index-foreman-el.html#installing-a-smart-proxy-server

Correct, we never completed https://github.com/theforeman/foreman-installer/pull/262 but I can see use cases for it. We've had questions about a Remote Execution-only Foreman Proxy, without all of the content.

src/playbooks/deploy/deploy.yaml
arvind4501: Ok, i just wanted to get to the reason of test failures on the pr and how can that effect current tests markers, so i can then look for new apporch, but yeah

evgeni: IMHO this should use the `role` attribute of our feature metadata, not some weird `when in` construct.

src/playbooks/deploy/deploy.yaml
evgeni, line 38: IMHO this should use the `role` attribute of our feature metadata, not some weird `when in` construct.

tests/feature/foreman-proxy/base_test.py
```diff
@@ -13,9 +13,12 @@ def test_foreman_proxy_features(curl_request, proxy_base_url, enabled_features):
     assert "logs" in features
     if 'remote-execution' in enabled_features:
         assert "script" in features
-        assert "dynflow" in features
     else:
         assert "script" not in features
+    if "dynflow" in enabled_features:
```
ekohl, line 18: This now cause a test failure that is addressed by https://github.com/theforeman/foremanctl/pull/781. However, perhaps it is good to also test that `dynflow` is present when REX is enabled.

PR #528, opened 2026-05-29
declare foreman-tasks feature

evgeni:

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/features.yaml
+++ b/src/features.yaml
@@ -10,6 +10,8 @@ katello:
   foreman:
     plugin_name: katello
   hammer: katello
+  dependencies:
+    - tasks
 google:
   description: Google Compute Engine plugin for Foreman
   hammer: foreman_google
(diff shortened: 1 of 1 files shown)


src/features.yaml
```diff
@@ -20,6 +22,11 @@ azure-rm:
   hammer: foreman_azure_rm
   foreman:
     plugin_name: foreman_azure_rm
+tasks:
```
ehelms, line 25: Should this be marked as internal?

evgeni: idk? For pure foreman it might not be.

ehelms: I guess it won't affect our use cases since it will be pulled in by dependencies. Just always been a weird one to list as a "feature" since it's not user facing.

evgeni: Updated

ehelms (approved):

PR #529, opened 2026-05-29
Require SSL in devel

jturel:

I want to understand why `require_ssl` defaults to false in development. It seems the devel settings.yaml was copied from puppet-katello_devel which has defaulted to `false` since the very beginning of the repo in 2014.

* Requires SSL in development environments

TBD

Steps to reproduce:

*

#### Checklist


ehelms:
Good question... here's another one: what does this setting do and what are the implications of it being true?

PR #531, opened 2026-06-01
Add KEA DHCP provider feature for foreman-proxy

archanaserver:
ISC DHCP reached end-of-life in October 2022 and has been removed from RHEL 10. This adds support for KEA DHCP as the replacement.

* Add `dhcp-kea-external` feature definition to `src/features.yaml`
* Create Smart Proxy configuration template `dhcp_kea_api.yml.j2`
* Add default Ansible variables for KEA server connection (localhost:8000, default subnet)
* Add integration tests for DHCP KEA feature validation

Steps to reproduce:

1. Verify feature is available: `./foremanctl features | grep dhcp-kea-external`
2. Deploy with KEA DHCP feature enabled: `./foremanctl deploy --add-feature dhcp-kea-external`
3. Verify feature is enabled: `./foremanctl features --list-enabled | grep dhcp-kea-external`
4. Check configuration file was created: `podman exec foreman-proxy ls -la /etc/foreman-proxy/settings.d/dhcp_kea_api.yml`
5. Run tests `python -m pytest tests/foreman_proxy_test.py::test_dhcp_kea_feature_present -v`

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -115,10 +115,6 @@ jobs:
       - name: Setup Kea DHCP server
         run: |
           ./forge kea
-      - name: Create installer certificates
-        if: contains(matrix.certificate_source, 'installer')
-        run: |
-          ./forge installer-certs
       - name: Create custom certificates
         if: matrix.certificate_source == 'custom_server'
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -112,6 +112,13 @@ jobs:
       - name: Configure repositories
         run: |
           ./forge setup-repositories
+      - name: Setup Kea DHCP server
+        run: |
+          ./forge kea
+      - name: Create installer certificates
+        if: contains(matrix.certificate_source, 'installer')
+        run: |
+          ./forge installer-certs
       - name: Create custom certificates
         if: matrix.certificate_source == 'custom_server'
         run: |
@@ -152,6 +159,9 @@ jobs:
             --add-feature google \
             --add-feature remote-execution \
             --add-feature bmc \
+            --add-feature dhcp \
+            --foreman-proxy-dhcp-provider dhcp_kea \
+            --foreman-proxy-dhcp-kea-url http://127.0.0.1:8000 \
             ${{ matrix.iop == 'enabled' && '--add-feature iop' || '' }}
       - name: Run tests
         run: |
@@ -245,6 +255,12 @@ jobs:
       - name: Configure repositories
         run: |
           ./forge setup-repositories
+      - name: Setup Kea DHCP server
+        run: |
+          ./forge kea
+      - name: Configure base version
+        run: |
+          sed -i '/container_tag_stream:/ s/:.*/: "${{ matrix.upgrade_from }}"/' src/vars/images.yml
       - name: Run image pull
         run: |
           ./foremanctl pull-images
@@ -262,7 +278,10 @@ jobs:
             --add-feature foreman-proxy \
             --add-feature azure-rm \
             --add-feature google \
-            --add-feature remote-execution
+            --add-feature remote-execution \
+            --add-feature dhcp \
+            --foreman-proxy-dhcp-provider dhcp_kea \
+            --foreman-proxy-dhcp-kea-url http://127.0.0.1:8000
       - name: Stop services
         run:
           vagrant ssh quadlet -- sudo systemctl stop foreman.target
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -160,7 +160,6 @@ jobs:
             --add-feature remote-execution \
             --add-feature bmc \
             --add-feature dhcp \
-            --foreman-proxy-dhcp-provider dhcp_kea \
             --foreman-proxy-dhcp-kea-url http://127.0.0.1:8000 \
             ${{ matrix.iop == 'enabled' && '--add-feature iop' || '' }}
       - name: Run tests
@@ -280,7 +279,6 @@ jobs:
             --add-feature google \
             --add-feature remote-execution \
             --add-feature dhcp \
-            --foreman-proxy-dhcp-provider dhcp_kea \
             --foreman-proxy-dhcp-kea-url http://127.0.0.1:8000
       - name: Stop services
         run:
--- a/development/playbooks/kea/kea.yaml
+++ b/development/playbooks/kea/kea.yaml
@@ -32,53 +32,18 @@
         group: root
 
     - name: Configure Kea DHCP4 server
-      ansible.builtin.copy:
+      ansible.builtin.template:
+        src: templates/kea-dhcp4.conf.j2
         dest: /etc/kea/kea-dhcp4.conf
-        content: |
-          {
-            "Dhcp4": {
-              "interfaces-config": {
-                "interfaces": [ "*" ]
-              },
-              "control-socket": {
-                "socket-type": "unix",
-                "socket-name": "/var/run/kea/kea4-ctrl-socket"
-              },
-              "subnet4": [
-                {
-                  "id": 1,
-                  "subnet": "192.168.122.0/24",
-                  "pools": [{ "pool": "192.168.122.100 - 192.168.122.200" }]
-                }
-              ],
-              "lease-database": {
-                "type": "memfile",
-                "persist": true,
-                "name": "/var/lib/kea/kea-leases4.csv"
-              }
-            }
-          }
         mode: '0644'
         owner: root
         group: root
       notify: Restart Kea DHCP4
 
     - name: Configure Kea Control Agent
-      ansible.builtin.copy:
+      ansible.builtin.template:
+        src: templates/kea-ctrl-agent.conf.j2
         dest: /etc/kea/kea-ctrl-agent.conf
-        content: |
-          {
-            "Control-agent": {
-              "http-host": "127.0.0.1",
-              "http-port": 8000,
-              "control-sockets": {
-                "dhcp4": {
-                  "socket-type": "unix",
-                  "socket-name": "/var/run/kea/kea4-ctrl-socket"
-                }
-              }
-            }
-          }
         mode: '0644'
         owner: root
         group: root
@@ -106,6 +71,21 @@
       ansible.builtin.systemd:
         name: kea-dhcp4
         state: started
+
+    - name: Check Kea DHCP4 service status
+      ansible.builtin.systemd:
+        name: kea-dhcp4
+      register: kea_dhcp4_status
+      failed_when: false
+
+    - name: Display Kea DHCP4 service status
+      ansible.builtin.debug:
+        var: kea_dhcp4_status
+
+    - name: Wait for DHCP4 control socket to be created
+      ansible.builtin.wait_for:
+        path: /var/run/kea/kea4-ctrl-socket
+        timeout: 30
 
     - name: Start Kea Control Agent service
       ansible.builtin.systemd:
--- a/development/playbooks/kea/kea.yaml
+++ b/development/playbooks/kea/kea.yaml
@@ -15,6 +15,22 @@
           - kea
         state: present
 
+    - name: Create Kea runtime directory
+      ansible.builtin.file:
+        path: /var/run/kea
+        state: directory
+        mode: '0750'
+        owner: root
+        group: root
+
+    - name: Create Kea lib directory
+      ansible.builtin.file:
+        path: /var/lib/kea
+        state: directory
+        mode: '0755'
+        owner: root
+        group: root
+
     - name: Configure Kea DHCP4 server
       ansible.builtin.copy:
         dest: /etc/kea/kea-dhcp4.conf
@@ -26,7 +42,7 @@
               },
               "control-socket": {
                 "socket-type": "unix",
-                "socket-name": "/run/kea/kea4-ctrl-socket"
+                "socket-name": "/var/run/kea/kea4-ctrl-socket"
               },
               "subnet4": [
                 {
@@ -58,7 +74,7 @@
               "control-sockets": {
                 "dhcp4": {
                   "socket-type": "unix",
-                  "socket-name": "/run/kea/kea4-ctrl-socket"
+                  "socket-name": "/var/run/kea/kea4-ctrl-socket"
                 }
               }
             }
@@ -68,14 +84,33 @@
         group: root
       notify: Restart Kea Control Agent
 
-    - name: Start Kea services
+    - name: Validate Kea DHCP4 configuration
+      ansible.builtin.command: kea-dhcp4 -t /etc/kea/kea-dhcp4.conf
+      changed_when: false
+      register: dhcp4_validation
+
+    - name: Validate Kea Control Agent configuration
+      ansible.builtin.command: kea-ctrl-agent -t /etc/kea/kea-ctrl-agent.conf
+      changed_when: false
+      register: ctrl_agent_validation
+
+    - name: Enable Kea services
       ansible.builtin.systemd:
         name: "{{ item }}"
-        state: started
         enabled: true
       loop:
         - kea-dhcp4
         - kea-ctrl-agent
+
+    - name: Start Kea DHCP4 service
+      ansible.builtin.systemd:
+        name: kea-dhcp4
+        state: started
+
+    - name: Start Kea Control Agent service
+      ansible.builtin.systemd:
+        name: kea-ctrl-agent
+        state: started
 
     - name: Wait for Kea Control Agent to be ready
       ansible.builtin.wait_for:
--- a/development/playbooks/kea/kea.yaml
+++ b/development/playbooks/kea/kea.yaml
@@ -1,0 +1,96 @@
+---
+- name: Kea DHCP
+  hosts:
+    - quadlet
+  become: true
+  tasks:
+    - name: Install EPEL repository
+      ansible.builtin.package:
+        name: epel-release
+        state: present
+
+    - name: Install Kea packages
+      ansible.builtin.package:
+        name:
+          - kea
+        state: present
+
+    - name: Configure Kea DHCP4 server
+      ansible.builtin.copy:
+        dest: /etc/kea/kea-dhcp4.conf
+        content: |
+          {
+            "Dhcp4": {
+              "interfaces-config": {
+                "interfaces": [ "*" ]
+              },
+              "control-socket": {
+                "socket-type": "unix",
+                "socket-name": "/run/kea/kea4-ctrl-socket"
+              },
+              "subnet4": [
+                {
+                  "id": 1,
+                  "subnet": "192.168.122.0/24",
+                  "pools": [{ "pool": "192.168.122.100 - 192.168.122.200" }]
+                }
+              ],
+              "lease-database": {
+                "type": "memfile",
+                "persist": true,
+                "name": "/var/lib/kea/kea-leases4.csv"
+              }
+            }
+          }
+        mode: '0644'
+        owner: root
+        group: root
+      notify: Restart Kea DHCP4
+
+    - name: Configure Kea Control Agent
+      ansible.builtin.copy:
+        dest: /etc/kea/kea-ctrl-agent.conf
+        content: |
+          {
+            "Control-agent": {
+              "http-host": "127.0.0.1",
+              "http-port": 8000,
+              "control-sockets": {
+                "dhcp4": {
+                  "socket-type": "unix",
+                  "socket-name": "/run/kea/kea4-ctrl-socket"
+                }
+              }
+            }
+          }
+        mode: '0644'
+        owner: root
+        group: root
+      notify: Restart Kea Control Agent
+
+    - name: Start Kea services
+      ansible.builtin.systemd:
+        name: "{{ item }}"
+        state: started
+        enabled: true
+      loop:
+        - kea-dhcp4
+        - kea-ctrl-agent
+
+    - name: Wait for Kea Control Agent to be ready
+      ansible.builtin.wait_for:
+        host: 127.0.0.1
+        port: 8000
+        state: started
+        timeout: 30
+
+  handlers:
+    - name: Restart Kea DHCP4
+      ansible.builtin.systemd:
+        name: kea-dhcp4
+        state: restarted
+
+    - name: Restart Kea Control Agent
+      ansible.builtin.systemd:
+        name: kea-ctrl-agent
+        state: restarted
--- a/src/features.yaml
+++ b/src/features.yaml
@@ -62,3 +62,5 @@ dhcp:
   description: DHCP management
   foreman_proxy:
     plugin_name: dhcp
+  dependencies:
+    - foreman-proxy
--- a/src/features.yaml
+++ b/src/features.yaml
@@ -58,3 +58,7 @@ bmc:
   description: Power management for bare metal hosts (IPMI, Redfish)
   foreman_proxy:
     plugin_name: bmc
+dhcp-kea-external:
+  description: KEA DHCP provider (external unmanaged server)
+  foreman_proxy:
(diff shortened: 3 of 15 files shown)


src/features.yaml
```diff
@@ -58,3 +48,7 @@ bmc:
   description: Power management for bare metal hosts (IPMI, Redfish)
   foreman_proxy:
     plugin_name: bmc
+dhcp-kea-external:
```
evgeni: I think the feature should be "dhcp" and then you add a parameter to select which provider to use, with kea being one of the answers

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -23,3 +23,8 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
 # BMC settings
 foreman_proxy_bmc_ipmi_implementation: ipmitool
 foreman_proxy_bmc_redfish_verify_ssl: true
+
+# KEA DHCP settings (external unmanaged server)
+foreman_proxy_dhcp_kea_server: localhost
```
evgeni: What are the chances that `localhost` is a place where a Kea will actually run (outside of development setups)?

I think we should leave that `undef` (like in https://github.com/theforeman/foremanctl/pull/523/changes) and force the user to set a sensible value.

src/roles/foreman_proxy/templates/settings.d/dhcp_kea_api.yml.j2
```diff
@@ -0,0 +1,9 @@
+---
+# KEA DHCP provider configuration
+# This configures Smart Proxy to use an external KEA DHCP server
+
+:enabled: {{ feature_enabled }}
+:use_provider: dhcp_kea_api
+:server: {{ foreman_proxy_dhcp_kea_server }}
```
evgeni: where does this come from? I don't see such an option in https://gitlab.surrey.ac.uk/sm0049/smart-proxy-dhcp-kea-api/

src/roles/foreman_proxy/templates/settings.d/dhcp_kea_api.yml.j2
```diff
@@ -0,0 +1,9 @@
...
+:kea_url: {{ foreman_proxy_dhcp_kea_url }}
```
evgeni: shouldn't this be `kea_api_url`?

src/roles/foreman_proxy/templates/settings.d/dhcp_kea_api.yml.j2
```diff
@@ -0,0 +1,9 @@
...
+:kea_subnet: {{ foreman_proxy_dhcp_kea_subnet }}
```
evgeni: where does this come from? I don't see such an option in https://gitlab.surrey.ac.uk/sm0049/smart-proxy-dhcp-kea-api/

src/roles/foreman_proxy/templates/settings.d/dhcp_kea_api.yml.j2
evgeni: also, we need `kea_api_username` and `kea_api_password`

evgeni:
Shall we package https://gitlab.surrey.ac.uk/sm0049/smart-proxy-dhcp-kea-api until we figure out whether we want to merge it into core or not?

archanaserver:
> Shall we package https://gitlab.surrey.ac.uk/sm0049/smart-proxy-dhcp-kea-api until we figure out whether we want to merge it into core or not?

Yeah, I was thinking we can package it for now, that will be independent of our decision about potentially merging it into the core later.

src/roles/foreman_proxy/templates/settings.d/dhcp_kea_api.yml.j2
archanaserver: yeah, i was just playing around with few things, needed to look again, the reason i added that in draft but thanks for reviews, also i tryna fix few things with the updated changes.

src/roles/foreman_proxy/templates/settings.d/dhcp_kea_api.yml.j2
archanaserver: yes def

src/features.yaml
archanaserver: makes way more sense to have a single `dhcp` feature with a provider parameter

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -23,3 +23,12 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
 # BMC settings
 foreman_proxy_bmc_ipmi_implementation: ipmitool
 foreman_proxy_bmc_redfish_verify_ssl: true
+
+# DHCP settings
+foreman_proxy_dhcp_provider: dhcp_kea_api
```
stejskalleos: Why not just:

```suggestion
foreman_proxy_dhcp_provider: dhcp_kea
```

src/roles/foreman_proxy/templates/settings.d/dhcp_kea.yml.j2
```diff
@@ -0,0 +1,18 @@
...
+:enabled: {{ feature_enabled }}
+:kea_api_url: {{ foreman_proxy_dhcp_kea_api_url }}
+{% if foreman_proxy_dhcp_kea_api_username is defined and foreman_proxy_dhcp_kea_api_username %}
+:kea_api_username: {{ foreman_proxy_dhcp_kea_api_username }}
+{% endif %}
+{% if foreman_proxy_dhcp_kea_api_password is defined and foreman_proxy_dhcp_kea_api_password %}
+:kea_api_password: {{ foreman_proxy_dhcp_kea_api_password }}
+{% endif %}
+{% if foreman_proxy_dhcp_kea_managed_subnets is defined and foreman_proxy_dhcp_kea_managed_subnets %}
+:managed_subnets:
```
stejskalleos: There is a definition of subnets in DHCP as well: https://github.com/theforeman/smart-proxy/blob/develop/config/settings.d/dhcp.yml.example

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -23,3 +23,12 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
...
+
+# KEA DHCP provider settings (when foreman_proxy_dhcp_provider is dhcp_kea_api)
+foreman_proxy_dhcp_kea_api_url: "{{ undef(hint='You must specify the KEA API URL') }}"
+foreman_proxy_dhcp_kea_api_username: ""
+foreman_proxy_dhcp_kea_api_password: ""
+foreman_proxy_dhcp_kea_managed_subnets: []
```
stejskalleos: I think this should be:
```suggestion
foreman_proxy_dhcp_subnets: []
```

as in https://github.com/theforeman/smart-proxy/blob/develop/config/settings.d/dhcp.yml.example

evgeni:
> > Shall we package https://gitlab.surrey.ac.uk/sm0049/smart-proxy-dhcp-kea-api until we figure out whether we want to merge it into core or not?
> 
> Yeah, I was thinking we can package it for now, that will be independent of our decision about potentially merging it into the core later.

Do you want to do it, or shall I?

src/features.yaml
evgeni: I played with that thought in #532 for the DNS side of things, would love to hear your opinion on the approach I took

src/features.yaml
```diff
@@ -58,3 +48,7 @@ bmc:
   description: Power management for bare metal hosts (IPMI, Redfish)
   foreman_proxy:
     plugin_name: bmc
+dhcp:
+  description: DHCP management
+  foreman_proxy:
+    plugin_name: dhcp
```
amolpati30, line 64: Do you think using `dhcp` is sufficient, or should it be more specific, such as `kea-dhcp`?

archanaserver:
> > > Shall we package https://gitlab.surrey.ac.uk/sm0049/smart-proxy-dhcp-kea-api until we figure out whether we want to merge it into core or not?
> > 
> > 
> > Yeah, I was thinking we can package it for now, that will be independent of our decision about potentially merging it into the core later.
> 
> Do you want to do it, or shall I?

I'll do it

src/features.yaml
archanaserver: `dhcp` is sufficient. As discussed earlier in [this comment](https://github.com/theforeman/foremanctl/pull/531#discussion_r3333064886), it makes way more sense to have a single `dhcp` feature with a provider parameter (foreman_proxy_dhcp_provider: dhcp_kea) rather than creating separate features for each DHCP provider.

archanaserver:
Update: Following the refinement session, we decided to integrate the KEA DHCP provider directly into smart-proxy core rather than packaging it as an external gem. I'm on it. So keeping this into draft till then.

Also looking at the current CI test failures, I believe they are coming from IOP tests failures and are unrelated to the DHCP feature changes in this PR.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,19 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_dhcp_kea_api_url:
+    parameter: --foreman-proxy-dhcp-kea-api-url
+    help: KEA Control Agent API URL (e.g., http://kea-server:8000/)
```
stejskalleos: Nit: https example
```suggestion
    help: KEA Control Agent API URL (e.g., https://kea-server)
```

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,19 @@ variables:
...
+  foreman_proxy_dhcp_kea_api_username:
+    parameter: --foreman-proxy-dhcp-kea-api-username
+    help: Optional HTTP Basic Auth username for KEA API
```
stejskalleos: Do we want to make it optional? It's not really secure IMO

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -22,3 +22,13 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
...
 foreman_proxy_bmc_ipmi_implementation: ipmitool
 foreman_proxy_bmc_redfish_verify_ssl: true
+
+# DHCP settings
+foreman_proxy_dhcp_provider: dhcp_kea
+foreman_proxy_dhcp_subnets: []
+
+# KEA DHCP provider settings (when foreman_proxy_dhcp_provider is dhcp_kea)
+foreman_proxy_dhcp_kea_api_url: "{{ undef(hint='You must specify the KEA API URL') }}"
+foreman_proxy_dhcp_kea_api_username: "{{ undef(hint='Optional HTTP Basic Auth username for KEA API') }}"
```
stejskalleos: If the parameter is optional, why raise an error when no value is provided?

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,19 @@ variables:
...
+  foreman_proxy_dhcp_kea_api_password:
+    parameter: --foreman-proxy-dhcp-kea-api-password
+    help: Optional HTTP Basic Auth password for KEA API
+  foreman_proxy_dhcp_kea_verify_ssl:
+    parameter: --foreman-proxy-dhcp-kea-verify-ssl
+    help: Verify SSL certificates when connecting to KEA API
```
stejskalleos: Mention the default value (`true`)

src/playbooks/deploy/metadata.obsah.yaml
stejskalleos: (`foreman_proxy_dhcp_kea_verify_ssl`) is true by default, so this should be https

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -22,3 +22,13 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
...
+foreman_proxy_dhcp_kea_api_password: "{{ undef(hint='Optional HTTP Basic Auth password for KEA API') }}"
```
stejskalleos: same as above

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,19 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_dhcp_kea_api_url:
```
stejskalleos: I don't see a parameter for DHCP subnets.

src/features.yaml
```diff
@@ -58,3 +58,7 @@ bmc:
   description: Power management for bare metal hosts (IPMI, Redfish)
   foreman_proxy:
     plugin_name: bmc
+dhcp:
+  description: DHCP management
```
stejskalleos, line 62: We should mention KEA

tests/foreman_proxy_test.py
```diff
@@ -69,3 +73,26 @@ def test_bmc_capabilities(proxy_v2_features):
 def test_bmc_default_provider(proxy_v2_features):
     settings = proxy_v2_features['bmc'].get('settings', {})
     assert settings.get('bmc_default_provider') == 'ipmitool'
+
+
+@pytest.mark.feature('dhcp')
```
stejskalleos, line 99: Similar to the [TFTP PR](https://github.com/theforeman/foremanctl/pull/533), we should have tests that spin up the KEA, make the DHCP calls, and verify that it works.

stejskalleos (changes requested):

src/features.yaml
evgeni: No, this is also for Infoblox

src/playbooks/deploy/metadata.obsah.yaml
archanaserver: I have updated it but auth is genuinely optional for kea control agent (many deployments run without auth on internal networks), but i agree we should encourage it for security. so fixed it

src/features.yaml
archanaserver: What I think of we can either keep the desc generic to avoid maintaining a provider list or changes it to something like this `DHCP management (supports ISC KEA, ISC DHCP, MS DHCP, Libvirt, Infoblox)` to include all of it here. not sure though?!

src/playbooks/deploy/metadata.obsah.yaml
archanaserver: yes, now we can specify this

tests/foreman_proxy_test.py
archanaserver: Yeah good point. So I was looking into this and I remember talking with @evgen @ekohl few days back, that the right approach would be to make the tests generic like https://github.com/theforeman/forklift/blob/master/bats/fb-proxy-dns.bats that just hit the foreman proxy API, not KEA-specific stuff. That way same tests would work for Infoblox or whatever DHCP backend.

And I created [SAT-45851](https://redhat.atlassian.net/browse/SAT-45851) for tracking this. Thinking separate PR since we'd need to set up KEA in CI and all that. WDYT?

src/roles/foreman_proxy/defaults/main.yaml
archanaserver: So I removed optional from the hint text to reduce any confusion but the `undef()` is never actually raised because the template checks if you see 

```
{% if foreman_proxy_dhcp_kea_api_username is defined and foreman_proxy_dhcp_kea_api_username %}
:dhcp_kea_api_username: {{ foreman_proxy_dhcp_kea_api_username }}
{% else %}
:dhcp_kea_api_username: ~
{% endif %}
```

is defined before using it. I added just to satisfy the linter earlier. do you have any other suggestons?

tests/foreman_proxy_test.py
ekohl: With DNS and TFTP we did include the tests in the PR itself because it proves it works. I'd prefer to include it.

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -22,3 +22,13 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
 # BMC settings
 foreman_proxy_bmc_ipmi_implementation: ipmitool
 foreman_proxy_bmc_redfish_verify_ssl: true
+
+# DHCP settings
+foreman_proxy_dhcp_provider: dhcp_kea
+foreman_proxy_dhcp_subnets: []
```
stejskalleos, line 28:
```suggestion
foreman_proxy_dhcp_subnets: "{{ undef(hint='You must specify subnets for KEA') }}"
```

stejskalleos: Or is there a scenario where you want to configure KEA without subnets?

src/roles/foreman_proxy/templates/settings.d/dhcp_kea.yml.j2
```diff
@@ -0,0 +1,17 @@
...
+{% else %}
+:dhcp_kea_username: ~
+{% endif %}
+{% if foreman_proxy_dhcp_kea_password is defined and foreman_proxy_dhcp_kea_password %}
+:dhcp_kea_password: {{ foreman_proxy_dhcp_kea_password }}
+{% else %}
+:dhcp_kea_password: ~
+{% endif %}
+:dhcp_kea_verify_ssl: {{ foreman_proxy_dhcp_kea_verify_ssl | default(true) | lower }}
+:dhcp_kea_lease_timeout: {{ foreman_proxy_dhcp_kea_lease_timeout | default(60) }}
```
stejskalleos: The default value should be set in the `src/roles/foreman_proxy/defaults/main.yaml`

stejskalleos: Question: Do we want to give users a chance to edit this value via foremanctl? If yes, then we need another parameter for it.

src/roles/foreman_proxy/templates/settings.d/dhcp_kea.yml.j2
```diff
@@ -0,0 +1,17 @@
...
+:dhcp_kea_username: {{ foreman_proxy_dhcp_kea_username }}
+{% else %}
+:dhcp_kea_username: ~
+{% endif %}
+{% if foreman_proxy_dhcp_kea_password is defined and foreman_proxy_dhcp_kea_password %}
+:dhcp_kea_password: {{ foreman_proxy_dhcp_kea_password }}
+{% else %}
+:dhcp_kea_password: ~
+{% endif %}
+:dhcp_kea_verify_ssl: {{ foreman_proxy_dhcp_kea_verify_ssl | default(true) | lower }}
```
stejskalleos: Is this needed when the default value is already set in `src/roles/foreman_proxy/defaults/main.yaml`?


```suggestion
:dhcp_kea_verify_ssl: {{ foreman_proxy_dhcp_kea_verify_ssl }}
```

tests/foreman_proxy_test.py
stejskalleos, line 99: Can we also add the `forge kea` command (similar to tftp` which will set up a KEA instance, so we can test it E2E?

stejskalleos (changes requested):

src/roles/foreman_proxy/defaults/main.yaml
archanaserver: My thinking on this is yes, because the `subnets` parameter is a filter that restricts here with which subnets smart proxy manages. and it's not required. ALso when omitted from smart proxy config, it manages all subnets, and this is the most common case - users typically want to manage all their kea subnets and only use this parameter if they need to restrict to a subset.

So if we use `undef()` here, IMO would force every user to specify subnets even for the default "manage all" scenario.

src/roles/foreman_proxy/templates/settings.d/dhcp_kea.yml.j2
archanaserver: Yes, both values are exposed as CLI parameters

tests/foreman_proxy_test.py
archanaserver: Yes I've implemented it, seeing some failure will update the changes next

archanaserver: @stejskalleos done, after many failures, with the foreman-proxy

development/playbooks/kea/kea.yaml
```diff
@@ -0,0 +1,131 @@
...
+        path: /var/lib/kea
+        state: directory
+        mode: '0755'
+        owner: root
+        group: root
+
+    - name: Configure Kea DHCP4 server
+      ansible.builtin.copy:
+        dest: /etc/kea/kea-dhcp4.conf
+        content: |
```
stejskalleos: Can we make it a file/template?

development/playbooks/kea/kea.yaml
```diff
@@ -0,0 +1,131 @@
...
+      ansible.builtin.systemd:
+        name: kea-dhcp4
+        state: started
+
+    - name: Start Kea Control Agent service
+      ansible.builtin.systemd:
+        name: kea-ctrl-agent
+        state: started
+
+    - name: Wait for Kea Control Agent to be ready
```
stejskalleos, line 95: This is failing for me, repeatedly.

```
TASK [Wait for Kea Control Agent to be ready] ****************************************************************************************************
[ERROR]: Task failed: Module failed: Timeout when waiting for 127.0.0.1:8000
Origin: /home/lstejska/projects/foremanctl/development/playbooks/kea/kea.yaml:115:7

113         state: started
114
115     - name: Wait for Kea Control Agent to be ready
          ^ column 7

fatal: [quadlet]: FAILED! => {
    "changed": false,
    "elapsed": 30
}

MSG:

Timeout when waiting for 127.0.0.1:8000

PLAY RECAP ***************************************************************************************************************************************
quadlet                    : ok=12   changed=9    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

before I ran:
```
./foremanctl deploy --initial-admin-password=changeme --tuning development --add-feature foreman-proxy --add-feature dhcp --foreman-proxy-dhcp-kea-url http://127.0.0.1:8000
```

.github/workflows/test.yml
```diff
@@ -112,6 +112,13 @@ jobs:
       - name: Configure repositories
         run: |
           ./forge setup-repositories
+      - name: Setup Kea DHCP server
+        run: |
+          ./forge kea
+      - name: Create installer certificates
+        if: contains(matrix.certificate_source, 'installer')
+        run: |
+          ./forge installer-certs
```
stejskalleos: Why is this needed? 

For me it's failing with:
```
forge: error: argument action: invalid choice: 'installer-certs' (choose from 'custom-certs', 'deploy-dev', 'kea', 'lock', 'mock-installer', 'remote-database', 'security', 'setup-repositories', 'smoker', 'sos', 'test', 'vms'
```

stejskalleos (changes requested):

development/playbooks/kea/kea.yaml
archanaserver: I see it for me as well, taking a look on it

archanaserver: Also you don't need to include `--add-feature foreman-proxy`, I have added it as a dependency in the dhcp with `expand_features` that automatically includes all dependencies.

.github/workflows/test.yml
archanaserver: I see that's a leftover dead code. I should have dropped earlier. Let's see now

PR #532, opened 2026-06-01
expose DNS feature with nsupdate provider and a dedicated DNS server

evgeni:

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/docs/user/parameters.md
+++ b/docs/user/parameters.md
@@ -112,6 +112,16 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--add-feature bmc` | Enable BMC feature | `--foreman-proxy-bmc` |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
+| `--add-feature dns` | Enable DNS feature | `--foreman-proxy-dns` |
+| `--dns-provider` | Select DNS provider | `--foreman-proxy-dns-provider` |
+| `--dns-server` | Configure the target DNS server | `--foreman-proxy-dns-server` |
+| `--dns-nsupdate-keyfile` | Configure the key for the target DNS server | `--foreman-proxy-keyfile` |
+
+### Unmapped
+
+| foreman-installer Parameter | Description | Reason |
+| --------------------------- | ----------- | ------ |
+| `--foreman-proxy-dns-managed` | Installer managed DNS | not supported |
 
 ### Undetermined
 
@@ -134,11 +144,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-proxy-dhcp-server` | | foreman_proxy | dhcp_server |
 | `--foreman-proxy-libvirt-network` | | foreman_proxy | libvirt_network |
 | `--foreman-proxy-libvirt-url` | | foreman_proxy | libvirt_url |
-| `--foreman-proxy-dns` | | foreman_proxy | dns |
-| `--foreman-proxy-dns-managed` | | foreman_proxy | dns_managed |
-| `--foreman-proxy-dns-provider` | | foreman_proxy | foreman_proxy::dns_provider |
-| `--foreman-proxy-dns-server` | | foreman_proxy | foreman_proxy::dns_server |
-| `--foreman-proxy-keyfile` | | foreman_proxy | foreman_proxy::keyfile |
 | `--foreman-proxy-plugin-dns-powerdns-rest-api-key` | | foreman_proxy::plugin::dns_powerdns | rest_api_key |
 | `--foreman-proxy-plugin-dns-powerdns-rest-url` | | foreman_proxy::plugin::dns_powerdns | rest_url |
 | `--foreman-proxy-plugin-dns-route53-aws-access-key` | | foreman_proxy::plugin::dns_route53 | aws_access_key |
--- a/src/roles/foreman_proxy/tasks/feature/dns.yaml
+++ b/src/roles/foreman_proxy/tasks/feature/dns.yaml
@@ -1,0 +1,7 @@
+---
+- name: Include tasks for {{ item }}
+  ansible.builtin.include_tasks: feature.yaml
+  vars:
+    feature_enabled: "{{ foreman_proxy_dns_provider_name == item }}"
+    feature_name: "{{ item }}"
+  loop: "{{ foreman_proxy_dns_provider_list }}"
--- a/src/roles/foreman_proxy/tasks/feature/dns_nsupdate.yaml
+++ b/src/roles/foreman_proxy/tasks/feature/dns_nsupdate.yaml
@@ -1,0 +1,22 @@
+---
+- name: Create RNDC Key podman secret
+  containers.podman.podman_secret:
+    state: present
+    name: foreman_proxy-dns_nsupdate-rndc_key
+    data: "{{ lookup('ansible.builtin.file', foreman_proxy_dns_nsupdate_keyfile) }}"
+  notify:
+    - Restart Foreman Proxy
+    - Refresh Foreman Proxy
+
+- name: Mount RNDC key
+  ansible.builtin.copy:
+    dest: /etc/containers/systemd/foreman-proxy.container.d/dns_nsupdate-rndc_key.conf
+    content: |
+      [Container]
+      Secret=foreman_proxy-dns_nsupdate-rndc_key,type=mount,target=/etc/foreman-proxy/rndc.key
+    mode: '0644'
+    owner: root
+    group: root
+  notify:
+    - Restart Foreman Proxy
+    - Refresh Foreman Proxy
--- a/tests/foreman_proxy_test.py
+++ b/tests/foreman_proxy_test.py
@@ -27,6 +27,10 @@ def test_foreman_proxy_features(curl_request, proxy_base_url, enabled_features):
         assert "bmc" in features
     else:
         assert "bmc" not in features
+    if 'dns' in enabled_features:
+        assert "dns" in features
+    else:
+        assert "dns" not in features
 
 
 def test_foreman_proxy_service(server):
@@ -65,3 +69,56 @@ def test_bmc_capabilities(proxy_v2_features):
 def test_bmc_default_provider(proxy_v2_features):
     settings = proxy_v2_features['bmc'].get('settings', {})
     assert settings.get('bmc_default_provider') == 'ipmitool'
+
+
+@pytest.mark.feature('dns')
+def test_dns_provider(proxy_v2_features):
+    settings = proxy_v2_features['dns'].get('settings', {})
+    assert settings.get('use_provider') == 'dns_nsupdate'
+
+
+@pytest.mark.feature('dns')
+@pytest.mark.parametrize('dns_name,dns_value,dns_type', [
+    ('test-v4.example.test', '192.0.2.100', 'A'),
+    ('test-v6.example.test', '2001:db8::1', 'AAAA'),
+    ('test-cname.example.test', 'test.example.test', 'CNAME'),
+    ('reverse-v4.example.test', '100.2.0.192.in-addr.arpa', 'PTR'),
+    ('reverse-v6.example.test', '0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa', 'PTR')
+])
+def test_dns_nsupdate(server, certificates, server_fqdn, dns_name, dns_value, dns_type):
+
+    def dns_cmd(curl_opts, extra_url=""):
+        return server.run(
+            f"curl --cacert {certificates['server_ca_certificate']} "
+            f"--cert {certificates['client_certificate']} "
+            f"--key {certificates['client_key']} "
+            f"--output /dev/null --write-out '%{{http_code}}' "
+            f"{curl_opts} "
+            f"https://{server_fqdn}:8443/dns/{extra_url}"
+        )
+
+    if dns_type == 'PTR':
+        expected = f"{dns_name}."
+        query = dns_value
+    elif dns_type == 'CNAME':
+        expected = f"{dns_value}."
+        query = dns_name
+    else:
+        expected = dns_value
+        query = dns_name
+
+    create_cmd = dns_cmd(f"--data 'fqdn={dns_name}&value={dns_value}&type={dns_type}'")
+    assert create_cmd.succeeded
+    assert create_cmd.stdout == '200'
+
+    verify_cmd = server.run(f"dig +short {query} {dns_type} @named.example.com")
+    assert verify_cmd.succeeded
+    assert verify_cmd.stdout.strip() == expected
+
+    delete_cmd = dns_cmd("-X DELETE", f"{query}/{dns_type}")
+    assert delete_cmd.succeeded
+    assert delete_cmd.stdout == '200'
+
+    verify_cmd = server.run(f"dig +short {query} {dns_type} @named.example.com")
+    assert verify_cmd.succeeded
+    assert verify_cmd.stdout.strip() == ""
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -104,17 +104,20 @@ jobs:
         run: ./setup-environment
       - name: Start VMs
         run: |
-          ./forge vms start --vms "quadlet client ${{ matrix.database == 'external' && 'database' || '' }}"
+          ./forge vms start --vms "quadlet client named ${{ matrix.database == 'external' && 'database' || '' }}"
       - name: Downgrade openssl
         if: matrix.box == 'centos/stream9'
         run: |
-          for vm in quadlet client ${{ matrix.database == 'external' && 'database' || '' }}; do
+          for vm in quadlet client named ${{ matrix.database == 'external' && 'database' || '' }}; do
             vagrant ssh "$vm" -- sudo dnf downgrade -y openssl openssl-libs
           done
       - name: Configure remote-database
         if: matrix.database == 'external'
         run: |
           ./forge remote-database
+      - name: Configure remote-named
+        run: |
+          ./forge remote-named
       - name: Configure repositories
         run: |
           ./forge setup-repositories
@@ -158,6 +161,8 @@ jobs:
             --add-feature google \
             --add-feature remote-execution \
             --add-feature bmc \
+            --add-feature dns \
+            --dns-provider nsupdate --dns-server named.example.com --dns-nsupdate-keyfile $(pwd)/.var/lib/foremanctl/rndc.key \
             ${{ matrix.iop == 'enabled' && '--add-feature iop' || '' }}
       - name: Run tests
         run: |
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -58,6 +58,15 @@ Vagrant.configure("2") do |config|
     end
   end
 
+  config.vm.define "named" do |override|
+    override.vm.box = "centos/stream9"
+    override.vm.hostname = "named.#{DOMAIN}"
+
+    override.vm.provider "libvirt" do |libvirt, provider|
+      libvirt.memory = 2048
+    end
+  end
+
   # Load user-local box definitions from boxes.yaml (gitignored)
   boxes_yaml = File.join(__dir__, 'boxes.yaml')
   if File.exist?(boxes_yaml)
--- a/development/playbooks/remote-named/metadata.obsah.yaml
+++ b/development/playbooks/remote-named/metadata.obsah.yaml
@@ -1,0 +1,2 @@
+---
+help: Setup a machine with BIND
--- a/development/playbooks/remote-named/remote-named.yaml
+++ b/development/playbooks/remote-named/remote-named.yaml
@@ -1,0 +1,13 @@
+---
+- name: Setup remote named
+  hosts:
+    - named
+  become: true
+  roles:
+    - role: bind
+  tasks:
+    - name: Fetch rndc key
+      ansible.builtin.fetch:
+        src: "/etc/named/rndc.key"
+        dest: "{{ obsah_state_path }}/rndc.key"
+        flat: true
--- a/development/playbooks/test/test.yaml
+++ b/development/playbooks/test/test.yaml
@@ -8,6 +8,7 @@
       ansible.builtin.package:
         name:
           - nmap
+          - bind-utils
 
 - name: Execute tests
   gather_facts: false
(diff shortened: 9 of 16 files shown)


ehelms:
Let's see how smart my review tool is getting, I ran it over this draft as it's a good playground for it:

  Ansible Role — copy module src path

  - development/roles/bind/tasks/main.yml:10 — src: files/rndc.key is wrong. Inside a role, copy resolves src relative to the role's files/ directory, so Ansible will look for
  <role>/files/files/rndc.key, which doesn't exist. Same bug on line 18 with src: files/example.test.db. Both should drop the files/ prefix: src: rndc.key and src: example.test.db.

  Ansible Role — service not enabled

  - development/roles/bind/tasks/main.yml:48 — The Start BIND task starts the service but doesn't set enabled: true. After a VM reboot the service won't come back up, which will break any
  test run that reboots the named VM.

  Missing obsah metadata

  - development/playbooks/remote-named/ — No metadata.obsah.yaml. Without it, forge (via obsah) won't discover the remote-named subcommand, so it can't be invoked.

evgeni:
> Let's see how smart my review tool is getting, I ran it over this draft as it's a good playground for it:
> 
> Ansible Role — copy module src path
> 
>     * development/roles/bind/tasks/main.yml:10 — src: files/rndc.key is wrong. Inside a role, copy resolves src relative to the role's files/ directory, so Ansible will look for
>       /files/files/rndc.key, which doesn't exist. Same bug on line 18 with src: files/example.test.db. Both should drop the files/ prefix: src: rndc.key and src: example.test.db.

And yet it works… https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbook_pathing.html#id3 is more complicated than "it will look in `files/`"

> Ansible Role — service not enabled
> 
>     * development/roles/bind/tasks/main.yml:48 — The Start BIND task starts the service but doesn't set enabled: true. After a VM reboot the service won't come back up, which will break any
>       test run that reboots the named VM.

Technically correct, but nothing reboots the VM.

> Missing obsah metadata
> 
>     * development/playbooks/remote-named/ — No metadata.obsah.yaml. Without it, forge (via obsah) won't discover the remote-named subcommand, so it can't be invoked.

And yet it works… ;)

(Yes, those are all valid points, but their criticality and impact is way off)

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,12 +56,28 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_dns_provider:
+    parameter: --dns-provider
+    help: DNS provider to use.
+    choices:
+      - nsupdate
```
evgeni: The provider names are really `dns_<name>`, so `dns_nsupdate`, but it felt rather cumbersome to me to make the user add the (imho redundant) prefix here. Instead whenever `foreman_proxy_dns_provider` is used, it needs to be prefixed with `dns_`

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,12 +56,28 @@ variables:
...
+    parameter: --dns-server
+    help: DNS server to be managed
+    type: FQDN
 
 constraints:
   required_together:
     - [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]
   forbidden_if:
     - [certificates_source, installer, [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]]
+  # FIXME: this only should trigger if the dns feature is on
```
evgeni: obsah can't do `required_if` on a feature, but it should?

src/roles/foreman_proxy/tasks/feature/dns.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Include tasks for {{ foreman_proxy_dns_provider }}
+  ansible.builtin.include_tasks: feature.yaml
```
evgeni: is this too much of a "clever" hack? a feature (dns) enabling a "sub" feature (dns_nsupdate)?

tests/foreman_proxy_test.py
```diff
@@ -69,3 +73,31 @@ def test_bmc_capabilities(proxy_v2_features):
...
+
+
+@pytest.mark.feature('dns')
+def test_dns_provider(proxy_v2_features):
+    settings = proxy_v2_features['dns'].get('settings', {})
+    assert settings.get('use_provider') == 'dns_nsupdate'
+
+
+@pytest.mark.feature('dns')
+def test_dns_nsupdate(server, certificates, server_fqdn):
```
ekohl: This effectively implements a subset of https://github.com/theforeman/forklift/blob/master/bats/fb-proxy-dns.bats. It's a good start but I'd love to see it expanded in the future

tests/foreman_proxy_test.py
```diff
@@ -69,3 +73,31 @@ def test_bmc_capabilities(proxy_v2_features):
...
+    create_cmd = server.run(
+        f"curl --cacert {certificates['server_ca_certificate']} "
+        f"--cert {certificates['client_certificate']} "
+        f"--key {certificates['client_key']} "
+        f"--output /dev/null --write-out '%{{http_code}}' "
+        f"--data 'fqdn={dns_name}&value={dns_value}&type={dns_type}' "
+        f"https://{server_fqdn}:8443/dns/"
+    )
+    assert create_cmd.succeeded
+    assert create_cmd.stdout == '200'
```
ekohl, line 112: Nit: I always try to minimize the number of asserts. Perhaps combine them?
```suggestion
    assert create_cmd.succeeded and create_cmd.stdout == '200'
```

src/playbooks/deploy/metadata.obsah.yaml
ehelms: Agreed.

src/roles/foreman_proxy/tasks/feature/dns.yaml
ekohl: In https://github.com/theforeman/puppet-foreman_proxy I made a distinction of [modules](https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/module.pp) and [providers](https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/provider.pp).

But what you're talking about here also has precedent: https://github.com/theforeman/puppet-foreman_proxy/blob/5e4fcc4200ec540cbe033e7258aecc60083554d8/manifests/config.pp#L19-L32. As the person who wrote that and is happy with it: not too clever in my book.

tests/foreman_proxy_test.py
evgeni: Won't that break pytest output? When it compares strings, it shows a diff, but now it's just a Boolean?

ekohl:
```python
from dataclasses import dataclass


@dataclass
class Output:
    succeeded: bool
    stdout: str


def test_output():
    create_cmd = Output(True, '404')
    assert create_cmd.succeeded and create_cmd.stdout == '200'
```

Then running it:
```console
$ pytest test_assert.py
============================= test session starts ==============================
platform linux -- Python 3.14.3, pytest-8.3.5, pluggy-1.6.0
rootdir: /home/ekohl
plugins: anyio-4.12.1
collected 1 item                                                               

test_assert.py F                                                         [100%]

=================================== FAILURES ===================================
_________________________________ test_output __________________________________

    def test_output():
        create_cmd = Output(True, '404')
>       assert create_cmd.succeeded and create_cmd.stdout == '200'
E       AssertionError: assert (True and '404' == '200'
E        +  where True = Output(succeeded=True, stdout='404').succeeded
E         
E         - 200
E         + 404)

test_assert.py:12: AssertionError
=========================== short test summary info ============================
FAILED test_assert.py::test_output - AssertionError: assert (True and '404' == '200'
============================== 1 failed in 0.05s ===============================
```

I'll leave it up to you to judge if this is clear enough.

tests/foreman_proxy_test.py
evgeni: You're saying we should test all supported DNS types and also their deletion etc? Can do, sure.

BTW, you don't happen to have an Python implementation of Foreman's ProxyAPI implementation at hand? ;)

tests/foreman_proxy_test.py
evgeni: Mhh. I think I still dislike it, but my concern is gone ;)

What do others think?

tests/foreman_proxy_test.py
ekohl: > You're saying we should test all supported DNS types and also their deletion etc? Can do, sure.

I think that would be great. And for PTR both IPv4 and IPv6.

> BTW, you don't happen to have an Python implementation of Foreman's ProxyAPI implementation at hand? ;)

Sorry, I don't.

evgeni: `git init proxyapy`

evgeni: Anyway, added all but PTR (was too lazy to add a .arpa domain to the setup)

src/playbooks/deploy/metadata.obsah.yaml
evgeni: https://github.com/theforeman/obsah/pull/119 WDYT

tests/foreman_proxy_test.py
archanaserver: The combined assertion seems fine from a pytest output perspective i believe, and so I don't think that concern applies here anymore. That said, I'd probably lean towards keeping the separate assertions in this case, looking at the code, `create_cmd.succeeded` and `create_cmd.stdout == '200'` are two different checks and separating makes it easier to pinpoint what failed, that's from my thought.

ekohl: > create_cmd.succeeded and create_cmd.stdout == '200' are two different checks and separating makes it easier to pinpoint what failed, that's from my thought.

This is where I see it as a single check: you assert that the command did exactly what you expected.

I did a bit more digging and in https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test has a very good discussion. In particular I liked https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test/171502#171502
> Tests should fail for one reason only, but that doesn't always mean that there should be only one Assert statement.

Following that logic, here what you want is the right information. If the command fails, do you have enough information to debug that? If you modify that and rerun the test, does it then pass?

And to be fair, that's also more about unit tests than integration tests. So with that in mind, I'm fine with 2 asserts here.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,13 +56,27 @@ variables:
...
+    type: FQDN
 
 constraints:
   required_together:
     - [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]
   forbidden_if:
     - [certificates_source, installer, [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]]
-
+  required_in_list:
+    - [[[features, dns], [foreman_proxy_dns_provider, nsupdate]], [foreman_proxy_dns_keyfile, foreman_proxy_dns_server]]
```
evgeni: this uses https://github.com/theforeman/obsah/pull/119, but I am not 100% satisfied with it. partially because this only works when the user passes both `--add-feature dns` and `--dns-provider nsupdate` explicitly. Relying on the default dns provider being `nsupdate` doesn't work, as then the "required" check doesn't trigger.

docs/user/parameters.md
```diff
@@ -133,11 +137,7 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-proxy-dhcp-server` | | foreman_proxy | dhcp_server |
 | `--foreman-proxy-libvirt-network` | | foreman_proxy | libvirt_network |
 | `--foreman-proxy-libvirt-url` | | foreman_proxy | libvirt_url |
-| `--foreman-proxy-dns` | | foreman_proxy | dns |
 | `--foreman-proxy-dns-managed` | | foreman_proxy | dns_managed |
```
ekohl: I think it's safe to say we won't implement this. Perhaps we need to introduce a third category besides mapped and unmapped: dropped.

tests/foreman_proxy_test.py
```diff
@@ -69,3 +73,53 @@ def test_bmc_capabilities(proxy_v2_features):
...
+            f"{curl_opts} "
+            f"https://{server_fqdn}:8443/dns/{extra_url}"
+        )
+
+    if dns_type == 'PTR':
+        expected = dns_name
+        query = dns_value
+    else:
+        expected = dns_value
+        query = dns_name
```
ekohl, line 108: This reminds me of https://projects.theforeman.org/issues/18398. Actually the inspiration for me to expose capabilities in the registration protocol: signify the Smart Proxy supports a DNS v2 API. That never happened, but we have used the registration protocol v2.

development/roles/bind/files/rndc.key
```diff
@@ -0,0 +1,4 @@
+key "rndc-key" {
+	algorithm hmac-sha256;
+	secret "+C5T2CKyChhLUQDpkOG5Z7PIGmnjFP+EPTvHBMUrh48=";
+};
```
shubhamsg199: Why are we pre-creating it? Can't we generate it on the run like in the foreman-installer ?

evgeni: We can, but given this is only the "devel" part I was rather on the lazy side :)

evgeni: implemented :)

src/roles/foreman_proxy/tasks/feature/dns.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Include tasks for {{ foreman_proxy_dns_provider_name }}
+  ansible.builtin.include_tasks: feature.yaml
+  vars:
+    feature_enabled: "true"
```
ekohl: How about making this value dependent on whether the provider is selected? Then you can list every provider and include disable steps. Not a blocker for this PR, but thinking about the pattern here.

ekohl (commented):
Not a blocker for this, but I wonder about downstream vs upstream supported providers. What if it differs? Perhaps a question for https://community.theforeman.org/t/container-based-downstreams/46492.

src/roles/foreman_proxy/tasks/feature/dns.yaml
evgeni: At first I was like "wth is he talking about", but then :bulb: came.

You mean something like:
```yaml
- name: Include tasks for {{ item }}
  ansible.builtin.include_tasks: feature.yaml
  vars:
    feature_enabled: "{{ foreman_proxy_dns_provider_name == item }}"
  loop:
    - every
    - possible
    - dns
    - provider
```

ekohl: I didn't even think about a loop, but that's even nicer.

evgeni:
> Not a blocker for this, but I wonder about downstream vs upstream supported providers. What if it differs? Perhaps a question for https://community.theforeman.org/t/container-based-downstreams/46492.

My naïve reply would be:
- `foreman_proxy` role defines a `foreman_proxy_supported_dns_providers` list in its `defaults`
- any downstream can override this by providing a `downstream.yaml` vars file that is included in all playbooks

(no, this is not implemented right now, but would be possible without too much of a headache)

Edit: you'd also have to override the params definition to skip the unsupported ones. Which we can't do today. Ugh.

docs/user/parameters.md
```diff
@@ -111,6 +111,16 @@ There are multiple use cases from the users perspective that dictate what parame
 | --------- | ----------- | ---------------------------- |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
+| `--add-feature dns` | Enable DNS feauture | `--foreman-proxy-dns` |
+| `--dns-provider` | Select DNS provider | `--foreman-proxy-dns-provider` |
+| `--foreman-proxy-dns-server` | Configure the target DNS server | `--foreman-proxy-dns-server` |
+| `--foreman-proxy-keyfile` | Confogure the key for the target DNS server | `--foreman-proxy-keyfile` |
```
shubhamsg199:
```suggestion
| `--dns-server` | Configure the target DNS server | `--foreman-proxy-dns-server` |
| `--dns-keyfile` | Configure the key for the target DNS server | `--foreman-proxy-keyfile` |
```

src/playbooks/deploy/metadata.obsah.yaml
shubhamsg199: That means only `foremanctl deploy --add-feature dns` would fail?

shubhamsg199: So I tested `foremanctl deploy --add-feature foreman-proxy --add-feature dns` it works `--dns-provider nsupdate`  is selected by default.

But it fails with
```
The task includes an option with an undefined variable. The error was: 'foreman_proxy_dns_server' is undefined. 'foreman_proxy_dns_server' is undefined. 'foreman_proxy_dns_server' is undefined. 'foreman_proxy_dns_server' is undefined
```
Which is expected since I didn't pass `--dns-server`. But i think this validation should happen earlier not in the later stage. What do you think?

Also, why there are 4 undefined statements in the error?

evgeni: > That means only `foremanctl deploy --add-feature dns` would fail?

No.

`--add-feature dns --dns-provider nsupdate` will fail validation with "dns_server is required" and "dns_keyfile" is required. But if you don't pass `--dns-provider` it will fail only later during actual execution as you've observed it:

> ```
> The task includes an option with an undefined variable. The error was: 'foreman_proxy_dns_server' is undefined. 'foreman_proxy_dns_server' is undefined. 'foreman_proxy_dns_server' is undefined. 'foreman_proxy_dns_server' is undefined
> ```
> 
> Which is expected since I didn't pass `--dns-server`. But i think this validation should happen earlier not in the later stage. What do you think?

Should it? Yes. But right now I have no good idea how to implement that :(

docs/user/parameters.md
```diff
@@ -111,6 +111,16 @@ There are multiple use cases from the users perspective that dictate what parame
 | --------- | ----------- | ---------------------------- |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
+| `--add-feature dns` | Enable DNS feauture | `--foreman-proxy-dns` |
```
ekohl: It surprises me we don't have a similar line for `bmc`.

src/roles/foreman_proxy/tasks/feature/dns_nsupdate.yaml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Create RNDC Key podman secret
+  containers.podman.podman_secret:
+    state: present
```
ekohl, line 4: Taking it a step further, this should probably look at the enabled state too. Ideally we would clean up items (secrets, systemd drop-ins).

src/roles/foreman_proxy/tasks/feature/dns.yaml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Include tasks for {{ item }}
+  ansible.builtin.include_tasks: feature.yaml
+  vars:
+    feature_enabled: "{{ foreman_proxy_dns_provider_name == item }}"
```
ekohl, line 5: Should this also look at the dns feature being enabled?

src/roles/foreman_proxy/tasks/feature/dns_nsupdate.yaml
evgeni: Right now this file is not included when the feature is disabled, but we certainly can extend this.

docs/user/parameters.md
evgeni: bc7cacab3ba23327aed2925eb89e78117b051981

src/playbooks/deploy/metadata.obsah.yaml
evgeni: Okay, after bit of talking with @ekohl, we ended up declaring `dns_provider` a required parameter (without a default value, as the `nsupdate` default wouldn't work out of the box anyway), which solves this particular issue.

evgeni:
@coderabbitai review

evgeni:
@coderabbitai review

docs/user/parameters.md
```diff
@@ -112,6 +112,16 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--add-feature bmc` | Enable BMC feature | `--foreman-proxy-bmc` |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
+| `--add-feature dns` | Enable DNS feature | `--foreman-proxy-dns` |
+| `--dns-provider` | Select DNS provider | `--foreman-proxy-dns-provider` |
+| `--dns-server` | Configure the target DNS server | `--foreman-proxy-dns-server` |
+| `--dns-keyfile` | Configure the key for the target DNS server | `--foreman-proxy-keyfile` |
```
ekohl: `keyfile` is specific to nsupdate and doesn't apply to other providers, such as Infoblox. I always thought that was confusing in our old installer. Perhaps a good chance to fix that?

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,13 +56,28 @@ variables:
...
+    help: DNS provider to use.
+    choices:
+      - nsupdate
+  foreman_proxy_dns_keyfile:
+    parameter: --dns-keyfile
+    help: Key file to be used to authenticate with the DNS server
+    type: AbsolutePath
+  foreman_proxy_dns_server:
+    parameter: --dns-server
+    help: DNS server to be managed
```
ekohl: I wonder how descriptive we can be. I think this helps:
```suggestion
    help: Hostname of the DNS server to be managed
```

I'm worried about `managed` implying something more because in our old installer we fully managed the server. Now we just manage the zone contents.

evgeni: I first wanted to do "to send nsupdate requests to", but infoblox uses the same setting for its server, so ended up with "to communicate with".

src/roles/foreman_proxy/tasks/feature/dns.yaml
evgeni, line 5: Coderabbit (in a nitpick) pointed out that this is now a boolean, while the original intent of the variable was to map the whole spectrum of `true/false/http/https` that foreman proxy supports and with a boolean the `!= 'false'` will be wrong. Sounds like a very valid concern to me, will think about it.

(Mostly commenting so it's not lost in the rabbit hole, hah)

ehelms:
This needs a rebase

PR #533, opened 2026-06-02
TFTP feature

stejskalleos:
Support the TFTP Smart Proxy feature.

* `foremanctl deploy ---add-feature tftp`

```shell
./foremanctl deploy --help

  --tftp-root FOREMAN_PROXY_TFTP_ROOT
                        Directory to serve TFTP files from. (persisted)
  --tftp-servername FOREMAN_PROXY_TFTP_SERVERNAME
                        Server name to use for TFTP. (persisted)
```

```shell
./foremanctl deploy --add-feature foreman-proxy --add-feature tftp \
  --tftp-root /var/lib/foreman-custom-tftp \
  --tftp-servername something-else.example.com \
```
Go to https://quadlet.example.com/smart_proxies/2-quadlet-example-com, TFTP should be enabled.

```
vagrant ssh quadlet
sudo podman exec -it foreman-proxy /bin/bash
```

```shell
cat /etc/foreman-proxy/settings.d/tftp.yml
```

```yaml
---
:enabled: false
:tftproot: /var/lib/foreman-custom-tftp
:tftp_servername: something-else.example.com
:tftp_connect_timeout: 10
:verify_server_cert: true
```

```
./foremanctl deploy --remove-feature tftp
./foremanctl features
=>

tftp                      available    Enable TFTP feature on Foreman Proxy
```

`cat /etc/foreman-proxy/settings.d/tftp.yml`
```yaml
---
:enabled: false
:tftproot: /var/lib/foreman-custom-tftp
:tftp_servername: something-else.example.com
:tftp_connect_timeout: 10
:verify_server_cert: true
```

Go to https://quadlet.example.com/smart_proxies/2-quadlet-example-com, TFTP should be disabled.

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -120,6 +120,9 @@ jobs:
         if: matrix.security != 'none'
         run: |
           ./forge security --mode ${{ matrix.security }}
+      - name: Setup TFTP server
+        run: |
+          ./forge tftp
       - name: Apply fapolicyd workarounds
         # https://access.redhat.com/solutions/7072618 / https://issues.redhat.com/browse/RHEL-37912
         # https://github.com/theforeman/foreman-fapolicyd/blob/develop/15-foreman-container.rules
@@ -152,6 +155,8 @@ jobs:
             --add-feature google \
             --add-feature remote-execution \
             --add-feature bmc \
+            --add-feature tftp \
+            --tftp-servername 127.0.0.1 \
             ${{ matrix.iop == 'enabled' && '--add-feature iop' || '' }}
       - name: Run tests
         run: |
--- a/development/playbooks/tftp/metadata.obsah.yaml
+++ b/development/playbooks/tftp/metadata.obsah.yaml
@@ -1,0 +1,3 @@
+---
+help: |
+  Setup TFTP server
--- a/development/playbooks/tftp/tftp.yaml
+++ b/development/playbooks/tftp/tftp.yaml
@@ -1,0 +1,21 @@
+---
+- name: TFTP
+  hosts:
+    - quadlet
+  become: true
+  tasks:
+    - name: Install packages
+      ansible.builtin.package:
+        name:
+          - tftp-server
+          - tftp
+        state: present
+
+    - name: Start TFTP services
+      ansible.builtin.systemd:
+        name: "{{ item }}"
+        state: started
+        enabled: true
+      with_items:
+        - tftp.socket
+        - tftp.service
--- a/development/playbooks/tftp/tftp.yaml
+++ b/development/playbooks/tftp/tftp.yaml
@@ -17,5 +17,4 @@
         state: started
         enabled: true
       with_items:
-        - tftp.socket
         - tftp.service
--- a/docs/user/parameters.md
+++ b/docs/user/parameters.md
@@ -112,6 +112,15 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--add-feature bmc` | Enable BMC feature | `--foreman-proxy-bmc` |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
+| `--add-feature tftp` | Enable TFTP feature | `--foreman-proxy-tftp` |
+| `--tftp-servername` | IP address of the TFTP server (required when enabling TFTP) | `--foreman-proxy-tftp-servername` |
+| `--tftp-root` | Directory to serve TFTP files from | `--foreman-proxy-tftp-root` |
+
+### Unmapped
+
+| foreman-installer Parameter | Description | Reason |
+| --------------------------- | ----------- | ------ |
+| `--foreman-proxy-tftp-managed` | Installer managed TFTP | not supported |
 
 ### Undetermined
 
@@ -144,10 +153,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-proxy-plugin-dns-route53-aws-access-key` | | foreman_proxy::plugin::dns_route53 | aws_access_key |
 | `--foreman-proxy-plugin-dns-route53-aws-secret-key` | | foreman_proxy::plugin::dns_route53 | aws_secret_key |
 | `--foreman-proxy-httpboot` | | foreman_proxy | httpboot |
-| `--foreman-proxy-tftp` | | foreman_proxy | tftp |
-| `--foreman-proxy-tftp-servername` | | foreman_proxy | tftp_servername |
-| `--foreman-proxy-tftp-managed` | | foreman_proxy | tftp_managed |
-| `--foreman-proxy-tftp-root` | | foreman_proxy | tftp_root |
 | `--foreman-proxy-plugin-dhcp-remote-isc-dhcp-config` | | foreman_proxy::plugin::dhcp_remote_isc | config |
 | `--foreman-proxy-plugin-dhcp-remote-isc-dhcp-leases` | | foreman_proxy::plugin::dhcp_remote_isc | dhcp_config |
 | `--foreman-proxy-plugin-dhcp-remote-isc-key-name` | | foreman_proxy::plugin::dhcp_remote_isc | key_name |
--- a/src/playbooks/deploy/metadata.obsah.yaml
+++ b/src/playbooks/deploy/metadata.obsah.yaml
@@ -56,10 +56,21 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_tftp_root:
+    parameter: --tftp-root
+    help: Directory to serve TFTP files from.
+    type: AbsolutePath
+  foreman_proxy_tftp_servername:
+    parameter: --tftp-servername
+    help: Server name or IP address for TFTP. TFTP clients typically do not have DNS available, so an IP address is recommended.
 
 constraints:
   required_together:
     - [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]
+  forbidden_if:
+    - [certificates_source, installer, [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]]
+  required_in_list:
+    - [[[features, tftp]], [foreman_proxy_tftp_servername]]
 
 include:
   - _certificate_source
--- a/src/playbooks/deploy/metadata.obsah.yaml
+++ b/src/playbooks/deploy/metadata.obsah.yaml
@@ -58,7 +58,7 @@ variables:
     type: Boolean
   foreman_proxy_tftp_root:
     parameter: --tftp-root
-    help: Directory to serve TFTP files from.
+    help: Directory to serve TFTP files from, defaults to /var/lib/tftpboot if not specified.
     type: AbsolutePath
   foreman_proxy_tftp_servername:
     parameter: --tftp-servername
--- a/src/roles/foreman_proxy/defaults/main.yaml
+++ b/src/roles/foreman_proxy/defaults/main.yaml
@@ -22,3 +22,7 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
 # BMC settings
 foreman_proxy_bmc_ipmi_implementation: ipmitool
 foreman_proxy_bmc_redfish_verify_ssl: true
+
+# TFTP settings
+foreman_proxy_tftp_root: /var/lib/tftpboot
+foreman_proxy_tftp_servername: "{{ undef(hint='You must specify a TFTP servername') }}"
--- a/src/roles/foreman_proxy/tasks/feature/tftp.yaml
+++ b/src/roles/foreman_proxy/tasks/feature/tftp.yaml
@@ -1,0 +1,21 @@
+---
+- name: TFTP root directory
+  notify:
+    - Restart Foreman Proxy
+    - Refresh Foreman Proxy
+  block:
+    - name: Create TFTP root directory
+      ansible.builtin.file:
+        path: "{{ foreman_proxy_tftp_root }}"
+        state: directory
+        mode: "0777"
+
+    - name: Mount the directory into foreman-proxy container
+      ansible.builtin.copy:
+        dest: /etc/containers/systemd/foreman-proxy.container.d/tftp-root.conf
+        content: |
+          [Container]
+          Volume={{ foreman_proxy_tftp_root }}:/var/lib/tftpboot:rw
+        mode: '0644'
+        owner: root
+        group: root
--- a/src/roles/foreman_proxy/tasks/feature/tftp.yaml
+++ b/src/roles/foreman_proxy/tasks/feature/tftp.yaml
@@ -8,7 +8,9 @@
       ansible.builtin.file:
         path: "{{ foreman_proxy_tftp_root }}"
         state: directory
-        mode: "0777"
+        mode: "0755"
+        owner: 998
+        group: 998
 
     - name: Mount the directory into foreman-proxy container
       ansible.builtin.copy:
--- a/src/roles/foreman_proxy/tasks/main.yaml
+++ b/src/roles/foreman_proxy/tasks/main.yaml
@@ -16,6 +16,8 @@
     sdnotify: true
     network: host
     hostname: "{{ ansible_facts['hostname'] }}.local"
+    security_opt:
+      - "label=disable"
     secrets:
       - 'foreman-proxy-settings-yml,type=mount,target=/etc/foreman-proxy/settings.yml'
       - 'foreman-proxy-ssl-ca,type=mount,target=/etc/foreman-proxy/ssl_ca.pem'
--- a/src/roles/foreman_proxy/templates/settings.d/tftp.yml.j2
+++ b/src/roles/foreman_proxy/templates/settings.d/tftp.yml.j2
@@ -1,0 +1,4 @@
+---
+:enabled: {{ feature_enabled }}
+:tftproot: /var/lib/tftpboot
+:tftp_servername: {{ foreman_proxy_tftp_servername }}
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,7 @@ from requests.adapters import HTTPAdapter
 SSH_CONFIG = './.tmp/ssh-config'
 OBSAH_STATE = os.environ.get('OBSAH_STATE', '.var/lib/foremanctl')
 PARAMETERS_FILE = os.path.join(OBSAH_STATE, 'parameters.yaml')
+FOREMAN_PROXY_PORT = 8443
 
 
 class UserParameters:
@@ -269,6 +270,26 @@ class ResolveAdapter(HTTPAdapter):
         conn.host = self.target_ip
 
         return conn
+
+
+@pytest.fixture(scope="module")
+def proxy_request(server, certificates, server_fqdn):
+    def _request(path, method=None, data=None, return_body=False):
+        curl_opts = (
+            f"--cacert {certificates['server_ca_certificate']} "
+            f"--cert {certificates['client_certificate']} "
+            f"--key {certificates['client_key']} "
+            f"--silent "
+        )
+        if not return_body:
+            curl_opts += "--output /dev/null --write-out '%{http_code}' "
+        if method:
+            curl_opts += f"-X {method} "
+        if data:
+            curl_opts += f"-d '{data}' "
+        return server.run(f"curl {curl_opts}https://{server_fqdn}:{FOREMAN_PROXY_PORT}/{path}")
+
+    return _request
 
 
 @pytest.fixture(scope="module")
--- a/tests/foreman_proxy_test.py
+++ b/tests/foreman_proxy_test.py
@@ -3,23 +3,18 @@ import json
 
 import pytest
 
-FOREMAN_PROXY_PORT = 8443
+from tests.conftest import FOREMAN_PROXY_PORT
 
 
 @pytest.fixture(scope="module")
-def proxy_v2_features(server, certificates, server_fqdn):
-    cmd = server.run(
-        f"curl --cacert {certificates['server_ca_certificate']} "
-        f"--cert {certificates['client_certificate']} "
-        f"--key {certificates['client_key']} "
-        f"--silent https://{server_fqdn}:{FOREMAN_PROXY_PORT}/v2/features"
-    )
+def proxy_v2_features(proxy_request):
+    cmd = proxy_request("v2/features", return_body=True)
     assert cmd.succeeded, f"Failed to query /v2/features: {cmd.stderr}"
     return json.loads(cmd.stdout)
 
 
-def test_foreman_proxy_features(server, certificates, server_fqdn, enabled_features):
-    cmd = server.run(f"curl --cacert {certificates['server_ca_certificate']} --silent https://{server_fqdn}:{FOREMAN_PROXY_PORT}/features")
+def test_foreman_proxy_features(proxy_request, enabled_features):
+    cmd = proxy_request("features", return_body=True)
     assert cmd.succeeded
     features = json.loads(cmd.stdout)
     assert "logs" in features
@@ -29,6 +24,10 @@ def test_foreman_proxy_features(server, certificates, server_fqdn, enabled_featu
         assert "bmc" in features
     else:
         assert "bmc" not in features
+    if 'tftp' in enabled_features:
+        assert "tftp" in features
+    else:
+        assert "tftp" not in features
 
 
 def test_foreman_proxy_service(server):
@@ -54,6 +53,18 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
     )
     assert cmd.succeeded
     assert cmd.stdout == '201'
+
+
+@pytest.mark.feature('tftp')
+def test_tftp_write_and_fetch(proxy_request, server):
+    test_mac = "aa:bb:cc:dd:ee:ff"
+    cmd = proxy_request(f"tftp/{test_mac}", data="syslinux_config=foremanctl+test+probe")
+    assert cmd.succeeded
+    assert cmd.stdout == '200', f"Expected HTTP 200 when creating TFTP PXE config, got {cmd.stdout}"
+
+    cmd = server.run("tftp 127.0.0.1 -c get pxelinux.cfg/01-aa-bb-cc-dd-ee-ff /tmp/foremanctl_tftp_test_download")
+    assert cmd.succeeded, f"TFTP get failed: {cmd.stdout}"
+    assert server.file("/tmp/foremanctl_tftp_test_download").content_string == "foremanctl test probe"
 
 
 @pytest.mark.feature('bmc')
--- a/src/features.yaml
+++ b/src/features.yaml
@@ -58,3 +58,7 @@ bmc:
   description: Power management for bare metal hosts (IPMI, Redfish)
   foreman_proxy:
     plugin_name: bmc
+tftp:
+  description: Enable TFTP feature on Foreman Proxy
+  foreman_proxy:
+    plugin_name: tftp
--- a/src/plugins/modules/migrate_answers.py
+++ b/src/plugins/modules/migrate_answers.py
@@ -32,6 +32,11 @@ PARAMETER_MAP = {
     ('foreman', 'server_ssl_key'): 'IGNORE',
     ('foreman', 'server_ssl_ca'): 'IGNORE',
 
+    # Foreman Proxy TFTP
+    ('foreman_proxy', 'tftp_managed'): 'IGNORE',
+    ('foreman_proxy', 'tftp_root'): 'foreman_proxy_tftp_root',
+    ('foreman_proxy', 'tftp_servername'): 'foreman_proxy_tftp_servername',
+
     # TODO: Add more mappings as discovered
 }
 


docs/user/parameters.md
```diff
@@ -111,6 +111,11 @@ There are multiple use cases from the users perspective that dictate what parame
 | --------- | ----------- | ---------------------------- |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
+| `--add-feature tftp` | Enable TFTP feature | `--foreman-proxy-tftp` |
+| `--tftp-servername` | | `--foreman-proxy-tftp-servername` |
+| `--tftp-root` | Directory for TFTP | `--foreman-proxy-tftp-root` |
```
stejskalleos: I just realized that I forgot to update the quadlet to mount this directory!
Will add in a follow-up commit.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,20 @@ variables:
...
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_tftp_root:
+    parameter: --tftp-root
+    help: Directory to serve TFTP files from.
+    type: AbsolutePath
+  foreman_proxy_tftp_servername:
+    parameter: --tftp-servername
+    help: Server name to use for TFTP.
+  foreman_proxy_tftp_connect_timeout:
```
stejskalleos: foreman_proxy_tftp_connect_timeout & foreman_proxy_tftp_verify_server_cert cannot be set in the foreman-installer, see:

```shell
satellite-installer --full-help | grep tftp
    --foreman-proxy-tftp                                                                 Enable TFTP feature (current: true)
    --reset-foreman-proxy-tftp                                                           Reset tftp to the default value (false)
    --foreman-proxy-tftp-dirs                                                            Directories to be create in $tftp_root (current: UNDEF)
    --reset-foreman-proxy-tftp-dirs                                                      Reset tftp_dirs to the default value (UNDEF)
    --foreman-proxy-tftp-listen-on                                                       TFTP proxy to listen on https, http, or both (current: "https")
    --reset-foreman-proxy-tftp-listen-on                                                 Reset tftp_listen_on to the default value ("https")
    --foreman-proxy-tftp-manage-wget                                                     If enabled will install the wget package (current: true)
    --reset-foreman-proxy-tftp-manage-wget                                               Reset tftp_manage_wget to the default value (true)
    --foreman-proxy-tftp-managed                                                         The TFTP daemon is managed by this module. (current: true)
    --reset-foreman-proxy-tftp-managed                                                   Reset tftp_managed to the default value (true)
    --foreman-proxy-tftp-replace-grub2-cfg                                               Determines if grub2.cfg will be replaced (current: false)
    --reset-foreman-proxy-tftp-replace-grub2-cfg                                         Reset tftp_replace_grub2_cfg to the default value (false)
    --foreman-proxy-tftp-root                                                            TFTP root directory (current: "/var/lib/tftpboot")
    --reset-foreman-proxy-tftp-root                                                      Reset tftp_root to the default value ("/var/lib/tftpboot")
    --foreman-proxy-tftp-servername                                                      Defines the TFTP Servername to use, overrides the name in the subnet declaration (current: UNDEF)
    --reset-foreman-proxy-tftp-servername                                                Reset tftp_servername to the default value (UNDEF)
    --foreman-proxy-plugin-discovery-tftp-root                                           TFTP root directory where extracted discovery image will be installed (current: "/var/lib/tftpboot")
    --reset-foreman-proxy-plugin-discovery-tftp-root                                     Reset tftp_root to the default value ("/var/lib/tftpboot")
```

I don't know if we want to introduce new params, or we can just keep them as they were by default.

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -23,3 +23,9 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
 # BMC settings
 foreman_proxy_bmc_ipmi_implementation: ipmitool
 foreman_proxy_bmc_redfish_verify_ssl: true
+
+# TFTP settings
+foreman_proxy_tftp_root: /var/lib/tftpboot
+foreman_proxy_tftp_servername: "{{ ansible_facts['fqdn'] }}"
```
ekohl: Does this need a default? IIRC the code has a fallback on the Smart Proxy's hostname already so this doesn't add anything. In a remote setup it will also be guaranteed wrong.

src/playbooks/deploy/metadata.obsah.yaml
ekohl: IIRC the module itself depends on `wget`, but now that I look at https://github.com/theforeman/smart-proxy/commit/3d87c6feaa8caaf9b8c040f96061c3b8c33a39ef then we switched to curl a long time ago. IMHO you should make sure the container contains the command.

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -23,3 +23,9 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
...
+foreman_proxy_tftp_connect_timeout: 10
+foreman_proxy_tftp_verify_server_cert: false
```
evgeni: The default in https://github.com/theforeman/smart-proxy/blob/3e249dd019dd36f41b546fe44c045ab055c1da6c/modules/tftp/tftp_plugin.rb#L11 is `true`, and we didn't expose that parameter at all in https://github.com/theforeman/puppet-foreman_proxy/blob/master/templates/tftp.yml.erb -- why changing it to `false` here?

evgeni (changes requested):
Can we get a test that sets up a `tftp-server` (on the quadlet machine), instructs the proxy to create a file in the tftp root and uses `tftp` to fetch that file?

src/playbooks/deploy/metadata.obsah.yaml
evgeni: IMHO we should not add new parameters, just allow for feature parity.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,20 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_tftp_root:
+    parameter: --tftp-root
+    help: Directory to serve TFTP files from.
```
evgeni: Ewouds comment in [jira](https://redhat.atlassian.net/browse/SAT-39234) made me realize that this description is wrong.
We also use `/var/lib/tftpboot` for the `httpboot` module (https://github.com/theforeman/smart-proxy/blob/3e249dd019dd36f41b546fe44c045ab055c1da6c/modules/httpboot/httpboot_plugin.rb#L13) and do not allow users to configure that in Puppet.

It is  unclear to me how the HTTPBoot thing obtains the files (seems there is no code for that), so maybe it relies on the TFTP module to place the files and just exposes them over HTTP instead of TFTP?

(This is not blocking the feature here, just food for further thought going forward)

arvind4501: I also faced similar doubts while just scaffolding httpboot feature, and i realised its use `/var/lib/tftp` which is not created by httpboot, which makes me think why we don't add tftp as dependency for httpboot. for full context you can refer https://github.com/theforeman/foremanctl/pull/518#issuecomment-4562107756

src/roles/foreman_proxy/defaults/main.yaml
lzap: Hey, this is completely random (I was just browsing this repository) but:

TFTP clients usually do not have DNS stack available, therefore, this must be IP address if anything. I see `fqdn` so I assume this would be a hostname. I remember there was some FQDN>IP hack somewhere but using IP was always the best and the safest thing to do.

This has always been terrible name, `servername`. What it does is sets up `next-server` which is probably the root cause of all of this terminology. But what users typically need to set is an IP address.

ekohl: You're right, that is a better value. Ideally this value defaults to empty and we force users to provide the value when they enable the TFTP feature. Then input validation forces a valid IPv4 address.

src/playbooks/deploy/metadata.obsah.yaml
ekohl: I'd go even further and say that we don't need feature parity. We won't manage a TFTP server (removes `--foreman-proxy-tftp-managed ` and we shouldn't manage `grub2.cfg` (removes `--foreman-proxy-tftp-replace-grub2-cfg`). And I was already hinting at not needing `--foreman-proxy-tftp-manage-wget`.

evgeni: Sure, we don't need implement things we don't need :)

src/playbooks/deploy/metadata.obsah.yaml
ekohl: > which makes me think why we don't add tftp as dependency for httpboot.

You may find https://github.com/theforeman/smart-proxy/commit/db33bc6780e28b5d694a1be6642e637aff59936f interesting.

Perhaps we should call it the `netboot` directory? I already had https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/tftp/netboot.pp in the old installer.

evgeni: "interesting" as in "wait, what!?"? ;-)

How is the directory populated, when the TFTP feature is off, but HTTPBoot is on?

ekohl: Magic? In https://github.com/theforeman/puppet-foreman_proxy/commit/e4b7763e45fecc5fbaa5d88cc066906d6caf7bf7 I wrote:
> To get the netboot files, the TFTP feature must still be enabled.

It was a long discussion that was rather tedious. I'm not sure the whole use case of netbooting with HTTPBoot without TFTP was ever really well supported.

arvind4501: > How is the directory populated, when the TFTP feature is off, but HTTPBoot is on?

AFAIK, no population happens, just configuration, As of today, by default httpboot enablement will add :root_dir: /var/lib/tftpboot in config, and still be indirectly dependent on tftp to create and populate that directory

src/playbooks/deploy/metadata.obsah.yaml
stejskalleos: I kept `servername` and `tftproot` , the rest of the parameters is removed.

stejskalleos:
> Can we get a test that sets up a tftp-server (on the quadlet machine), instructs the proxy to create a file in the tftp root and uses tftp to fetch that file?

I'm still working on this test; I will add it in the upcoming commit.

src/roles/foreman_proxy/defaults/main.yaml
stejskalleos: We could at least rename the parameter in foremactl.
Change it to `tftp-ip`  and update the description.
WDYT?

.github/workflows/test.yml
```diff
@@ -159,6 +159,8 @@ jobs:
             --add-feature google \
             --add-feature remote-execution \
             --add-feature bmc \
+            --add-feature tftp \
+            --tftp-servername quadlet.example.com \
```
ehelms: Are we setting a bad example here based on -- https://github.com/theforeman/foremanctl/pull/533/changes#diff-0e3f528c6f4a769e4f3ce63a81c6cdaae734ddea4f6efac9430bd38b92bd5e4cR9

src/roles/foreman_proxy/tasks/feature/tftp.yaml
```diff
@@ -0,0 +1,33 @@
...
+- name: Create TFTP root directory
+  ansible.builtin.file:
+    path: "{{ foreman_proxy_tftp_root }}"
+    state: directory
+    mode: '0750'
+    owner: root
+    group: root
+  notify:
+    - Restart Foreman Proxy
+    - Refresh Foreman Proxy
```
ehelms, line 5: I would think the next task should be sufficient for this, or the two actions should be put into a `block`.

src/roles/foreman_proxy/tasks/feature/tftp.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: Mount TFTP root directory
+  ansible.builtin.copy:
+    dest: /etc/containers/systemd/foreman-proxy.container.d/tftp-root.conf
+    content: |
+      [Container]
+      Volume={{ foreman_proxy_tftp_root }}:{{ foreman_proxy_tftp_root }}:Z
```
ehelms: See https://github.com/theforeman/foremanctl/pull/375 -- that could help get ahead of the game by applying it here.

src/roles/foreman_proxy/tasks/feature/tftp.yaml
```diff
@@ -0,0 +1,33 @@
...
+    path: "{{ foreman_proxy_tftp_root }}"
+    state: directory
+    mode: '0750'
+    owner: root
+    group: root
+  notify:
+    - Restart Foreman Proxy
+    - Refresh Foreman Proxy
+
+- name: Mount TFTP root directory
```
ehelms: This action technically needs a daemon-reload as well I think.

evgeni: Yes, but that is not implemented as a handler. We always do one once are features are configured:
https://github.com/theforeman/foremanctl/blob/c0019f246c6efd4277f9545e347be25ce5e72b0a/src/roles/foreman_proxy/tasks/main.yaml#L59-L61

stejskalleos:
@ehelms applied comments from your review.
Can't add the tests for the feature because of a broken local environment.
I'm trying to figure it out in the meantime.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,13 @@ variables:
...
+    type: AbsolutePath
+  foreman_proxy_tftp_servername:
+    parameter: --tftp-servername
+    help: Server name to use for TFTP.
```
ehelms: Given the validation message below, I'd include this line here in some form:

>       This should be the IP address of the TFTP server (TFTP clients typically do not have DNS available).

ehelms: Which I will also point out is slightly ironic we call it servername and then state it should be the IP address.

ekohl: Technically we're talking about https://www.rfc-editor.org/info/rfc2132/#section-9.4. DHCP option 66 is called TFTP server name. Sadly it doesn't describe at all what the content of it should be at the protocol level.

If we look at what we do for ISC DHCP (https://github.com/theforeman/smart-proxy/blob/3e249dd019dd36f41b546fe44c045ab055c1da6c/modules/dhcp_common/isc/omapi_provider.rb#L205-L217) then that's basically trying to send an IP, but ISC DHCP itself can also be configured with a hostname where the service does a lookup.

Kea's `all-options.json` actually has an example that is a hostname: https://github.com/isc-projects/kea/blob/f79874a1f22c29b96ee543f5bc905960818991f7/doc/examples/kea4/all-options.json#L787-L798.

So I'm not sure we need to be so strict in only accepting an IP. Apologies for not digging into that deep enough before.

tests/foreman_proxy_test.py
```diff
@@ -56,6 +60,37 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
     assert cmd.stdout == '201'
 
 
+@pytest.mark.feature('tftp')
```
stejskalleos, line 58: Leftover, will remove

development/playbooks/tftp/tftp.yaml
```diff
@@ -0,0 +1,33 @@
...
+      ansible.builtin.file:
+        path: /var/lib/tftpboot
+        state: directory
+        mode: '0755'
+        owner: root
+        group: root
+
+    - name: Change file SELinux security context
+      ansible.builtin.command:
+        cmd: chcon -t tftpdir_t /var/lib/tftpboot
```
evgeni: You can set the type using `setype` of the `file` module directly: https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/file_module.html#parameter-setype

evgeni: But I wonder what's wrong with `system_u:object_r:tftpdir_rw_t:s0` that this folder should get by default.

src/roles/foreman_proxy/tasks/feature/tftp.yaml
```diff
@@ -0,0 +1,32 @@
+---
+- name: Validate TFTP servername is provided
```
evgeni: This should be validated by `obsah` on the parameter level, not deep in the Ansible execution. See https://github.com/theforeman/foremanctl/pull/532 where I used https://github.com/theforeman/obsah/pull/119 to achieve that.

tests/foreman_proxy_test.py
```diff
@@ -56,6 +60,32 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
...
+        f"curl --cacert {certificates['server_ca_certificate']} "
+        f"--cert {certificates['client_certificate']} "
+        f"--key {certificates['client_key']} "
+        f"--silent --output /dev/null --write-out '%{{http_code}}' "
+        f"-d 'syslinux_config=foremanctl+test+probe' "
+        f"https://{server_fqdn}:{FOREMAN_PROXY_PORT}/tftp/{test_mac}"
+    )
+    assert cmd.succeeded
+    assert cmd.stdout == '200', f"Expected HTTP 200 when creating TFTP PXE config, got {cmd.stdout}"
+
```
evgeni, line 64: there is no validation that the file is actually served via tftp correctly

src/roles/foreman_proxy/templates/settings.d/tftp.yml.j2
```diff
@@ -0,0 +1,6 @@
+---
+:enabled: {{ feature_enabled }}
+:tftproot: {{ foreman_proxy_tftp_root }}
```
evgeni: You mount the directory statically at `/var/lib/tftpboot` but use the (dynamic) host path here. This won't work if the user passes anything else than `/var/lib/tftpboot` to `foremanctl`. You can use the static path instead.

```suggestion
:tftproot: /var/lib/tftpboot
```

tests/foreman_proxy_test.py
```diff
@@ -56,6 +60,32 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
     assert cmd.stdout == '201'
 
 
+@pytest.mark.feature('tftp')
+def test_tftp_write(server, certificates, server_fqdn):
+    test_mac = "aa:bb:cc:dd:ee:ff"
+    cmd = server.run(
```
evgeni: not blocking, but seeing I've written very similar invocations in #532, I think we should create a "do an HTTP request against the Proxy API using cert-auth" helper and use it (similar how we have the `foremanapi` fixture to talk to the Foreman API).

can be solved in a follow up, just noting my thoughts

evgeni:
@coderabbitai review

development/playbooks/tftp/metadata.obsah.yaml
```diff
@@ -0,0 +1,3 @@
+---
```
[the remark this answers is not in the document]
evgeni: In which world is this a playbook?

src/roles/foreman_proxy/handlers/main.yml
```diff
@@ -11,3 +11,8 @@
     oauth1_consumer_key: "{{ foreman_proxy_oauth_consumer_key }}"
     oauth1_consumer_secret: "{{ foreman_proxy_oauth_consumer_secret }}"
     ca_path: "{{ foreman_proxy_foreman_ca_certificate | default(omit) }}"
+
+- name: Change TFTP root directory SELinux security context
```
evgeni: Why?

src/roles/foreman_proxy/tasks/feature/tftp.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - Restart Foreman Proxy
+    - Refresh Foreman Proxy
+  block:
+    - name: Create TFTP root directory
+      ansible.builtin.file:
+        path: "{{ foreman_proxy_tftp_root }}"
+        state: directory
+        mode: "0776"
+
+    - name: Restore SELinux security context
```
evgeni: Who is breaking it?

tests/foreman_proxy_test.py
```diff
@@ -56,6 +55,33 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
...
+
+
+@pytest.mark.feature('tftp')
+def test_tftp_service(server):
+    tftp = server.service("tftp")
+    assert tftp.is_running
+
+
+@pytest.mark.feature('tftp')
+def test_tftp_upload_and_download(server):
```
evgeni: This doesn't test that the file written by the proxy could be loaded?

tests/foreman_proxy_test.py
```diff
@@ -56,6 +55,33 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
...
+    assert cmd.stdout == '200', f"Expected HTTP 200 when creating TFTP PXE config, got {cmd.stdout}"
+
+    cmd = proxy_request(f"tftp/{test_mac}", method="DELETE")
+    assert cmd.succeeded
+    assert cmd.stdout == '200', f"Expected HTTP 200 when deleting TFTP PXE config, got {cmd.stdout}"
+
+
+@pytest.mark.feature('tftp')
+def test_tftp_service(server):
+    tftp = server.service("tftp")
```
evgeni: Given this is not set up by foremanctl, I don't think we should be testing it (directly).

tests/conftest.py
```diff
@@ -268,6 +268,28 @@ def get_connection_with_tls_context(self, request, verify, proxies=None, cert=No
         return conn
 
 
+@pytest.fixture(scope="module")
+def proxy_request(server, certificates, server_fqdn):
```
evgeni, line 276: Cool!

src/roles/foreman_proxy/templates/settings.d/tftp.yml.j2
```diff
@@ -0,0 +1,6 @@
+---
+:enabled: {{ feature_enabled }}
+:tftproot: /var/lib/tftpboot
+:tftp_servername: {{ foreman_proxy_tftp_servername }}
+:tftp_connect_timeout: 10
+:verify_server_cert: true
```
ekohl: If these are hardcoded, what use are they? They match the defaults: https://github.com/theforeman/smart-proxy/blob/3e249dd019dd36f41b546fe44c045ab055c1da6c/modules/tftp/tftp_plugin.rb#L10-L11

tests/conftest.py
```diff
@@ -268,6 +268,28 @@ def get_connection_with_tls_context(self, request, verify, proxies=None, cert=No
...
+    port = 8443
```
ekohl: We have a constant already:
https://github.com/theforeman/foremanctl/blob/d61b189145a35bb7b5c47cdd1e108a6462a36362/tests/foreman_proxy_test.py#L6
Is there a way to align these? Perhaps a fixture of its own?

tests/foreman_proxy_test.py
```diff
@@ -56,6 +55,22 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
...
+def test_tftp_write(proxy_request):
+    test_mac = "aa:bb:cc:dd:ee:ff"
+    cmd = proxy_request(f"tftp/{test_mac}", data="syslinux_config=foremanctl+test+probe")
+    assert cmd.succeeded
+    assert cmd.stdout == '200', f"Expected HTTP 200 when creating TFTP PXE config, got {cmd.stdout}"
+
+
+@pytest.mark.feature('tftp')
+def test_tftp_upload_and_download(server):
+    cmd = server.run("tftp 127.0.0.1 -c get pxelinux.cfg/01-aa-bb-cc-dd-ee-ff /tmp/foremanctl_tftp_test_download")
```
ekohl, line 65: This test depends on the previous test to write it. I think it's better not to rely on that.

tests/foreman_proxy_test.py
```diff
@@ -56,6 +55,22 @@ def test_foreman_proxy_client_auth_to_foreman(server, certificates, server_fqdn)
...
+    assert cmd.succeeded, f"TFTP get failed: {cmd.stdout}"
+    cmd = server.run("cat /tmp/foremanctl_tftp_test_download")
+    assert cmd.stdout.strip() == "foremanctl test probe"
```
ekohl: Perhaps use the existing helper?
```suggestion
    assert server.file("/tmp/foremanctl_tftp_test_download").content_string == "foremanctl test probe"
```

docs/user/parameters.md
```diff
@@ -144,10 +147,7 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-proxy-plugin-dns-route53-aws-access-key` | | foreman_proxy::plugin::dns_route53 | aws_access_key |
 | `--foreman-proxy-plugin-dns-route53-aws-secret-key` | | foreman_proxy::plugin::dns_route53 | aws_secret_key |
 | `--foreman-proxy-httpboot` | | foreman_proxy | httpboot |
-| `--foreman-proxy-tftp` | | foreman_proxy | tftp |
-| `--foreman-proxy-tftp-servername` | | foreman_proxy | tftp_servername |
 | `--foreman-proxy-tftp-managed` | | foreman_proxy | tftp_managed |
```
ekohl: In https://github.com/theforeman/foremanctl/pull/532 there's also an `Unmapped` category for things we don't support.

src/roles/foreman_proxy/templates/settings.d/tftp.yml.j2
stejskalleos: Sorry, I don't understand. What's the concern here?

ekohl: You might as well remove them, right?
```suggestion
```

stejskalleos: :+1:

tests/conftest.py
```diff
@@ -268,6 +269,28 @@ def get_connection_with_tls_context(self, request, verify, proxies=None, cert=No
...
+            f"--key {certificates['client_key']} "
+            f"--silent "
+        )
+        if not return_body:
+            curl_opts += "--output /dev/null --write-out '%{http_code}' "
+        if method:
+            curl_opts += f"-X {method} "
+        if data:
+            curl_opts += f"-d '{data}' "
+        return server.run(f"curl {curl_opts}https://{server_fqdn}:{port}/{path}")
```
evgeni:
```suggestion
        return server.run(f"curl {curl_opts}https://{server_fqdn}:{FOREMAN_PROXY_PORT}/{path}")
```

then you don't need to do `port = FOREMAN_PROXY_PORT` above

src/roles/foreman_proxy/templates/settings.d/tftp.yml.j2
```diff
@@ -0,0 +1,4 @@
+---
+:enabled: {{ feature_enabled }}
+:tftproot: /var/lib/tftpboot
+:tftp_servername: {{ foreman_proxy_tftp_servername }}
```
evgeni, line 4: this fails to render (due to the `undef`) when the `foreman-proxy` feature is present, but `tftp` is not. that's a bug in how we render the templates and I propose https://github.com/theforeman/foremanctl/pull/566 to fix it (also ran into this when doing #532 )

evgeni (changes requested):

src/roles/foreman_proxy/templates/settings.d/tftp.yml.j2
evgeni: this has been merged, so a rebase should make CI green

evgeni (approved):

development/playbooks/tftp/tftp.yaml
```diff
@@ -0,0 +1,21 @@
...
+        state: present
+
+    - name: Start TFTP services
+      ansible.builtin.systemd:
+        name: "{{ item }}"
+        state: started
+        enabled: true
+      with_items:
+        - tftp.socket
+        - tftp.service
```
Gauravtalreja1: systemd config for `tftp.service` has `Requires=tftp.socket`, so when we just start `tftp.service` both will be started, right? so do we really need this loop?

evgeni:
@stejskalleos would you mind rebasing this one? there is a conflict :/

evgeni:
And another rebase is in order.

@Gauravtalreja1 you wanted to have a look here before we merge? Or can we merge to save Leos from rebasing as a service?

evgeni:
(rebased as leos is busy)

src/roles/foreman_proxy/tasks/feature/tftp.yaml
```diff
@@ -0,0 +1,21 @@
...
+- name: TFTP root directory
+  notify:
+    - Restart Foreman Proxy
+    - Refresh Foreman Proxy
+  block:
+    - name: Create TFTP root directory
+      ansible.builtin.file:
+        path: "{{ foreman_proxy_tftp_root }}"
+        state: directory
+        mode: "0777"
```
Gauravtalreja1: Isn't this too much open permissions for a dir? I checked previously we had 755, is this change intended?

src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -16,6 +16,8 @@
     sdnotify: true
     network: host
     hostname: "{{ ansible_facts['hostname'] }}.local"
+    security_opt:
+      - "label=disable"
```
Gauravtalreja1, line 20: IIUC, `label=disable` effectively disables SELinux for the container. Could you explain the rationale behind this change and the underlying issue it is meant to resolve?

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,10 +56,21 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_tftp_root:
+    parameter: --tftp-root
+    help: Directory to serve TFTP files from.
```
Gauravtalreja1: Should we be explicit and mention what we'll use default if not specified?

And should we add a validation to check if its valid directory and not file or something else? I tried passing a file I manually created for --tftp-root and it failed at very later stage, so such validation would help when failed earlier
```suggestion
    help: Directory to serve TFTP files from, defaults to /var/lib/tftpboot if not specified.
```

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,10 +56,21 @@ variables:
...
+    parameter: --tftp-servername
+    help: Server name or IP address for TFTP. TFTP clients typically do not have DNS available, so an IP address is recommended.
 
 constraints:
   required_together:
     - [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]
+  forbidden_if:
+    - [certificates_source, installer, [certificates_custom_server_certificate, certificates_custom_server_key, certificates_custom_server_ca_certificate]]
+  required_in_list:
+    - [[[features, tftp]], [foreman_proxy_tftp_servername]]
```
Gauravtalreja1, line 73: Should we add a foreman-proxy feature as a requirement here as well?

I tried to enable tftp without foreman-proxy and it was enabled without any errors, but I know ideally it wouldn't work, so should we add a better validation here to check if tftp is present in required_in_list or do we want to enable foreman-proxy feature as dependancy to TFTP?

docs/user/parameters.md
```diff
@@ -112,6 +112,15 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--add-feature bmc` | Enable BMC feature | `--foreman-proxy-bmc` |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
+| `--add-feature tftp` | Enable TFTP feature | `--foreman-proxy-tftp` |
+| `--tftp-servername` | IP address of the TFTP server (required when enabling TFTP) | `--foreman-proxy-tftp-servername` |
+| `--tftp-root` | Directory to serve TFTP files from | `--foreman-proxy-tftp-root` |
```
Gauravtalreja1, line 117: Optional: Should we update `PARAMETER_MAP` too for this new options, so we're better positioned for the migration when we take up the upgrade work in future?

https://github.com/theforeman/foremanctl/blob/master/src/plugins/modules/migrate_answers.py#L14

tests/conftest.py
```diff
@@ -271,6 +272,26 @@ def get_connection_with_tls_context(self, request, verify, proxies=None, cert=No
...
+            f"--key {certificates['client_key']} "
+            f"--silent "
+        )
+        if not return_body:
+            curl_opts += "--output /dev/null --write-out '%{http_code}' "
+        if method:
+            curl_opts += f"-X {method} "
+        if data:
+            curl_opts += f"-d '{data}' "
+        return server.run(f"curl {curl_opts}https://{server_fqdn}:{FOREMAN_PROXY_PORT}/{path}")
```
Gauravtalreja1, line 290: I tried this manually, and whenever I invoke this API, it creates the directory structure on the Foreman side rather than on the TFTP server.

Additionally, when I run the `tftp` command from the test you added, it fails with a `File not found` error because the expected files are not present on the TFTP server.

Could you elaborate on the expected workflow here and how this test is supposed to work? Am I missing a step that synchronizes or publishes the generated content to the TFTP server?

src/roles/foreman_proxy/tasks/feature/tftp.yaml
stejskalleos: It doesn't work with `755`. You need to have all the permissions for the quadlet service.

The fix might be to create a new dedicated user or user group and run the container with it, but I'm not sure if it is the correct approach.

src/roles/foreman_proxy/tasks/main.yaml
stejskalleos: That's the only way to make the container read, write, and list data in the directory.

tests/conftest.py
stejskalleos: There is nothing else needed; just replicate the test steps, and it should work. The CI is passing, so it's working.

development/playbooks/tftp/tftp.yaml
stejskalleos: > so do we really need this loop?

TBH I don't know, but if it is like you say, we can remove it.

src/playbooks/deploy/metadata.obsah.yaml
Gauravtalreja1: @stejskalleos I see help text is updated as per this, but could you take a look at the validation part for tfpt-root option?

development/playbooks/tftp/tftp.yaml
ekohl: Technically it's a bit redundant. In puppet-tftp we used to only activate `tftp.socket`. Later we added `tftp.service` because otherwise it wasn't restarted properly. For a development server either will probably suffice. Doesn't really matter which one.

src/playbooks/deploy/metadata.obsah.yaml
stejskalleos: @ekohl What is the correct behavior in this scenario:

User runs `foremanctl deploy --add-feature tftp` without the `--add-feature foreman-proxy` (in the current run or in the previous one).

What's the expected behavior?
a. Foremanctl should fail with an error about the required feature
b. Foremanctl should automatically add the smart proxy feature

src/playbooks/deploy/metadata.obsah.yaml
stejskalleos: > And should we add a validation to check if its valid directory and not file or something else? I tried passing a file I manually created for --tftp-root and it failed at very later stage, so such validation would help when failed earlier

I mean, we could, but it is kinda the user's fault for passing a file instead of a directory, and the overall effort to do it early just isn't worth it IMO. We have validation, and thats important IMO.

src/playbooks/deploy/metadata.obsah.yaml
ekohl: I don't think `tftp` should silently pull in `foreman-proxy`. Adding a whole separate service is something that must be done explicitly IMHO because you may also need to open up firewall ports. So I'd lean more to an error that a required feature is not enabled.

This will also apply to all other Smart Proxy features, like DNS and DHCP. I lean to tracking that as a follow up task because I doubt we fixed in other places.

docs/user/parameters.md
ekohl: Good call, also applies to other places. We should do that on every change to the file to keep them in sync.

tests/conftest.py
ekohl: The core design is that it will be created on the Foreman Proxy side (which may also run a Foreman). It's expected that the admin makes sure the TFTP server also uses the same directory. For example, using NFS as done in https://docs.theforeman.org/3.19/Integrating_Provisioning_Infrastructure_Services/index-foreman-el.html#integrating-a-generic-tftp-server.

Ideally we'd have an accompanying docs PR that explains the end-to-end workflow.

src/roles/foreman_proxy/tasks/feature/tftp.yaml
ekohl: I think mode `0777` is not acceptable from a security perspective. If it doesn't work, we need to find another solution.

Note the `smart-proxy` process runs as user `foreman-proxy` (https://github.com/theforeman/foreman-oci-images/blob/776ba097225f0308d34eb48f6b4ccdb8a63b72c5/images/foreman-proxy/Containerfile#L17) so that user will need to have write permission.

We may need to guarantee the same UID is used within the container and the host, or on the remote server.

In short, agreed with @Gauravtalreja1 this needs to be addressed before we merge it.

src/playbooks/deploy/metadata.obsah.yaml
stejskalleos: Created https://redhat.atlassian.net/browse/SAT-46972

src/roles/foreman_proxy/tasks/feature/tftp.yaml
stejskalleos: Updated with `0755`, the owner is now foreman-proxy user (`998`)

stejskalleos:
I don't think failed CI is related to my changes:
```
TASK [theforeman.forklift.sos_report : Find sosreport file] ********************
Error: : Task failed: Failed to create temporary directory. In some cases, you may have been able to authenticate and did not have permissions on the target directory. Consider changing the remote tmp path in ansible.cfg to a path rooted in "/tmp", for more error information use -vvv. Failed command was: ( umask 77 && mkdir -p "` echo ~/.ansible/tmp `"&& mkdir "` echo ~/.ansible/tmp/ansible-tmp-1782200501.2489645-21366-92798372804531 `" && echo ansible-tmp-1782200501.2489645-21366-92798372804531="` echo ~/.ansible/tmp/ansible-tmp-1782200501.2489645-21366-92798372804531 `" ), exited with result 1
Origin: /home/runner/work/foremanctl/foremanctl/build/collections/forge/ansible_collections/theforeman/forklift/roles/sos_report/tasks/sosreport_fetch_results.yml:2:3

1 ---
2 - name: 'Find sosreport file'
    ^ column 3

fatal: [quadlet]: UNREACHABLE! => {
    "changed": false,
    "unreachable": true
}

MSG:

Task failed: Failed to create temporary directory. In some cases, you may have been able to authenticate and did not have permissions on the target directory. Consider changing the remote tmp path in ansible.cfg to a path rooted in "/tmp", for more error information use -vvv. Failed command was: ( umask 77 && mkdir -p "` echo ~/.ansible/tmp `"&& mkdir "` echo ~/.ansible/tmp/ansible-tmp-1782200501.2489645-21366-92798372804531 `" && echo ansible-tmp-1782200501.2489645-21366-92798372804531="` echo ~/.ansible/tmp/ansible-tmp-1782200501.2489645-21366-92798372804531 `" ), exited with result 1
```

evgeni:
it's not, I've just reported it to fapolicyd upstream: https://github.com/linux-application-whitelisting/fapolicyd/issues/413

src/roles/foreman_proxy/tasks/feature/tftp.yaml
Gauravtalreja1: @stejskalleos I tried re-running the deploy with `--tftp-root /var/lib/my_tftpboot/` and permissions are updated as `0755` but owner is not set to `foreman-proxy` instead it is `polkitd`
```
# ls -lZ /var/lib/ | grep my_tftpboot
drwxr-xr-x. 4 polkitd        polkitd        unconfined_u:object_r:var_lib_t:s0                45 Jun 23 13:31 my_tftpboot
```

ekohl: The UID on the host and in the container may not be the same. We don't have a reserved UID. This may actually prove to be a problem for NFS mounts and on the next container rebuild that `998` may be `997`.

Can't podman map the user ID?

stejskalleos: My assumption was that the ID would always be the same, which is why the values were hardcoded.
Why would the ID change anyway? What if I hardcode it in the container image?

I'm thinking about the correct solution here. Adding `z`,`Z`, or `U` flags to the volume breaks either smart proxy access or TFTP access, so that's no use.

rahulc07:
Hi can I ask what the status on this PR is?

stejskalleos:
> Hi can I ask what the status on this PR is?

So far, it's been postponed and will be delivered later.
I will reopen it when it is updated.

PR #544, opened 2026-06-08
Fix IOP tests to properly validate service endpoints

ehelms:

The IOP HTTP endpoint tests had several issues that prevented them from actually validating services:
- Tests curled `localhost` for ports not published to the host (only the gateway publishes 24443)
- `|| echo` fallbacks made `rc` always 0, so tests passed whether services were up or down
- `if result.succeeded` guards silently skipped assertions when `podman run` failed due to missing `--network`
- Endpoint paths didn't match the actual health/liveness endpoints exposed by the services

* Add `--network=iop-core-network` to all `podman run` curl commands so container DNS resolves
* Remove all `|| echo` fallbacks and `if result.succeeded` guards so failures surface
* Use correct service endpoints from puppet-iop acceptance specs: `/api/insights/v1/status/live/` for advisor, `/health` for remediations

Steps to reproduce:

* Deploy IOP with `./foremanctl deploy`
* Run IOP tests: `./forge test --pytest_args="tests/iop/"`
* Verify all HTTP endpoint tests actually reach services and assert real status codes

#### Checklist

### DIFF
--- a/tests/iop/conftest.py
+++ b/tests/iop/conftest.py
@@ -5,6 +5,8 @@ ROLE_MAP = {
     "iop-ingress": "iop_ingress",
     "iop-inventory": "iop_inventory",
     "iop-advisor": "iop_advisor",
+    "iop-gateway": "iop_gateway",
+    "iop-remediation": "iop_remediation",
 }
 
 
--- a/tests/iop/test_advisor.py
+++ b/tests/iop/test_advisor.py
@@ -138,5 +138,5 @@ def test_advisor_fdw_permissions_on_view(server):
 
 
 def test_advisor_api_endpoint(server, iop_image):
-    result = server.run(f"podman run --network=iop-core-network --rm {iop_image('iop-advisor')} curl -s -o /dev/null -w '%{{http_code}}' http://iop-service-advisor-backend-api:8000/ 2>/dev/null || echo '000'")
-    assert result.stdout.strip() != "000"
+    result = server.run(f"podman run --network=iop-core-network --rm {iop_image('iop-advisor')} curl --fail -s -o /dev/null http://iop-service-advisor-backend-api:8000/api/insights/v1/status/live/")
+    assert result.succeeded
--- a/tests/iop/test_integration.py
+++ b/tests/iop/test_integration.py
@@ -17,9 +17,9 @@ def test_iop_core_ingress_service(server):
         assert service.is_enabled
 
 
-def test_iop_ingress_endpoint(server):
-    result = server.run("curl -f http://localhost:8080/ 2>/dev/null || echo 'Ingress not yet responding'")
-    assert result.rc == 0
+def test_iop_ingress_endpoint(server, iop_image):
+    result = server.run(f"podman run --network=iop-core-network --rm {iop_image('iop-gateway')} curl --fail -s -o /dev/null http://iop-core-ingress:8080/")
+    assert result.succeeded
 
 
 def test_iop_core_puptoo_service(server):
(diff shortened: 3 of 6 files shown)


tests/iop/conftest.py
```diff
@@ -26,3 +28,5 @@ def _get(name):
         return cache[name]
 
     return _get
+
+
```
evgeni: wat

tests/iop/test_advisor.py
```diff
@@ -138,5 +138,5 @@ def test_advisor_fdw_permissions_on_view(server):
 
 
 def test_advisor_api_endpoint(server, iop_image):
-    result = server.run(f"podman run --network=iop-core-network --rm {iop_image('iop-advisor')} curl -s -o /dev/null -w '%{{http_code}}' http://iop-service-advisor-backend-api:8000/ 2>/dev/null || echo '000'")
-    assert result.stdout.strip() != "000"
+    result = server.run(f"podman run --network=iop-core-network --rm {iop_image('iop-advisor')} curl -s -o /dev/null -w '%{{http_code}}' http://iop-service-advisor-backend-api:8000/api/insights/v1/status/live/")
+    assert result.succeeded and "200" in result.stdout
```
evgeni: So "Copyright 2006 a drunken Sailor" makes this succeed? ;)

More so, you could pass `--fail` to curl and that would automatically assert that the http code was < 400, so you only need to test `result.succeeded` unless you explicitly want a specific http status.

tests/iop/test_integration.py
```diff
@@ -65,18 +65,18 @@ def test_iop_core_gateway_service(server):
 
 
 def test_iop_gateway_endpoint(server):
-    result = server.run("curl -f http://localhost:24443/ 2>/dev/null || echo 'Gateway not yet responding'")
-    assert result.rc == 0
+    result = server.run("curl -s -o /dev/null -w '%{http_code}' http://localhost:24443/")
+    assert result.succeeded and result.stdout.strip() in ("200", "400")
```
evgeni: Why is a bad request OK here?

tests/iop/test_inventory.py
```diff
@@ -27,15 +27,13 @@ def test_inventory_service_dependencies(server):
...
+    result = server.run(f"podman run --network=iop-core-network --rm {iop_image('iop-inventory')} curl -s -o /dev/null -w '%{{http_code}}' http://iop-core-host-inventory-api:8081/health")
+    assert result.succeeded and "200" in result.stdout
 
 
 def test_inventory_hosts_endpoint(server, iop_image):
-    result = server.run(f"podman run --rm {iop_image('iop-inventory')} curl -s -o /dev/null -w '%{{http_code}}' http://iop-core-host-inventory-api:8081/api/inventory/v1/hosts")
-    if result.succeeded:
-        assert "200" in result.stdout
+    result = server.run(f"podman run --network=iop-core-network --rm {iop_image('iop-inventory')} curl -s -o /dev/null -w '%{{http_code}}' http://iop-core-host-inventory-api:8081/api/inventory/v1/hosts")
+    assert result.succeeded and result.stdout.strip() in ("200", "401")
```
evgeni: 401 is not authorized?

tests/iop/conftest.py
ehelms: huh

ehelms:
Comments addressed

evgeni (changes requested):
One last missing piece, otherwise ACK

tests/iop/test_integration.py
```diff
@@ -65,18 +65,18 @@ def test_iop_core_gateway_service(server):
...
-    assert result.rc == 0
+    result = server.run("curl -s -o /dev/null http://localhost:24443/")
+    assert result.succeeded
 
 
 def test_iop_gateway_api_ingress_endpoint(server):
-    result = server.run("curl -f http://localhost:24443/api/ingress 2>/dev/null || echo 'Gateway API ingress not yet responding'")
-    assert result.rc == 0
+    result = server.run("curl -s -o /dev/null http://localhost:24443/api/ingress")
+    assert result.succeeded
```
evgeni, line 69: These two don't use `--fail`, which I assume is intentional, but there is no explanation why and it will be confusing the next time we touch this, can you add a comment for that please?

evgeni (approved):

evgeni:
Migration failures unrelated (see https://github.com/theforeman/obsah/pull/118#issuecomment-4749091661)

Merging

PR #548, opened 2026-06-09
Add molecule tests to certificates role

ShimShtein:
This is the PR that implements https://community.theforeman.org/t/molecule-testing-in-foremanctl/46760.

It is sill draft and more of a POC than actual viable PR at this point.

### DIFF
--- a/.config/molecule/config.yml
+++ b/.config/molecule/config.yml
@@ -1,0 +1,45 @@
+---
+dependency:
+  name: galaxy
+  options:
+    requirements-file: ${MOLECULE_PROJECT_DIRECTORY}/../../requirements.yml
+
+driver:
+  name: podman
+
+platforms:
+  - name: quadlet
+    image: registry.access.redhat.com/ubi9/ubi
+    volumes:
+      - "${MOLECULE_PROJECT_DIRECTORY}/../../../:/vagrant:Z"
+    tmpfs:
+      /tmp: exec
+      /run: rw,noexec,nosuid,nodev
+
+provisioner:
+  name: ansible
+  env:
+    ANSIBLE_COLLECTIONS_PATH: ${MOLECULE_PROJECT_DIRECTORY}/../../../build/collections/foremanctl
+    ANSIBLE_COLLECTIONS_SCAN_SYS_PATH: "false"
+  inventory:
+    group_vars:
+      all:
+        certificates_ca_password: molecule-test-password
+  config_options:
+    defaults:
+      roles_path: ${MOLECULE_PROJECT_DIRECTORY}/..
+
+verifier:
+  name: ansible
+
+scenario:
+  test_sequence:
+    - cleanup
+    - destroy
+    - syntax
+    - create
+    - converge
+    - idempotence
+    - verify
+    - cleanup
+    - destroy
--- a/src/roles/certificates/molecule/common/tasks/verify_custom_server.yml
+++ b/src/roles/certificates/molecule/common/tasks/verify_custom_server.yml
@@ -1,0 +1,111 @@
+---
+- name: Read internal CA certificate subject
+  ansible.builtin.command:
+    cmd: >-
+      openssl x509 -in {{ certificates_ca_directory_certs }}/ca.crt
+      -noout -subject
+  register: certificates_internal_ca_subject
+  changed_when: false
+
+- name: Read custom server CA certificate subject
+  ansible.builtin.command:
+    cmd: >-
+      openssl x509 -in {{ certificates_ca_directory_certs }}/server-ca.crt
+      -noout -subject
+  register: certificates_server_ca_subject
+  changed_when: false
+
+- name: Verify custom server CA differs from internal CA
+  ansible.builtin.assert:
+    that:
+      - certificates_internal_ca_subject.stdout != certificates_server_ca_subject.stdout
+    fail_msg: Custom server CA should have a different subject than the internal CA
+
+- name: Read server certificate issuer
+  ansible.builtin.command:
+    cmd: >-
+      openssl x509 -in {{ certificates_ca_directory_certs }}/{{ certificates_test_hostname }}.crt
+      -noout -issuer
+  register: certificates_server_issuer
+  changed_when: false
+
+- name: Verify server certificate is issued by custom server CA
+  ansible.builtin.assert:
+    that:
+      - certificates_server_issuer.stdout == certificates_server_ca_subject.stdout.replace('subject=', 'issuer=')
+    fail_msg: Server certificate should be issued by the custom server CA
+
+- name: Read client certificate issuer
+  ansible.builtin.command:
+    cmd: >-
+      openssl x509 -in {{ certificates_ca_directory_certs }}/{{ certificates_test_hostname }}-client.crt
+      -noout -issuer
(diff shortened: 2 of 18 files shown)


ehelms:
It would be good to see this PoC hooked into the Github Actions.

src/roles/certificates/molecule/common/tasks/verify_custom_server.yml
```diff
@@ -0,0 +1,111 @@
+---
```
ehelms, line 1: Can tests with  molecule be written in pytest or do they have to be written in Ansible?

tests/molecule_roles/test_roles.py
```diff
@@ -0,0 +1,63 @@
...
+    if {"role", "scenario", "role_dir"} <= set(metafunc.fixturenames):
+        params = discover_molecule_scenarios()
+        metafunc.parametrize(
+            ("role", "scenario", "role_dir"),
+            params,
+            ids=[f"{role}[{scenario}]" for role, scenario, _ in params],
+        )
+
+
+def test_molecule_role(role: str, scenario: str, role_dir: Path) -> None:
```
evgeni, line 44: how does this compare to `pytest-ansible` that already has molecule discovery and testing?

.config/molecule/config.yml
```diff
@@ -0,0 +1,45 @@
...
+    requirements-file: ${MOLECULE_PROJECT_DIRECTORY}/../../requirements.yml
+
+driver:
+  name: podman
+
+platforms:
+  - name: quadlet
+    image: registry.access.redhat.com/ubi9/ubi
+    volumes:
+      - "${MOLECULE_PROJECT_DIRECTORY}/../../../:/vagrant:Z"
```
evgeni, line 14: why is this mounting things at `/vagrant`? this should run in a container, right?

src/roles/certificates/molecule/common/tasks/verify_custom_server.yml
ShimShtein: Currently molecule supports ansible and testinfra verifiers. 
If that's not enough, we can use ansible to run any arbitrary script on the machine and test its result.

tests/molecule_roles/test_roles.py
ShimShtein: This is the integration with pytest-ansible. As you can see, this is a thin wrapper, and pytest-ansible does the heavy lifting of running the full molecule "lifecycle" as part of the test.

.config/molecule/config.yml
ShimShtein: Because I was lazy when switched between drivers. I need the sources mounted into the container to do things like copying the custom certificates: https://github.com/theforeman/foremanctl/pull/548/changes#diff-8e0370b95ab38b1aa4d27ac77915e805f1832f8adba342ef74dda35f550660a1R6.
Since I was lazy to change the folder names, and if I use vagrant as the driver, I have synced the folder structure between the drivers.

tests/molecule_roles/test_roles.py
evgeni: No it is not. It uses `subprocess.run` to call `python -m molecule`.

ShimShtein:
> It would be good to see this PoC hooked into the Github Actions.

@ehelms right now it is a matter of adding the `--molecule` flag to the `pytest` command: https://github.com/theforeman/foremanctl/blob/377474a25564dd91d9baaf5e8f460e08908673ba/.github/workflows/test.yml#L168

I wanted to showcase the method before we hook it up into the CI

tests/molecule_roles/test_roles.py
ShimShtein: I see. I had some weird double requirement error and I let AI to fix it, somehow I have missed the fact that it just calls it in a subprocess. At least theoretically the plugin should do the discovery automagically. I suppose it would be harder to control it, if we want to keep standalone pytests side by side with molecule.

Although in a broader picture, we can do the opposite: let molecule to be the main driver instead of forge. This is how molecule is actually intended to be used. It has the options to spin up a dev machine/container, then you can work on it with `molecule converge` and in the end it will destroy the machine on its own.

ShimShtein: I mean instead of running `./forge test ....` we can call `molecule test` that will do the same thing.

src/roles/certificates/molecule/common/tasks/verify_custom_server.yml
ehelms: testinfra verifiers are pytest based right?

tests/molecule_roles/test_roles.py
ehelms: Using pytest-ansible would allow the test driver to remain pytest right? All `./forge test` is doing is preventing manual work to generate the ssh-config from Vagrant to then run pytest with some basic initial variables. You can still run `pytest` directly if needed. In fact, if switching to pytest or make would be cleaner I have no problem with that. We'd still need something to do that generating of the ssh-config.

ShimShtein: Molecule has all the framework laid out for setup. It has the specific subcommand to create the environment. Under the hood it knows how to work with vagrant and containers (there are also other drivers, but I didn't explore too much in that direction).
The TL;DR version of molecule goes like this:
under the hood `molecule test` executes the following [stages](https://docs.ansible.com/projects/molecule/workflow/#overview):
```
dependency → cleanup → destroy → syntax → create → prepare → converge → idempotence → side_effect → verify → cleanup → destroy
```
each one of he stages has an out of the box playbook, or you can override it with a custom one. For example for `create` they already have support for containers and vagrant (there are more, but I didn't explore in this direction).
I would say that we can replace the whole `forge` concept since it supports also a role dev cycle as create, converge, validate, iterate, destroy.
Anything that `forge` gives us beyond that?

src/roles/certificates/molecule/common/tasks/verify_custom_server.yml
ShimShtein: yes I think so.

ehelms:
```
INFO     Found config file /home/ehelms/workspace/upstream/installer/foremanctl-2/.config/molecule/config.yml
ERROR    Failed to find driver podman. Please ensure that the driver is correctly installed.
```

After I installed the molecule-podman package I get:

```
TASK [Set async_dir for HOME env] **********************************************
[ERROR]: Task failed: Conditional result (True) was derived from value of type 'str' at "<environment variable 'HOME'>". Conditionals must have a boolean result.

Task failed.
Origin: /home/ehelms/workspace/upstream/installer/foremanctl-2/.venv/lib64/python3.14/site-packages/molecule_podman/playbooks/destroy.yml:11:7

 9     podman_exec: "{{ lookup('env','MOLECULE_PODMAN_EXECUTABLE')|default('podman',true) }}"
10   tasks:
11     - name: Set async_dir for HOME env
         ^ column 7

<<< caused by >>>

Conditional result (True) was derived from value of type 'str' at "<environment variable 'HOME'>". Conditionals must have a boolean result.
Origin: /home/ehelms/workspace/upstream/installer/foremanctl-2/.venv/lib64/python3.14/site-packages/molecule_podman/playbooks/destroy.yml:14:13
```

PR #549, opened 2026-06-09
Add foremanctl restore command - Complete offline backup restore

Chyenne8:
# Summary

  Implements `foremanctl restore` to restore Foreman instances from backups created by `foremanctl backup`.

  ## Usage
  
  ```bash
  # Validate backup without making changes
  foremanctl restore /path/to/backup --validate

  # Restore (requires --force on existing deployments)
  foremanctl restore /path/to/backup --force

  What It Does

  1. Validates - Checks backup integrity, disk space, and system requirements
  2. Restores - Databases, Pulp content, encryption keys, OAuth secrets, and passwords
  3. Deploys - Calls foremanctl deploy to configure and start services
  4. Verifies - Confirms services are healthy and API is responding

  Key Features

  - Safety check - Requires --force flag on existing deployments to prevent accidental data loss
  - Automated validation - Checks disk space, database mode compatibility, and file existence
  - Ownership transfer - Properly sets database, table, and sequence ownership for migrations
  - Error recovery - Stops services on failure and provides clear error messages
  - Idempotent - Safe to retry after fixing issues

  Security

  - Verifies encryption keys after restore
  - Restores OAuth secrets and database passwords
  - No secrets logged
  - Backups should be encrypted at rest (contain sensitive data)

  Testing

  # Create backup
  foremanctl backup /var/backup

  # Validate
  foremanctl restore /var/backup/foreman-backup-TIMESTAMP --validate

  # Restore
  foremanctl restore /var/backup/foreman-backup-TIMESTAMP --force

  # Verify
  systemctl status foreman.target
  curl -k https://$(hostname -f)/api/status


src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+- name: Restore databases from dump files
+  ansible.builtin.command:
+    cmd: >
+      pg_restore
+      -h {{ database_host }}
+      -p {{ database_port }}
+      -U {{ item.owner }}
+      -d {{ item.database_name }}
+      --no-owner
```
sjha4: Is the next task required because of the --no-owner restore here?

src/roles/restore/tasks/restore_foremanctl_state.yaml
```diff
@@ -0,0 +1,72 @@
+---
+# Restore foremanctl state (OAuth keys, passwords, parameters)
+# CRITICAL: Must be restored before starting services
+
+- name: Set foremanctl state path
+  ansible.builtin.set_fact:
+    foremanctl_state_path: /root/foremanctl/.var/lib/foremanctl
```
sjha4: This will be different for deployments..Use the obsah_state_path..Something similar to backup does for taking the backup..

src/roles/restore/tasks/main.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - name: Phase 4b - Restore foremanctl state
+      ansible.builtin.include_tasks:
+        file: restore_foremanctl_state.yaml
+
+    - name: Phase 4b complete
+      ansible.builtin.debug:
+        msg: |
+          ═══════════════════════════════════════════════════════════════
+          Phase 4b Complete: OAuth keys and passwords restored!
+          ═══════════════════════════════════════════════════════════════
```
sjha4: I like the debug messages here..The wording about `Phases` is not needed for user facing messages. They can just be name of the task.

src/roles/restore/tasks/deploy_and_verify.yaml
```diff
@@ -0,0 +1,94 @@
...
+
+      All services are running:
+      ✓ foreman.target - active
+      ✓ foreman.service - active
+      ✓ postgresql.service - active
+
+      Foreman API: https://{{ ansible_fqdn }}/api/status - {{ restore_api_status }} ✓
+
+      Your Foreman instance has been successfully restored!
+      ═══════════════════════════════════════════════════════════════
```
sjha4: We probably need a foremanctl deploy in these steps somewhere after the foremanctl state is restored for everything to take effect.

src/roles/restore/tasks/deploy_and_verify.yaml
```diff
@@ -0,0 +1,94 @@
...
+    - foreman.target
+    - foreman.service
+    - postgresql.service
+  changed_when: false
+
+- name: Display restore success
+  ansible.builtin.debug:
+    msg: |
+      ═══════════════════════════════════════════════════════════════
+      ✅ RESTORE COMPLETE!
```
sjha4: We probably want to avoid non-ASCII characters..Also sentence casing..

src/roles/restore/tasks/restore_databases.yaml
Chyenne8: Yes the --no-owner flag tells pg_restore not to restore ownership from the dump file. This was the recommended approach for pg_restore when you want to ensure correct ownership regardless of the backup source. Is there a more efficient way to do this?

src/roles/restore/tasks/restore_foremanctl_state.yaml
Chyenne8: updated to use obsah_state_path instead of a hardcoded path.

src/roles/restore/tasks/main.yaml
Chyenne8: I have removed the "phases" terminology in the messages.

src/roles/restore/tasks/deploy_and_verify.yaml
Chyenne8: The emojis and unnecessary sentence casing have been removed.

src/roles/restore/tasks/deploy_and_verify.yaml
Chyenne8: Added foremanctl deploy and tested it in foremanctl install environment.

src/roles/backup/tasks/main.yaml
```diff
@@ -0,0 +1,226 @@
...
+- name: Remove write test file
+  ansible.builtin.file:
+    path: "{{ backup_dir }}/.write_test"
+    state: absent
+  when: backup_write_test is succeeded
+  changed_when: false
+
+- name: Perform backup operations
+  block:
+    - name: Create timestamped backup directory
```
ianballou, line 37: Should we run the preflight checks before creating the backup directory? That way we don't have empty files left behind.

Edit: let me be more clear - I realize this is a CP from @sjha4 's PR - but a better question would be if this was a purposeful change.

src/playbooks/restore/metadata.obsah.yaml
```diff
@@ -0,0 +1,18 @@
+---
+help: |
+  Restore Foreman from an offline backup
```
ehelms:
```suggestion
  Restore Foreman from a backup
```

src/playbooks/restore/metadata.obsah.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+variables:
+  backup_dir:
+    parameter: backup_dir
+    help: Directory containing the backup files
+    type: AbsolutePath
+    persist: false
+
+  dry_run:
+    help: Validate backup without making any changes
```
ehelms, line 16: Given this, should this maybe be a parameter named `--validate` ?

src/playbooks/restore/restore.yaml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Restore from offline backup
```
ehelms:
```suggestion
- name: Restore from a backup
```

src/roles/restore/tasks/deploy_and_verify.yaml
```diff
@@ -0,0 +1,101 @@
+---
+# Deploy and verify
+# Run foremanctl deploy to regenerate podman secrets from restored credentials
+
+- name: Stop PostgreSQL
```
ehelms: Are we assuming that on restore services might already exist and therefore be running? If that is the case, I would suggest stopping all services (if they exist).

src/roles/restore/tasks/deploy_and_verify.yaml
```diff
@@ -0,0 +1,101 @@
...
+  ansible.builtin.systemd:
+    name: postgresql.service
+    state: stopped
+  when:
+    - database_mode == 'internal'
+    - restore_postgresql_started | default(false)
+
+- name: Mark PostgreSQL as stopped
```
ehelms: I do not think this is needed.

src/roles/restore/tasks/deploy_and_verify.yaml
```diff
@@ -0,0 +1,101 @@
...
+  ansible.builtin.set_fact:
+    restore_postgresql_started: false
+  when: database_mode == 'internal'
+
+- name: Display deployment status
+  ansible.builtin.debug:
+    msg: |
+      Running foremanctl deploy to regenerate configuration...
+      All data has been restored:
```
ehelms: The data hasn't been restored yet has it?

src/roles/restore/tasks/deploy_and_verify.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - Databases restored
+      - Pulp content and encryption keys restored
+      - OAuth keys and passwords restored
+      - Foremanctl state restored
+
+      Deploy will regenerate podman secrets from restored credentials.
+
+- name: Run foremanctl deploy
+  ansible.builtin.command:
+    cmd: foremanctl deploy
```
ehelms: This is a bad idea. I think this needs to be built into the playbook rather than buried in the role. And it should make use of the existing deploy playbook.

ehelms:
@Chyenne8 could you add documentation for restore, similar to @sjha4 backup documentation, I think it will help to see this documented from the users perspective when reviewing the code.

src/roles/restore/tasks/main.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - name: Stop PostgreSQL on failure
+      ansible.builtin.systemd:
+        name: postgresql.service
+        state: stopped
+      when:
+        - database_mode == 'internal'
+        - restore_postgresql_started | default(false)
+      failed_when: false
+
+    - name: Restart Foreman services on failure
```
ehelms: If restore fails, then most likely the services won't start. I think with the rescue on a restore is, what should the state of the system be:

- Revert the restore
- Leave it in the broken state for further investigation and re-run

src/roles/restore/tasks/main.yaml
```diff
@@ -0,0 +1,80 @@
+---
+# Main restore orchestration
+
+- name: Run validation checks
+  ansible.builtin.include_tasks:
+    file: validate.yaml
+
+- name: Perform restore operations
```
ehelms, line 6: I don't think we need all the debug messages in here, this is not a pattern we use anywhere else right and we let the Ansible tasks speak for themselves.

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,45 @@
+---
+# Phase 1: Basic validation - check required files exist
+# This runs BEFORE any destructive actions
```
ehelms: This is not strictly true, it shouldbe run, but there is nothing enforcing that. I would drop these comments.

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: Verify required backup files exist
+  ansible.builtin.stat:
+    path: "{{ backup_dir }}/{{ item }}"
+  register: restore_required_files
+  failed_when: not restore_required_files.stat.exists
+  loop:
+    - foreman.dump
+    - candlepin.dump
+    - pulp.dump
```
ehelms: This is going to fail when there are flavors that don't have these databases. Perhaps these should be derived from the backup metadata? @sjha4

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: Display validation success
+  ansible.builtin.debug:
+    msg: |
+      Backup validation passed
+      Backup directory exists: {{ backup_dir }}
+      Metadata file found
+      Required files present (foreman.dump, candlepin.dump, pulp.dump)
```
ehelms: See comment above, I would not tie this output to those specific files

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: Stop here if dry-run mode
+  ansible.builtin.meta: end_play
+  when: dry_run | default(false)
```
ehelms: I would do this via a `when` condition in the `main.yml` rather than this method.

src/roles/restore/tasks/prepare_system.yaml
```diff
@@ -0,0 +1,49 @@
+---
+# Phase 2: Prepare system for database restore
+# Stop services and ensure PostgreSQL is ready
```
ehelms: If restoring over top of an already existing system, then all services should be stopped and not just postgresql. I would use `foreman.target` as the thing to stop if it exists already.

src/roles/restore/tasks/restore_foremanctl_state.yaml
```diff
@@ -0,0 +1,68 @@
+---
+# Restore foremanctl state (OAuth keys, passwords, parameters)
+# CRITICAL: Must be restored before starting services
+
+- name: Check if foremanctl-state archive exists
```
ehelms: Should the validate.yml handle this?

src/roles/restore/tasks/restore_foremanctl_state.yaml
```diff
@@ -0,0 +1,68 @@
...
+  register: restore_state_files
+  failed_when: not restore_state_files.stat.exists
+  loop:
+    - parameters.yaml
+    - foreman-oauth-consumer-key
+    - foreman-oauth-consumer-secret
+    - postgresql-admin-password
+    - foreman-db-password
+    - candlepin-db-password
+    - pulp-db-password
```
ehelms: I'd consider deriving these form the metadata file instead of hard-coding them.

src/roles/restore/tasks/deploy_and_verify.yaml
Chyenne8: Good catch, we should handle the case where services might not exist yet. I can make the updates

src/roles/restore/tasks/deploy_and_verify.yaml
Chyenne8: Removing this task

src/roles/restore/tasks/deploy_and_verify.yaml
Chyenne8: I reverted it because the pulp database migration failed. I will update and attempt a different approach to build it in the playbook.

src/roles/restore/tasks/restore_foremanctl_state.yaml
Chyenne8: I corrected the  hard coded forms here and throughout the code.

src/roles/restore/tasks/main.yaml
Chyenne8: I removed the redundant debug messages and simplified others through out the code.

src/roles/restore/tasks/deploy_and_verify.yaml
Chyenne8: Data has been restored by this point but I can clarify this message and confirm whats happening at this stage.

src/roles/restore/tasks/main.yaml
Chyenne8: I will update the rescue to keep the broken state for investigation.

src/roles/restore/tasks/restore_foremanctl_state.yaml
Chyenne8: Yes, I moved this check into validate.yml.

src/roles/restore/tasks/validate.yaml
Chyenne8: These comments have been removed.

src/roles/restore/tasks/validate.yaml
Chyenne8: Removed the meta: end_play from validate.yaml and added a when: not dry_run | default(false) condition to the "Perform restore operations" block in main.yaml.

src/roles/restore/tasks/validate.yaml
Chyenne8: Updated  this to check based on the backup metadata.

docs/user/restore.md
```diff
@@ -0,0 +1,409 @@
...
+
+| Argument | Description |
+|----------|-------------|
+| `BACKUP_DIR` | Path to the backup directory created by `foremanctl backup`. This should be the timestamped directory (e.g., `/var/backup/foreman-backup-20260617T104115`), not the parent directory. |
+
+### Optional
+
+| Option | Description |
+|--------|-------------|
+| `--dry-run` | Validate the backup without performing the restore. Checks that all required files exist and the backup metadata is valid. |
```
ehelms: I am leaning towards a name that indicates validation like `--validate` or `--validate-backup`.

src/playbooks/restore/restore.yaml
```diff
@@ -0,0 +1,81 @@
+---
+- name: Restore from a backup
+  hosts: quadlet
+  become: true
+  gather_facts: true
+  vars_files:
```
ehelms, line 6: Can you verify that all of these variable files are needed?

src/playbooks/restore/restore.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - "../../vars/flavors/{{ flavor }}.yml"
+    - "../../vars/{{ certificates_source }}_certificates.yml"
+    - "../../vars/images.yml"
+    - "../../vars/tuning/{{ tuning }}.yml"
+    - "../../vars/foreman.yml"
+    - "../../vars/database.yml"
+    - "../../vars/database_iop.yml"
+    - "../../vars/base.yaml"
+    - "../../roles/pulp/defaults/main.yaml"
+  roles:
```
ehelms, line 10: Is there a way to import / call the deploy playbook?

src/roles/restore/tasks/prepare_system.yaml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Stop Foreman services
+  ansible.builtin.systemd:
+    name: foreman.target
+    state: stopped
+  failed_when: false
+
+- name: Stop PostgreSQL explicitly
```
ehelms: Why do we stop postgres here when we stopped it before using the `foreman.target` ? And also, to then start it again right after we stop it.

docs/user/backup-restore.md
```diff
@@ -0,0 +1,409 @@
...
+foremanctl restore /var/backup/foreman-backup-20260617T104115 --dry-run
+```
+
+This checks:
+- Backup directory exists
+- `metadata.yml` is present and valid
+- All expected database dumps exist
+- `foremanctl-state.tar.gz` exists
+
+## Prerequisites
```
ehelms, line 272: Are the things listed in this section things we can check for instead of relying on potential human error in assuring them? e.g. can we validate the OS matches the requirement or that we have sufficient space?

docs/user/backup-restore.md
```diff
@@ -0,0 +1,409 @@
...
+
+3. **Network connectivity**
+   - For external database mode, connection to database server
+   - Access to container registries (for deployment phase)
+
+4. **Backup compatibility**
+   - Backup must be from the same or previous version of foremanctl
+   - Forward compatibility is not guaranteed
+
+## What Gets Restored
```
ehelms, line 279: I do not think we need to list everything that get's restored here in the docs.

docs/user/backup-restore.md
```diff
@@ -0,0 +1,409 @@
...
+### Pulp Content
+
+If `pulp-content.tar.gz` exists in the backup:
+- Repository content files
+- Database encryption keys (`database_fields.symmetric.key`)
+- Django secret key (`django_secret_key`)
+
+If the archive is missing (backup used `--skip-pulp-content`), Pulp content restore is skipped.
+
+## Restore Process
```
ehelms, line 286: I think this section is useful, but it could be more concise and less detailed as some of these details are not relevant to the user or can be figured out from the role itself.

docs/user/backup-restore.md
```diff
@@ -0,0 +1,409 @@
...
+After deployment:
+
+1. Wait for services to stabilize (30 seconds)
+2. Poll Foreman API until responding (up to 5 minutes)
+3. Verify service health:
+   - `foreman.target` is active
+   - `foreman.service` is active
+   - `postgresql.service` is active
+
+## Error Handling
```
ehelms, line 296: I like this section, however it's a bit list heavy and could be written in a more user centric way so that users understand the general idea of what happens if an error occurs and what their next step(s) should be.

docs/user/restore.md
```diff
@@ -0,0 +1,409 @@
...
+
+### Check Service Status
+
+```bash
+systemctl status foreman.target
+systemctl status foreman.service
+systemctl status postgresql.service
+```
+
+### Check Container Status
```
ehelms: Shouldn't the step above be sufficient?

docs/user/restore.md
```diff
@@ -0,0 +1,409 @@
...
+
+3. **Recovery:**
+   - Review error message
+   - Fix underlying issue
+   - Re-run restore command
+   - Or manually start services: `systemctl start foreman.target`
+
+## Post-Restore Verification
+
+After restore completes, verify your system:
```
ehelms: Which of this can be automated through the Ansible?

docs/user/restore.md
```diff
@@ -0,0 +1,409 @@
...
+- All current databases are dropped and recreated
+- All Pulp content is replaced
+- All configuration is replaced
+- There is NO undo operation
+
+**Always verify you have the correct backup before proceeding.**
+
+### Same-Host Restore
+
+Restoring on the same host where the backup was created:
```
ehelms: Should we warn users when they are restoring on an existing system and have a flag to force it?

docs/user/restore.md
```diff
@@ -0,0 +1,409 @@
...
+- Monitor available disk space
+- Consider bandwidth for network-based storage
+
+## Troubleshooting
+
+### Restore Fails with "Database already exists"
+
+**Cause:** Previous restore attempt left databases in inconsistent state.
+
+**Solution:** Manually drop databases before retrying:
```
ehelms: Should restore handle this for the user?

docs/user/backup-restore.md
```diff
@@ -0,0 +1,409 @@
...
+podman exec postgresql psql -U postgres -c "ALTER DATABASE foreman OWNER TO foreman;"
+
+# Fix candlepin database
+podman exec postgresql psql -U postgres -c "ALTER DATABASE candlepin OWNER TO candlepin;"
+
+# Fix pulp database
+podman exec postgresql psql -U postgres -c "ALTER DATABASE pulp OWNER TO pulp;"
+```
+
+## Best Practices
```
ehelms, line 384: This section feels rather AI heavy in it's use of bold and lists. Consider making this more user-focused and explain, in some cases, why it's a best practice.

src/roles/restore/tasks/main.yaml
```diff
@@ -0,0 +1,56 @@
...
+      ansible.builtin.include_tasks:
+        file: deploy_and_verify.yaml
+
+  rescue:
+    - name: Stop PostgreSQL on failure
+      ansible.builtin.systemd:
+        name: postgresql.service
+        state: stopped
+      when: database_mode == 'internal'
+      failed_when: false
```
sjha4: We don't need to stop postgresql, the foreman.target will cover that.

src/roles/restore/defaults/main.yaml
```diff
@@ -0,0 +1,11 @@
...
+# Timeout and retry settings for restore operations
+restore_postgresql_ready_retries: 10
+restore_postgresql_ready_delay: 2
+restore_postgresql_stop_retries: 30
+restore_postgresql_stop_delay: 1
+
+# State tracking variables
+restore_service_stopped: false
+restore_postgresql_started: false
+restore_destructive_action_taken: false
```
sjha4: Don't need to track states..We reverted this pattern in the backup PR and rely on ansible to do it's thing.

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,130 @@
+---
+# Phase 3: Restore databases from backup
```
sjha4: Remove the comments on these tasks, here and in other files.

src/playbooks/restore/restore.yaml
```diff
@@ -0,0 +1,81 @@
...
+        certificate_checks_key: "{{ server_key }}"
+        certificate_checks_ca: "{{ server_ca_certificate }}"
+    - role: postgresql
+      when:
+        - database_mode == 'internal'
+    - redis
+    - candlepin
+    - httpd
+    - pulp
+    - foreman
```
sjha4: Are these roles here to mimic the deploy playbook? Could we use deploy directly?

docs/user/restore.md
Chyenne8: The --validate name is clearer than --dry-run for what it actually does. I have updated the name.

src/playbooks/restore/restore.yaml
Chyenne8: I've verified and reduced the vars_files from 10 down to 3.

src/playbooks/restore/restore.yaml
Chyenne8: Changed from shelling out to `foremanctl deploy` to using `import_playbook: ../deploy/deploy.yaml`.

 The `when: not validate | default(false)` condition on the import ensures the deploy playbook is skipped during `--validate` runs.

Tested both `--validate` (skips deploy) and full restore (runs deploy) - both working correctly.

src/roles/restore/tasks/prepare_system.yaml
Chyenne8: The file now only stops foreman.target and nothing else. All the PostgreSQL stop/start logic was removed from prepare_system.yaml.

src/roles/backup/tasks/main.yaml
Chyenne8: This should have not been pushed I was testing out this change locally. I have corrected this.

src/playbooks/restore/metadata.obsah.yaml
Chyenne8: Changed dry_run to validate.

docs/user/backup-restore.md
Chyenne8: I've added automated checks for the things we can programmatically validate.

Chyenne8: Also updated the prerequisites list the user should check.

docs/user/backup-restore.md
Chyenne8: Agreed, I simplified the list.

docs/user/backup-restore.md
Chyenne8: I condensed this section.

docs/user/backup-restore.md
Chyenne8: Changed from a numbered list of technical details to a user-focused narrative explaining what happens and what their next steps should be.

src/playbooks/restore/metadata.obsah.yaml
ehelms: I assume you haven't pushed the change yet?

docs/user/restore.md
Chyenne8: Simplified from multiple detailed verification sections down to just checking foreman.target status and accessing the UI.

docs/user/restore.md
Chyenne8: I've automated the verification that can be done through Ansible.

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,118 @@
+---
+- name: Check if foremanctl is already deployed
+  ansible.builtin.stat:
+    path: /var/lib/foremanctl
```
sjha4:
```suggestion
    path: "{{ obsah_state_path }}"
```

src/roles/restore/tasks/restore_pulp_content.yaml
```diff
@@ -0,0 +1,56 @@
...
+      ansible.builtin.stat:
+        path: "{{ pulp_storage_path }}/media"
+      register: restore_pulp_media_dir
+
+    - name: Backup existing pulp media directory
+      ansible.builtin.command:
+        cmd: mv {{ pulp_storage_path }}/media {{ pulp_storage_path }}/media.backup-{{ ansible_date_time.epoch }}
+      when: restore_pulp_media_dir.stat.exists
+      changed_when: true
+      failed_when: false
```
sjha4: Do we need this? We don't clean this up later so this will add additional disk space usage to restore.

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,112 @@
...
+    login_password: "{{ postgresql_admin_password }}"
+  loop: "{{ restore_databases_to_restore }}"
+  loop_control:
+    label: "{{ item.database_name }}"
+
+- name: Create empty databases
+  community.postgresql.postgresql_db:
+    name: "{{ item.database_name }}"
+    state: present
+    owner: postgres
```
sjha4: If we set these to item.owner, we can skip the ownership transfer and granting privileges tasks..Then we can restore without the --no-owner and --no-acl flags in pg_restore..We can assume the same users exist since we have had one run of foremanctl deploy on this machine to do that..

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,110 @@
...
+        database_name: "{{ foreman_database_name }}"
+        owner: "{{ foreman_database_user }}"
+      - dump_file: candlepin.dump
+        database_name: "{{ candlepin_database_name }}"
+        owner: "{{ candlepin_database_user }}"
+        when_db: candlepin
+      - dump_file: pulp.dump
+        database_name: "{{ pulp_database_name }}"
+        owner: "{{ pulp_database_user }}"
+        when_db: pulp
```
sjha4: We need to expand this to include IOP DBs as well..

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,118 @@
...
+    path: "{{ backup_dir }}/metadata.yml"
+  register: restore_metadata_content
+
+- name: Parse backup metadata
+  ansible.builtin.set_fact:
+    restore_backup_metadata: "{{ restore_metadata_content['content'] | b64decode | from_yaml }}"
+
+- name: Build list of expected dump files
+  ansible.builtin.set_fact:
+    restore_expected_dump_files: "{{ restore_backup_metadata.databases | map('regex_replace', '$', '.dump') | list }}"
```
sjha4: Looking at this, I think we need to change [src/roles/backup/tasks/database_dumps.yaml ](https://github.com/theforeman/foremanctl/blob/df8114acdc248a2f1ae99893f2704868008bbe59/src/roles/backup/tasks/database_dumps.yaml#L10) to 
```
--file={{ backup_dir_full }}/{{ item.database }}.dump
```

src/roles/restore/tasks/restore_foremanctl_state.yaml
```diff
@@ -0,0 +1,32 @@
...
+    src: "{{ backup_dir }}/foremanctl-state.tar.gz"
+    dest: "{{ obsah_state_path | dirname }}"
+    remote_src: true
+
+- name: Build list of expected state files
+  ansible.builtin.set_fact:
+    restore_expected_state_files: >-
+      {{
+        ['parameters.yaml', 'foreman-oauth-consumer-key', 'foreman-oauth-consumer-secret', 'postgresql-admin-password'] +
+        (restore_backup_metadata.databases | map('regex_replace', '$', '-db-password') | list)
```
sjha4: This conflicts for IOP..The db names are like advisor_db whereas password files are like iop-advisor-db-password ..

src/roles/restore/tasks/prepare_system.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Stop Foreman services
+  ansible.builtin.systemd:
+    name: foreman.target
+    state: stopped
+  failed_when: false
```
sjha4, line 6: We need a check for when postgresql is actually stopped. This returns immediately leading to intermitten errors like below when restoring..
```
stderr: 'pg_restore: error: connection to database "candlepin" failed: FATAL:  the
        database system is shutting down'
```

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,127 @@
...
+        owner: "{{ foreman_database_user }}"
+      - name: candlepin
+        dump_file: "{{ candlepin_database_name }}.dump"
+        database_name: "{{ candlepin_database_name }}"
+        owner: "{{ candlepin_database_user }}"
+      - name: pulp
+        dump_file: "{{ pulp_database_name }}.dump"
+        database_name: "{{ pulp_database_name }}"
+        owner: "{{ pulp_database_user }}"
+      - name: iop_advisor
```
sjha4: Can we make this conditional based on what metadata.yml contains in list of enabled features..
```
enabled_features:
- foreman
- katello
- content/ansible
- content/container
- content/deb
- content/python
- content/rpm
- hammer
- foreman-proxy
- remote-execution
- iop
```

This is the section..We can guard candlepin/pulp DBs on katello feature and iop dbs on iop feature status here.

sjha4: Let's not guard pulp on katello, foreman-proxy will also need pulp for backup/restore to be supported on proxies.. For reference, metadata.yml on proxy backup will look like this:
```
databases:
- pulp
enabled_features:
- foreman-proxy
- content/rpm
- content/deb
- content/container
- content/ansible
- content/python
```

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,137 @@
...
+
+- name: Restore databases from dump files
+  ansible.builtin.command:
+    cmd: >
+      pg_restore
+      -h {{ database_host }}
+      -p {{ database_port }}
+      -U {{ item.owner }}
+      -d {{ item.database_name }}
+      {{ backup_dir }}/{{ item.dump_file }}
```
sjha4, line 90: Apparently the 3 dbs: advisor_db ,inventory_db, vulnerability_db have an extension postgres_fdw which requires superuser privilege to create..A simple solution would be to create those ourselves..But a better solution is to reuse whatever iop db deploy is doing and not have to duplicate this..
The error I saw is this on the 3 db restores for debugging:
```
pg_restore: while PROCESSING TOC:
        pg_restore: from TOC entry 2; 3079 33314 EXTENSION postgres_fdw (no owner)
        pg_restore: error: could not execute query: ERROR:  permission denied to create extension "postgres_fdw"
        HINT:  Must be superuser to create this extension.
        Command was: CREATE EXTENSION IF NOT EXISTS postgres_fdw WITH SCHEMA public;
```

sjha4: Wonder if it's better to restore as superuser and then grant privileges the way you had before..The IOP Dbs are weird with the postgres_fdw extension. If we isolate the postgres_fdw tasks we can probably reuse that here..Or just revert to what you had before.. Will have to rely on @ehelms to answer that..

sjha4: The last push to create these extensions as postrges_admin works with restoring IOP DBs correctly.. 👍🏼 Only concern is borrowing that task from IOP but I can't think of a good way to avoid that..

sjha4 (approved):
Functionally this is working well..Code LGTM overall..Acking it and we'll have Eric and maintainers take a final look and we can go from there..

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,151 @@
...
+  ansible.builtin.set_fact:
+    restore_backup_metadata: "{{ restore_metadata_content['content'] | b64decode | from_yaml }}"
+
+- name: Build database restore configuration
+  ansible.builtin.set_fact:
+    restore_databases_config:
+      - name: foreman
+        dump_file: "{{ foreman_database_name }}.dump"
+        database_name: "{{ foreman_database_name }}"
+        owner: "{{ foreman_database_user }}"
```
sjha4: We have a flavor foreman-proxy coming up in an open PR: https://github.com/theforeman/foremanctl/pull/571 We will want to be mindful of changes there and the fact that foreman db is not available in foreman-proxy flavor..There will be changes to this config building with those changes.

sjha4: Locally I tested backup/restore with the proxy flavor and changes from that PR..Here's a patch which gets it working: https://gist.github.com/sjha4/65e43e7ad8147b40d32ed02c7ba92710

Don't want to worry about it in this PR since the other PR is in flight..We can revisit this as part of backup/restore on proxy which is being tracked separately.

sjha4:
@ehelms : I have acked this from my end with some questions around the IOP extensions in postgres. We have a card tracking proxy backup/restore so I have left a note here..In prelim debugging, it works with a few changes overall but it'd be nice to keep that separate from here.

src/playbooks/restore/metadata.obsah.yaml
```diff
@@ -0,0 +1,23 @@
+---
+help: |
+  Restore Foreman from a backup
```
ehelms:
```suggestion
  Restore from a backup
```

src/playbooks/restore/metadata.obsah.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+  Validates backup contents, extracts configuration files, restores databases,
+  restores Pulp content, and redeploys the system.
```
ehelms, line 6:
```suggestion
  restores content, and redeploys the system.
```

src/playbooks/restore/restore.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - "../../vars/database.yml"
+    - "../../vars/base.yaml"
+  roles:
+    - restore
+
+- name: Deploy services after restore
+  import_playbook: ../deploy/deploy.yaml
+  when: not validate | default(false)
+
+- name: Verify restore completion
```
ehelms: Is this needed? Shouldn't the fact that `deploy.yaml` was ran handle that for us?

src/roles/restore/tasks/prepare_system.yaml
ehelms: That is likely worth a Github issue to investigate what our expected behaviour of foreman.target is and if it can handle this scenario properly or we need something higher level.

src/roles/restore/tasks/prepare_system.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+- name: Wait for PostgreSQL to fully stop
+  ansible.builtin.systemd:
+    name: postgresql.service
+  register: restore_postgres_status
+  until: restore_postgres_status.status.ActiveState in ['inactive', 'failed']
+  retries: "{{ restore_postgresql_stop_retries }}"
+  delay: "{{ restore_postgresql_stop_delay }}"
+  when: database_mode == 'internal'
```
ehelms: This is leaking a higher level variable into the role.

src/roles/restore/tasks/main.yaml
```diff
@@ -0,0 +1,45 @@
...
+    file: validate.yaml
+
+- name: Perform restore operations
+  when: not validate | default(false)
+  block:
+    - name: Restore foremanctl state
+      ansible.builtin.include_tasks:
+        file: restore_foremanctl_state.yaml
+
+    - name: Prepare system for restore
```
ehelms, line 9: Why does prepare happen after restoring `foremanctl state` ? Can that ordering cause problems?

src/roles/restore/tasks/main.yaml
```diff
@@ -0,0 +1,45 @@
...
+        state: stopped
+      failed_when: false
+
+    - name: Report failure
+      ansible.builtin.fail:
+        msg: |
+          Restore failed: {{ ansible_failed_result.msg | default('Unknown error') }}
+
+          System state: Services stopped
+          The system has been left in a stopped state for investigation.
```
ehelms, line 42: Maybe combine these two statements into one.

src/roles/restore/tasks/main.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+          To recover:
+          1. Review the error above
```
ehelms:
```suggestion
          1. Review errors
```

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,151 @@
+---
+- name: Create PostgreSQL admin password secret
+  ansible.builtin.shell: |
+    cat {{ obsah_state_path }}/postgresql-admin-password | podman secret create postgresql-admin-password -
+  when: database_mode == 'internal'
```
ehelms: I'd consider this gating here -- when a user uses an external database they are responsible for the database backup. Therefore, you could instead gate this whole task file on this condition (using a role variable instead of a global one)

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,151 @@
+---
+- name: Create PostgreSQL admin password secret
```
ehelms: This should be done the same way as the postgresl role.

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,151 @@
+---
```
ehelms, line 1: This whole process of restoring postgres database makes me wonder if it should use parts of or be incoporated into the postgresql role. As there is a lot of over lap.

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,151 @@
...
+    owner: "{{ item.owner }}"
+    login_host: "{{ database_host }}"
+    login_port: "{{ database_port }}"
+    login_user: postgres
+    login_password: "{{ postgresql_admin_password }}"
+  loop: "{{ restore_databases_to_restore }}"
+  loop_control:
+    label: "{{ item.database_name }}"
+
+- name: Enable postgres_fdw extension for IOP databases
```
ehelms, line 69: Does this need to be done at this point? Or can the `deploy` step handle this?

src/roles/restore/tasks/restore_foremanctl_state.yaml
```diff
@@ -0,0 +1,32 @@
...
+  ansible.builtin.file:
+    path: "{{ obsah_state_path | dirname }}"
+    state: directory
+    mode: '0755'
+
+- name: Extract foremanctl state archive
+  ansible.builtin.unarchive:
+    src: "{{ backup_dir }}/foremanctl-state.tar.gz"
+    dest: "{{ obsah_state_path | dirname }}"
+    remote_src: true
```
ehelms, line 11: Why remote_src ?

src/roles/restore/tasks/restore_foremanctl_state.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+- name: Build list of expected state files
+  ansible.builtin.set_fact:
+    restore_expected_state_files: >-
```
ehelms: This make me wonder if they should be part of the metadata instead of hard-coded. Or at least, store these as a variable to make it easier in the future to add items to the list.

src/roles/restore/tasks/restore_pulp_content.yaml
```diff
@@ -0,0 +1,49 @@
...
+        path: "{{ pulp_storage_path }}/django_secret_key"
+      register: restore_django_secret_key
+      failed_when: not restore_django_secret_key.stat.exists
+
+    - name: Count restored media files
+      ansible.builtin.find:
+        paths: "{{ pulp_storage_path }}/media"
+        file_type: file
+        recurse: true
+      register: restore_pulp_media_files
```
ehelms: Is this used anywhere?

src/roles/restore/tasks/restore_pulp_content.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+    - name: Get archive size
+      ansible.builtin.stat:
+        path: "{{ backup_dir }}/pulp-content.tar.gz"
+      register: restore_pulp_content_size
```
ehelms: Is this used anywhere? if not, get rid of it

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,136 @@
...
+  ansible.builtin.stat:
+    path: /var/lib/foremanctl
+  register: restore_existing_deployment
+
+- name: Fail if restoring over existing system without --force
+  ansible.builtin.fail:
+    msg: |
+      DESTRUCTIVE OPERATION WARNING:
+
+      An existing Foreman deployment was detected on this system.
```
ehelms:
```suggestion
      An existing deployment was detected on this system.
```

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,136 @@
...
+- name: Read backup metadata
+  ansible.builtin.slurp:
+    path: "{{ backup_dir }}/metadata.yml"
+  register: restore_metadata_content
+
+- name: Parse backup metadata
+  ansible.builtin.set_fact:
+    restore_backup_metadata: "{{ restore_metadata_content['content'] | b64decode | from_yaml }}"
+
+- name: Build name to database mapping
```
ehelms: I'm not clear why this is needed. Shouldn't the metadata have this information already and thus hard-coding this list be unnecessary?

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,136 @@
...
+  loop: "{{ restore_expected_dump_files }}"
+
+- name: Check for foremanctl-state archive
+  ansible.builtin.stat:
+    path: "{{ backup_dir }}/foremanctl-state.tar.gz"
+  register: restore_state_archive
+
+- name: Fail if foremanctl-state is missing
+  ansible.builtin.fail:
+    msg: "CRITICAL: foremanctl-state.tar.gz not found in backup! Cannot restore OAuth keys and passwords."
```
ehelms: Why is this one CRITICAL but the others are just normal messages? I think everything is critical in the backup if it's missing. I'd just keep this like the others.

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+- name: Get total size of database dumps
+  ansible.builtin.shell: du -sB1 {{ backup_dir }}/*.dump 2>/dev/null | awk '{sum+=$1} END {print int(sum/1024/1024/1024)+1}'
+  register: restore_dumps_size
+  changed_when: false
+  failed_when: false
+
+- name: Warn if insufficient PostgreSQL disk space
+  ansible.builtin.debug:
+    msg: "WARNING: Available space ({{ restore_pgsql_space.stdout }}GB) may be insufficient for database dumps (~{{ restore_dumps_size.stdout }}GB needed)"
```
ehelms: This says "may be" -- under what conditions will the dumps being bigger than the available postgres space actually install properly?

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+- name: Get Pulp content archive size
+  ansible.builtin.stat:
+    path: "{{ backup_dir }}/pulp-content.tar.gz"
+  register: restore_pulp_archive
+  when: restore_pulp_archive is defined
+
+- name: Warn if insufficient Pulp disk space
+  ansible.builtin.debug:
+    msg: "WARNING: Available space ({{ restore_pulp_space.stdout }}GB) may be insufficient for Pulp content ({{ (restore_pulp_archive.stat.size / 1024 / 1024 / 1024) | int + 1 }}GB needed)"
```
ehelms: This says "may be" -- under what conditions will the pulp content being bigger than the available space actually install properly?

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,136 @@
...
+  when:
+    - restore_pulp_archive.stat.exists | default(false)
+    - restore_pulp_space.stdout | int < (restore_pulp_archive.stat.size / 1024 / 1024 / 1024) | int + 1
+
+- name: Check database mode matches backup
+  ansible.builtin.debug:
+    msg: "WARNING: Current database_mode ({{ database_mode }}) may not match backup configuration. Verify compatibility before proceeding."
+  when: restore_backup_metadata.database_mode is defined and restore_backup_metadata.database_mode != database_mode
```
ehelms: Right now, the user does not specify the `database_mode` when running restore. That means, the only place to get that information is from the backup itself. For the design overall I think that means two things:

1. This check is not necessary
2. The database_mode should be pulled from the metadata and then used as the source of what type of restore is happening.

sjha4:
@Chyenne8 : With your next push, could you update commit description to be a concise description ideally a one or two line summary..Right now the commit description kind of fills the screen with git log..

src/roles/restore/tasks/validate.yaml
sjha4: This got reverted it seems..

src/roles/restore/tasks/restore_databases.yaml
sjha4: The issue here is postgres user needs to enable these but we are restoring as db_owner to maintain ownership of restored DBs..pg_restore expects to have the extension enabled to succeed..

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,141 @@
...
+- name: Fail if foremanctl-state is missing
+  ansible.builtin.fail:
+    msg: "CRITICAL: foremanctl-state.tar.gz not found in backup! Cannot restore OAuth keys and passwords."
+  when: not restore_state_archive.stat.exists
+
+- name: Check available disk space for PostgreSQL
+  ansible.builtin.shell: df -BG /var/lib/pgsql | awk 'NR==2 {print $4}' | sed 's/G//'
+  register: restore_pgsql_space
+  changed_when: false
+  when: restore_database_mode == 'internal'
```
sjha4: `restore_database_mode` is set later at the end in the playbook..That needs to be before this step..

src/roles/restore/tasks/validate.yaml
sjha4: Ignore this.. /var/lib/foremanctl is correct..

inventories/localhost
```diff
@@ -1 +1,2 @@
+[quadlet]
```
sjha4: Pushed local change here I think..

src/roles/restore/tasks/validate.yaml
ehelms: path: "{{ obsah_state_path }}" is the correct one to have

src/roles/restore/tasks/main.yaml
```diff
@@ -0,0 +1,49 @@
+---
+- name: Run validation checks
+  ansible.builtin.include_tasks:
+    file: validate.yaml
```
ehelms, line 4: Should this task be skipped if validate is false?

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,129 @@
...
+    restore_name_to_db_map:
+      foreman: "{{ foreman_database_name }}"
+      candlepin: "{{ candlepin_database_name }}"
+      pulp: "{{ pulp_database_name }}"
+      iop_advisor: "{{ iop_advisor_database_name }}"
+      iop_inventory: "{{ iop_inventory_database_name }}"
+      iop_remediation: "{{ iop_remediation_database_name }}"
+      iop_vmaas: "{{ iop_vmaas_database_name }}"
+      iop_vulnerability: "{{ iop_vulnerability_database_name }}"
+  when: "'database_mapping' not in restore_backup_metadata"
```
ehelms: Why wouldn't it be there? I think this whole task can be removed now.

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+- name: Use database mapping from backup metadata
+  ansible.builtin.set_fact:
+    restore_name_to_db_map: "{{ restore_backup_metadata.database_mapping }}"
+  when: "'database_mapping' in restore_backup_metadata"
```
ehelms: I don't think you need this, you can set this a `var` on the task that needs it.

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+- name: Verify database dump files exist
+  ansible.builtin.stat:
+    path: "{{ backup_dir }}/{{ restore_name_to_db_map[item] }}.dump"
+  register: restore_dump_files_check
+  failed_when: not restore_dump_files_check.stat.exists
+  loop: "{{ restore_backup_metadata.databases }}"
+  when: "'databases' in restore_backup_metadata.backed_up_components | default(['databases'])"
```
ehelms: Is there a time when this won't be true?

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+- name: Verify foremanctl state archive exists
+  ansible.builtin.stat:
+    path: "{{ backup_dir }}/foremanctl-state.tar.gz"
+  register: restore_state_archive
+  failed_when: not restore_state_archive.stat.exists
+  when: "'foremanctl_state' in restore_backup_metadata.backed_up_components | default(['foremanctl_state'])"
```
ehelms, line 102: WHy is this when necessary?

src/roles/restore/tasks/restore_foremanctl_state.yaml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Remove existing foremanctl state directory
+  ansible.builtin.file:
+    path: "{{ obsah_state_path }}"
+    state: absent
+
+- name: Ensure state directory parent exists
```
ehelms: The previous operation would fail if this were not already true, so i think you can omit it?

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,114 @@
...
+  loop: "{{ restore_databases_to_restore }}"
+  loop_control:
+    label: "{{ item.database }}"
+  when: item.name in ['iop_advisor', 'iop_inventory', 'iop_vulnerability']
+
+- name: Restore databases from dump files
+  ansible.builtin.command:
+    cmd: >
+      pg_restore
+      -h {{ database_host }}
```
ehelms: Please use full option names for command e.g. `--host`

src/roles/restore/tasks/restore_databases.yaml
ehelms: Which task is it?

src/roles/restore/tasks/validate.yaml
sjha4: Can't think of anything outside user meddling with backup tar to cause this to fail. Safe to remove but also an argument to be made about a good defensive check to have..

sjha4:
Testing last restore commit and unrelated,I rebased this on master and it actually ran the whole postgres13 to 16 as part of deploy from a backup I had from an older postgres..I had some trouble with the upgrade playbook with stopping postgresql..Had to add an extra step to stop postgresql to get step: `Wait for port 5432 to be completely freed by Podman` here https://github.com/theforeman/foremanctl/blob/master/src/roles/postgresql/tasks/upgrade.yaml#L29 to not timeout..
```
TASK [postgresql : Stop foreman target] **********************************************************************************************************
ok: [quadlet]

TASK [postgresql : Stop PostgreSQL service] ******************************************************************************************************
changed: [quadlet]

TASK [postgresql : Wait for port 5432 to be completely freed by Podman] **************************************************************************
ok: [quadlet]
```

src/roles/restore/tasks/validate.yaml
ehelms: If the user messed with the backup tar, and these value does not exist in the metadata, I would prefer it to error. As that should indicate "something is wrong with this backup, do not continue".

src/roles/restore/tasks/validate.yaml
ehelms: If the user messed with the backup tar, and these value does not exist in the metadata, I would prefer it to error. As that should indicate "something is wrong with this backup, do not continue".

src/roles/restore/tasks/validate.yaml
ehelms: Should this only be done if the mode is internal?

ehelms: You could define a `block` and put all of the databases checks within that block gated by checking the database mode. That makes it easier to read.

src/roles/restore/tasks/restore_databases.yaml
sjha4: Missed this comment here..This was a question around enabling postgres_fdw extension for IOP databases step as part of restore..

sjha4:
The lint failures are related..The backup test failures are also related to the dump file name change here..

sjha4:
Can we add the documentation for this in the existing backup.md and rename the file to backup-restore.md .. Makes sense to have it in one doc.

tests/backup_test.py
```diff
@@ -130,11 +130,11 @@ def test_iop_database_dumps_created(server, backup_result):
     backup_dir = backup_result['backup_dir']
 
     expected_iop_dumps = [
-        'iop_advisor.dump',
-        'iop_inventory.dump',
-        'iop_remediation.dump',
-        'iop_vmaas.dump',
-        'iop_vulnerability.dump',
```
sjha4, line 137: Makes sense to me..+1 to changing DB dumps to the database name instead of name.

ianballou:
@ehelms what permission do we need to give @Chyenne8 so that the CI workflow runs at push time? I am unable to start it since it requires a maintainer.

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,105 @@
...
+- name: Build database restore configuration
+  ansible.builtin.set_fact:
+    restore_databases_config: "{{ restore_databases_config | default([]) + [db_entry] }}"
+  vars:
+    db_entry:
+      name: "{{ item.name }}"
+      dump_file: "{{ restore_backup_metadata.database_mapping[item.name] }}.dump"
+      database: "{{ item.database }}"
+      user: "{{ item.user }}"
+  loop: "{{ databases }}"
```
sjha4, line 30:
```
TASK [restore : Report failure]                                                                                                                 
  ******************************************************************************************************************                              
      changed: false                                                                                                                              
      msg: |-                                                                                                                                     
          Restore failed: The task includes an option with an undefined variable. The error was: {'name': '{{ item.name }}', 'dump_file': '{{     
  restore_backup_metadata.database_mapping[item.name] }}.dump', 'database': '{{ item.database }}', 'user': '{{ item.user }}'}: 'dict object' has  
  no attribute 'iop_advisor'. 'dict object' has no attribute 'iop_advisor'. {'name': '{{ item.name }}', 'dump_file': '{{                          
  restore_backup_metadata.database_mapping[item.name] }}.dump', 'database': '{{ item.database }}', 'user': '{{ item.user }}'}: 'dict object' has  
  no attribute 'iop_advisor'. 'dict object' has no attribute 'iop_advisor'                                                                        
                                                                                                                                                  
          The error appears to be in '/usr/share/foremanctl/src/roles/restore/tasks/restore_databases.yaml': line 21, column 3, but may           
          be elsewhere in the file depending on the exact syntax problem.                                                                         
                                                                                                                                                  
          The offending line appears to be:                                                                                                       
                                                                                                                                                  
                                                                                                                                                  
          - name: Build database restore configuration                                                                                            
            ^ here                                                                
```

Getting this from this task.. It is looping over databases which contains all DBs, including IOP even on boxes with IOP disabled.

sjha4: We should either use all_databases (with the flavor vars loaded), or loop over the backup metadata's databases list instead of the databases variable which is unfiltered.

sjha4: This worked for me locally:

```
31    loop_control:
32      label: "{{ item.name }}"
33 +  when: item.name in restore_backup_metadata.database_mapping
```

sjha4: If you make that change, the step below becomes a no-op which we can remove.

sjha4:
Not suggesting changes here but saw some weirdness with systems with different features enabled. For flow like 
`Take backup on IOP disabled -> Enable IOP -> Restore from pre-IOP backup` leaves IOP bits in the system ex DB, systemctl units etc.. We can document this by specifying that the enabled features on backup and restore machines need to be the same for proper functionality. Even with a restore above, the services come back up correctly. The IOP related DBs, services are basically orphaned at this point which would need extra cleanup. This will be the case for other features enabled/disabled between backup/restore.

ehelms:
> @ehelms what permission do we need to give @Chyenne8 so that the CI workflow runs at push time? I am unable to start it since it requires a maintainer.

Just needs to be a member of `theforeman` organization. I have sent an invite.

src/roles/restore/tasks/restore_foremanctl_state.yaml
sjha4: @Chyenne8 : 
For backup/restore to work on localhost and quadlets, the backup role:
  1. Archives state on the controller (delegate_to: localhost)
  2. Copies the archive to the remote target
  3. Cleans up on the controller

The restore role needs to do the reverse — fetch the archive from the remote and extract it on the controller. Currently it's trying to extract on the remote host where .var/lib doesn't exist.

sjha4: You can use this diff as a pointer:

```
diff --git a/src/roles/restore/tasks/restore_foremanctl_state.yaml b/src/roles/restore/tasks/restore_foremanctl_state.yaml
index 7de3d90..76a7aa8 100644
--- a/src/roles/restore/tasks/restore_foremanctl_state.yaml
+++ b/src/roles/restore/tasks/restore_foremanctl_state.yaml
@@ -1,11 +1,41 @@
 ---
+- name: Fetch foremanctl state archive from target
+  ansible.builtin.fetch:
+    src: "{{ backup_dir }}/foremanctl-state.tar.gz"
+    dest: "{{ obsah_state_path | dirname }}/foremanctl-state.tar.gz"
+    flat: true
+
 - name: Remove existing foremanctl state directory
   ansible.builtin.file:
     path: "{{ obsah_state_path }}"
     state: absent
+  delegate_to: localhost
+  become: false
+
+- name: Ensure foremanctl state parent directory exists
+  ansible.builtin.file:
+    path: "{{ obsah_state_path | dirname }}"
+    state: directory
+  delegate_to: localhost
+  become: false
 
-- name: Extract foremanctl state archive
+- name: Extract foremanctl state archive on controller
   ansible.builtin.unarchive:
-    src: "{{ backup_dir }}/foremanctl-state.tar.gz"
+    src: "{{ obsah_state_path | dirname }}/foremanctl-state.tar.gz"
     dest: "{{ obsah_state_path | dirname }}"
-    remote_src: true
+  delegate_to: localhost
+  become: false
+
+- name: Set permissions on foremanctl state directory
+  ansible.builtin.file:
+    path: "{{ obsah_state_path }}"
+    mode: '0700'
+  delegate_to: localhost
+  become: false
+
+- name: Clean up foremanctl state archive on controller
+  ansible.builtin.file:
+    path: "{{ obsah_state_path | dirname }}/foremanctl-state.tar.gz"
+    state: absent
+  delegate_to: localhost
+  become: false

```

sjha4:
The proxy error is unrelated..Will probably go away on a rerun..We do have a separate PR addressing proxy restore here: https://github.com/theforeman/foremanctl/pull/684 to keep this PR focused on foreman server..

sjha4:
PS: @vijaysawant has a robottelo PR for coverage: https://github.com/SatelliteQE/robottelo/pull/22300

@ehelms : This could use your re-review eyes. Think the PR is stable now and tests have run fine.

ehelms (approved):
Looks like a good starting state for restore, there may be some elements to follow up with but I think this is good. 

@sjha4 as you had a few comments can you also approve?

sjha4 (approved):
Thanks @ehelms!

My comments were mostly findings from the churn in foremanctl..I am good with the PR..Ack 👍🏼

PR #554, opened 2026-06-10
Warn about unrecognized smart proxy features

adamruzicka:
Companion to https://github.com/theforeman/foreman/pull/11017

Currently if the proxy offers more features than Foreman knows, those would go silently ignored (unless all of the proxy's features are unknown to Foreman), which makes it hard to spot misconfigured deployments.

After registering a proxy, fctl will check if all the proxy's features are recognized by Foreman and warn if not.

Steps to reproduce:
1) Somehow force a deployment where the proxy has features which Foreman doesn't know
2) See the warning

#### Checklist
  * [ ] figure out how to do this
  * this doesn't really need docs

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -325,3 +325,12 @@
   register: foreman_tasks_status
   when:
     - "'katello' in foreman_status.json['results']"
+
+- name: Warn about unrecognized Pulp Smart Proxy features
+  ansible.builtin.debug:
+    msg: >-
+      Proxy {{ ansible_facts['fqdn'] }}-pulp has features not recognized by Foreman:
+      {{ _foreman_pulp_proxy_registration.entity.smart_proxies[0].unrecognized_features | join(', ') }}.
+      If these features come from a Smart Proxy plugin, make sure Foreman has the plugin installed too.
+  when:
+    - _foreman_pulp_proxy_registration.entity.smart_proxies[0].unrecognized_features | default([]) | length > 0
--- a/src/roles/foreman_proxy/tasks/main.yaml
+++ b/src/roles/foreman_proxy/tasks/main.yaml
@@ -74,16 +74,6 @@
     oauth1_consumer_key: "{{ foreman_proxy_oauth_consumer_key }}"
     oauth1_consumer_secret: "{{ foreman_proxy_oauth_consumer_secret }}"
     ca_path: "{{ foreman_proxy_foreman_ca_certificate | default(omit) }}"
-  register: _foreman_proxy_registration
-
-- name: Warn about unrecognized Smart Proxy features
-  ansible.builtin.debug:
-    msg: >-
-      Proxy {{ foreman_proxy_name }} has features not recognized by Foreman:
-      {{ _foreman_proxy_registration.entity.smart_proxies[0].unrecognized_features | join(', ') }}.
-      If these features come from a Smart Proxy plugin, make sure Foreman has the plugin installed too.
-  when:
-    - _foreman_proxy_registration.entity.smart_proxies[0].unrecognized_features | default([]) | length > 0
 
 - name: Flush handlers to restart services
   ansible.builtin.meta: flush_handlers
--- a/src/roles/foreman_proxy/tasks/main.yaml
+++ b/src/roles/foreman_proxy/tasks/main.yaml
@@ -74,6 +74,16 @@
     oauth1_consumer_key: "{{ foreman_proxy_oauth_consumer_key }}"
     oauth1_consumer_secret: "{{ foreman_proxy_oauth_consumer_secret }}"
     ca_path: "{{ foreman_proxy_foreman_ca_certificate | default(omit) }}"
+  register: _foreman_proxy_registration
+
+- name: Warn about unrecognized Smart Proxy features
+  ansible.builtin.debug:
+    msg: >-
+      Proxy {{ foreman_proxy_name }} has features not recognized by Foreman:
+      {{ _foreman_proxy_registration.entity.smart_proxies[0].unrecognized_features | join(', ') }}.
+      If these features come from a Smart Proxy plugin, make sure Foreman has the plugin installed too.
+  when:
+    - _foreman_proxy_registration.entity.smart_proxies[0].unrecognized_features | default([]) | length > 0
 
 - name: Flush handlers to restart services
   ansible.builtin.meta: flush_handlers
--- a/src/roles/post_install/tasks/main.yaml
+++ b/src/roles/post_install/tasks/main.yaml
@@ -1,3 +1,6 @@
 ---
 - name: Run post install tasks
   ansible.builtin.include_tasks: "{{ flavor }}.yaml"
+
+- name: Check for unrecognized Smart Proxy features
+  ansible.builtin.include_tasks: smart_proxy_features_check.yaml
--- a/src/roles/post_install/tasks/main.yaml
+++ b/src/roles/post_install/tasks/main.yaml
@@ -4,3 +4,6 @@
 
 - name: Check for unrecognized Smart Proxy features
   ansible.builtin.include_tasks: smart_proxy_features_check.yaml
+  vars:
+    foreman_server_url: "{{ foreman_url }}"
+    foreman_ca_path: "{{ foreman_ca_certificate }}"
--- a/src/roles/post_install/tasks/main.yaml
+++ b/src/roles/post_install/tasks/main.yaml
@@ -5,5 +5,7 @@
 - name: Check for unrecognized Smart Proxy features
   ansible.builtin.include_tasks: smart_proxy_features_check.yaml
   vars:
-    foreman_server_url: "{{ foreman_url }}"
-    foreman_ca_path: "{{ foreman_ca_certificate }}"
+    post_install_foreman_server_url: "{{ foreman_url }}"
+    post_install_foreman_ca_path: "{{ foreman_ca_certificate }}"
+    post_install_foreman_oauth_consumer_key: "{{ foreman_oauth_consumer_key }}"
+    post_install_foreman_oauth_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
--- a/src/roles/post_install/tasks/main.yaml
+++ b/src/roles/post_install/tasks/main.yaml
@@ -4,8 +4,3 @@
 
 - name: Check for unrecognized Smart Proxy features
   ansible.builtin.include_tasks: smart_proxy_features_check.yaml
-  vars:
-    post_install_foreman_server_url: "{{ foreman_url }}"
-    post_install_foreman_ca_path: "{{ foreman_ca_certificate }}"
-    post_install_foreman_oauth_consumer_key: "{{ foreman_oauth_consumer_key }}"
-    post_install_foreman_oauth_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
--- a/src/roles/post_install/tasks/smart_proxy_features_check.yaml
+++ b/src/roles/post_install/tasks/smart_proxy_features_check.yaml
@@ -1,10 +1,10 @@
 ---
 - name: Get all registered Smart Proxies
   theforeman.foreman.resource_info:
-    server_url: "{{ foreman_server_url }}"
-    oauth1_consumer_key: "{{ foreman_oauth_consumer_key }}"
-    oauth1_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
-    ca_path: "{{ foreman_ca_path }}"
+    server_url: "{{ post_install_foreman_server_url }}"
+    oauth1_consumer_key: "{{ post_install_foreman_oauth_consumer_key }}"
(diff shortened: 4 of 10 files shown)


adamruzicka:
> figure out how to do this [tests]

This would require a test that would intentionally break things just to check that this would then emit a warning. I'm open to idea, but at this point I'm leaning towards not having tests for this here.

src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -74,6 +74,16 @@
...
     oauth1_consumer_secret: "{{ foreman_proxy_oauth_consumer_secret }}"
     ca_path: "{{ foreman_proxy_foreman_ca_certificate | default(omit) }}"
+  register: _foreman_proxy_registration
+
+- name: Warn about unrecognized Smart Proxy features
+  ansible.builtin.debug:
+    msg: >-
+      Proxy {{ foreman_proxy_name }} has features not recognized by Foreman:
+      {{ _foreman_proxy_registration.entity.smart_proxies[0].unrecognized_features | join(', ') }}.
+      If these features come from a Smart Proxy plugin, make sure Foreman has the plugin installed too.
```
ehelms: I don't think a user can act on this cleanly. Plugins are not a thing we expose here in foremanctl. We present the user with `features`. I like the general idea of where you are going with this, but it feels like a `post_install` check that should happen once at the end of `deploy`.

adamruzicka: Done. The only thing I'm not exactly happy about is that right now it checks all the proxies, not just those that were registered (or affected) during this installation run. Probably not a problem right now, but it could become somewhat confusing later

src/roles/post_install/tasks/smart_proxy_features_check.yaml
```diff
@@ -0,0 +1,19 @@
...
+    resource: smart_proxies
+    full_details: true
+  register: _post_install_smart_proxies
+
+- name: Warn about unrecognized Smart Proxy features
+  ansible.builtin.debug:
+    msg: >-
+      Proxy {{ item.name }} has features not recognized by Foreman:
+      {{ item.unrecognized_features | join(', ') }}.
+      If these features come from a Smart Proxy plugin, make sure Foreman has the plugin installed too.
```
ehelms: I think the `plugin` part of this is still something I do not think users will understand, as they cannot affect this. I am not sure what the right output is. Maybe use "Feature enabled" ?

src/roles/post_install/tasks/smart_proxy_features_check.yaml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Get all registered Smart Proxies
+  theforeman.foreman.resource_info:
+    server_url: "{{ foreman_url }}"
+    username: "{{ foreman_initial_admin_username }}"
+    password: "{{ foreman_initial_admin_password }}"
```
ehelms: You'll need to pass these in instead of using them from another role.

ehelms: Also, using initial variables will fail inevitably when the user changes them.

src/roles/post_install/tasks/smart_proxy_features_check.yaml
adamruzicka: Is it worse than referring to two different things with the word `feature` in the same sentence?

ehelms: Think in terms of deployment. This should only ever happen between an external Foreman Proxy/Capsule and Foreman. And the user will be thinking in terms of the `foremanctl` deployment tool, and using the `features` command to see enabled features and then possibly the `--add-feature` option to enable it.

adamruzicka: Isn't that an oversimplified view of things? There is no guarantee that the strings that can possibly show up as unrecognized features will in any way map to foremanctl's features, no matter how people may or may not think about them.

> And the user will be thinking in terms of the foremanctl deployment tool, and using the features command to see enabled features and then possibly the --add-feature option to enable it.

Putting the "a user has a custom proxy image with bogus features" edge case aside, let's have a model situation. The user installs vanilla foreman. Then the user tries to set up an external proxy with remote execution. This check will warn the user about unrecognized feature `script`. The user could be tempted to do `--add-feature script` somewhere, but that will fail as well.

I can reword it to say something like `Proxy $name has features not recognized by Foreman: $feature_list . Make sure Foreman and the proxy have the same set of features enabled`, but that doesn't really help anything

ofedoren: _Just to cents, might be dumb:_

If we don't want to confuse user, let's not assume what they'll think, we never know. As with docs, let's give them something unified as the source that cannot be vague.

Since foremanctl abstracts plugins into features, foreman and proxy has plugins which compose features and we know about it (what is used to compose what), can't we just tell the user something like (based on Adams example): "Proxy $name has features not recognized by Foreman: `script`, `openscap`. Use `--add-feature remote_execution` for `script`, `--add-feature openscap` for `openscap`.".

Sure, it makes us to either find a way for dynamic/auto mapping or maintain a static list with mapping. AFAIU, we don't support installation of features that foremanctl is not aware of anyway, so we already have something static to maintain, so it shouldn't be a problem...

adamruzicka: @ehelms could you chime in please? For the time being, I went with `Proxy {{ item.name }} reports unrecognized features: {{ unrecognized_features_list }}. Enable matching features in foremanctl or disable them on the proxy.`. If that's not sufficient, we could go with what Oleh proposed, even though that would be slightly bigger

ehelms:
Generally looks good now, you'll have to deal with the variables that are not scoped to the role name.

adamruzicka:
🍏

src/roles/post_install/tasks/smart_proxy_features_check.yaml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Get all registered Smart Proxies
+  theforeman.foreman.resource_info:
+    server_url: "{{ foreman_server_url }}"
+    oauth1_consumer_key: "{{ foreman_oauth_consumer_key }}"
+    oauth1_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+    ca_path: "{{ foreman_ca_path }}"
```
ehelms: To follow our rules, all of the variables in this role need to be role scoped (e.g. post_intsall_foreman_ca_path). Yes, I realize that creates a lot of overhead for this particular case, however, thems the rules currently.

adamruzicka: I was under the impression there was a lint that was enforcing this? Anyway, added

ehelms: I think it's enforcing role variables being namespaced but not the use of variables which I realize might sound like the same thing but it's not.

src/roles/post_install/tasks/main.yaml
```diff
@@ -14,3 +14,11 @@
     mode: '0640'
   delegate_to: localhost
   become: false
+
+- name: Check for unrecognized Smart Proxy features
+  ansible.builtin.include_tasks: smart_proxy_features_check.yaml
+  vars:
+    post_install_foreman_server_url: "{{ foreman_url }}"
```
ehelms: This is effectively the same issue -- these are variables not scoped to the current role. These would need to be passed into `post_install` role when it's declared in the `deploy` playbook for example.

Would we benefit from global variables? Perhaps. I think that should be a broader discussion.

adamruzicka: So the use of `foreman_url` on L8 is also wrong?

src/roles/post_install/tasks/main.yaml
```diff
@@ -5,12 +5,15 @@
       - "{{ _post_install_url_msg }}"
       - "{{ _post_install_cred_msg if (post_install_done_flag is not exists) else '' }}"
   vars:
-    _post_install_url_msg: "Foreman is running at {{ foreman_url }}"
-    _post_install_cred_msg: "Admin credentials: {{ foreman_initial_admin_username }}:{{ foreman_initial_admin_password }}"
+    _post_install_url_msg: "Foreman is running at {{ post_install_foreman_server_url }}"
+    _post_install_cred_msg: "Admin credentials: {{ post_install_foreman_initial_admin_username }}:{{ post_install_foreman_initial_admin_password }}"
```
ehelms: Thanks for handling those!

ehelms (approved):

ehelms:
@adamruzicka rebase needed -- I merged a big set of changes with the `deploy-proxy` command

src/roles/foreman/tasks/main.yaml
```diff
@@ -325,3 +325,12 @@
   register: foreman_tasks_status
   when:
     - "'katello' in foreman_status.json['results']"
+
+- name: Configure Foreman Proxy
```
ehelms: This exists over here -- https://github.com/theforeman/foremanctl/blob/master/src/roles/pulp/tasks/main.yaml#L317

ehelms (approved):

evgeni:
This now creates a rather huge skipped block at the end of every run:
```
TASK [post_install : Warn about unrecognized Smart Proxy features] ********************************************************************************************************************************************************************
skipping: [proxy] => (item={'created_at': '2026-07-09 15:08:02 UTC', 'updated_at': '2026-07-09 15:08:02 UTC', 'hosts_count': 0, 'unrecognized_features': [], 'name': 'proxy.example.com', 'id': 3, 'url': 'https://proxy.example.com:8443', 'remote_execution_pubkey': None, 'remote_execution_ca_pubkey': None, 'download_policy': 'on_demand', 'supported_pulp_types': [], 'lifecycle_environments': [], 'features': [{'capabilities': [], 'name': 'Logs', 'id': 13}], 'locations': [{'id': 2, 'name': 'Default Location', 'title': 'Default Location', 'description': None}], 'organizations': [{'id': 1, 'name': 'Default Organization', 'title': 'Default Organization', 'description': None}]}) 
skipping: [proxy] => (item={'created_at': '2026-07-09 15:07:32 UTC', 'updated_at': '2026-07-09 15:07:32 UTC', 'hosts_count': 0, 'unrecognized_features': [], 'name': 'proxy.example.com-pulp', 'id': 2, 'url': 'https://proxy.example.com/pulp/api/v3/smart_proxy', 'remote_execution_pubkey': None, 'remote_execution_ca_pubkey': None, 'download_policy': 'on_demand', 'supported_pulp_types': ['ansible_collection', 'deb', 'docker', 'file', 'python', 'yum'], 'lifecycle_environments': [], 'features': [{'capabilities': ['core', 'certguard', 'file', 'ansible', 'python', 'deb', 'smart_proxy', 'container', 'rpm'], 'name': 'Pulpcore', 'id': 3}], 'locations': [{'id': 2, 'name': 'Default Location', 'title': 'Default Location', 'description': None}], 'organizations': [{'id': 1, 'name': 'Default Organization', 'title': 'Default Organization', 'description': None}]}) 
skipping: [proxy] => (item={'created_at': '2026-07-09 14:56:46 UTC', 'updated_at': '2026-07-09 14:56:46 UTC', 'hosts_count': 0, 'unrecognized_features': [], 'name': 'quadlet.example.com-pulp', 'id': 1, 'url': 'https://quadlet.example.com/pulp/api/v3/smart_proxy', 'remote_execution_pubkey': None, 'remote_execution_ca_pubkey': None, 'download_policy': 'on_demand', 'supported_pulp_types': ['ansible_collection', 'deb', 'docker', 'file', 'python', 'yum'], 'lifecycle_environments': [{'id': 1, 'name': 'Library', 'description': None, 'library': True, 'organization_id': 1, 'created_at': '2026-07-09 14:53:46 UTC', 'updated_at': '2026-07-09 14:53:46 UTC', 'label': 'Library', 'registry_name_pattern': None, 'registry_unauthenticated_pull': False, 'prior': None, 'prior_id': None, 'organization': 'Default Organization'}], 'features': [{'capabilities': ['core', 'certguard', 'file', 'ansible', 'python', 'deb', 'smart_proxy', 'container', 'rpm'], 'name': 'Pulpcore', 'id': 3}], 'locations': [{'id': 2, 'name': 'Default Location', 'title': 'Default Location', 'description': None}], 'organizations': [{'id': 1, 'name': 'Default Organization', 'title': 'Default Organization', 'description': None}]}) 
skipping: [proxy]
```

especially: it makes the `post_install` message we have not the last thing printed.
```
TASK [post_install : Admin credentials] **********************************************
ok: [quadlet] => 
    msg:
    - Foreman is running at https://quadlet.example.com
```

Could Claude come up with a better solution for our users?

ehelms:
> This now creates a rather huge skipped block at the end of every run:

How does one avoid this with Ansible?



> especially: it makes the `post_install` message we have not the last thing printed.

That's fair, it was not obvious from the code that the "message" part is tied into the flavor element. We could make that code more obvious.

ehelms:
I took a shot at both through a re-factor of the post_install role: https://github.com/theforeman/foremanctl/pull/641

PR #536, opened 2026-06-03
Fixes #535 - Add httpd to foreman.target

qcjames53:
#### Why are you introducing these changes? (Problem description, related links)
Apache was not a part of the foreman.target systemd target, leading to more difficult server administration and testing.

#### What are the changes introduced in this pull request?
- Added httpd to the foreman.target.
- Created a unit test to verify httpd is wanted by and part of the foreman.target.

#### How to test this pull request

Steps to reproduce:
1. Run `./forge vms start`.
2. Run `vagrant ssh quadlet` after VM boot.
3. Wait for services to spin up, then run `systemctl list-dependencies foreman.target` to verify httpd made the list.
4. Validate the unit test and source for quality.

#### Checklist
* [X] Tests added/updated (if applicable)
* [ ] Documentation updated (if applicable) <- Nothing to document imho

PR #561, opened 2026-06-11
Fixes #560 - Add systemctl wants to foreman.target for httpd

### LINKED ISSUE #560: Apache does not start on `systemctl start` with the rest of foreman.target
When running `systemctl start foreman.target`, Apache (httpd) does not start on it's own. I attempted to include this as a part of the fix in https://github.com/theforeman/foremanctl/pull/536 but biffed the implementation.

Correct `src/roles/httpd/tasks/main.yml` such that Apache remains part of the foreman.target, but remove the faulty `[Install]` block ignored by systemd. Use `systemctl add-wants foreman.target httpd.service` as a replacement, as this is standard practice. Ensure that systemd actions work on the foreman.target.

qcjames53:
When running systemctl start foreman.target, Apache (httpd) does not start on it's own. I attempted to include this as a part of the fix in https://github.com/theforeman/foremanctl/pull/536 but biffed the implementation.

* Corrects `foreman.target`'s wants to properly include httpd. The old add-wants block was skipped by systemctl due to httpd being in the `multi-user.target` (higher prio). This was fine unless the user manually wanted to run `systemctl stop foreman.target` and `systemctl start foreman.target`, which might sometimes be the case.

Steps to reproduce:
1. Pull changes and rebuild the VMs.
2. Run `vagrant ssh quadlet` after VM boot.
3. Wait for services to spin up, then run `systemctl list-dependencies foreman.target` to verify httpd made the list.
4. Run `systemctl stop foreman.target` and `systemctl status httpd` to ensure it stopped.
5. Run `systemctl start foreman.target` and `systemctl status httpd` to ensure it is running.
6. Validate unit test changes.

#### Checklist
* [ ] Documentation updated (if applicable) N/A

### DIFF
--- a/src/roles/httpd/tasks/main.yml
+++ b/src/roles/httpd/tasks/main.yml
@@ -103,12 +103,10 @@
     dest: /etc/systemd/system/httpd.service.d/foreman-target.conf
     mode: "0644"
     content: |
+      [Install]
+      WantedBy=default.target foreman.target
       [Unit]
       PartOf=foreman.target
-
-      [Install]
-      WantedBy=foreman.target
-  notify: Restart httpd
 
 - name: Reload systemd daemon
   ansible.builtin.systemd:
--- a/tests/httpd_test.py
+++ b/tests/httpd_test.py
@@ -140,9 +140,13 @@ def test_httpd_headers_use_dashes(server):
     assert cmd.stdout.strip() == '', f"HTTP header names should use dashes, not underscores:\n{cmd.stdout}"
 
 
-def test_httpd_foreman_target_drop_in(server):
+def test_httpd_foreman_target_config(server):
     drop_in = server.file("/etc/systemd/system/httpd.service.d/foreman-target.conf")
     assert drop_in.exists
     assert drop_in.is_file
     assert drop_in.contains("PartOf=foreman.target")
-    assert drop_in.contains("WantedBy=foreman.target")
+    assert drop_in.contains(r"WantedBy=default\.target foreman\.target")
+
+    wants_link = server.file("/etc/systemd/system/foreman.target.wants/httpd.service")
+    assert wants_link.exists
+    assert wants_link.is_symlink


src/roles/httpd/tasks/main.yml
```diff
@@ -106,9 +106,11 @@
       [Unit]
       PartOf=foreman.target
 
-      [Install]
-      WantedBy=foreman.target
-  notify: Restart httpd
+- name: Add httpd to foreman.target wants
```
ehelms: This is how we do it for postgres -- https://github.com/theforeman/foremanctl/blob/master/src/roles/postgresql/tasks/main.yml#L37-L41 I think we are missing the default part.

qcjames53: Funnily enough that is where I stole the original approach from. I'm not a systemd expert by any means but Claude is indicating there's a difference between postgresql and httpd because the former is a quadlet and the latter is an rpm installed drop-in service; allegedly systemd does not merge "[Install]" blocks for drop-in services.

Is httpd intended to also be a quadlet? If so that's a bigger fish to fry.

ehelms: systemd is systemd. And yes, it won't merge the sections, drop-ins **overwrite** the previous instance of it.

Have you tried this as the drop-in?

```
        [Install]
        WantedBy=default.target foreman.target
        [Unit]
        PartOf=foreman.target
```

qcjames53: I'll report back on adding default.

qcjames53: I rebuilt a box from scratch and lo and behold `default.target` fixed everything (`/etc/systemd/system/foreman.target.wants/httpd.service` was properly created on deploy). I've adjusted the role to match this. Thanks Eric!

src/playbooks/deploy/deploy.yaml
```diff
@@ -24,6 +24,7 @@
         certificate_checks_certificate: "{{ server_certificate }}"
         certificate_checks_key: "{{ server_key }}"
         certificate_checks_ca: "{{ server_ca_certificate }}"
+    - role: systemd_target
```
ehelms: Why did this need to move?

qcjames53: The foreman.target needed to be created before the httpd task used it for my previous `add-wants` approach. I just moved it back in the newest push.

ehelms (approved):

tests/httpd_test.py
```diff
@@ -140,9 +140,13 @@ def test_httpd_headers_use_dashes(server):
...
 
 
-def test_httpd_foreman_target_drop_in(server):
+def test_httpd_foreman_target_config(server):
     drop_in = server.file("/etc/systemd/system/httpd.service.d/foreman-target.conf")
     assert drop_in.exists
     assert drop_in.is_file
     assert drop_in.contains("PartOf=foreman.target")
-    assert drop_in.contains("WantedBy=foreman.target")
+    assert not drop_in.contains("WantedBy=foreman.target")
```
ehelms: Err sorry, this is kind of a weird assertion. I think if we are going to test anything, then we should test that the `httpd.service` has what we expect.

qcjames53: Agreed. Asked Claude to fix the unit test but missed that it changed this line to something weird. It should assert on the actual string now.

ianballou (approved):
Working nicely for me:

```
[root@katello-production ~]# systemctl list-dependencies foreman.target 
foreman.target
● ├─candlepin.service
● ├─dynflow-sidekiq@orchestrator.service
● ├─dynflow-sidekiq@worker-hosts-queue.service
● ├─dynflow-sidekiq@worker.service
● ├─foreman-proxy.service
× ├─foreman-recurring@daily.service
● ├─foreman-recurring@daily.timer
× ├─foreman-recurring@hourly.service
● ├─foreman-recurring@hourly.timer
○ ├─foreman-recurring@monthly.service
● ├─foreman-recurring@monthly.timer
× ├─foreman-recurring@weekly.service
● ├─foreman-recurring@weekly.timer
● ├─foreman.service
● ├─httpd.service
● ├─postgresql.service
● ├─pulp-api.service
● ├─pulp-content.service
● ├─pulp-worker@1.service
● ├─pulp-worker@2.service
● ├─pulp-worker@3.service
● ├─pulp-worker@4.service
● ├─redis.service
● └─pulp-worker.target
●   ├─pulp-worker@1.service
●   ├─pulp-worker@2.service
●   ├─pulp-worker@3.service
●   └─pulp-worker@4.service
```

ehelms (approved):

ehelms:
The EL10 failure is unrelated.

qcjames53:
Thank you for the merge, Eric :heart:

PR #568, opened 2026-06-12
Add --preserve-plugin-branches flag to deploy-dev

jeremylenz:
## Summary
- Adds `--preserve-plugin-branches` flag to `forge deploy-dev`
- When passed, skips `git checkout` for all plugins, preserving local branches and uncommitted changes across deploy-dev runs
- Useful when developing on a plugin and re-running deploy-dev without losing your working branch

## Usage
```bash
forge deploy-dev --preserve-plugin-branches
```

## Test plan
- [ ] Run `forge deploy-dev` without the flag — plugins should be checked out to upstream HEAD as before
- [ ] Run `forge deploy-dev --preserve-plugin-branches` with a plugin on a local branch — branch should be preserved

### DIFF
--- a/development/playbooks/deploy-dev/metadata.obsah.yaml
+++ b/development/playbooks/deploy-dev/metadata.obsah.yaml
@@ -29,6 +29,10 @@ variables:
   foreman_development_github_username:
     help: GitHub username to add as additional remote for git checkouts
     action: store
+  foreman_development_preserve_plugin_branches:
+    parameter: --preserve-plugin-branches
+    help: Skip git checkout for plugins, preserving local branches and changes.
+    type: Boolean
 
 include:
   - _flavor_features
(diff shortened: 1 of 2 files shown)


jeremylenz:
@ehelms :eyes:  (I can't request your review here, so just tagging you instead)

jeremylenz:
closing: `b44e9e7` on master added `--manage-repos` — when set to false, it skips both the Foreman repo clone/checkout AND all plugin repo clones/checkouts. It also respects the per-plugin
  manage_repo config.

  `--manage-repos false` is a superset of `--preserve-plugin-branches` — it achieves the same goal (don't blow away local plugin branches) and goes further by also skipping the Foreman repo
  checkout. The implementation on master is also cleaner, using a boolean AND with the per-plugin manage_repo setting.

PR #569, opened 2026-06-12
Add cloud-connector as a native foremanctl feature

jeremylenz:
## Summary
- Re-implements the upstream `satellite_operations.cloud_connector` role natively in foremanctl (SAT-45966 / SAT-44641)
- New `cloud-connector` feature gated behind `--add-feature cloud-connector`, with `rh-cloud` dependency
- Installs `rhc` and `yggdrasil-worker-forwarder`, configures the worker, starts `rhcd`, sets `rhc_instance_id` via the Foreman API, and announces to Sources
- Works with both `foremanctl deploy` and `forge deploy-dev`
- Uses OAuth for all Foreman API calls during setup, matching the pattern used by other roles (foreman_proxy, iop_core, pulp, backup)
- Creates a dedicated `cloud_connector_user` service user with limited `dispatch_cloud_requests` permission for runtime worker authentication
- Adds a reusable `foremanctl_api` custom module for OAuth-authenticated API calls where no FAM module exists
- Early pre-checks validate consumer certificate presence and mutual exclusion with the `iop` feature
- Adds Foreman CA to system trust store so the worker binary can verify Foreman's certificate
- Optional `--cloud-connector-http-proxy` flag for environments without direct internet access

**Companion PRs:** 

https://github.com/theforeman/foreman_rh_cloud/pull/1214 (adds the `announce_to_sources` API endpoint)
https://github.com/theforeman/yggdrasil-worker-forwarder/pull/5 (allows specifying a certificate to use)

## Test plan
- [ ] `foremanctl deploy --add-feature cloud-connector` completes successfully
- [ ] `foremanctl features --list-enabled` includes `cloud-connector`
- [ ] `systemctl status rhcd` shows active/running
- [ ] `/etc/rhc/workers/foreman_rh_cloud.toml` has correct content with service user credentials
- [ ] `hammer settings info --name rhc_instance_id` shows the consumer cert CN
- [ ] `hammer user info --login cloud_connector_user` shows the service user with Cloud Connector role
- [ ] Second `foremanctl deploy` is idempotent
- [ ] `forge deploy-dev --add-feature cloud-connector` completes successfully
- [ ] Enabling both `cloud-connector` and `iop` fails early with a clear error


src/roles/check_cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Check cloud-connector prerequisites
+  when: "'cloud-connector' in enabled_features"
+  block:
+    - name: Verify cloud-connector is not used with iop
+      ansible.builtin.assert:
+        that:
+          - "'iop' not in enabled_features"
```
ehelms: We should be able to enforce this at the parameter level that these two features cannot co-exist.

src/roles/check_cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,38 @@
...
+        fail_msg: >-
+          The cloud-connector feature cannot be used together with the iop feature.
+          Remove one of them with --remove-feature before deploying.
+
+    - name: Check that consumer certificate exists
+      ansible.builtin.stat:
+        path: /etc/pki/consumer/cert.pem
+      register: __cloud_connector_consumer_cert
+
+    - name: Verify consumer certificate exists
```
ehelms: Is this sufficient? What if the system is registered to somewhere else for example? What if the system has a consumer cert but it's stale or cannot reach `console.redhat.com` ?

src/roles/check_cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,38 @@
...
+          - __cloud_connector_consumer_cert.stat.exists
+        fail_msg: >-
+          /etc/pki/consumer/cert.pem not found.
+          The system must be registered with subscription-manager.
+
+    - name: Check that yggdrasil-worker-forwarder package is available
+      ansible.builtin.command: dnf info yggdrasil-worker-forwarder
+      changed_when: false
+      failed_when: false
+      register: __cloud_connector_pkg_check
```
ehelms: Is it one or two underscores for internal variables? I thought we had used one before. @evgeni do you know what our standard should be?

src/roles/cloud_connector/defaults/main.yaml
```diff
@@ -0,0 +1,5 @@
+---
+cloud_connector_url: "{{ foreman_url }}"
+cloud_connector_user: "{{ foreman_initial_admin_username }}"
+cloud_connector_password: "{{ foreman_initial_admin_password }}"
```
ehelms: Avoid referencing other role variables in a role.

src/roles/cloud_connector/handlers/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Restart rhcd
+  ansible.builtin.service:
+    name: rhcd
+    state: restarted
+    daemon_reload: true
```
ehelms, line 6: Does this work?

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,103 @@
+---
+- name: Install rhc and yggdrasil-worker-forwarder
+  ansible.builtin.package:
+    name:
+      - rhc
```
ehelms, line 5: Is this available in CentOS stream 9 now?

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - yggdrasil-worker-forwarder
+    disable_plugin: foreman-protector
+
+- name: Create workers directory
+  ansible.builtin.file:
+    state: directory
+    path: /etc/rhc/workers
```
ehelms: I am surprised the `rhc` package doesn't create this.

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,103 @@
...
+    owner: root
+    group: root
+    mode: '0755'
+
+- name: Configure rhc-cloud-connector-worker
```
ehelms: This name seems inconsistent with the next step that references foreman-rh-cloud-worker

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,103 @@
...
+    dest: /usr/libexec/rhc/foreman-rh-cloud-worker
+    content: |
+      #!/bin/bash
+
+      CONFIG_FILE="{{ cloud_connector_config_file }}" exec /usr/libexec/yggdrasil-worker-forwarder
+    owner: root
+    group: root
+    mode: '0755'
+
+- name: Add Foreman CA to system trust store
```
ehelms: Is this how we do it today? This is an anti-pattern we have been trying to avoid.

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,103 @@
...
+    name: rhcd
+    state: started
+    enabled: true
+
+- name: Read client ID from CN of consumer certificate
+  ansible.builtin.command: openssl x509 -in /etc/pki/consumer/cert.pem -subject -noout
+  register: __cloud_connector_cert_output
+  changed_when: false
+
+- name: Set rhc_instance_id in Foreman
```
ehelms, line 54: Is there a Foreman ansible module for managing settings?

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,103 @@
...
+    password: "{{ cloud_connector_password }}"
+    body:
+      setting:
+        value: true
+    method: PUT
+    ca_path: "{{ foreman_ca_certificate }}"
+    force_basic_auth: true
+    body_format: json
+
+- name: Announce Satellite to Sources
```
ehelms: Satellite? 😁

src/roles/cloud_connector/handlers/main.yaml
jeremylenz: Yes, it seems to work fine in my tests.

src/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
```diff
@@ -0,0 +1,8 @@
+exec = "/usr/libexec/yggdrasil-worker-forwarder"
+protocol = "grpc"
+env = [
+  "FORWARDER_USER={{ cloud_connector_user }}",
+  "FORWARDER_PASSWORD={{ cloud_connector_password }}",
```
ehelms: This is the part we still need to work out how to properly handle, that is, the authentication.

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: I am not sure. My dev box is RHEL.

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: `foreman-installer` does the same thing via `puppet-certs`:

- [templates/rhsm-katello-reconfigure.erb](https://github.com/theforeman/puppet-certs/blob/master/templates/rhsm-katello-reconfigure.erb) — copies to `/etc/pki/ca-trust/source/anchors/` and runs `update-ca-trust`
- [manifests/katello.pp](https://github.com/theforeman/puppet-certs/blob/master/manifests/katello.pp) — uses `trusted_ca::ca` to do the same

We need it here because `yggdrasil-worker-forwarder` is a Go binary that uses the OS trust store — no way to pass it a CA path. That said, we could move this into the `certificates` role so it's done once globally instead of per-feature. Would that be cleaner?

jeremylenz: I was apprehensive about this too. I figured we can change it later..
Claude reply above

src/roles/check_cloud_connector/tasks/main.yaml
jeremylenz: What exactly do you mean by parameter? I attemped to solve this with a new PR which I am now realizing I hadn't raised yet, stand by..

jeremylenz: https://github.com/theforeman/foremanctl/pull/570

src/roles/check_cloud_connector/tasks/main.yaml
jeremylenz: Apparently I had an old version of ansible-lint locally which didn't catch this. Updated.

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: If FAM has one, should we be using it? Would we expect foremanctl to come with FAM installed?

src/roles/check_cloud_connector/tasks/main.yaml
jeremylenz: Addressed in 342bcdc.

src/roles/cloud_connector/defaults/main.yaml
jeremylenz: Addressed in 342bcdc.

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: Addressed in 342bcdc.

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: Addressed in 342bcdc.

src/roles/check_cloud_connector/tasks/main.yaml
jeremylenz: All of those things will cause it to break later, either during deploy or after. But this is the same thing the original role does. (I actually added this check so that it will fail earlier than it would otherwise.)

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: It's from the upstream role — we ported it 1:1. The `rhc` package does create it now, but older versions may not have. It's idempotent so I'd lean toward keeping it as a safety net.

src/roles/check_cloud_connector/tasks/main.yaml
ehelms: Would a `subscription-manager status` tell us more?

src/roles/cloud_connector/handlers/main.yaml
ehelms: I do not see `daemon_reload` on: https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/service_module.html

i do see it on https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/systemd_service_module.html#ansible-collections-ansible-builtin-systemd-service-module

I think then, if we use a property of a specific service manager, we should use that service managers module and not the generic one. Everywhere else we reference `systemd` when do this.

src/roles/cloud_connector/tasks/main.yaml
ehelms: Welp, this has to work in both places or use constraints to limit where it can be enabled.

src/roles/cloud_connector/tasks/main.yaml
ehelms: If `rhc` on the platforms we support (EL9) creates this directory via the RPM, then I'd prefer we drop this and let the RPM handle it.

src/roles/cloud_connector/tasks/main.yaml
ehelms: It already does.

src/roles/cloud_connector/tasks/main.yaml
ehelms: > * [templates/rhsm-katello-reconfigure.erb](https://github.com/theforeman/puppet-certs/blob/master/templates/rhsm-katello-reconfigure.erb) — copies to `/etc/pki/ca-trust/source/anchors/` and runs `update-ca-trust`
> * [manifests/katello.pp](https://github.com/theforeman/puppet-certs/blob/master/manifests/katello.pp) — uses `trusted_ca::ca` to do the same

These are both 404s because they were moved quite a bit ago. I think we need to consider udpating yggdrasil-worker-forwarder vs. starting this trend of relying on the system store. Let's phone a friend for another opinion. @evgeni :phone:

jeremylenz: We need to update it anyway to use DBUS (so it can run on RHEL 10), so we can probably just tack that change on there.

src/roles/check_cloud_connector/tasks/main.yaml
jeremylenz: `subscription-manager status` gives us this:
```
# subscription-manager status
+-------------------------------------------+
   System Status Details
+-------------------------------------------+
Overall Status: Registered
Content Access Mode is set to Simple Content Access. This host has access to content, regardless of subscription status.
```

`subscription-manager identity` gives us the actual consumer uuid:
```
# subscription-manager identity
system identity: cf46f013-6455-4ceb-913e-b1af651f098b
name: ip-10-0-198-13.rhos-01.prod.psi.rdu2.redhat.com
org name: 11949999
org ID: 11949999
```
But that command does make an API call.

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: Seems `rhc` is available in Centos Stream 9 - AppStream:

```
$ vagrant ssh quadlet
Last login: Mon Jun 15 16:57:33 2026 from 192.168.122.1
[vagrant@quadlet ~]$ dnf info rhc
CentOS Stream 9 - BaseOS                                                                                                                                     742 kB/s | 8.9 MB     00:12    
CentOS Stream 9 - AppStream                                                                                                                                  3.5 MB/s |  28 MB     00:07    
CentOS Stream 9 - Extras packages                                                                                                                            133 kB/s |  21 kB     00:00    
Foreman nightly                                                                                                                                               13 MB/s | 2.0 MB     00:00    
Foreman plugins nightly                                                                                                                                       15 MB/s | 1.9 MB     00:00    
Katello Nightly                                                                                                                                              318 kB/s | 118 kB     00:00    
Candlepin: an open source entitlement management system.                                                                                                     375 kB/s |  32 kB     00:00    
pulpcore: Fetch, Upload, Organize, and Distribute Software Packages.                                                                                         780 kB/s | 306 kB     00:00    
Puppet 8 Repository el 9 - x86_64                                                                                                                            7.7 MB/s | 3.7 MB     00:00    
Available Packages
Name         : rhc
Epoch        : 1
Version      : 0.2.7
Release      : 1.el9
Architecture : x86_64
Size         : 11 M
Source       : rhc-0.2.7-1.el9.src.rpm
Repository   : appstream
```

jeremylenz:
Pushed updates addressing review feedback (1bfe640):

- **Handler**: switched from `ansible.builtin.service` to `ansible.builtin.systemd_service` since we use `daemon_reload`
- **Workers directory**: removed the task — confirmed `rhc` package on EL9 creates `/etc/rhc/workers/`
- **Settings API**: replaced raw `uri` calls with `theforeman.foreman.setting` module for `rhc_instance_id` and `allow_auto_inventory_upload`
- **Lint fix**: added `noqa: no-static-secrets` on the defaults password (same pattern as `foreman_development`)

src/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
jeremylenz: As I see it our options are
1 - Keep things as they are here, feeding admin/changeme to rhc - Not acceptable because the admin password ends up in the worker config.toml.
2 - Create a service user with limited permissions - Doable in the existing PRs here and in foreman_rh_cloud.
3 - Create a service user + personal access token and use that, same as the previous architecture - FAM doesn't have a module for personal access tokens, so this would be a fair bit of work

I think Option 2 seems promising and am pushing that update now.

src/roles/check_cloud_connector/tasks/main.yaml
jeremylenz: Is https://github.com/theforeman/foremanctl/pull/570 what you had in mind?

jeremylenz:
rebased & fixed conflicts

src/roles/check_cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Verify consumer certificate exists
+      ansible.builtin.assert:
+        that:
+          - check_cloud_connector_consumer_cert.stat.exists
+        fail_msg: >-
+          /etc/pki/consumer/cert.pem not found.
+          The system must be registered with subscription-manager.
+
+    - name: Check that yggdrasil-worker-forwarder package is available
+      ansible.builtin.command: dnf info yggdrasil-worker-forwarder
```
qcjames53: `yggdrasil-worker-forwarder` - Is this available in centos9? Where is it shipped? I know it won't be a problem for shipping pre-built images but for dev environments and such we should really document a pathway for getting this installed.

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,114 @@
...
+    dest: /etc/pki/ca-trust/source/anchors/foreman-ca.pem
+    remote_src: true
+    owner: root
+    group: root
+    mode: '0644'
+  notify: Restart rhcd
+
+- name: Update system CA trust
+  ansible.builtin.command: update-ca-trust
+  changed_when: true
```
qcjames53: Why is this always true? Doesn't this break idempotency?

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,114 @@
...
+- name: Ensure rhcd started and enabled
+  ansible.builtin.service:
+    name: rhcd
+    state: started
+    enabled: true
+
+- name: Read client ID from CN of consumer certificate
+  ansible.builtin.command: openssl x509 -in /etc/pki/consumer/cert.pem -subject -noout
+  register: cloud_connector_cert_output
+  changed_when: false
```
qcjames53: Same as above but opposite. Idempotency?

src/roles/checks/tasks/main.yml
```diff
@@ -3,6 +3,7 @@
   ansible.builtin.include_tasks: execute_check.yml
   loop:
     - check_features
+    - check_cloud_connector
```
qcjames53: I'd say we should follow the format of `check_database_index` and include a task with the when `"'cloud-connector' in enabled_features"` clause as a condition. This would let you get rid of the whole role-sized block in the check file. Feels more ~~pythonic~~ ansible-like

qcjames53 (changes requested):
I'm not enough in the loop to give this a full review but had a few comments that might be helpful. Consider this 80% good points 20% ass-talking. Here's what I was able to test directly: 

- The iop conflict and missing consumer cert checks both work on the positive and negative cases (used manual subscription manager commands for the latter). 
- The `yggdrasil-worker-forwarder` check works in the negative (was unable to figure out package install on centos9).
- The cloud connector role works.

Beyond these tests, I have questions about the whole approach of cloud connector even after reading the planning docs and such. 

1. How are we splitting RHEL and Centos 9 deploys? I think the differentiation needs to be clearer since (best I can tell) `yggdrasil-worker-forwarder` is RHEL only. If the packaged container images are centos9 and the rpms needed for cloud connector are RHEL-only, that seems like a big problem to me!
2. `deploy-dev` needs to work with cloud connector. At minimum a procedure but ideally out of the box. I think we really need a procedure for setting up a centos9 deploy-dev environment to build/test cloud connector. If that's not possible, then we need a good procedure for setting up a RHEL environment as deploy-dev.
3. If it's possible to pass in subscription manager credentials safely, I would greatly prefer if we could deploy an environment with cloud connector immediately instead of requiring manual registration for the quadlet VM, THEN adding the feature. I understand this won't be an issue for customers but this would be HUGE for testing purposes.

I have to gracefully bow out of reviewing this for a few reasons: capacity + artemis stuff + this baby could come **any day**! Sorry to not complete testing but I put in an honest effort.

jeremylenz:
I'll respond with thoughts off the top of my head :)

> 1. How are we splitting RHEL and Centos 9 deploys? I think the differentiation needs to be clearer since (best I can tell) `yggdrasil-worker-forwarder` is RHEL only. If the packaged container images are centos9 and the rpms needed for cloud connector are RHEL-only, that seems like a big problem to me!

In order for cloud connector to work you need available on your base system:
1. `ansible-runner` command - Downstream, from the Satellite repo, e.g. `satellite-6.19-for-rhel-9-x86_64-rpms`. Upstream, from the Foreman Plugins repo, e.g. https://yum.theforeman.org/plugins/3.19/el9/x86_64/ 
2. `yggdrasil-worker-forwarder` - Downstream, from the Satellite repo. Upstream, from the Foreman Plugins repo, e.g. https://yum.theforeman.org/plugins/3.19/el9/x86_64/
3.  `rhc` - from AppStream (CentOS or RHEL)

In the previous architecture, these weren't even listed as prerequisites, even in upstream docs. Downstream, you'd already have all of these since the Satellite system must be registered and have those repos enabled. But upstream, anything can happen. It seems this is a documentation gap. cc @Lennonka

> 2. `deploy-dev` needs to work with cloud connector. At minimum a procedure but ideally out of the box. I think we really need a procedure for setting up a centos9 deploy-dev environment to build/test cloud connector. If that's not possible, then we need a good procedure for setting up a RHEL environment as deploy-dev.

I was testing with deploy-dev on RHEL (with `--target-host localhost`). I think it works fine, as long as you have the repos available that I mentioned above.

src/roles/check_cloud_connector/tasks/main.yaml
jeremylenz: I checked and it's in https://yum.theforeman.org/plugins/3.19/el9/x86_64/

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: This is correct — `changed_when: false` means "this command never changes anything," which is right for a read-only `openssl x509` call. Standard Ansible pattern for info-gathering tasks.

jeremylenz:
@qcjames53 Addressed 2 comments and replied to the third

src/roles/cloud_connector/defaults/main.yaml
```diff
@@ -0,0 +1,7 @@
+---
+cloud_connector_url: "https://{{ ansible_facts['fqdn'] }}"
+cloud_connector_admin_user: admin
+cloud_connector_admin_password: changeme # noqa: no-static-secrets
+cloud_connector_service_user: cloud_connector_admin_user
```
lfu: Should this be: `cloud_connector_user`?

jeremylenz: We are creating a new user with limited permissions, e.g. a "service user."

lfu: When creating the service user ([tasks/main.yaml:22-24](vscode-webview://0tmje2tc8aic0n40p7usskrc52nac0i0usqt2f630mnajn6e3343/src/roles/cloud_connector/tasks/main.yaml#L22-L24)):

```
- name: Create cloud connector service user
  theforeman.foreman.user:
    login: "{{ cloud_connector_service_user }}"  # ← Expands to what?
    user_password: "{{ cloud_connector_service_password }}"
```
With current defaults: Creates user with login "cloud_connector_admin_user" ❌
With base.yaml override: Creates user with login "cloud_connector_user" ✅

jeremylenz: good catch! Updated :+1:

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: https://github.com/theforeman/yggdrasil-worker-forwarder doesn't seem to use Github Issues, so I will create an internal Jira for that.

jeremylenz: I updated the description of https://redhat.atlassian.net/browse/SAT-27307. However, ideally I would like this PR to be merged and use the "incorrect" approach temporarily, to move things forward,.

lfu:
Tested cloud connector feature in development mode. All core functionality works correctly. Found 3 necessary configuration changes for development environment.

---

## 🔍 Testing Methodology

**Environment:** Development mode using `forge deploy-dev` with Rails source code
**Test Scope:** 
- Cloud connector deployment
- Service user authentication
- API endpoint functionality
- Worker configuration

## 🔗 Related/Dependent PRs

### Foreman RH Cloud Plugin
- **PR:** https://github.com/theforeman/foreman_rh_cloud/pull/1214
- **Purpose:** Cloud connector service user and role support
- **Status:** Pending
- **Dependency:** Required for cloud connector service user authentication functionality

### Foremanctl Branch Preservation
- **PR:** https://github.com/theforeman/foremanctl/pull/568
- **Purpose:** Preserve plugin branches during development deployment
- **Status:** Pending
- **Dependency:** Needed for `--preserve-plugin-branches` flag used in testing

**Note:** These PRs should be merged before or alongside this cloud connector feature PR.

---

## ✅ What Works

### 1. Cloud Connector Deployment
- ✓ Cloud connector role created successfully
- ✓ Required packages present (`rhc`, `yggdrasil-worker-forwarder`) - installed as prerequisite
- ✓ Worker config file created: `/etc/rhc/workers/foreman_rh_cloud.toml`
- ✓ rhcd service running and enabled

**Note:** The cloud connector role does not install packages - it assumes `rhc` and `yggdrasil-worker-forwarder` are already present (installed manually or by prerequisite roles).

### 2. Service User Authentication
Created service user `cloud_connector_user` with proper credentials:
```
User ID: 5
Role: Cloud Connector
Auth Source: Internal
Last Login: Successfully authenticated via API
```

### 3. API Endpoint Verification
Manual test of `/api/v2/rh_cloud/cloud_request` endpoint:

**Request:**
```bash
curl -u cloud_connector_user:$PASSWORD \
  -X POST \
  http://localhost:3000/api/v2/rh_cloud/cloud_request \
  -H 'Content-Type: application/json' \
  -d '{"directive":"foreman_rh_cloud", ...}'
```

**Result:**
```
✓ Authenticated user cloud_connector_user against INTERNAL authentication source
✓ Authorized user cloud_connector_user(cloud_connector_user)
✓ Request processed (playbook URL validation occurred as expected)
```

### 4. Configuration Validation
- ✓ `rhc_instance_id` setting configured: `5fc35187-ee43-48a8-b61a-01c0153d0562`
- ✓ Worker URL correct: `http://localhost:3000/api/v2/rh_cloud/cloud_request`
- ✓ Credentials properly stored in worker config

---

## 🔧 Required Code Changes

Found 3 necessary changes for development environment support:

### Change 1: Development URL Configuration
**File:** `development/playbooks/deploy-dev/deploy-dev.yaml`

**Issue:** Development uses Rails dev server on port 3000, not containerized Foreman

**Fix:**
```yaml
vars:
  cloud_connector_url: "http://localhost:3000"
  cloud_connector_admin_user: "{{ foreman_development_admin_user }}"
  cloud_connector_admin_password: "{{ foreman_development_admin_password }}"
```

**Why:** Points worker to the correct development Rails server endpoint

---

### Change 2: Disable Certificate Validation for HTTP
**File:** `src/roles/cloud_connector/tasks/main.yaml`

**Issue:** Development uses HTTP (no SSL), certificate validation fails

**Fix:**
```yaml
- name: Create cloud connector service user
  theforeman.foreman.user:
    # ... existing parameters ...
    validate_certs: false  # <-- Add this
```

**Why:** HTTP connections don't have valid certificates in development

---

### Change 3: Make CA Certificate Path Optional
**File:** `src/roles/cloud_connector/tasks/main.yaml`

**Issue:** No CA certificate exists for HTTP connections, undefined variable error

**Fix:**
```yaml
- name: Create cloud connector service user
  theforeman.foreman.user:
    # ... existing parameters ...
    ca_path: "{{ foreman_ca_certificate | default(omit) }}"  # <-- Change this
```

**Why:** Prevents Ansible error when `foreman_ca_certificate` is undefined (HTTP mode)

---

## ⚠️ Known Limitations

### Automated pytest Suite Skipped

**Issue:** All 6 cloud connector tests were skipped:
```
tests/cloud_connector_test.py::test_rhc_package_installed SKIPPED
tests/cloud_connector_test.py::test_yggdrasil_worker_forwarder_package_installed SKIPPED
tests/cloud_connector_test.py::test_workers_directory_exists SKIPPED
tests/cloud_connector_test.py::test_worker_config_exists SKIPPED
tests/cloud_connector_test.py::test_worker_script_exists SKIPPED
tests/cloud_connector_test.py::test_rhcd_service_running SKIPPED
```

**Root Cause:** 
- pytest hardcoded to check `./foremanctl features` (production state)
- Development uses `./forge deploy-dev` (separate state)
- Feature flags from `--add-feature` are temporary, not persisted

**Impact:** 
- ❌ Automated test suite cannot verify development deployments
- ✅ Manual verification confirms all test criteria pass

---

**Testing completed with:** Claude Code

jeremylenz:
Re: Change 1 - cloud_connector_url defaults to foreman_url which is https://{{ fqdn }},
  and httpd proxies to the Rails server. The HTTPS URL through httpd works fine in dev — we tested it. Using http://localhost:3000 directly would bypass httpd and also break the ca_path on
  all the FAM module calls. The worker TOML would also have an HTTP URL, which means the runtime cloud requests would go over unencrypted HTTP.

Change 2 is not necessary unless I were to do Change 1, which I think is not needed for the reason above.

Updating now with Change 3.

jeremylenz:
Btw, Our role does install the packages — line 2-7 of tasks/main.yaml:

```yaml
  - name: Install rhc and yggdrasil-worker-forwarder
    ansible.builtin.package:
      name:
        - rhc
        - yggdrasil-worker-forwarder
```

src/roles/check_cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+  changed_when: false
+  failed_when: false
+  register: check_cloud_connector_pkg_check
+
+- name: Verify yggdrasil-worker-forwarder is available
+  ansible.builtin.assert:
+    that:
+      - check_cloud_connector_pkg_check.rc == 0
+    fail_msg: >-
+      The yggdrasil-worker-forwarder package is not available.
```
ehelms: The role will try to install `yggdrasil-worker-forwarder` which, if not available, will fail on it's own. Therefore, I do not think we need this check here. Let's just let the `package` module do it's job.

src/roles/checks/tasks/main.yml
```diff
@@ -7,6 +7,11 @@
     - check_database_connection
     - check_system_requirements
 
+- name: Run cloud connector checks
+  ansible.builtin.include_role:
+    name: check_cloud_connector
+  when: "'cloud-connector' in enabled_features"
```
ehelms:
```suggestion
  when: enabled_features | has_feature('cloud-connector')
```

src/roles/cloud_connector/tasks/http_proxy.yaml
```diff
@@ -0,0 +1,17 @@
...
+    group: root
+    mode: '0755'
+
+- name: Deploy HTTP proxy systemd drop-in for rhcd
+  ansible.builtin.template:
+    src: proxy.conf.j2
+    dest: /etc/systemd/system/rhcd.service.d/proxy.conf
+    owner: root
+    group: root
+    mode: '0644'
```
ehelms:
```suggestion
    mode: '0640'
```

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,110 @@
...
+- name: Create rhcd worker script
+  ansible.builtin.copy:
+    dest: /usr/libexec/rhc/foreman-rh-cloud-worker
+    content: |
+      #!/bin/bash
+
+      CONFIG_FILE="{{ cloud_connector_config_file }}" exec /usr/libexec/yggdrasil-worker-forwarder
+    owner: root
+    group: root
+    mode: '0755'
```
ehelms, line 41: Do we want this world-readable and executable?

src/roles/cloud_connector/tasks/main.yaml
ehelms: The problem is -- if this makes this change, and then another feature relies on it, we end up with a snowball effect of things relying on this paradigm. I think we need to come to a consensus about whether we configure the system trust with our CA or not.

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,110 @@
...
+  when: foreman_ca_certificate is defined
+
+- name: Ensure rhcd started and enabled
+  ansible.builtin.service:
+    name: rhcd
+    state: started
+    enabled: true
+
+- name: Read client ID from CN of consumer certificate
+  ansible.builtin.command: openssl x509 -in /etc/pki/consumer/cert.pem -subject -noout
```
ehelms: Try this -- https://docs.ansible.com/projects/ansible/latest/collections/community/crypto/x509_certificate_info_module.html#ansible-collections-community-crypto-x509-certificate-info-module

src/roles/cloud_connector/defaults/main.yaml
```diff
@@ -0,0 +1,7 @@
+---
+cloud_connector_url: "https://{{ ansible_facts['fqdn'] }}"
+cloud_connector_admin_user: admin
+cloud_connector_admin_password: changeme # noqa: no-static-secrets
```
ehelms: Relying on this won't scale, because an admin can change this at any time and there is no way to input this value right now. We either need to add that as a paradigm or change the auth method for `foremanctl` which is a broader problem than just this PR. Let me start a discussion on this.

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -57,6 +62,12 @@
       vars:
         iop_core_foreman_oauth_consumer_key: "{{ foreman_oauth_consumer_key }}"
         iop_core_foreman_oauth_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+    - role: cloud_connector
+      when:
+        - "'cloud-connector' in enabled_features"
+      vars:
+        cloud_connector_admin_user: "{{ foreman_development_admin_user }}"
+        cloud_connector_admin_password: "{{ foreman_development_admin_password }}"
```
nofaralfasi: Should we also enable `foreman_rh_cloud` in `foreman_development_enabled_plugins` when cloud-connector is enabled (similar to the iop block above)?

I want to confirm deploy-dev won’t run the `cloud_connector` role before the plugin is available in the dev Foreman setup.

src/roles/cloud_connector/tasks/http_proxy.yaml
```diff
@@ -0,0 +1,17 @@
...
+  notify: Restart rhcd
```
nofaralfasi, line 17: Do we support unsetting `cloud_connector_http_proxy` after it was set once? If yes, we should add removal of `proxy.conf`. If not, can we document that this value is persisted/sticky?

src/roles/cloud_connector/defaults/main.yaml
jeremylenz: Ok, see what you think of the latest commit

jeremylenz:
We decided that foremanctl should not rely on the system CA trust store; rather, the applications consuming the certificate must be configured to point to the certificate directly. This means the consuming application (yggdrasil in this case) must be _able_ to be configured in this way.

Converting to draft while we wait for that to be implemented in yggdrasil.

src/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
jeremylenz: @ehelms Updated to use OAuth, but we still need the service user because yggdrasil-worker-forwarder only supports HTTP basic auth, and doesn't support OAuth.

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -14,6 +14,11 @@
     httpd_foreman_backend: "http://localhost:3000"
     pulp_register_foreman_proxy: false
   pre_tasks:
+    - name: Check cloud-connector prerequisites
+      ansible.builtin.include_role:
+        name: check_cloud_connector
+      when: "'cloud-connector' in enabled_features"
```
ehelms: You can use the `has_feature` filter.

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,122 @@
...
+- name: Enable automatic inventory upload
+  theforeman.foreman.setting:
+    name: allow_auto_inventory_upload
+    value: true
+    server_url: "{{ cloud_connector_url }}"
+    oauth1_consumer_key: "{{ foreman_oauth_consumer_key }}"
+    oauth1_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+    ca_path: "{{ foreman_ca_certificate | default(omit) }}"
+
+- name: Announce to Sources
```
ehelms, line 63: What does announce to sources do?

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,122 @@
...
+    name: rhc_instance_id
+    value: "{{ cloud_connector_cert_info.subject.commonName }}"
+    server_url: "{{ cloud_connector_url }}"
+    oauth1_consumer_key: "{{ foreman_oauth_consumer_key }}"
+    oauth1_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+    ca_path: "{{ foreman_ca_certificate | default(omit) }}"
+
+- name: Enable automatic inventory upload
+  theforeman.foreman.setting:
+    name: allow_auto_inventory_upload
```
ehelms: Why do we have to set this? Is the default value false?

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: It's an API call made to the cloud to identify this Foreman instance as a source in a service that was previously called Sources - I'm not sure what the name of the service is now. My understanding is that it's needed in order for cloud remediations to show your registered host as "connected"

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: Setting up cloud connector has always required turning on this setting, and in the previous architecture there was a message in the UI to that effect. The default is true, but the user may have turned it off and then tried to enable cloud connector.

jeremylenz: cc @ShimShtein in case you can add more color here

src/roles/cloud_connector/tasks/main.yaml
evgeni: Yeah, seems my ☎️ was broken and I never seen this ping.

But I do agree with Eric, we do not want to add certs to the system trust, as
- it's not inherited by containers, so they need the explicit file passed anyway
- it makes tools we do not manage trust this CA
- it's a pattern we are trying to avoid as it makes things inconsistent (because of the above)

jeremylenz:
> We decided that foremanctl should not rely on the system CA trust store; rather, the applications consuming the certificate must be configured to point to the certificate directly. This means the consuming application (yggdrasil in this case) must be _able_ to be configured in this way.

https://github.com/theforeman/yggdrasil-worker-forwarder/pull/5

jeremylenz:
this also required a change to https://github.com/theforeman/foreman_rh_cloud/pull/1214, so be sure to pull that too

jeremylenz:
Rebased & updated to use the new "mutually exclusive features" feature.

jeremylenz:
test failure seems unrelated

lfu:
## Test Results - Round 2

**Environment:** Development mode (`forge deploy-dev`) on `jeremylenz_cloud-connector-feature` branch
**Date:** 2026-07-22

### Automated Tests: 5 passed, 1 failed

```
tests/cloud_connector_test.py::test_rhc_package_installed PASSED
tests/cloud_connector_test.py::test_yggdrasil_worker_forwarder_package_installed PASSED
tests/cloud_connector_test.py::test_workers_directory_exists PASSED
tests/cloud_connector_test.py::test_worker_config_exists PASSED
tests/cloud_connector_test.py::test_worker_script_exists FAILED
tests/cloud_connector_test.py::test_rhcd_service_running PASSED
```

---

## Issues Found

### 1. Worker script mode mismatch (test failure)

**Files:**
- `src/roles/cloud_connector/tasks/main.yaml` line 53: sets `mode: '0750'`
- `tests/cloud_connector_test.py` line 31: asserts `script.mode == 0o755`

The task creates the worker script with mode `0750` but the test expects `0755`. Either the task or the test needs to be updated to match.

---

### 2. deploy-dev `httpd_foreman_backend` override is silently ignored

**File:** `development/playbooks/deploy-dev/deploy-dev.yaml`

Line 7 sets `httpd_foreman_backend: "http://localhost:3000"` under `vars:`, but `src/vars/base.yaml` line 27 sets it to `unix:///run/httpd.foreman.sock|http://foreman` and is loaded via `vars_files:` at line 16. In Ansible, `vars_files` takes higher precedence than `vars`, so the dev override is silently ignored.

This causes httpd to return 503 for all requests in dev mode because the Unix socket does not exist (Foreman runs as a direct Rails process, not in a container).

**Workaround:** `--extra-vars httpd_foreman_backend=http://localhost:3000`

**Fix:** Move the override to a higher-precedence location such as `set_fact` in `pre_tasks`.

---

### 3. Greedy directory removal for HTTP proxy drop-in

**File:** `src/roles/cloud_connector/tasks/main.yaml` lines 107-111

```yaml
- name: Remove systemd drop-in directory if empty
  ansible.builtin.file:
    path: /etc/systemd/system/rhcd.service.d
    state: absent
  notify: Restart rhcd
```

`ansible.builtin.file` with `state: absent` on a directory removes it and all its contents unconditionally — it does not check if the directory is empty despite the task name. If another role or package adds files to `/etc/systemd/system/rhcd.service.d/`, they will be deleted.

Additionally, the `notify: Restart rhcd` on this task is redundant — the preceding task (removing `proxy.conf`) already notifies the same handler.

jeremylenz:
> 2. deploy-dev httpd_foreman_backend override is silently ignored
development/playbooks/deploy-dev/deploy-dev.yaml

> Line 7 sets httpd_foreman_backend: "http://localhost:3000" under vars:, but src/vars/base.yaml line 27 sets it to unix:///run/httpd.foreman.sock|http://foreman and is loaded via vars_files: at line 16. In Ansible, vars_files takes higher precedence than vars, so the dev override is silently ignored.

> This causes httpd to return 503 for all requests in dev mode because the Unix socket does not exist (Foreman runs as a direct Rails process, not in a container).

> Workaround: --extra-vars httpd_foreman_backend=http://localhost:3000

> Fix: Move the override to a higher-precedence location such as set_fact in pre_tasks.

This issue is unrelated to these changes. Opened https://github.com/theforeman/foremanctl/pull/696

jeremylenz:
@lfu Updated to address issue 1 and 3.

lfu:
## Test Results

**Branch:** `jeremylenz_cloud-connector-feature`
**Date:** 2026-07-22
**Environments tested:** Development (`forge deploy-dev`) and Production (`foremanctl deploy --tuning development`)

### Automated Tests: 6 passed, 0 failed (after fixes)

```
tests/cloud_connector_test.py::test_rhc_package_installed PASSED
tests/cloud_connector_test.py::test_yggdrasil_worker_forwarder_package_installed PASSED
tests/cloud_connector_test.py::test_workers_directory_exists PASSED
tests/cloud_connector_test.py::test_worker_config_exists PASSED
tests/cloud_connector_test.py::test_worker_script_exists PASSED
tests/cloud_connector_test.py::test_rhcd_service_running PASSED
```

### Production Deploy

`foremanctl deploy --initial-admin-password=changeme --add-feature cloud-connector --tuning development` completed successfully except for the `Announce to Sources` task, which returned 404 on `/api/v2/rh_cloud/announce_to_sources`. This endpoint requires the dependent [foreman_rh_cloud#1214](https://github.com/theforeman/foreman_rh_cloud/pull/1214) PR to be merged.

**Note:** The stock `yggdrasil-worker-forwarder` RPM does not include `FORWARDER_CA_FILE` support. We replaced it with a build from [yggdrasil-worker-forwarder](https://github.com/theforeman/yggdrasil-worker-forwarder) `main` (which includes the merged [#5](https://github.com/theforeman/yggdrasil-worker-forwarder/pull/5)). The worker config correctly sets `FORWARDER_CA_FILE` pointing to the Foreman CA bundle, and rhcd starts without errors. End-to-end request forwarding could not be tested because the `cloud_request` endpoint requires [foreman_rh_cloud#1214](https://github.com/theforeman/foreman_rh_cloud/pull/1214).

---

## Test Plan Coverage

| Criterion | Status | Notes |
|-----------|--------|-------|
| `foremanctl deploy --add-feature cloud-connector` completes | PASS | Completed except `announce_to_sources` (expected, depends on foreman_rh_cloud#1214) |
| `foremanctl features --list-enabled` includes cloud-connector | PASS | cloud-connector listed as enabled |
| `systemctl status rhcd` shows active/running | PASS | `test_rhcd_service_running` passed in both environments |
| `/etc/rhc/workers/foreman_rh_cloud.toml` has correct content | PASS | `test_worker_config_exists` passed in both environments |
| `hammer settings info --name rhc_instance_id` shows consumer cert CN | PASS | Shows `d8ae42ed-2714-4338-ba23-eb76ed69d218` |
| `hammer user info --login cloud_connector_user` shows service user | PASS | User ID 5, Cloud Connector role assigned, Internal auth |
| Second `foremanctl deploy` is idempotent | PASS | Same result as first deploy (fails only at `announce_to_sources`) |
| `forge deploy-dev --add-feature cloud-connector` completes | PASS | Completed with httpd workaround (see issue #2) |
| Enabling both cloud-connector and iop fails early | PASS | Fails with clear error: "cloud-connector conflicts with iop" |

lfu:
## Round 3: End-to-End Testing with console.redhat.com

Successfully validated the full cloud connector end-to-end flow in **development mode** on a remote RHEL 9.8 host. A remediation triggered from console.redhat.com was received by Foreman, dispatched via remote execution, and executed successfully on a managed host.

---

## Issues Encountered During Testing

### 1. SELinux blocks httpd to port 3000

httpd cannot connect to the Rails dev server on port 3000 due to SELinux policy.

**Fix:** `setsebool -P httpd_can_network_connect on`

### 2. yggdrasil-worker-forwarder RPM not available

The stock `yggdrasil-worker-forwarder` package is not in any standard repo. Must be built from the [theforeman/yggdrasil-worker-forwarder](https://github.com/theforeman/yggdrasil-worker-forwarder) `main` branch (which includes `FORWARDER_CA_FILE` support).

The cloud_connector role's package install task fails. Added `ignore_errors: true` as a workaround for testing.

### 3. pulp-worker container lost after disk full

When the disk filled up, the pulp-worker container was lost and had to be restarted with `systemctl start pulp-worker@1`.

### 4. smart-proxy-development service not started

The smart proxy service was disabled and not running. Required `systemctl enable --now smart-proxy-development` for remote execution to work.

---


## Conclusion

The cloud connector feature works end-to-end in development mode. A remediation initiated from console.redhat.com was successfully delivered to Foreman via the cloud connector (rhcd + yggdrasil-worker-forwarder), dispatched to a managed host via remote execution, and completed successfully.

The foremanctl `cloud-connector` feature successfully replaces the "Configure Cloud Connector" function in the Foreman UI.

---

**Testing completed with:** Claude Code
**Deployment mode:** Development (`forge deploy-dev`) on remote RHEL 9.8 host

ehelms:
Here is the RPM update for yggdrasil-worker-forwarder: https://github.com/theforeman/foreman-packaging/pull/13772

src/roles/cloud_connector/library
```diff
@@ -0,0 +1 @@
+../../plugins/modules
```
ehelms: This shouldn't be necessary.

jeremylenz: Removed the symlink and added `library = ../src/plugins/modules` to `development/ansible.cfg` instead — same pattern it already uses for `roles_path` and `filter_plugins`. The development lint needs this to resolve `foremanctl_api` since `deploy-dev` imports the `cloud_connector` role from `src/roles/`.

src/plugins/modules/foremanctl_api.py
ekohl, line 1: Why implement this if we already rely on [apypie](https://github.com/Apipie/apypie) that already implements it (since https://github.com/Apipie/apypie/commit/69b898bd05ae6286c79465f08d443f90f76903b3)?

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,117 @@
...
+  theforeman.foreman.role:
+    name: Cloud Connector
+    filters:
+      - permissions:
+          - dispatch_cloud_requests
+    server_url: "{{ cloud_connector_url }}"
+    oauth1_consumer_key: "{{ foreman_oauth_consumer_key }}"
+    oauth1_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+    ca_path: "{{ foreman_ca_certificate | default(omit) }}"
+    state: present
```
ekohl: Shouldn't `foreman_rh_cloud` seed this role to exist?

src/plugins/modules/foremanctl_api.py
jeremylenz: FAM doesn't have a module that allows arbitrary API requests with a POST. But I can refactor our custom module here to use apypie.

tests/feature/cloud-connector/cloud_connector_test.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+
+pytestmark = pytest.mark.feature("cloud-connector")
```
ekohl, line 3: Please use the feature directory functionality and move this to `tests/feature/cloud-connector`.

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,117 @@
...
+  theforeman.foreman.setting:
+    name: rhc_instance_id
+    value: "{{ cloud_connector_cert_info.subject.commonName }}"
+    server_url: "{{ cloud_connector_url }}"
+    oauth1_consumer_key: "{{ foreman_oauth_consumer_key }}"
+    oauth1_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+    ca_path: "{{ foreman_ca_certificate | default(omit) }}"
+
+- name: Enable automatic inventory upload
+  theforeman.foreman.setting:
```
ekohl: foremanctl shouldn't use `theforeman.foreman.setting` because a user can flip it back. If you want to mandate it, set it via `settings.yaml` so it becomes read-only via the UI/API.

jeremylenz: In the previous architecture it was never actually mandated; it was just a warning in the UI ("Cloud Connector has been configured however the inventory auto-upload is disabled, it's recommended to enable it"). So I don't think I want to mandate it here either.

ekohl: I want to be consistent: architecturally if foremanctl configures something, it must stay that way. The old foreman_rh_cloud never fit into the rest of the architecture so I don't want to keep that as a reference.

ekohl: And to elaborate: we will tell the user to run `foremanctl` to make changes. That will again change the value to `true`. That is a poor surprise for the user. Because of that there's IMHO a hard rule: either foremanctl takes complete control over a setting, or doesn't touch it at all.

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: Updated the foreman_rh_cloud PR to seed the role. So now all we do here is assign the user. :+1:

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: In that case I would say we just remove the setting, uh, setting here, and document that you should probably have it turned on. Thoughts?

jeremylenz: ~~Or what if the feature deploy failed if the setting is off; that's another option~~ thinking about it some more, I think just removing it is better. The consequence if the setting is off is that hosts that are added afterward won't show up in cloud inventory, and thus you won't be able to run cloud-initiated remediations thru cloud connector for those hosts. So it doesn't strictly break _everything_ if you turn it off - just newly added hosts.

jeremylenz:
@ehelms @ekohl @lfu Any more comments here?

src/plugins/modules/foremanctl_announce_to_sources.py
ekohl: I'd prefer a specific module for the `announce_to_sources` API call. Internally it can call [wait_for_task](https://github.com/Apipie/apypie/blob/01ab54706c7aae2a5e7593cf11407bcd809993d4/apypie/foreman.py#L101-L119) to ensure it actually completed successfully (or failed). That way we clearly manage the state instead of it being a best effort thing.

src/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
```diff
@@ -0,0 +1,10 @@
+exec = "/usr/libexec/yggdrasil-worker-forwarder"
+protocol = "grpc"
+env = [
+  "FORWARDER_USER={{ cloud_connector_service_user }}",
+  "FORWARDER_PASSWORD={{ cloud_connector_service_password }}",
+  "FORWARDER_URL={{ cloud_connector_url }}/api/v2/rh_cloud/cloud_request",
+  "FORWARDER_HANDLER=foreman_rh_cloud"{% if foreman_ca_certificate is defined %},
+  "FORWARDER_CA_FILE={{ foreman_ca_certificate }}"{% endif %}
```
ekohl: Perhaps just use trailing commas to keep it simple?
```suggestion
  "FORWARDER_HANDLER=foreman_rh_cloud",
  {% if foreman_ca_certificate is defined %}
  "FORWARDER_CA_FILE={{ foreman_ca_certificate }}",
  {% endif %}
```

ekohl (changes requested):
Marking as request changes because I think this needs some fundamental things:
* Stop using `theforeman.foreman.setting`
* Resolve the registration task discussion

src/plugins/modules/foremanctl_announce_to_sources.py
```diff
@@ -78,31 +62,34 @@ def run_module():
...
-        changed=method != 'get',
-        url=url,
-        body=response_body,
-    )
+        # apypie caches the apidoc on disk, keyed by an Apipie-Checksum header that
+        # doesn't reliably change when a plugin adds/renames an action (observed:
+        # the checksum stayed identical after announce_to_sources was added to the
+        # rh_cloud inventory controller). Force a fresh fetch so a stale on-disk
+        # apidoc from before this feature was installed can't hide the new action.
+        api.clean_cache()
```
ekohl: I think in production we should be able to rely on the cache to be consistent. Also, please file a file bug to [apipie-rails](https://github.com/Apipie/apipie-rails).

jeremylenz:
> Marking as request changes because I think this needs some fundamental things:
> 
> * Stop using `theforeman.foreman.setting`
> * Resolve the registration task discussion

This no longer uses `theforeman.foreman.setting` so that part is resolved IMO. See [comment above](https://github.com/theforeman/foremanctl/pull/569#discussion_r3736755346)

Added a new commit with the purpose-built module for announce to sources. Tested again and deploy succeeds, so I'm happy with it if you are.

src/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
jeremylenz: updated :+1:

jeremylenz:
rebased

jeremylenz:
:green_apple:

jeremylenz:
@ekohl Are your concerns addressed for now? @ehelms Can we get this in soon? I'll be away starting 5 September so would like to get it merged before then; also having it in will help us with the DBUS rewrite of yggdrasil-worker-forwarder.

src/roles/cloud_connector/defaults/main.yaml
```diff
@@ -0,0 +1,5 @@
+---
+cloud_connector_url: "https://{{ ansible_facts['fqdn'] }}"
+cloud_connector_service_user: cloud_connector_user
+cloud_connector_service_password: changeme # noqa: no-static-secrets
```
ehelms: I would argue there is no sane default for this, so we should just omit this and enforce that it must be set to use this role.

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,93 @@
...
+    user_password: "{{ cloud_connector_service_password }}"
+    mail: "{{ cloud_connector_service_user }}@localhost"
+    auth_source: Internal
+    roles:
+      - Cloud Connector
+    admin: false
+    server_url: "{{ cloud_connector_url }}"
+    oauth1_consumer_key: "{{ foreman_oauth_consumer_key }}"
+    oauth1_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+    ca_path: "{{ foreman_ca_certificate | default(omit) }}"
```
ehelms: These three will need to be role scoped variables.

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,93 @@
...
+  register: cloud_connector_cert_info
+
+- name: Set rhc_instance_id in Foreman
+  theforeman.foreman.setting:
+    name: rhc_instance_id
+    value: "{{ cloud_connector_cert_info.subject.commonName }}"
+    server_url: "{{ cloud_connector_url }}"
+    oauth1_consumer_key: "{{ foreman_oauth_consumer_key }}"
+    oauth1_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+    ca_path: "{{ foreman_ca_certificate | default(omit) }}"
```
ehelms: Same here with the role scoped variables

src/roles/cloud_connector/tasks/main.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+- name: Configure HTTP proxy for rhcd
+  ansible.builtin.include_tasks: http_proxy.yaml
+  when: cloud_connector_http_proxy is defined
+
+- name: Remove HTTP proxy configuration for rhcd
```
ehelms: Can this live in the http_proxy.yaml to keep it all together?

src/vars/base.yaml
```diff
@@ -73,3 +73,10 @@ backup_foreman_oauth_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
 backup_foreman_ca_certificate: "{{ foreman_ca_certificate }}"
 backup_foreman_client_certificate: "{{ foreman_client_certificate }}"
 backup_foreman_client_key: "{{ foreman_client_key }}"
+
+cloud_connector_url: "{{ foreman_url }}"
+cloud_connector_service_user: cloud_connector_user
```
ehelms: Not necessary if it matches the default.

src/roles/cloud_connector/tasks/main.yaml
jeremylenz: Done. Moved the "Remove HTTP proxy configuration" block out of tasks/main.yaml and into
     tasks/http_proxy.yaml, so it's now a single when: cloud_connector_http_proxy is defined / is not defined pair of blocks in one file. main.yaml just does one unconditional include_tasks:
     http_proxy.yaml.

src/vars/base.yaml
jeremylenz: Done. Removed cloud_connector_service_user: cloud_connector_user from vars/base.yaml, since it was identical to the
     role default already in defaults/main.yaml.

jeremylenz:
> * Stop using `theforeman.foreman.setting`

I realized this isn't entirely accurate - we did stop using it to set `allow_auto_inventory_upload`, but we still use it to set `rhc_instance_id`. However, `rhc_instance_id` isn't really a user preference; it's derived from the consumer certificate and is overwritten by foremanctl on every deploy. So I think that satisfies the "controlled by foremanctl" requirement.

jeremylenz:
Found a way to make it even better, stand by

jeremylenz:
@ekohl Okay, found a way to not use `theforeman.foreman.setting` module at all (see the latest commit). :+1:

ehelms (approved):

PR #570, opened 2026-06-12
Add --manage-repos parameter for forge deploy-dev

jeremylenz:

When iterating on Foreman or plugin code during development, `forge deploy-dev` clones and checks out the default branch for each repository, discarding any local branches or changes. There is no way to disable this behavior for subsequent deploys.

* Add `--manage-repos` CLI parameter for `forge deploy-dev` that controls whether git repositories are cloned and checked out. Set to `false` to skip repository management for both Foreman and plugins.
* When `--manage-repos=false`, the global setting takes precedence but individual plugin `manage_repo` config is still respected when the global flag is not set.

Steps to reproduce:

* Run `forge deploy-dev` to do an initial deploy
* Check out a custom branch in Foreman or a plugin directory
* Run `forge deploy-dev --manage-repos=false` and verify repositories are not touched

#### Checklist

### DIFF
--- a/development/playbooks/deploy-dev/metadata.obsah.yaml
+++ b/development/playbooks/deploy-dev/metadata.obsah.yaml
@@ -29,6 +29,10 @@ variables:
   foreman_development_github_username:
     help: GitHub username to add as additional remote for git checkouts
     action: store
+  foreman_development_manage_repos:
+    parameter: --manage-repos
+    help: Manage git repositories for Foreman and plugins. Set to false to skip cloning and checking out repositories.
+    type: Boolean
 
 include:
   - _flavor_features
--- a/development/roles/foreman_development/tasks/main.yaml
+++ b/development/roles/foreman_development/tasks/main.yaml
@@ -95,6 +95,7 @@
     git_repository_repository_name: "foreman"
     git_repository_revision: "{{ foreman_development_git_revision }}"
     git_repository_secondary_remote_owner: "{{ foreman_development_github_username }}"
+  when: foreman_development_manage_repos | default(true)
 
 - name: Create database configuration
   ansible.builtin.template:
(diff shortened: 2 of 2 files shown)


ehelms:
@jeremylenz I extracted your conflicts idea from this PR and modified it a bit -- https://github.com/theforeman/foremanctl/pull/610

jeremylenz:
Updated this PR to include only the ~`--preserve-plugin-branches`~ ~`--preserve-branches`~ `--manage-repos` functionality. I will also open a new PR for the answers persistence

jeremylenz:
> I will also open a new PR for the answers persistence

https://github.com/theforeman/foremanctl/pull/612

development/playbooks/deploy-dev/metadata.obsah.yaml
```diff
@@ -29,6 +29,10 @@ variables:
   foreman_development_github_username:
     help: GitHub username to add as additional remote for git checkouts
     action: store
+  foreman_development_preserve_branches:
+    parameter: --preserve-branches
```
ehelms: Down in https://github.com/theforeman/foremanctl/pull/570/changes#diff-298320288bbee9edf592c4fb437fe1e71276c3a877b2e3290a2fa26dbd6b9916R157 this uses `manage_repo` naming but does it on a per plugin basis. What do you think about calling this something like that? e.g. `--manage-repos=False` 

As you aren't simply "preserving branches" you are actually disabling management of the repositories by the tool.

If your intent was just to simply avoid it over-writing the checkout, then I think you'd want to actually set https://github.com/theforeman/foremanctl/blob/master/development/roles/git_repository/tasks/main.yml#L6 in which case it would preserve.

Also, we maybe drop https://github.com/theforeman/foremanctl/blob/master/development/roles/git_repository/tasks/main.yml#L5C18-L5C41 ? or set it to omit unless that variable is set so that it doesn't enforce check out of a particular version.

jeremylenz: My understanding is the Ansible git module can [overwrite local commits](https://github.com/ansible/ansible-modules-core/issues/614), and [discard uncommitted changes](https://github.com/ansible/ansible/issues/19809). So skipping the task still seems safest to me.

ehelms (approved):

PR #571, opened 2026-06-14
Flavor: foreman-proxy-content

arvind4501:

*  Add `deploy-proxy` sub-command  to deploy proxy specific flavors(ex: `foreman-proxy-content`)
*  Add proxy specific parameters to `deploy-proxy` ex: certs tar file and foreman fqdn
*  Make health checks, post-install tasks, and `pull-images` flavor-aware instead of hardcoded to katello
*  Support certificate extraction from tar bundles on proxy nodes
*  Add httpd vhosts for proxying Pulp content and RHSM requests
*  Add flavor-specific test directories,` --flavor` pytest option, and proxy CI job
*  Refactor deploy metadata into reusable obsah includes (_foreman, _foreman_proxy, _pulp, _flavors/katello)

- `./foremanctl deploy` 
- `./foremanctl certificate-bundle proxy.example.com` to generate bundle

- ` ./foremanctl deploy-proxy
            --flavor foreman-proxy-content
            --certificate-bundle /path-to-tar
            --foreman-fqdn quadlet.example.com 
`

- Observe only relevent services are deployed

Steps to reproduce:

* `./foremanctl deploy-proxy
            --flavor foreman-proxy-content
            --certificate-bundle /path-to-tar
            --foreman-fqdn quadlet.example.com`

#### Checklist


src/playbooks/deploy/deploy.yaml
```diff
@@ -1,7 +1,7 @@
 ---
-- name: Setup quadlet demo machine
+- name: Setup quadlet machine
   hosts:
-    - quadlet
+    - "{{ foremanctl_target | default('quadlet') }}"
```
arvind4501: This is a workaround for development and CI, in production quadlet should just work fine

src/playbooks/deploy/deploy.yaml
```diff
@@ -1,7 +1,7 @@
 ---
-- name: Setup quadlet demo machine
+- name: Setup quadlet machine
```
ehelms: Might as well update this as a whole, we don't need to reference the quadlet part in the name here.

src/playbooks/deploy/deploy.yaml
```diff
@@ -17,21 +17,25 @@
...
-    - candlepin
+    - role: candlepin
+      when: enabled_features | has_feature('candlepin')
     - httpd
-    - pulp
-    - foreman
+    - role: pulp
+      when: enabled_features | has_feature('pulp')
+    - role: foreman
+      when: "'foreman' in enabled_features"
```
ehelms: Should this also use `has_feature` ?

src/playbooks/deploy/deploy.yaml
```diff
@@ -43,4 +47,5 @@
     - role: hammer
       when:
         - "'hammer' in enabled_features"
-    - post_install
+    - role: post_install
+      when: "'foreman' in enabled_features"
```
ehelms: I expect `post_install` to have elements that can affect other flavors, so I don't think gating it here is the right approach.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,24 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
+  foreman_proxy_foreman_server_url:
```
ehelms: I am starting to think `foreman_proxy_` should have it's own variable file.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,24 @@ variables:
...
+    parameter: --foreman-url
+    help: URL of the parent Foreman server this proxy connects to.
+  foreman_proxy_trusted_hosts:
+    help: FQDN of a host trusted by the smart proxy. Can be specified multiple times.
```
ehelms: This might be confusing to end users as to what should be specified. e.g. clients? foreman servers?

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,24 @@ variables:
...
+    action: append_unique
+  foreman_oauth_consumer_key:
+    parameter: --oauth-consumer-key
+    help: OAuth consumer key.
```
ehelms: oauth consumer key of what? foreman I assume, but it's ambiguous.

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -56,6 +56,24 @@ variables:
...
+  foreman_oauth_consumer_secret:
+    parameter: --oauth-consumer-secret
+    help: OAuth consumer secret.
+  pulp_rhsm_url:
+    parameter: --rhsm-url
+    help: RHSM URL for Pulp smart proxy content mirroring.
+  pulp_smart_proxy_auth_methods:
+    parameter: --smart-proxy-auth-methods
```
ehelms: This is confusing, I don't know what this applies to.

src/roles/checks/defaults/main.yml
```diff
@@ -1,2 +1,13 @@
 ---
 checks_databases: []
+checks_per_flavor:
```
ehelms: Should the flavor define this instead of the checks role defining it?

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -7,18 +7,25 @@ foreman_proxy_https_port: 8443
...
-foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
+# In case of standalone proxy, we need to add foreman server URL not host fqdn
+foreman_proxy_with_foreman: "{{ 'foreman' in enabled_features }}"
+foreman_proxy_foreman_server_url: "{{ foreman_proxy_with_foreman | ternary('https://{{ ansible_facts['fqdn'] }}', foreman_proxy_foreman_server_url) }}"
 
 # BMC settings
 foreman_proxy_bmc_ipmi_implementation: ipmitool
 foreman_proxy_bmc_redfish_verify_ssl: true
+
+foreman_proxy_with_pulp_mirror: "{{ pulp_mirror | default(false) }}"
```
ehelms, line 25: Where does `pulp_mirror` come from? That reads like a non-role variable.

ehelms: Does foreman_proxy know about Pulp? Pulp has a built in smart-proxy.

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -7,18 +7,25 @@ foreman_proxy_https_port: 8443
 foreman_proxy_url: "https://{{ foreman_proxy_name }}:{{ foreman_proxy_https_port }}"
 
 # Settings
-foreman_proxy_trusted_hosts:
+foreman_proxy_default_trusted_hosts:
   - "{{ foreman_proxy_name }}"
 
+foreman_proxy_trusted_hosts: []
+foreman_proxy_all_trusted_hosts: "{{ (foreman_proxy_default_trusted_hosts + foreman_proxy_trusted_hosts) | unique | list }}"
```
ehelms: Is this useful here in the defaults? Could you move this directly into the template?

src/roles/httpd/defaults/main.yml
```diff
@@ -5,6 +5,10 @@ httpd_foreman_backend: http://localhost:3000
 httpd_pub_dir: /var/www/html/pub
 httpd_enabled_pulp_snippets: []
 
+httpd_proxy_foreman_url: "{{ foreman_proxy_foreman_server_url }}"
+httpd_client_certificate: "{{ client_certificate }}"
```
ehelms: What is `client_certificate`? This role doesn't know about this variable. Be mindful, throughout this PR you are blending role and non-role variables.

ehelms:
There are definitely some good nuggets of changes that would make for good, go-ahead, stand-alone PRs to get added. This also points to the need for `flavor` specific parameters that are only shown for a given flavor.

src/playbooks/deploy/deploy.yaml
arvind4501: Here, i am only using `has_feature('x')` when x feature is not directly in the  `enabled_features` list and for the features that are directly available in `enabled_features` i just use it to gate.

src/playbooks/deploy/deploy.yaml
arvind4501: Agree, i just wanted to have a working setup first and then optimize it for better but you are right, will update it

src/roles/foreman_proxy/defaults/main.yaml
arvind4501: > Where does pulp_mirror come from? That reads like a non-role variable.

`pulp_mirror` is defaults to false in role defaults, which make it easy for other flavors where pulp is not in mirror mode. its defined as true only in the foreman-proxy-content flavor's var file

> Does foreman_proxy know about Pulp? Pulp has a built in smart-proxy.

are you talking about https://github.com/theforeman/pulp_smart_proxy? AFAIK there is one defined in foreman https://github.com/theforeman/foremanctl/blob/master/src/roles/foreman/tasks/main.yaml#L316

src/roles/pulp/defaults/main.yaml
```diff
@@ -69,7 +71,10 @@ pulp_settings_other_env:
   PULP_REMOTE_USER_ENVIRON_NAME: "HTTP_REMOTE_USER"
   PULP_REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES: >-
     ['rest_framework.authentication.SessionAuthentication', 'pulpcore.app.authentication.PulpRemoteUserAuthentication']
-  PULP_SMART_PROXY_PULP_URL: "{{ pulp_pulp_url }}"
+  SMART_PROXY_PULP_URL: "{{ pulp_pulp_url }}"
```
arvind4501: @evgeni , i am bit confused on this, i wanted a replacement to pulpcore config(https://github.com/theforeman/puppet-foreman_proxy/blob/master/templates/plugin/pulpcore.yml.erb), as we are now using https://github.com/theforeman/pulp_smart_proxy/blob/develop/pulp_smart_proxy/app/settings.py, is this appropriate, thoughts?

src/playbooks/deploy/metadata.obsah.yaml
arvind4501:
```
foreman-installer --scenario foreman-proxy-content \
--certs-tar-file "/root/smart-proxy_cert/smartproxy.example.com-certs.tar" \
--foreman-proxy-foreman-base-url "https://foreman.example.com" \
--foreman-proxy-trusted-hosts "foreman.example.com" \
--foreman-proxy-trusted-hosts "smartproxy.example.com" \
--foreman-proxy-oauth-consumer-key "s97QxvUAgFNAQZNGg4F9zLq2biDsxM7f" \
--foreman-proxy-oauth-consumer-secret "6bpzAdMpRAfYaVZtaepYetomgBVQ6ehY"
```
updated it to be as before, except `foreman-base-url `, i think we can get rid of `foreman-proxy` prefix for all of these params

src/playbooks/_foreman_proxy/metadata.obsah.yaml
```diff
@@ -0,0 +1,22 @@
...
+    choices:
+      - freeipmi
+      - ipmitool
+  foreman_proxy_bmc_redfish_verify_ssl:
+    parameter: --bmc-redfish-verify-ssl
+    help: Verify SSL certificates for Redfish BMC connections.
+    type: Boolean
+  foreman_proxy_foreman_server_url:
+    parameter: --foreman-base-url
+    help: URL of Base Foreman server this proxy connects to.
```
ehelms: I'm not a fan of the `base` terminology here. Implies there are other Foreman servers that a user can be using. `URL of the Foreman server`

src/playbooks/_foreman_proxy/metadata.obsah.yaml
```diff
@@ -0,0 +1,22 @@
...
+  foreman_proxy_trusted_hosts:
+    help: FQDN of a host trusted by the proxy.
+    action: append_unique
+  foreman_proxy_oauth_consumer_key:
+    help: OAuth key to be used for REST interaction.
```
ehelms:
```suggestion
    help: OAuth key to be used for communication with Foreman.
```

src/playbooks/deploy/deploy.yaml
ehelms: That distinction isn't obvious when reading the code. I think a consistent method would be better.

src/roles/foreman_proxy/defaults/main.yaml
ehelms: Correct. I would expect us to be using `pulp_smart_proxy` plugin for Pulp mirrors.

src/vars/flavors/foreman-proxy-content.yml
```diff
@@ -0,0 +1,14 @@
+flavor_features:
```
ehelms, line 1: I forget, can a user remove a flavor feature? Or do they represent the minimum set of enforced, always on, features?

Vagrantfile
```diff
@@ -47,4 +47,14 @@ Vagrant.configure("2") do |config|
       libvirt.memory = 2048
     end
   end
+
+  config.vm.define "proxy" do |override|
```
ehelms, line 51: Let's be more specific on the kind of proxy to reduce confusion.

src/vars/flavors/foreman-proxy-content.yml
arvind4501: I think they can only remove optional features, yes, not the flavor defaults

ehelms: That raises the question then if we really want all these features enabled and non-removable. An admin may not want content types to be present that they do not manage for example.

arvind4501: Ah... Yes, you are right, but that goes against our current deployment design, https://github.com/theforeman/foremanctl/blob/master/docs/developer/deployment.md#features-and-flavors we mention that 

> A flavor is a set of features that are enabled by default and can not be disabled. This is to allow common deployment types like "vanilla foreman", "katello", "satellite" and similar.

but i think that is something we solved in foreman-installer with the use of `custom-hiera` to override default options/parameters and we should solve it in foremanctl as well where we should give admin ability to override flavor defaults

arvind4501:
I have been facing one issue which i already have a workaround implemented but i wanted to know if we should solve it properly?
 
so we have a static `quadlet` target in our playbooks(which may be fine for today as it points to localhost), but in CI and development we need to deploy on different servers not localhost. currently the workaround is to pass a extra var in command and use that to determine the host.

What about exposting a `--target` obsah option to manage that, but the issue is its visible to users

arvind4501:
> This also points to the need for flavor specific parameters that are only shown for a given flavor.


facce79197bee78a5aa413f5141114c55e72dfdb  does it, but there is a catch, that we must give flavor in order to see help, as we have done in foreman-installer, other way would to add a default attribute for variables to make them aware about default value of a variable, which does require changes to obsah

src/vars/flavors/foreman-proxy-content.yml
arvind4501: https://github.com/theforeman/foremanctl/issues/572

arvind4501:
With https://github.com/theforeman/obsah/pull/120 the recursive metadata inclusion will work, which is currently causing the test falures

src/playbooks/deploy/deploy.yaml
```diff
@@ -17,30 +17,34 @@
...
         certificate_checks_key: "{{ server_key }}"
         certificate_checks_ca: "{{ server_ca_certificate }}"
     - role: postgresql
-      when:
-        - database_mode == 'internal'
+      when: database_mode == 'internal'
     - redis
-    - candlepin
+    - role: candlepin
+      when: enabled_features | has_feature('candlepin')
```
evgeni: Should this be handled by the `role` attribute of the feature? (Handling that is not implemented yet, but the idea was to allow for cases like this)
(obviously non blocking, as it would require additional work here)

arvind4501: looking at https://github.com/theforeman/foremanctl/blob/master/docs/developer/feature-metadata.md , i read role attribute as `when defining a feature if foreman/foreman-proxy part does not belong to their respective plugin then execute this role`

in case of gating roles, we need to know the features for a flavor. so i don't see it being used for gating roles or am i misinterpreting.

evgeni: I am not sure what you mean by "gating role", but my intent when writing the metadata spec was to say:
- if your feature is (or has a part in) Foreman, set the `foreman` property
- if your feature is (or has a part in) Foreman Proxy, set the `foreman_proxy` property
- if you feature is a standalone service, describe it in a dedicated role

And I'd argue `candlepin`, `pulp`, `iop` fall into this "standalone" bucket.

arvind4501: by "gating role", i mean `for what condition we run a particular role in deploy playbook`

Also if we consider `role` metadata defined for a feature in features.yaml, how does that not require a when condition on roles defined in deploy playbook? sorry i am too confused

src/roles/pulp/defaults/main.yaml
evgeni: all pulp setting variables need to be prefixed with `PULP_` otherwise they won't be applied.

src/roles/pulp/defaults/main.yaml
```diff
@@ -69,7 +71,10 @@ pulp_settings_other_env:
...
+  SMART_PROXY_MIRROR: "{{ pulp_mirror }}"
+  SMART_PROXY_RHSM_URL: "{{ pulp_rhsm_url }}"
+  SMART_PROXY_AUTH_METHODS: ['client_certificate']
```
evgeni: I don't think you need to set these two

src/roles/pulp/defaults/main.yaml
arvind4501: sorry i couldn't find, where is `PULP_`  prefix is enforced?

src/roles/foreman_proxy/tasks/main.yaml
```diff
@@ -75,5 +75,15 @@
     oauth1_consumer_secret: "{{ foreman_proxy_oauth_consumer_secret }}"
     ca_path: "{{ foreman_proxy_foreman_ca_certificate | default(omit) }}"
 
+- name: Register Pulp Proxy to Foreman
```
evgeni: we have a similar thing in https://github.com/theforeman/foremanctl/blob/78542496954373661540bd852e4420127dc9765d/src/roles/foreman/tasks/main.yaml#L316-L323 and I think that both places are wrong 🙈 

(but to answer you question: yes, we will have to register the pulp proxy, and your current code does that.)

arvind4501: for me the confusing part is having two proxy records in db, one for pulp and other for other features.  i don't have much idea about the pulp proxy, but yeah it's confusing

ehelms:
For me, seeing this PR highlights a feeling I've continued to have that we would benefit by having two distinct, core deployment types: server and external-infrastructure-capsule-node. Fundamentally, our architecture always deploys either a Foreman server or an external node that contains at least one smart-proxy that connect back to Foreman. 

The flavors we are defining will deploy onto one of these two but will not apply to both. And, we know there are some fundamental truths about each that we could optimize for by having a top-level command for each (e.g. `deploy`, `deploy-node`). For example, the foreman role will never be on an infra-node, or we know that certificates will not be generated on an infra-node. We also know that certain flavors we build will only apply to a Foreman server or an infra node. 

Please note that naming is hard, we do not have a great name for the "infra node that is not the Foreman server". 

To recap what I am proposing:

- Two top-level commands: `deploy` and `deploy-node-capsule-infra-thing` 
- Flavors target and apply to one of the two deployment types
- Flavors continue to represent a set of features that are hard-coded, and any other features can be added if they apply to the deployment type

I believe this would make understanding what is being deployed easier for the user and developer. And the code would require less paths and conditionals to unwind when trying to understand what will be deployed in what scenario.

arvind4501:
> Two top-level commands: deploy and deploy-node-capsule-infra-thing

By these you mean we would be having two seprate deployment playbooks? then we need to also think what flavors we expose in each of them, for ex: foreman server can't deploy proxy flavor or do we support deploying all flavors?

This would also affect how we structure tests corresponding to each deployment type(server or proxy).

ehelms:
> By these you mean we would be having two seprate deployment playbooks? then we need to also think what flavors we expose in each of them, for ex: foreman server can't deploy proxy flavor or do we support deploying all flavors?

Correct, two separate commands, two separate playbooks. And yes, certain flavors would only target one of the two.

src/roles/pulp/defaults/main.yaml
evgeni: https://pulpproject.org/pulpcore/docs/admin/guides/configure-pulp/#through-environment-variables

docs/developer/deployment.md
```diff
@@ -1,5 +1,30 @@
...
+./foremanctl deploy
+```
+
+### Proxy
+
+Deploys a Foreman Smart Proxy node that connects to an foreman server.
+
+```bash
+./foremanctl deploy-proxy \
+  --flavor foreman-proxy-content \
```
evgeni: should this be the default flavour (for now) - just like `katello` is for `deploy`?

src/playbooks/deploy-proxy/deploy-proxy.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - "../../vars/database.yml"
+    - "../../vars/foreman.yml"
+    - "../../vars/base.yaml"
+  roles:
+    - role: pre_install
+    - role: checks
+      vars:
+        checks_databases: "{{ all_databases }}"
+    - role: certificates
+      when: "certificates_source in ['default', 'custom_server']"
```
evgeni:
```suggestion
```
we don't have any other sources anymore

src/playbooks/deploy-proxy/metadata.obsah.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+variables:
+  flavor:
+    help: Base flavor to use in this deployment.
+    choices:
+      - foreman-proxy-content
+  certificates_bundle:
+    help: Path to the certificate bundle tar file.
+    type: AbsolutePath
+    parameter: --certs-tar-file
```
evgeni:
```suggestion
    parameter: --certificate-bundle
```

All other params use `certificate`, not `certs`, so I'd use that.

src/playbooks/deploy-proxy/metadata.obsah.yaml
```diff
@@ -0,0 +1,30 @@
...
+    persist: false
+  foreman_proxy_foreman_server_url:
+    parameter: --foreman-url
+    help: URL of Foreman server this proxy connects to.
+  foreman_proxy_trusted_hosts:
+    help: FQDN of a host trusted by the proxy.
+    action: append_unique
```
evgeni: In #583 I went with "foreman name" and derived URL and trusted hosts from it, with the reasoning that we never should *need* to add any other trusted host but the main foreman instance (to a prod deployment).

this makes the user flow nicer, IMHO. what do you think?

src/roles/certificates/tasks/extract.yml
```diff
@@ -0,0 +1,18 @@
...
+    msg: "Path to certificate tar file not found: {{ certificates_bundle }}"
+  when: not _certificates_tar_file_path.stat.exists
+
+- name: Extract certificate bundle
+  ansible.builtin.unarchive:
+    src: "{{ certificates_bundle }}"
+    dest: "{{ certificates_ca_directory }}"
+    remote_src: true
+    extra_opts:
+      - --strip-components=1
```
evgeni, line 18: @ehelms shall we drop the first (folder) component from the generated tarball? I used the same hack in #583 and it seems that if it's just there to be stripped, why adding it in the first place

src/roles/certificates/tasks/main.yml
```diff
@@ -29,23 +29,34 @@
...
-  ansible.builtin.assemble:
-    src: "{{ certificates_ca_directory_certs }}"
-    dest: "{{ certificates_ca_directory_certs }}/ca-bundle.crt"
-    regexp: '(ca|server-ca)\.crt$'
-    mode: '0444'
-  when:
-    - certificates_source == 'custom_server'
-    - certificates_custom_server_certificate is defined
+- name: 'Generate certificates'
+  when: certificates_bundle is not defined
```
evgeni: I used the same `when` in #583 but then realized this will break if on second invocation when we do not expect the user to pass the bundle again.

docs/developer/deployment.md
arvind4501: Yes, I need to pass flavor here to bring flavor vars in playbooks like features, where it used default vars(https://github.com/theforeman/foremanctl/blob/master/src/playbooks/features/features.yaml#L10) which will alway point to flavor katello.  also used for finding `flavor` in tests

src/playbooks/deploy-proxy/metadata.obsah.yaml
arvind4501: if we don't need more than `main foreman instance` in trusted proxies, then i think it's better to use single parameter. but in place of`--foreman-name` can we use `--foreman-fqdn`?

sjha4:
Wanted to test this..The production flow here would be to have foremanctl in 2 servers and deploying the foreman services in one server and foreman-proxy-content in another..Is copying certs over and running `foremanctl  deploy-proxy` the suggested way to do that?

arvind4501:
> Wanted to test this..The production flow here would be to have foremanctl in 2 servers and deploying the foreman services in one server and foreman-proxy-content in another..Is copying certs over and running `foremanctl deploy-proxy` the suggested way to do that?

Yes, so you can follow as below:

- `foremanctl deploy` on a vm1(server)
- `foremanctl certificate-bundle proxy.example.com` to generate cert bundle on same vm1
-  have a second vm2(proxy), which has foremanctl with the PR changes(via packit)
-  copy bundle from vm1 to vm2
-  `foremanctl deploy-proxy --flavor foreman-proxy-content --certificate-bundle /path/to/bundle --foreman-fqdn <fqdn> ` on vm2 (make sure you have changes  from the PR)

src/roles/certificates/tasks/main.yml
arvind4501: thanks for catching that, also i liked https://github.com/theforeman/foremanctl/pull/583/changes#diff-c07ec34ab81de2fa803843b62af316d5167c343f26374331c6a748df54717c90 which decouples pulp proxy from foreman role, so ended up updating same here

src/playbooks/deploy-proxy/deploy-proxy.yaml
```diff
@@ -0,0 +1,34 @@
+---
+- name: Setup proxy machine
+  hosts:
+    - proxy
```
sjha4, line 4: Will this need an inventory file update here : https://github.com/theforeman/foreman-packaging/blob/5efbbc888988c088f3b1d7c4f861374340d5e224/packages/foreman/foremanctl/foremanctl.spec#L33 to point proxy to localhost?

arvind4501: I think yes,  we are pointing to localhost for both cases so that should be fine for now, i think that should be updated when we release this

src/roles/pulp/tasks/main.yaml
```diff
@@ -311,3 +312,12 @@
...
     - pulp_existing_workers | length > 0
     - (item | regex_replace('^' + pulp_worker_container_name + '@(\\d+)\\.service$', '\\1') | int) > (pulp_worker_count | int)
+
+- name: Configure Foreman Proxy
+  theforeman.foreman.smart_proxy:
+    name: "{{ ansible_facts['fqdn'] }}-pulp"
+    url: "https://{{ ansible_facts['fqdn'] }}/pulp/api/v3/smart_proxy"
+    server_url: "{{ pulp_foreman_url }}"
+    oauth1_consumer_key: "{{ pulp_foreman_oauth_consumer_key }}"
+    oauth1_consumer_secret: "{{ pulp_foreman_oauth_consumer_secret }}"
```
sjha4, line 323: Will these key and secret files need to be copied over to proxy server to authenticate communication?

arvind4501: Yes, for production they should be provided in order to authenticate against foreman server(for now by manually copying), previously foreman-installer used to spit them as output of `foreman-proxy-certs-generate` but i think we need make that part easy

sjha4: Ack..In some good news, with the above 2 changes for production setup, I was able to get a proxy deployed and sync some repos from foreman -> proxy 🥳

.github/workflows/test.yml
```diff
@@ -372,6 +372,75 @@ jobs:
...
+            --foreman-fqdn quadlet.example.com
+      - name: Run tests
+        run: |
+          ./forge test --pytest-args="--server-hostname=proxy"
+      - name: Archive smoker report
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: smoker-proxy-content
+          path: "/home/runner/smoker/report/"
```
evgeni: you don't run smoker, so no need to archive its results

(and running it explodes for me due to low memory on the GitHub runners)

src/vars/database.yml
```diff
@@ -138,7 +138,7 @@ databases:
     feature: iop
 
 all_databases: >-
-  {{ databases | selectattr('feature', 'in', enabled_features) | list }}
+  {{ databases | databases_for_features(enabled_features) }}
```
evgeni, line 141: this is just cosmetic, right? the result remains the same?

tests/flavor/katello/images_test.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+
+FlAVOR_CORE_IMAGES = [
```
evgeni:
```suggestion
FLAVOR_CORE_IMAGES = [
```

tests/conftest.py
```diff
@@ -44,19 +45,34 @@ def enabled_features(self):
...
 @pytest.fixture(scope="module")
 def fixture_dir():
     return py.path.local(__file__).realpath() / '..' / 'fixtures'
 
 
 @pytest.fixture(scope="module")
-def server_hostname():
-    return 'quadlet'
+def server_hostname(pytestconfig):
+    return pytestconfig.getoption("server_hostname")
```
evgeni, line 76: wouldn't it be cleaner to have `proxy_hostname` and `proxy` fixtures, which are then used by the proxy-specific tests?

or is that required for tests that run on both, server and proxy?

tests/foreman_api_test.py
evgeni: this is also katello specific

evgeni: I wonder if we should split the files into foreman_api and katello_api

tests/httpd_test.py
```diff
@@ -63,6 +67,7 @@ def test_https_pulp_auth(server, certificates, server_fqdn):
     assert cmd.stdout == '200'
 
 
+@pytest.mark.feature('katello')
```
evgeni: technically pulp(/python)

tests/postgresql_test.py
```diff
@@ -35,7 +21,9 @@ def test_postgresql_password_encryption(database):
 
     reader = csv.reader(result.stdout.splitlines())
     for row in reader:
-        assert ("SCRAM-SHA-256" in row[6])
+        if row[0] == 'postgres':
+            continue
```
evgeni: why was this neccessary?

src/roles/httpd/tasks/main.yml
```diff
@@ -89,6 +104,16 @@
     src: foreman-ssl-vhost.conf.j2
     dest: /etc/httpd/conf.d/foreman-ssl.conf
     mode: "0644"
+  when: httpd_with_foreman
+  notify:
+    - Restart httpd
+
+- name: Configure rhsm-pulpcore vhost
```
evgeni: I think this should be called `pulpcore-ssl`? I was confused looking at pulpcore one and only seein port 80, and then everything with ssl in here while called "rhsm"

src/roles/httpd/templates/rhsm-pulpcore-vhost.conf.j2
```diff
@@ -0,0 +1,88 @@
+<VirtualHost *:443>
```
evgeni: maybe not for now, but I think we should try to merge this and the foreman-ssl template, as the only real difference is the set of services we proxy for.

(same for non-ssl)

src/roles/httpd/templates/rhsm-pulpcore-vhost.conf.j2
```diff
@@ -0,0 +1,88 @@
...
+    RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT
+    ProxyPass {{ httpd_pulp_content_backend }}/pulp/content disablereuse=on timeout=600
+    ProxyPassReverse {{ httpd_pulp_content_backend }}/pulp/content
+  </Location>
+
+  <Location "/pulp/api/v3">
+    RequestHeader unset REMOTE_USER
+    RequestHeader unset REMOTE-USER
+    RequestHeader set REMOTE-USER "admin" "expr=%{SSL_CLIENT_S_DN_CN} == '{{ ansible_facts['fqdn'] }}'"
+    RequestHeader set REMOTE-USER "admin" "expr=%{SSL_CLIENT_S_DN_CN} == '{{ foreman_name }}'"
```
evgeni: I think the `fqdn` one is not needed (it wasn't in #583 at least)

@ianballou do we ever need to talk to the pulp api *locally* on a proxy? or are all invocations always triggered by the foreman server?

src/roles/pulp/defaults/main.yaml
```diff
@@ -70,6 +73,9 @@ pulp_settings_other_env:
   PULP_REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES: >-
     ['rest_framework.authentication.SessionAuthentication', 'pulpcore.app.authentication.PulpRemoteUserAuthentication']
   PULP_SMART_PROXY_PULP_URL: "{{ pulp_pulp_url }}"
+  PULP_SMART_PROXY_MIRROR: "{{ pulp_mirror }}"
+  PULP_SMART_PROXY_RHSM_URL: "{{ pulp_rhsm_url }}"
+  PULP_SMART_PROXY_AUTH_METHODS: ['client_certificate']
```
evgeni: this is the default: https://github.com/theforeman/pulp_smart_proxy/blob/53894170bd37b8f012dbee9aa09bf82137020e7b/pulp_smart_proxy/app/settings.py#L4

no need to set it explicitly, IMHO

src/vars/database.yml
arvind4501: Nope, since `enabled_features` knows only katello(not candlepin and pulp), its need in https://github.com/theforeman/foremanctl/pull/571/changes#diff-86c1ec8597f39a3017aa1e93deffa4e641ae247282349d20fd80fe66e69178d9R120 for filtering pulp and candlepin with the help of `has_feature`

src/roles/pulp/defaults/main.yaml
```diff
@@ -24,9 +24,12 @@ pulp_content_origin: "http://{{ ansible_facts['fqdn'] }}:24816"
 pulp_pulp_url: "http://{{ ansible_facts['fqdn'] }}:24817"
 
 pulp_enable_analytics: false
+pulp_mirror: false
+pulp_register_foreman_proxy: true
+pulp_rhsm_url: "https://{{ ansible_facts['fqdn'] }}/rhsm"
```
evgeni, line 29:
```suggestion
pulp_rhsm_url: "https://{{ foreman_proxy_name }}/rhsm"
```

that way we only need to override it once (via `foreman_proxy_name`)

evgeni: but you probably will need to actually do `pulp_name` here and then set `pulp_name: "{{ foreman_proxy_name }}"` in the vars files

tests/conftest.py
arvind4501: I think that would lead to more complications since we have more tests which works on both the vms

src/roles/pulp/defaults/main.yaml
evgeni: oh, just seen you do override it below already, nvm

src/roles/foreman_proxy/templates/settings.yml.j2
```diff
@@ -1,7 +1,8 @@
 :settings_directory: /etc/foreman-proxy/settings.d
 
 :foreman_url: {{ foreman_proxy_foreman_server_url }}
-:trusted_hosts: {{ foreman_proxy_trusted_hosts }}
+:trusted_hosts:
+  - {{ foreman_proxy_trusted_hosts }}
```
evgeni: won't this now be a list of lists?

arvind4501: i don't think so, since we only expect foreman server in trusted hosts, which is a single entry,

src/roles/pulp/tasks/main.yaml
evgeni, line 27:
```suggestion
  loop: "{{ pulp_default_import_paths + pulp_import_paths }}"
  when: not pulp_mirror
```

src/vars/base.yaml
```diff
@@ -45,6 +51,7 @@ foreman_proxy_foreman_server_url: "{{ foreman_url }}"
 foreman_proxy_foreman_ca_certificate: "{{ foreman_ca_certificate }}"
 foreman_proxy_oauth_consumer_key: "{{ foreman_oauth_consumer_key }}"
 foreman_proxy_oauth_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+foreman_proxy_trusted_hosts: "{{ foreman_name }}"
```
evgeni:
```suggestion
foreman_proxy_trusted_hosts:
  - "{{ foreman_name }}"
```

now I get it why you made it a list above.

tests/postgresql_test.py
arvind4501: Earlier it failed for some reason related to row formattiong, but after looking at it now it passes, thanks will remove this

src/requirements.txt
```diff
@@ -1,3 +1,3 @@
-obsah >= 1.8.1
+obsah >= 1.9.0
```
evgeni:
```suggestion
obsah >= 1.9.1
```

.0 was a tad buggy :)

tests/foreman_api_test.py
arvind4501: yes, sounds reasonable to me

tests/flavor/foreman-proxy-content/httpd_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+def test_rhsm_pulpcore_vhost_exists(server):
+    conf = server.file("/etc/httpd/conf.d/rhsm-pulpcore.conf")
+    assert conf.exists
+    assert conf.is_file
+
+
+def test_https_pulp_api_with_client_cert(server, certificates, server_fqdn):
+    cmd = server.run(
+        f"curl --silent --cacert {certificates['server_ca_certificate']} "
```
evgeni: in https://github.com/theforeman/foremanctl/pull/533 Leos has written a nice proxy_request helper, which we could extract and make work with the server too.

src/roles/pulp/tasks/main.yaml
arvind4501: you mean no import directory, are we ditching the old one `/var/lib/pulp/sync_imports`, also 
```
pulp_default_import_paths: "{{ pulp_mirror | ternary(['/var/lib/pulp/sync_imports'], ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']) }}"
pulp_default_export_paths: "{{ pulp_mirror | ternary([], ['/var/lib/pulp/exports']) }}"
```
and with use of this in defaults i can just get rid of the whole `when: not pulp_mirror`

evgeni: yeah, imports should not be needed on a mirror (at least for now)

src/roles/foreman_proxy/templates/settings.yml.j2
evgeni: your default value is `[]` which is a list and the variable name also suggests a list being host*S*.
and *technically* the proxy can have many, so I really think we should allow it in the role, but in the actual deployment pass a single-entity list.

src/roles/httpd/templates/rhsm-pulpcore-vhost.conf.j2
arvind4501: Without the fqdn entry proxy /v2/features api gave 403 error

evgeni: oooh, because you use that in tests while running them from the machine directly

makes sense

src/features.d/content.yaml
evgeni, line 1: any reason to place this under features.d and not the features.yaml directly?

arvind4501: mostly for readability, since it bundled all `content/` featrures in single file

src/playbooks/_flavors/foreman-proxy-content/metadata.obsah.yaml
```diff
@@ -0,0 +1,4 @@
+---
+include:
+  - _pulp
```
evgeni, line 3: this adds import/export paths, which we don't want?

src/roles/certificates/defaults/main.yml
```diff
@@ -17,3 +17,4 @@ certificates_ca_validity_days: 7300
 certificates_validity_days: 7300
 certificates_renew: false
 certificates_renew_ca: false
+certificates_generate: true
```
evgeni, line 20: why was this needed?

evgeni: oooh, to skip it on the proxy when no bundle? I see

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -46,6 +50,13 @@
 
   # SSL Proxy directives
   SSLProxyEngine On
+{% if not httpd_with_foreman %}
+  SSLProxyCACertificateFile "{{ httpd_ssl_dir }}/certs/katello-default-ca.crt"
```
evgeni, line 53: shouldn't this be outside the if?

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.pytest.ini_options]
+addopts = "--import-mode=importlib"
```
evgeni: what does this do?

evgeni: (and should it go to a `pytest.ini` or so ,given that we're not really a python project using pyproject style data?)

src/playbooks/deploy-proxy/metadata.obsah.yaml
evgeni: I don't mind either naming, pick one you like more :)

src/roles/certificates/tasks/extract.yml
```diff
@@ -0,0 +1,18 @@
...
+- name: Fail if path to certificate tar file does not exist
+  ansible.builtin.fail:
+    msg: "Path to certificate tar file not found: {{ certificates_bundle }}"
+  when: not _certificates_tar_file_path.stat.exists
+
+- name: Extract certificate bundle
+  ansible.builtin.unarchive:
+    src: "{{ certificates_bundle }}"
+    dest: "{{ certificates_ca_directory }}"
+    remote_src: true
```
evgeni: this means we need to copy the bundle to the node vs having it on the system where `foremanctl` runs.

in prod deployments (today) those two systems are the same, so it shouldn't make too much difference, but e.g. in CI it's not (and I still have the vision to be able to deploy multiple systems from one controller, some time in the future).

given the other certificate-style inputs are "local to foremanctl controller", I'm slightly leaning to also do it here, but wouldn't force anyone who disagrees with me :)

docs/developer/deployment.md
```diff
@@ -1,5 +1,29 @@
 # Deployment Design
 
+## Deployment Types
+
+foremanctl supports two deployment types: **server** and **proxy**. Each has its own sub-command, flavor, and set of services.
```
evgeni, line 5: @ehelms was exploring whether we could give the external proxy a better name ("node", "capsule", whatever).
I am not sure how far he got with that (and I am sure we shouldn't block/wait on it), but that will probably produce some follow up work.

Additionally the current PR uses "Foreman Proxy", "node" etc rather inconsistently in the docs/help output.
Should probably cleaned up, but obviously doesn't break the workflow.

pyproject.toml
arvind4501: Without this we need to have a unique test file name so `tests/http_test.py` and `tests/flavor/foreman-proxy-content/httpd_test.py` will be considered as duplicates and pytest will fails, for refrence: https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html#choosing-an-import-mode

evgeni:
I've posted a few (IMHO final) comments, which aren't blocking (besides the `SSLProxyCACertificateFile` one).
Overall this is working nicely and unblocks a lot of follow up work, so I think we should get it in, even if there are still rough edges we might want to clean up later.

src/roles/certificates/defaults/main.yml
arvind4501: since in certificates role only using `when certificates_bundle is defined` is not sufficient as https://github.com/theforeman/foremanctl/pull/571#discussion_r3467655286 , i needed a another var to put condition and since `foreman-proxy-content` flavor var has `certificates_generate: false`, it wo'nt generate certificates but by defaullt it is true to generate certificate in deploy playbook

pyproject.toml
evgeni: TIL
(not sure I *wanted* to learn such details about the internals of Pytest, but I did)

src/roles/certificates/tasks/extract.yml
arvind4501: Agree, also for Ci i copied your fetch bundle thing from https://github.com/theforeman/foremanctl/pull/583/changes#diff-8fd8b7abb2fafbb8053e42bb545fd37199b2ce3e8cff03e611ab93f8d42d6bb4

.github/workflows/test.yml
```diff
@@ -372,6 +372,68 @@ jobs:
...
+            --tuning development \
+            --add-feature foreman-proxy
+      - name: Remove old parameters.yaml
+        run: rm -f .var/lib/foremanctl/parameters.yaml
+      - name: Generate certificates bundle on quadlet
+        run: |
+          ./foremanctl certificate-bundle proxy.example.com
+      - name: Fetch certificates bundle from quadlet
+        run: |
+          ./foremanctl fetch-bundle proxy.example.com
```
evgeni:
```suggestion
          ./forge fetch-bundle proxy.example.com
```

arvind4501: my bad :face_with_head_bandage:

evgeni (approved):

stejskalleos (changes requested):
I found one blocking issue:

On master, running `./foremanctl deploy --help` will list all parameters.

```
gh pr checkout 571 
Switched to branch 'flavor/foreman-proxy-content'
(.venv) foremanctl flavor/foreman-proxy-content $ git pull           
From https://github.com/arvind4501/foremanctl
 * branch            flavor/foreman-proxy-content -> FETCH_HEAD
Already up to date.

./foremanctl deploy --help

usage: foremanctl deploy [-h] [-v] [--external-authentication {ipa,ipa_with_api}]
                         [--external-authentication-pam-service EXTERNAL_AUTHENTICATION_PAM_SERVICE] [--add-feature FEATURES]
                         [--flavor {katello}] [--content-export-path PULP_EXPORT_PATHS] [--content-import-path PULP_IMPORT_PATHS]
                         [--remove-feature FEATURES]

Install

options:
  -h, --help            show this help message and exit
  -v, --verbose         verbose output
  --external-authentication {ipa,ipa_with_api}
                        External authentication method to use (persisted)
  --external-authentication-pam-service EXTERNAL_AUTHENTICATION_PAM_SERVICE
                        Name of the PAM service to use for IPA authentication (persisted)
  --add-feature FEATURES
                        Additional features to enable in this deployment. (persisted)
  --flavor {katello}    Base flavor to use in this deployment. (persisted)
  --content-export-path PULP_EXPORT_PATHS
                        Extra file path that Pulp can use for content exports. Argument may be used more than once. (persisted)
  --content-import-path PULP_IMPORT_PATHS
                        Extra file path that Pulp can use for content imports. Argument may be used more than once. (persisted)
  --remove-feature FEATURES
                        Additional features to disable in this deployment. (persisted)

Parameters marked as (persisted) can be reset by --reset-<parameter-name>.
```

No parameters :(

stejskalleos:
Please ignore; I forgot to run the setup-environment script :/

.github/workflows/test.yml
```diff
@@ -372,6 +372,83 @@ jobs:
...
+      - name: Deploy Katello on quadlet
+        run: |
+          ./foremanctl deploy \
+            --certificate-source=${{ matrix.certificate_source }} \
+            ${{ matrix.certificate_source == 'custom_server' && '--certificate-server-certificate /root/custom-certificates/certs/quadlet.example.com.crt --certificate-server-key /root/custom-certificates/private/quadlet.example.com.key --certificate-server-ca-certificate /root/custom-certificates/certs/ca.crt' || '' }} \
+            --initial-admin-password=changeme \
+            --tuning development \
+            --add-feature foreman-proxy
+      - name: Remove old parameters.yaml
+        run: rm -f .var/lib/foremanctl/parameters.yaml
```
ehelms, line 418: Why do we have to do this?

evgeni: To simulate deployment on a different box basically

arvind4501: Because since earlier `foremanctl deploy` writes its params and they share same `parameters.yaml` and as obsah resuses existing params, so its good to use a clean empty params file for `deploy-proxy` 
and as evgeni mentioned(i didn't refresh earlier)

ehelms: This is going to be a pain locally. Obviously doesn't block this PR but we either need to namespace this or what I have been feeling is that this should live on the target. I don't think that would break our design to be able to run `foremanctl` from a remote location. And, the parameters.yaml feels like information that should exist on the machine itself given the level of important information that it stores. Really, the whole directory should.

ehelms (approved):

ehelms:
@arvind4501 Can you please rebase this?

docs/developer/deployment.md
```diff
@@ -1,5 +1,29 @@
...
+
+### Proxy
+
+Deploys a Foreman Smart Proxy node that connects to an foreman server.
+
+```bash
+./foremanctl deploy-proxy \
+  --flavor foreman-proxy-content \
+  --certificate-bundle /root/proxy.example.com-certs.tar \
+  --foreman-fqdn quadlet.example.com
```
sjha4: Should we add more details here like steps..Run `foremanctl certificate-bundle` on foreman..Copy it over to proxy VM..Then run the deploy-proxy..

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -17,8 +16,12 @@ foreman_proxy_features: "{{ foreman_proxy_base_features + foreman_proxy_plugins
 foreman_proxy_available_features: "{{ [] | available_foreman_proxy_plugins }}"
 foreman_proxy_disabled_features: "{{ foreman_proxy_available_features | difference(foreman_proxy_features) }}"
 
-foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
+# In case of standalone proxy, we need to add foreman server URL not host fqdn
+foreman_proxy_with_foreman: "{{ 'foreman' in enabled_features }}"
+foreman_proxy_foreman_server_url: "{{ foreman_proxy_with_foreman | ternary('https://{{ ansible_facts['fqdn'] }}', foreman_proxy_foreman_server_url) }}"
```
Gauravtalreja1: Why is this variable self-referencing? Isn't that a bit confusing?

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -7,8 +7,7 @@ foreman_proxy_https_port: 8443
 foreman_proxy_url: "https://{{ foreman_proxy_name }}:{{ foreman_proxy_https_port }}"
 
 # Settings
-foreman_proxy_trusted_hosts:
-  - "{{ foreman_proxy_name }}"
+foreman_proxy_trusted_hosts: []
```
Gauravtalreja1, line 10: No --foreman-proxy-trusted-hosts parameter exposed for proxy, is that intended change?

arvind4501: Yes, thats intentional, https://github.com/theforeman/foremanctl/pull/571#discussion_r3467638974 we agreed that we should derive the trusted host from fqdn as foreman main server should be the only one in that list

src/roles/foreman_proxy/defaults/main.yaml
arvind4501: Correct, that's bit confusing, `foreman_proxy_foreman_server_url` from base.yaml should already override this

src/playbooks/deploy-proxy/metadata.obsah.yaml
```diff
@@ -0,0 +1,27 @@
...
+    choices:
+      - foreman-proxy-content
+  certificates_bundle:
+    help: Path to the certificate bundle tar file.
+    type: AbsolutePath
+    parameter: --certificate-bundle
+    persist: false
+  foreman_name:
+    parameter: --foreman-fqdn
+    help: FQDN of the Foreman server this proxy connects to.
```
Gauravtalreja1, line 17: When I ran `foremanctl deploy-proxy` without `--certificate-bundle` or `--foreman-fqdn` options, it fails with below error, instead i think it should have thrown a validation error that this are mandatory options, right?
```
[root@foreman-proxy-server ~]# foremanctl deploy-proxy
.
..
TASK [pulp : Configure Foreman Proxy] *********************************************************************************************************************************************************
    changed: false
    msg: |-
        Failed to connect to Foreman server: DocLoadingError: Could not load data from https://foreman-proxy-server.example.com: 503 Server Error: Service Unavailable for url: https://foreman-proxy-server.example.com/apidoc/v2.json
                          - is your server down?

PLAY RECAP ************************************************************************************************************************************************************************************
localhost                  : ok=126  changed=82   unreachable=0    failed=1    skipped=24   rescued=0    ignored=0

```

src/playbooks/deploy-proxy/metadata.obsah.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+variables:
+  flavor:
+    help: Base flavor to use in this deployment.
+    choices:
+      - foreman-proxy-content
+  certificates_bundle:
+    help: Path to the certificate bundle tar file.
+    type: AbsolutePath
+    parameter: --certificate-bundle
```
Gauravtalreja1, line 13: When I ran foremanctl deploy-proxy with random dir/nonexistnet `--certificate-bundle`  it fails with below error, instead i think it should have thrown a validation error file not exist or with a proper error, right?
```
[root@foreman-proxy-server ~]# foremanctl deploy-proxy --certificate-bundle ~/mycerts
.
..
TASK [pulp : Configure Foreman Proxy] *********************************************************************************************************************************************************
    changed: false
    msg: |-
        Failed to connect to Foreman server: DocLoadingError: Could not load data from https://foreman-proxy-server.example.com: 503 Server Error: Service Unavailable for url: https://foreman-proxy-server.example.com/apidoc/v2.json
                          - is your server down?

PLAY RECAP ************************************************************************************************************************************************************************************
localhost                  : ok=126  changed=82   unreachable=0    failed=1    skipped=24   rescued=0    ignored=0
```

src/playbooks/deploy-proxy/metadata.obsah.yaml
arvind4501: Yes, that might be concern, looking at https://obsah.readthedocs.io/en/latest/development.html#constraints there is no way to say `give these params otherwise deployment won't work` currently. but having said that i also expect user to follow steps for deployment for at least first time. i also wonder if we had such validations in `foreman-installer`?

src/playbooks/deploy-proxy/metadata.obsah.yaml
arvind4501: I think thats the issue of not passing fqdn, rather than certificate bundle. and validation exist for certificate bundle https://github.com/theforeman/foremanctl/pull/571/changes#diff-01885bb15c1a59c6f49bbaa057e4d11075238b4bf9ce1968f4cc4c47de259690R8

src/playbooks/deploy-proxy/metadata.obsah.yaml
arvind4501: Also, we would not want to make `--certificate-bundle` or `--foreman-fqdn` options as hard validation to be present each time we do `deploy-proxy`. for next runs if i want to add any feature i would just do `foremanctl deploy-proxy --add-feature bmc` and expect it to work without any issue

ehelms: I tend to agree -- parameters that are required the first time it ever runs but not necessarily after that is tricky. I think we should look into solving that but as a follow up and not in this initial PR. Please file a Github issue with that concern.

src/playbooks/deploy-proxy/metadata.obsah.yaml
ehelms: I think we could look into some better pre-install checks that helps fail fast but as a follow up instead of in this PR. From your testing, checks I think we should consider as a follow up:

- validating certificate bundle actually exists (and possibly that it's valid)
- that the foreman-fqdn is reachable

src/roles/certificate_bundle/tasks/main.yml
```diff
@@ -61,7 +61,7 @@
         src: "{{ certificate_bundle_server_key }}"
         dest: "{{ certificate_bundle_build_directory.path }}/certs/private/{{ certificate_bundle_hostname }}.key"
         remote_src: true
-        mode: '0600'
+        mode: '0440'
```
Gauravtalreja1, line 64: Optional: Since it would change permissions for `/var/lib/foremanctl/certs/private/$(hostname).key` to `0440`, just curious shouldn't we update the same permissions for `/var/lib/foremanctl/certs/private/$(hostname)-client.key` which is still `0600`?

src/playbooks/deploy-proxy/metadata.obsah.yaml
Gauravtalreja1: When I pass `--flavor foreman-proxy-content --foreman-fqdn foreman.example.com` along with  `--certificate-bundle ~/mycerts`, the validation works to check if certificate bundle actually exists and deploy-proxy fails with below error, but validation for foreman-fqdn is reachable still makes sense
```
TASK [certificates : Check path to certificate tar file exists] *******************************************************************************************************************************
ok: [localhost]

TASK [certificates : Fail if path to certificate tar file does not exist] *********************************************************************************************************************
    changed: false
    msg: 'Path to certificate tar file not found: /root/mycerts'
```

src/playbooks/deploy-proxy/metadata.obsah.yaml
Gauravtalreja1: Claude has created a followup issue for these discussion https://github.com/theforeman/foremanctl/issues/614

src/playbooks/deploy-proxy/metadata.obsah.yaml
Gauravtalreja1: Claude has created a followup issue for these discussion https://github.com/theforeman/foremanctl/issues/614

Gauravtalreja1 (approved):
ACK, tested with `foremanctl-3.0.0.develop-2.20260701164348391310.pr571.89.g56349d9.el9.noarch` 🍏 

Tested deploy-proxy with foreman-proxy-content flavor on a two-node setup (quadlet + proxy).

Certificate sources tested:
- Default (internal CA)
- Custom server certificates (generated via Robottelo scripts)

Validated:
- Proxy deploys successfully with both cert sources
- Only expected services running on proxy (pulp-api, pulp-content, pulp-worker, postgresql, valkey, httpd, foreman-proxy) (no foreman, candlepin, or dynflow)
- PostgreSQL has only pulp database and user on proxy (no foreman/candlepin)
- HTTPS endpoints working: /pulp/api/v3/status/ (200), /pulp/content/ (200), /rhsm (proxied to server), /katello/api/v2/repositories/ (proxied to server)
- Pulp mirror mode enabled on proxy (mirror: true, client_authentication: ["client_certificate"])
- Smart proxy capabilities include all expected features (core, smart_proxy, rpm, deb, ansible, python, container, file, certguard)
- Trusted hosts includes only foreman server FQDN

Issues found:
- #613
- #614

PR #395, opened 2026-03-03
Use PostgreSQL and Valkey from CentOS 10 Stream

evgeni:
!!! DO NOT MERGE !!!
THIS IS A TEST

PR #577, opened 2026-06-18
Switch to PostgreSQL 16 image

### LINKED ISSUE #66: How will Postgresql upgrades work if they are running inside containers?
We are versed in how Postgresql upgrades happen when using operating system packages but don't know how it is handled when Postgresql is running in containers. This issue is to research and understand how that process occurs and if it affects the design.

### LINKED ISSUE #465: Support PostgreSQL 16
PostgreSQL 13 is blocking Foreman in numerous ways:
* EL10 support
* Django 5 support (Pulp & Katello)
* Red Hat Lightspeed integration

foremanctl needs to automate the migration to PG16, fresh installation of it, and test its compatibility with our existing tools.

ShimShtein:

As part of being prepared to run on RHEL 10, we need to switch to postgres 16.

* Switch of the container to postgres 16 instead of 13.

If the system spins up successfully and works with content - it's good to go.

### DIFF
--- a/development/playbooks/remote-database/remote-database.yaml
+++ b/development/playbooks/remote-database/remote-database.yaml
@@ -7,7 +7,6 @@
     - "../../../src/vars/defaults.yml"
     - "../../../src/vars/flavors/{{ flavor }}.yml"
     - "../../../src/vars/database.yml"
-    - "../../../src/vars/images.yml"
   vars:
     certificates_hostnames:
       - "{{ ansible_facts['fqdn'] }}"
--- a/development/playbooks/remote-database/remote-database.yaml
+++ b/development/playbooks/remote-database/remote-database.yaml
@@ -7,6 +7,7 @@
     - "../../../src/vars/defaults.yml"
     - "../../../src/vars/flavors/{{ flavor }}.yml"
     - "../../../src/vars/database.yml"
+    - "../../../src/vars/images.yml"
   vars:
     certificates_hostnames:
       - "{{ ansible_facts['fqdn'] }}"
--- a/src/roles/postgresql/tasks/main.yml
+++ b/src/roles/postgresql/tasks/main.yml
@@ -17,6 +17,9 @@
   notify:
     - Restart postgresql
 
+- name: Upgrade PostgreSQL database
+  ansible.builtin.include_tasks: upgrade.yaml
+
 - name: Deploy PostgreSQL container
   containers.podman.podman_container:
     name: "{{ postgresql_container_name }}"
--- a/src/roles/postgresql/tasks/main.yml
+++ b/src/roles/postgresql/tasks/main.yml
@@ -107,18 +107,6 @@
     name: "{{ postgresql_container_name }}"
     state: started
 
-# SCRAM-SHA-256 is default for PostgreSQL 14+,
-# after the upgrade, we can drop this task.
-- name: Use scram-sha-256 for password encryption
-  community.postgresql.postgresql_set:
-    login_user: postgres
-    login_password: "{{ postgresql_admin_password }}"
-    login_host: localhost
-    name: password_encryption
-    value: "scram-sha-256"
-  notify:
-    - Restart postgresql
-
 - name: Create PostgreSQL users
   community.postgresql.postgresql_user:
     name: "{{ item.name }}"
--- a/src/roles/postgresql/tasks/upgrade.yaml
+++ b/src/roles/postgresql/tasks/upgrade.yaml
@@ -33,11 +33,12 @@
         state: stopped
         timeout: 30
 
-    - name: Upgrade the database from 13 to 16 but don't start the server
+    - name: Deploy PostgreSQL upgrade container
       containers.podman.podman_container:
         name: "{{ postgresql_container_name }}-upgrade"
-        image: "{{ postgresql_container_image }}:{{ postgresql_container_tag }}"
-        state: started
+        image: postgresql.image
+        state: quadlet
+        sdnotify: false
         volumes:
           - "{{ postgresql_data_dir }}:/var/lib/pgsql/data:rw,Z"
         secrets:
@@ -50,11 +51,36 @@
           # see https://github.com/sclorg/postgresql-container/issues/661
           POSTGRESQL_PREV_VERSION: "{{ postgresql_version }}"
           POSTGRESQL_UPGRADE: "{{ postgresql_upgrade }}"
-        detach: false
         command: "run-postgresql -V"
+        quadlet_options:
+          - |
+            [Unit]
+            Description=PostgreSQL Database Upgrade from 13 to 16
+            [Service]
+            Type=oneshot
+            RemainAfterExit=yes
+            TimeoutStartSec=30m
+
+    - name: Run daemon reload to make Quadlet create the upgrade service file
+      ansible.builtin.systemd:
+        daemon_reload: true
+
+    - name: Upgrade the database from 13 to 16 but don't start the server
+      ansible.builtin.systemd:
+        name: "{{ postgresql_container_name }}-upgrade.service"
+        state: restarted
 
     - name: Unmask PostgreSQL systemd service (Quadlet)
       ansible.builtin.systemd:
         name: postgresql.service
         state: stopped
         masked: false
+
+    - name: Remove the quadlet file
+      ansible.builtin.file:
+        path: "/etc/containers/systemd/{{ postgresql_container_name }}-upgrade.container"
+        state: absent
+
+    - name: Reload systemd
+      ansible.builtin.systemd_service:
+        daemon_reload: true
--- a/src/roles/postgresql/tasks/upgrade.yaml
+++ b/src/roles/postgresql/tasks/upgrade.yaml
@@ -21,11 +21,10 @@
     - "postgresql_version is defined"
     - "postgresql_version == '13'"
   block:
-    - name: Stop PostgreSQL systemd service (Quadlet)
+    - name: Stop foreman target
       ansible.builtin.systemd:
-        name: postgresql.service
+        name: foreman.target
         state: stopped
-        masked: true
 
     - name: Wait for port 5432 to be completely freed by Podman
       ansible.builtin.wait_for:
@@ -69,12 +68,6 @@
       ansible.builtin.systemd:
         name: "{{ postgresql_container_name }}-upgrade.service"
         state: restarted
-
-    - name: Unmask PostgreSQL systemd service (Quadlet)
-      ansible.builtin.systemd:
-        name: postgresql.service
-        state: stopped
-        masked: false
 
     - name: Remove the quadlet file
       ansible.builtin.file:
--- a/src/roles/postgresql/tasks/upgrade.yaml
+++ b/src/roles/postgresql/tasks/upgrade.yaml
@@ -1,0 +1,60 @@
+---
+- name: Check if PG_VERSION file exists
+  ansible.builtin.stat:
+    path: "{{ postgresql_data_dir }}/userdata/PG_VERSION"
+  register: postgresql_version_file
+
+- name: Set postgresql_version version fact
+  when: postgresql_version_file.stat.exists
+  block:
+    - name: Read PG_VERSION file
+      ansible.builtin.slurp:
+        src: "{{ postgresql_data_dir }}/userdata/PG_VERSION"
+      register: postgresql_version_encoded
+
+    - name: Register PG version fact
+      ansible.builtin.set_fact:
+        postgresql_version: "{{ postgresql_version_encoded.content | b64decode | trim }}"
+
+- name: Upgrade PostgreSQL 13 to 16
+  when:
+    - "postgresql_version is defined"
+    - "postgresql_version == '13'"
+  block:
+    - name: Stop PostgreSQL systemd service (Quadlet)
+      ansible.builtin.systemd:
+        name: postgresql.service
+        state: stopped
+        masked: true
+
+    - name: Wait for port 5432 to be completely freed by Podman
+      ansible.builtin.wait_for:
+        port: 5432
+        state: stopped
+        timeout: 30
+
+    - name: Upgrade the database from 13 to 16 but don't start the server
+      containers.podman.podman_container:
+        name: "{{ postgresql_container_name }}-upgrade"
+        image: "{{ postgresql_container_image }}:{{ postgresql_container_tag }}"
+        state: started
+        volumes:
+          - "{{ postgresql_data_dir }}:/var/lib/pgsql/data:rw,Z"
+        secrets:
+          - 'postgresql-admin-password,target=POSTGRESQL_ADMIN_PASSWORD,type=env'
+        env:
+          POSTGRESQL_MAX_CONNECTIONS: "{{ postgresql_max_connections }}"
+          POSTGRESQL_SHARED_BUFFERS: "{{ postgresql_shared_buffers }}"
+          POSTGRESQL_EFFECTIVE_CACHE_SIZE: "{{ postgresql_effective_cache_size }}"
+          # explicitly set the previous version to avoid the upgrade failure
+          # see https://github.com/sclorg/postgresql-container/issues/661
+          POSTGRESQL_PREV_VERSION: "{{ postgresql_version }}"
+          POSTGRESQL_UPGRADE: "{{ postgresql_upgrade }}"
+        detach: false
+        command: "run-postgresql -V"
+
+    - name: Unmask PostgreSQL systemd service (Quadlet)
+      ansible.builtin.systemd:
+        name: postgresql.service
+        state: stopped
+        masked: false
--- a/src/roles/pre_install/tasks/main.yaml
+++ b/src/roles/pre_install/tasks/main.yaml
@@ -3,6 +3,17 @@
   ansible.builtin.include_role:
     name: debug_tools
 
+- name: 'Enable postgresql:16 module'
+  ansible.builtin.command: dnf module enable -y postgresql:16
+  # can't use the `dnf` module for modules without a default stream
+  # https://github.com/ansible/ansible/issues/56504
+  # https://github.com/ansible/ansible/issues/64852
+  args:
+    creates: /etc/dnf/modules.d/postgresql.module
+  when:
+    - ansible_facts['os_family'] == 'RedHat'
+    - ansible_facts['distribution_major_version'] == '9'
+
 - name: Install podman and utilities
   ansible.builtin.package:
     name:
--- a/src/roles/postgresql/defaults/main.yml
+++ b/src/roles/postgresql/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
-postgresql_container_image: quay.io/sclorg/postgresql-13-c9s
+postgresql_container_image: quay.io/sclorg/postgresql-16-c10s
 postgresql_container_tag: "latest"
 postgresql_container_name: postgresql
 postgresql_network: host
@@ -12,3 +12,5 @@ postgresql_admin_password: "{{ undef(hint='Set a secure database password') }}"
 postgresql_max_connections: 500
 postgresql_shared_buffers: 512MB
 postgresql_effective_cache_size: 1GB
+
+postgresql_upgrade: "hardlink"
--- a/src/vars/images.yml
+++ b/src/vars/images.yml
@@ -9,7 +9,7 @@ foreman_container_tag: "{{ container_tag_stream }}"
 foreman_proxy_container_image: "quay.io/foreman/foreman-proxy"
 foreman_proxy_container_tag: "{{ container_tag_stream }}"
 
-postgresql_container_image: quay.io/sclorg/postgresql-13-c9s
+postgresql_container_image: quay.io/sclorg/postgresql-16-c10s
 postgresql_container_tag: "latest"
 pulp_container_image: quay.io/foreman/pulp
 pulp_container_tag: "foreman-{{ container_tag_stream }}"


ShimShtein:
Extracted out of https://github.com/theforeman/foremanctl/pull/395

evgeni:
> Extracted out of #395

What for? (#395 already identified the problems with it)

ShimShtein:
Because #395 has both postgres and valkey in the same PR?

ShimShtein:
I see the tests are failing on password encryption algorithm detection: https://github.com/theforeman/foremanctl/actions/runs/27768839212/job/82162895170?pr=577#step:17:643
That's weird. Locally the same setup gives me:
```
[vagrant@quadlet ~]$ sudo podman exec postgresql psql -U postgres -c 'SHOW password_encryption'
 password_encryption 
---------------------
 scram-sha-256
(1 row)
```
Which is weird.

evgeni:
> I see the tests are failing on password encryption algorithm detection: https://github.com/theforeman/foremanctl/actions/runs/27768839212/job/82162895170?pr=577#step:17:643 That's weird. Locally the same setup gives me:
> 
> ```
> [vagrant@quadlet ~]$ sudo podman exec postgresql psql -U postgres -c 'SHOW password_encryption'
>  password_encryption 
> ---------------------
>  scram-sha-256
> (1 row)
> ```
> 
> Which is weird.

It's not weird, as that was the ext DB and that wasn't using our image definition properly. Fixed.

src/roles/pre_install/tasks/main.yaml
```diff
@@ -3,6 +3,17 @@
   ansible.builtin.include_role:
     name: debug_tools
 
+- name: 'Enable postgresql:16 module'
```
ShimShtein, line 6: @evgeni, let me ask this question again: wouldn't it be more resilient to run the PG commands from inside a PG container instead of installing the client tooling directly on the host?

evgeni: Would, sure. I just had none that I could integrate easily into the various workflows where we need Postgres tooling on the host.

ekohl: https://github.com/theforeman/foremanctl/issues/605

evgeni:
The upgrade complains:
```
Jun 28 14:37:27 quadlet.example.com postgresql[123049]: Incompatible data directory.  This container image provides
Jun 28 14:37:27 quadlet.example.com postgresql[123049]: PostgreSQL '16', but data directory is of
Jun 28 14:37:27 quadlet.example.com postgresql[123049]: version '13'.
Jun 28 14:37:27 quadlet.example.com postgresql[123049]: 
Jun 28 14:37:27 quadlet.example.com postgresql[123049]: This image supports automatic data directory upgrade from
Jun 28 14:37:27 quadlet.example.com postgresql[123049]: '15', please _carefully_ consult image documentation
Jun 28 14:37:27 quadlet.example.com postgresql[123049]: about how to use the '$POSTGRESQL_UPGRADE' startup option.
```

The fact its supporting 15→16 only is encoded here:
https://github.com/sclorg/postgresql-container/blob/f1c89bca212c6a829ab74fdfdeec264f0d08a992/16/Dockerfile.c10s#L13-L14

Funnily, the container actually contains 13 binaries:
```
% podman run --rm -ti quay.io/sclorg/postgresql-16-c10s bash
…
bash-5.2$ rpm -q postgresql-upgrade
postgresql-upgrade-16.13-2.el10.x86_64
bash-5.2$ rpm -qf /usr/lib64/pgsql/postgresql-13/bin/postgres
postgresql-upgrade-16.13-2.el10.x86_64
bash-5.2$ /usr/lib64/pgsql/postgresql-13/bin/postgres --version
postgres (PostgreSQL) 13.23
```

I'd argue this is a bug in the image we should ask for a fix.

src/roles/pre_install/tasks/main.yaml
ehelms: You'd likely want to write a module that does this to be re-used because there are postgres actions that are run in a lot more places. I am not sure I see the value of it at the moment but it might be worth exploring after we get everything else.

ekohl: @ehelms Debian 13 has PostgreSQL 17. That probably means you can't run `pg_dump` on the PostgreSQL 16 container we run, which breaks backups.

development/playbooks/remote-database/remote-database.yaml
```diff
@@ -7,6 +7,7 @@
     - "../../../src/vars/defaults.yml"
     - "../../../src/vars/flavors/{{ flavor }}.yml"
     - "../../../src/vars/database.yml"
+    - "../../../src/vars/images.yml"
```
ehelms: Why is this needed now?

src/roles/postgresql/tasks/main.yml
```diff
@@ -17,6 +17,72 @@
   notify:
     - Restart postgresql
 
+- name: Check if PG_VERSION file exists
```
ehelms: This set of upgrade tasks look good. Could they go in a dedicated task file?

src/roles/postgresql/tasks/main.yml
```diff
@@ -17,6 +17,72 @@
...
+  when:
+    - "postgresql_version is defined"
+    - "postgresql_version == '13'"
+  register: postgresql_upgrade_result
+
+- name: Stop PostgreSQL systemd service (Quadlet)
+  ansible.builtin.systemd:
+    name: postgresql.service
+    masked: false
+  when:
```
ehelms: As this `when` pattern is repeated, you can put it in a `block`.

src/roles/postgresql/tasks/main.yml
```diff
@@ -17,6 +17,72 @@
...
+    state: started
+    volumes:
+      - "{{ postgresql_data_dir }}:/var/lib/pgsql/data:rw,Z"
+    secrets:
+      - 'postgresql-admin-password,target=POSTGRESQL_ADMIN_PASSWORD,type=env'
+    env:
+      POSTGRESQL_MAX_CONNECTIONS: "{{ postgresql_max_connections }}"
+      POSTGRESQL_SHARED_BUFFERS: "{{ postgresql_shared_buffers }}"
+      POSTGRESQL_EFFECTIVE_CACHE_SIZE: "{{ postgresql_effective_cache_size }}"
+      POSTGRESQL_PREV_VERSION: "{{ postgresql_version }}"
```
ekohl: Please link why this workaround is needed
```suggestion
      # https://github.com/sclorg/postgresql-container/issues/661
      POSTGRESQL_PREV_VERSION: "{{ postgresql_version }}"
```

src/roles/postgresql/tasks/main.yml
```diff
@@ -17,6 +17,72 @@
...
+      POSTGRESQL_UPGRADE: "hardlink"
```
ekohl: Can you please make this a variable so we could easily switch to `copy`?

src/roles/postgresql/tasks/main.yml
```diff
@@ -17,6 +17,72 @@
...
+    detach: false
+    debug: true
+    log_level: debug
+    command: "run-postgresql -V"
+  when:
+    - "postgresql_version is defined"
+    - "postgresql_version == '13'"
+  register: postgresql_upgrade_result
+
+- name: Stop PostgreSQL systemd service (Quadlet)
```
ekohl: Why is it needed to stop the service again?

ekohl (commented):
Wouldn't it be easier to unconditionally set `POSTGRESQL_UPGRADE` so it will always run when needed?

Thinking about the logging aspect: where do the logs of the upgrade go now? Can we somehow debug it when it fails? When it's part of the regular service then you'll have it the journal and sosreport will pick it up already.

ShimShtein:
> Wouldn't it be easier to unconditionally set POSTGRESQL_UPGRADE so it will always run when needed?

This will throw a warning: https://github.com/sclorg/postgresql-container/blob/134bb39ab5c435087487c3422522bddf160f727d/16/root/usr/share/container-scripts/postgresql/common.sh#L415 that will be visible in the logs (and if I am not mistaken it exits immediately, but at this point I am not sure if my environment was clean when I tested it). That was my original thought, but I had to work around the message.

> Thinking about the logging aspect: where do the logs of the upgrade go now? Can we somehow debug it when it fails? When it's part of the regular service then you'll have it the journal and sosreport will pick it up already.

Had to debug the process quite a lot. You will have limited logging in the playbook itself (I will remove the debug log level from the playbook), but the more interesting log will be in the data folder `{{ postgresql_data_dir }}/userdata/log/` and `{{ postgresql_data_dir }}/userdata-new/log/` depending on the failure stage. We do not collect these files explicitly in the sosreport, it is a bit tricky - I would say that it should be part of the postgresql plugin, since it's a standard postgresql action. On the other hand, the files are created there due to our specific postgresql configuration, so we can argue that we should collect this data. Let's create an issue and track postgresql upgrade logs separately.

src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -0,0 +1,77 @@
...
+        state: stopped
+        masked: true
+
+    - name: Wait for port 5432 to be completely freed by Podman
+      ansible.builtin.wait_for:
+        port: 5432
+        state: stopped
+        timeout: 30
+
+    - name: Deploy PostgreSQL upgrade container
```
evgeni, line 35: We need to clean this up after it was executed, otherwise it will try the upgrade again on next boot?

ShimShtein: I have tested, and after the reboot, the service is not there at all:
```
[vagrant@quadlet ~]$ systemctl status "postgres*" | grep service
● postgresql.service
     CGroup: /system.slice/postgresql.service
● postgresql-image.service
```
And the status shows:
```
[vagrant@quadlet ~]$ systemctl status postgresql-upgrade.service 
○ postgresql-upgrade.service - PostgreSQL Database Upgrade from 13 to 16
     Loaded: loaded (/etc/containers/systemd/postgresql-upgrade.container; generated)
     Active: inactive (dead)
```

evgeni: So, it *is* there, just not started by default. And can be started by the user.

I still think it needs cleanup.

ShimShtein: Added quadlet removal bits.

src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -0,0 +1,86 @@
...
+      containers.podman.podman_container:
+        name: "{{ postgresql_container_name }}-upgrade"
+        image: postgresql.image
+        state: quadlet
```
ekohl, line 39: Does it need to be a quadlet? In favor of it means you get logs in the journal, but it's also fragile: if Ansible somehow starts, it can leave the system in an unpredictable state.

evgeni (approved):

ShimShtein:
packit error seems unrelated:
```
2026-07-06 12:17:47.459 logging.py        INFO   HTTP ERROR response 429  [https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/develop/packages/foreman/foremanctl/foremanctl.spec]
```

ekohl (commented):
This comment is not blocking, but something to consider. Possibly in a follow up.

src/roles/postgresql/tasks/upgrade.yaml
ehelms: > if Ansible somehow starts, it can leave the system in an unpredictable state.

I don't follow this logic. However, I assume you are asking the question of a quadlet which generates a systemd service vs. just running the container directly with the equivalent of a `podman run`?

ShimShtein: >Does it need to be a quadlet? 

I had the same question. I decided to go on with a quadlet to reuse the `postgres.image` container. If I spin it up as a separate container, it means I cannot reuse the image definition and any drop-ins that will be defined there. Originally it was a regular container.

ekohl: And I messed up: I intended to write "Ansible somehow stops in the middle". Not start.

ShimShtein: I don't think it is disastrous:
* if the playbook stops before defining the service - no harm done
* if it stops after the service is defined, but before it executes the service - it will be executed at first reboot
* if it stops after the execution - it may leave some leftovers, but they would be harmless

There are options to add the upgrade as a dependency for the `postgres.container`, so it will pull the service before the execution and then push the condition into the service startup, so it will just silently finish with noop in case the DB was already migrated.

I think it's too complex for a one time only step.

src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -0,0 +1,79 @@
...
+          POSTGRESQL_EFFECTIVE_CACHE_SIZE: "{{ postgresql_effective_cache_size }}"
+          # explicitly set the previous version to avoid the upgrade failure
+          # see https://github.com/sclorg/postgresql-container/issues/661
+          POSTGRESQL_PREV_VERSION: "{{ postgresql_version }}"
+          POSTGRESQL_UPGRADE: "{{ postgresql_upgrade }}"
+        command: "run-postgresql -V"
+        quadlet_options:
+          - |
+            [Unit]
+            Description=PostgreSQL Database Upgrade from 13 to 16
```
ekohl, line 57: Perhaps make sure that systemd knows it's 1 or the other? Or will this introduce too much uncertainty?
```suggestion
            Description=PostgreSQL Database Upgrade from 13 to 16
            Conflicts=postgresql.service
```

src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -0,0 +1,79 @@
...
+            [Service]
+            Type=oneshot
+            RemainAfterExit=yes
```
ekohl, line 60: This means the service will remain as active. Is that useful?

ekohl (commented):
I would be OK with tracking the items in a follow up task when someone actually does proper testing of upgrades. Right now I consider this "good enough" to make CI pass.

src/roles/postgresql/tasks/upgrade.yaml
ShimShtein: It will mark the service as `active(exited)` when it will stop the execution. In the happy path it won't matter, since we remove the service completely, but if it fails, we will have indication whether it was run or not.

src/roles/postgresql/tasks/upgrade.yaml
ShimShtein: According to Gemini, the directive is symmetric, so it won't protect us from someone that tries to start the server. Actually it will just cause the upgrade container to stop (even more problematic in the middle of an upgrade).

src/roles/pre_install/tasks/main.yaml
```diff
@@ -3,6 +3,17 @@
...
+  ansible.builtin.command: dnf module enable -y postgresql:16
+  # can't use the `dnf` module for modules without a default stream
+  # https://github.com/ansible/ansible/issues/56504
+  # https://github.com/ansible/ansible/issues/64852
+  args:
+    creates: /etc/dnf/modules.d/postgresql.module
+  when:
+    - ansible_facts['os_family'] == 'RedHat'
+    - ansible_facts['distribution_major_version'] == '9'
```
Gauravtalreja1, line 15: I think it will make sense when database_mode is internal, right? 
```suggestion
  when:
    - ansible_facts['os_family'] == 'RedHat'
    - ansible_facts['distribution_major_version'] == '9'
    - database_mode == 'internal'
```

src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -0,0 +1,79 @@
...
+      register: postgresql_version_encoded
+
+    - name: Register PG version fact
+      ansible.builtin.set_fact:
+        postgresql_version: "{{ postgresql_version_encoded.content | b64decode | trim }}"
+
+- name: Upgrade PostgreSQL 13 to 16
+  when:
+    - "postgresql_version is defined"
+    - "postgresql_version == '13'"
```
Gauravtalreja1, line 22: Instead of hardcoding `13` here, how about we define `postgresql_target_version = 16` in role defaults (or fetch from image name?), so whenever in future we want to update PG 16->17 then we just update `postgresql_target_version` variable and it should work without many changes
```suggestion
  when:
    - postgresql_version is defined
    - postgresql_version | int < postgresql_target_version | int
```

src/roles/postgresql/tasks/main.yml
```diff
@@ -17,6 +17,9 @@
   notify:
     - Restart postgresql
 
+- name: Upgrade PostgreSQL database
+  ansible.builtin.include_tasks: upgrade.yaml
```
Gauravtalreja1, line 21: I think we should include this play only when `database_mode` is `internal`. For an external database_mode, we'd expect it to be managed explicitly by the user, right?
```suggestion
- name: Upgrade PostgreSQL database
  ansible.builtin.include_tasks: upgrade.yaml
  when: database_mode == 'internal'
```

src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -0,0 +1,79 @@
...
+          POSTGRESQL_UPGRADE: "{{ postgresql_upgrade }}"
+        command: "run-postgresql -V"
+        quadlet_options:
+          - |
+            [Unit]
+            Description=PostgreSQL Database Upgrade from 13 to 16
+            [Service]
+            Type=oneshot
+            RemainAfterExit=yes
+            TimeoutStartSec=30m
```
Gauravtalreja1, line 61: Nit: Instead of hardcoding timeout here, how about move it to roles default var like `postgresql_upgrade_timeout` and using here that var?

src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -0,0 +1,79 @@
...
+        state: restarted
+
+    - name: Remove the quadlet file
+      ansible.builtin.file:
+        path: "/etc/containers/systemd/{{ postgresql_container_name }}-upgrade.container"
+        state: absent
+
+    - name: Reload systemd
+      ansible.builtin.systemd_service:
+        daemon_reload: true
```
Gauravtalreja1, line 79: I tested upgrade from PG 13 ->16 and upgrade completed smoothly, but `systemctl list-units` shows a stale entry:
```
# systemctl list-units | grep postgresql-upgrade
● postgresql-upgrade.service                                                                                                           not-found active exited    postgresql-upgrade.service
```
The quadlet file is removed and daemon-reload runs, but systemd still holds the unit in memory since it ran during this boot session. Harmless (upgrade won't re-run since PG_VERSION is now 16), but it would look broken to anyone inspecting systemctl list-units until server is rebooted, right?

```
# systemctl status postgresql-upgrade
● postgresql-upgrade.service
     Loaded: not-found (Reason: Unit postgresql-upgrade.service not found.)
     Active: active (exited) since Thu 2026-07-09 20:50:04 EDT; 30min ago
   Main PID: 40013 (code=exited, status=0/SUCCESS)
        CPU: 20.621s

Jul 09 20:50:03 quadlet.example.com postgresql-upgrade[40187]: server stopped
Jul 09 20:50:03 quadlet.example.com postgresql-upgrade[40013]: waiting for server to shut down.... done
Jul 09 20:50:03 quadlet.example.com postgresql-upgrade[40013]: server stopped
Jul 09 20:50:03 quadlet.example.com postgresql-upgrade[40187]: Starting server...
Jul 09 20:50:03 quadlet.example.com postgresql-upgrade[40013]: Starting server...
Jul 09 20:50:03 quadlet.example.com postgresql-upgrade[40187]: postgres (PostgreSQL) 16.13
Jul 09 20:50:03 quadlet.example.com postgresql-upgrade[40013]: postgres (PostgreSQL) 16.13
Jul 09 20:50:03 quadlet.example.com podman[40013]: 2026-07-09 20:50:03.921408521 -0400 EDT m=+13.222458407 container died>
Jul 09 20:50:04 quadlet.example.com podman[40013]: 2026-07-09 20:50:04.157847588 -0400 EDT m=+13.458897484 container remo>
Jul 09 20:50:04 quadlet.example.com systemd[1]: Finished PostgreSQL Database Upgrade from 13 to 16.
```

Gauravtalreja1 (commented):
Tested with `foremanctl-3.0.0.develop-2.20260707081906812166.pr577.71.g8611011.el9.noarch` 🍏 

**Validated:**

* Fresh deployment with PostgreSQL 16: service starts successfully, port is reachable, `scram-sha-256` authentication is enabled, and all required databases are created.
* Upgrade from PostgreSQL 13 to 16: databases, users, and permissions are preserved, and the upgrade is correctly skipped on subsequent deployments.
* Minor issue identified: a stale `postgresql-upgrade.service` unit (`not-found`, `active (exited)`) remains visible in systemd until a reboot. This appears to be cosmetic only.
* Added a few inline code review comments.

src/roles/postgresql/tasks/upgrade.yaml
Gauravtalreja1: What's the intent behind `RemainAfterExit=yes` here? After the quadlet file is removed and systemd reloads, the unit shows as `not-found active exited` in `systemctl list-units`, it disappears after reboot anyway. Just curious what purpose does keeping it as `active(exited)` serve?

src/roles/postgresql/tasks/upgrade.yaml
ekohl: The container we use only supports version 13. I'd say it makes sense to hardcode it.

Gauravtalreja1: I'm not suggesting that the source version shouldn't stay hardcoded, in fact, it's already determined from the host's `PG_VERSION` file.

My suggestion is only to make `postgresql_target_version` configurable. That way, when we eventually move from PostgreSQL 16 to 17, we only need to update `postgresql_target_version` instead of modifying the upgrade logic in this playbook.

ekohl: > My suggestion is only to make `postgresql_target_version` configurable. That way, when we eventually move from PostgreSQL 16 to 17, we only need to update `postgresql_target_version` instead of modifying the upgrade logic in this playbook.

I think that's too premature. By the time we get there, we should look at a concrete plan so I'd prefer to keep the current implementation.

src/roles/postgresql/tasks/main.yml
ShimShtein: I am following the container creation task, and it does not have a `when` statement.

src/roles/postgresql/tasks/upgrade.yaml
ShimShtein: I don't think we will reuse this value. Can't see a reason to complicate the code for it.

src/roles/postgresql/tasks/upgrade.yaml
ShimShtein: I think it's a quite good state that we have:
`Loaded: not-found` means that I am cleaning up the service correctly, and it won't appear anymore in systemd.
`active (exited)` means that the service executed its work and succeeded.

So the current state is an in-memory last-minute trace that will go away after reboot. I think it's good to have it, in case the upgrade did something weird and we want the "paper trail" to investigate.

src/roles/postgresql/tasks/upgrade.yaml
ShimShtein: I have detailed it in my other comment: https://github.com/theforeman/foremanctl/pull/577#discussion_r3555788649
TL;DR: it's an in-memory paper trail for making sure the process succeeded, and it will go away once the system is rebooted.

This is besides the "unhappy" scenario, where we want the service to be there until the upgrade succeeds.

src/roles/postgresql/tasks/main.yml
evgeni: The whole role is not executed for `database_mode = external`

src/roles/pre_install/tasks/main.yaml
evgeni: No, we need the matching tools also for extDB

Gauravtalreja1 (approved):

PR #581, opened 2026-06-19
Don't leak PostgreSQL credentials when looping over DBs

evgeni:

The list we're looping over includes credentials, so we should not log any of that.

Fixes: ba1f1a500901f6f8c7347a63969e9e66c33438fb

* set no_log

Steps to reproduce:

*

#### Checklist


arvind4501 (approved):

ehelms:
We would benefit from having something that helps identify potential places where credentials might be leaked on PRs. This doesn't feel overly deterministic so tools like ansible-lint won't be able to catch it. Maybe this is a case for having an agentic review skill?

evgeni:
If you can teach the rabbit, go ahead.

PR #582, opened 2026-06-19
Switch to Valkey 8

evgeni:

We want to use Valkey 8

* 

Steps to reproduce:

*

#### Checklist


ehelms (approved):

archanaserver:
I tested the upgrade path and one thing I don't understand, Do we also need an upgrade cleanup task to stop/disable/remove the old `redis.service` quadlet/unit? (I think we should)
Because the existing deployments still have the old redis service enabled, and since both redis and valkey use port `6379`, that causing conflict with `valkey.service`. I see the old `redis.service` remains active/running from the previous deployment, and looking at it shows `redis-server` still listening on `:6379`. So how we are handling this, if I'm missing on some information?
```
  valkey[326282]: # Warning: Could not create server TCP listening socket *:6379: bind: Address already in use               
  valkey[326282]: # Failed listening on port 6379 (tcp), aborting.
```

evgeni:
Yeah, we will need that for the upgrades. That was precisely the reason why I did https://github.com/theforeman/foremanctl/pull/586 -- to quickly identify upgrade issues that come from the fact that the deployment details, not the images, change.

evgeni:
@archanaserver I have now included [src/roles/valkey/tasks/cleanup_redis.yaml](https://github.com/theforeman/foremanctl/pull/582/changes#diff-1f0b098c40aacc696d4d1f94f2526ee58454c3cc7ba9fe2a6e80e6991f4171c0) now, which should contain enough cleanup to upgrade cleanly

evgeni:
```
At least one activation key must have a lifecycle environment and content view assigned to it (HTTP error code 500: Internal Server Error)
```

Yay. Wat.

evgeni:
> ```
> At least one activation key must have a lifecycle environment and content view assigned to it (HTTP error code 500: Internal Server Error)
> ```
> 
> Yay. Wat.

#594

archanaserver:
> @archanaserver I have now included [src/roles/valkey/tasks/cleanup_redis.yaml](https://github.com/theforeman/foremanctl/pull/582/changes#diff-1f0b098c40aacc696d4d1f94f2526ee58454c3cc7ba9fe2a6e80e6991f4171c0) now, which should contain enough cleanup to upgrade cleanly

I see `redis.service` is no longer present so this fixed the issue, LGTM

archanaserver (approved):
LGTM!

src/roles/valkey/tasks/cleanup_redis.yaml
```diff
@@ -0,0 +1,23 @@
+---
```
shubhamsg199, line 1: I could be wrong but In case of migration, we would also need to migrate data from /var/lib/redis to /var/lib/valkey right?

evgeni: No. There is no valuable data in there. Only caches.
(We also do not back up /var/lib/redis on RPM based installations today)

src/roles/valkey/tasks/cleanup_redis.yaml
```diff
@@ -0,0 +1,23 @@
...
+  when: ansible_facts['services']['redis.service']['status'] | default('not-found') != 'not-found'
+
+- name: Remove Redis package
+  ansible.builtin.package:
+    name: redis
+    state: absent
+
+- name: Remove Redis container
+  ansible.builtin.file:
+    path: /etc/containers/systemd/redis.container
```
Gauravtalreja1, line 28: Should we clean up `/etc/containers/systemd/redis.image` and `/etc/containers/systemd/redis.image.d/`?

src/roles/valkey/tasks/cleanup_redis.yaml
```diff
@@ -0,0 +1,23 @@
+---
+- name: Populate service facts
+  ansible.builtin.service_facts:
+
+- name: Stop Redis service
+  ansible.builtin.service:
+    name: redis
+    state: stopped
+  when: ansible_facts['services']['redis.service']['status'] | default('not-found') != 'not-found'
```
Gauravtalreja1, line 9: Should we add this `when:` condition to other tasks of this playbook, else those will be running on every run?

evgeni: No, the other tasks look up the item in question and if it's not there do nothing.

Here we ensure a service to be stopped (not absent!) and that only works if the service actually exists, so we need to skip the execution if it does not.

src/roles/valkey/tasks/cleanup_redis.yaml
evgeni: Certainly won't hurt.

Gauravtalreja1: Also, should we remove the remnant container image for redis as part of this too?

src/roles/valkey/tasks/cleanup_redis.yaml
Gauravtalreja1: yeah, I get it, but if item is present then it will run and it would check every time on deploy followed by `daemon_reload`, right? if yes, then I think it would be better to skip entire block/play if service doesn't exist

src/roles/valkey/tasks/cleanup_redis.yaml
evgeni: You mean the one inside the podman storage? Not sure.

Given the number of users updating from containerized-redis to containerized-valkey will be small, I don't think it's worth it.

src/roles/valkey/tasks/cleanup_redis.yaml
evgeni: The presence/absence of the service does not *strictly* dictate the same for the files, so no.
But I've added a `when` to the daemon reload based on the removal of the `.container` file (as that's the one that is used for the service generation)

src/roles/valkey/tasks/cleanup_redis.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+- name: Remove Redis package
+  ansible.builtin.package:
+    name: redis
+    state: absent
+
+- name: Remove Redis image
```
ShimShtein, line 16: Should we use the ready-made [image module](https://docs.ansible.com/projects/ansible/latest/collections/containers/podman/podman_image_module.html) from Ansible instead of chasing the services e.t.c.

src/roles/valkey/tasks/main.yaml
```diff
@@ -1,25 +1,28 @@
...
-    path: /var/lib/redis
+    path: /var/lib/valkey
     state: directory
     owner: 1001
     group: 1001
     mode: '0755'
 
-- name: Run Redis as a container
+- name: Run Valkey as a container
   containers.podman.podman_container:
```
ShimShtein, line 17: Did we consider using the [podman_quadlet](https://docs.ansible.com/projects/ansible/latest/collections/containers/podman/podman_quadlet_module.html) module? It has much less options, so I am not sure it will do the trick here.

src/vars/images.yml
```diff
@@ -13,5 +13,5 @@ postgresql_container_image: quay.io/sclorg/postgresql-13-c9s
 postgresql_container_tag: "latest"
 pulp_container_image: quay.io/foreman/pulp
 pulp_container_tag: "foreman-{{ container_tag_stream }}"
-redis_container_image: quay.io/sclorg/redis-6-c9s
-redis_container_tag: "latest"
+valkey_container_image: quay.io/sclorg/valkey-8-c10s
```
ShimShtein, line 16: Why do we need to specify the image twice? There is also mention of the image in `valkey/defaults/main.yaml`

src/vars/images.yml
```diff
@@ -13,5 +13,5 @@ postgresql_container_image: quay.io/sclorg/postgresql-13-c9s
...
+valkey_container_tag: "latest"
```
ShimShtein, line 17: I suppose this means we're OK with whatever comes into RHEL 10?

src/roles/valkey/tasks/cleanup_redis.yaml
```diff
@@ -0,0 +1,36 @@
...
+- name: Populate service facts
+  ansible.builtin.service_facts:
+
+- name: Stop Redis service
+  ansible.builtin.service:
+    name: redis
+    state: stopped
+  when: ansible_facts['services']['redis.service']['status'] | default('not-found') != 'not-found'
+
+- name: Remove Redis package
```
ShimShtein, line 11: If I am not mistaken, this means we assume Redis package is installed directly on the host. In which scenario it would be installed? I was under the impression that it should be only installed as a container.

src/roles/valkey/tasks/main.yaml
```diff
@@ -1,25 +1,28 @@
...
+    name: valkey
+    image: valkey.image
     state: quadlet
     network: host
     sdnotify: true
-    command: ["run-redis", "--supervised", "systemd"]
+    command: ["run-valkey", "--supervised", "systemd"]
     volumes:
-      - /var/lib/redis:/data:rw,Z
+      - /var/lib/valkey:/data:rw,Z
```
ShimShtein, line 25: This assumes we don't need data migration. Seems OK to me, but just want to confirm.

Also I suppose we can delete the old folder, if we don't use it anymore.

src/roles/valkey/tasks/cleanup_redis.yaml
```diff
@@ -0,0 +1,36 @@
+---
+- name: Populate service facts
+  ansible.builtin.service_facts:
+
+- name: Stop Redis service
+  ansible.builtin.service:
+    name: redis
+    state: stopped
+  when: ansible_facts['services']['redis.service']['status'] | default('not-found') != 'not-found'
```
ShimShtein, line 9: According to Claude, we should rely on existence of `redis.service` under `services`. It says that the `status` field may not be there. Didn't check it further myself, so I would accept it's not correct if you say so.

src/roles/valkey/tasks/cleanup_redis.yaml
evgeni: I am not aware that this module can *remove* quadlet definitions (which is what we're removing here). But if you have an example how it can be done: yes, that'd be cleaner then.

src/roles/valkey/tasks/main.yaml
evgeni: We didn't, as that module didn't exist when we originally written this code. It's worth investigating whether we should switch the whole codebase to it. But not in this PR which merely renames Redis→Valkey.

src/vars/images.yml
evgeni: So that the role can work standalone, while this file being the canonical source of truth for the foremanctl-based deployment.

src/vars/images.yml
evgeni: Same as before. yes.

src/roles/valkey/tasks/cleanup_redis.yaml
evgeni: For upgrades from RPM-based installations, it will be a package.
Not necessary today, but will be in the future, and typing it out already was cheap enough.

src/roles/valkey/tasks/main.yaml
evgeni: Indeed, the data is not valuable: https://github.com/theforeman/foremanctl/pull/582#pullrequestreview-4568546105

And yes, we could delete the old folder in `cleanup_redis`.

src/roles/valkey/tasks/cleanup_redis.yaml
evgeni: I have no friends called Claude.

I do however have access to the Ansible documentation, where this exact line is used in an example how to do it: https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/service_facts_module.html#examples

stejskalleos (approved):
Running on the existing instance, the Redis service is not used anymore, and Foreman is working as expected.

```
systemctl list-units redis*
  UNIT                LOAD      ACTIVE SUB    DESCRIPTION        
● redis-image.service not-found active exited redis-image.service

LOAD   = Reflects whether the unit definition was properly loaded.
ACTIVE = The high-level unit activation state, i.e. generalization of SUB.
SUB    = The low-level unit activation state, values depend on unit type.
1 loaded units listed. Pass --all to see loaded but inactive units, too.
To show all installed unit files use 'systemctl list-unit-files'.
```

src/roles/valkey/tasks/cleanup_redis.yaml
ShimShtein:
```
- name: Remove an image
  containers.podman.podman_image:
    name: quay.io/bitnami/wildfly
    state: absent
```
From the example. I suppose it would be a follow-up

evgeni: Which removes an image, not an image definition (which is what we're removing here).

```console
# cat t.yaml 
---
- hosts: localhost
  tasks:
  - name: Generate image file
    containers.podman.podman_image:
      name: "quay.io/bitnami/wildfly"
      state: quadlet
      quadlet_filename: "wildfly"

  - name: Delete image definition
    containers.podman.podman_image:
      name: "quay.io/bitnami/wildfly"
      state: absent
      quadlet_filename: "wildfly"

# ansible-playbook t.yaml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] 

TASK [Gathering Facts] 
ok: [localhost]

TASK [Generate image file] 
changed: [localhost]

TASK [Delete image definition] 
ok: [localhost]

PLAY RECAP 
localhost                  : ok=3    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

# ls -alh /etc/containers/systemd/
total 12K
drwxr-xr-x. 2 root root 4.0K Jun 29 08:14 .
drwxr-xr-x. 7 root root 4.0K Jun 29 08:12 ..
-rw-r-----. 1 root root   38 Jun 29 08:14 wildfly.image
```

ShimShtein: I am just reading the podman module source...
Apparently the `podman_quadlet` module should remove all the leftovers (at least in some cases): https://github.com/containers/ansible-podman-collections/blob/efa337c5362f8603d4e1b105fc2d89d894bdd8fe/plugins/modules/podman_quadlet.py#L720
Still trying to understand how to properly use the module though.

evgeni: Yeah, I think using `podman_quadlet` might be a way for this.

evgeni: https://github.com/theforeman/foremanctl/issues/606

src/roles/valkey/tasks/main.yaml
evgeni: https://github.com/theforeman/foremanctl/issues/606

src/roles/valkey/tasks/cleanup_redis.yaml
Gauravtalreja1: yeah, from the podman storage using podman rmi command, but I agree, since we've very less users updating it would be auto removed when running `podman image prune --all`, so we're good! Thank you.

PR #583, opened 2026-06-19
My stab at a proxy deployment

evgeni:

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/playbooks/deploy-node/deploy-node.yaml
+++ b/src/playbooks/deploy-node/deploy-node.yaml
@@ -1,7 +1,7 @@
 ---
-- name: Setup quadlet demo machine
+- name: Setup proxy demo machine
   hosts:
-    - quadlet
+    - proxy
   become: true
   vars_files:
     - "../../vars/defaults.yml"
@@ -18,7 +18,6 @@
       vars:
         checks_databases: "{{ all_databases }}"
     - role: certificates
-      when: "certificates_source in ['default', 'custom_server']"
     - role: certificate_checks
       vars:
         certificate_checks_certificate: "{{ server_certificate }}"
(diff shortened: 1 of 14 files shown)


src/playbooks/deploy-node/deploy-node.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - "../../vars/flavors/{{ flavor }}.yml"
+    - "../../vars/certificates.yml"
+    - "../../vars/images.yml"
+    - "../../vars/tuning/{{ tuning }}.yml"
+    - "../../vars/database.yml"
+    - "../../vars/foreman.yml"
+    - "../../vars/base.yaml"
+  roles:
+    - role: pre_install
+    - role: checks
```
evgeni, line 17: this one should probably gain a check "is the target foreman instance reachable"

src/playbooks/deploy-node/metadata.obsah.yaml
```diff
@@ -0,0 +1,20 @@
...
+    help: Path to the certificates bundle
+    type: AbsolutePath
+    parameter: --certificate-bundle
+    persist: false
+  foreman_name:
+    help: Hostname of the main Foreman instance
+    type: FQDN
+
+include:
+  - _database_mode
```
evgeni, line 16: this would allow proxies with extdb, which we did not allow in the past -- I'd prefer dropping it, to make it less complex?

src/playbooks/deploy-node/metadata.obsah.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - _database_connection
+  - _tuning
+  - _flavor_features
+  - _pulp
```
evgeni, line 20: do we need import/export on the proxy?

evgeni:
This deploys 2 10GB VMs, and then wants that Chrome works… lol

ianballou:
What is the relationship between this PR and https://github.com/theforeman/foremanctl/pull/571 ?

evgeni:
571 is the thing, this is my playground for trying things out

evgeni:
#571 is where the fun is, closing here

PR #586, opened 2026-06-22
Use different foremanctl versions when testing upgrades

evgeni:

Instead of testing different container images using the same foremanctl version, actually deploy an older foremanctl version (and the images it is configured for), then upgrade to the "latest"

* check out `2.y-stable` branch, use that for the initial deployment, then switch to the PR/main branch

Steps to reproduce:

* CI

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -222,12 +222,13 @@ jobs:
       fail-fast: false
       matrix:
         upgrade_from:
-          - '3.18'
-        upgrade_to:
-          - 'nightly'
+          - '2.y-stable'
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7
+        with:
+          ref: ${{ matrix.upgrade_from }}
+          fetch-depth: 0
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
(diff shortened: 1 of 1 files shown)


evgeni:
This should show us if we have any issues when doing switches like #577 or #582

.github/workflows/test.yml
```diff
@@ -222,12 +222,13 @@ jobs:
       fail-fast: false
       matrix:
         upgrade_from:
-          - '3.18'
-        upgrade_to:
-          - 'nightly'
+          - '2.y-stable'
```
archanaserver, line 225: Do we want to keep `upgrade_from` as a matrix even if it only one path for now?

archanaserver (approved):

arvind4501 (approved):
Impressive, i now can foresee `upgrade_from` being extended to  3.y.stable and so on thanks

.github/workflows/test.yml
archanaserver: But I also believe that's fine, bcs it is very likely we may add more paths later

evgeni: it also was a single path in the past, but it seemed more natural to use the "matrix" section as the "configuration", instead of adjusting it in a step directly

PR #588, opened 2026-06-22
Support controller node and target node backup and add tests for backup

sjha4:
Add tests for foremanctl backup

* Add tests for foremanctl backup

The CI should run the new tests.
Steps to reproduce:

* Check CI

#### Checklist

PS: I had some issues with the OBSAH_STATE_PATH running these locally in a repo checkout. The backup is supposed to work on a host with services deployed and with path as /var/lib/foremanctl. The repo setup is different from production-like installs where the quadlet is mapped to localhost. On dev setup, quadlet is actual VM so the test finds OBSAH_STATE_PATH as /.var/lib/foremanctl inside the VM. Hence the conditional test for foremanctl state archive which archives this path.

### DIFF
--- a/src/playbooks/backup/metadata.obsah.yaml
+++ b/src/playbooks/backup/metadata.obsah.yaml
@@ -9,6 +9,11 @@ variables:
     type: AbsolutePath
     persist: false
 
+
+  target_host:
+    help: Target host to backup (defaults to quadlet)
+    persist: false
+
   skip_pulp_content:
     help: Skip Pulp content directory backup
     action: store_true
--- a/src/roles/backup/tasks/main.yaml
+++ b/src/roles/backup/tasks/main.yaml
@@ -86,37 +86,30 @@
 
     - name: Build database names list for display
       ansible.builtin.set_fact:
-        backup_databases_to_backup: "{{ backup_databases | map(attribute='database') | list }}"
+        backup_databases_to_backup: "{{ backup_databases | map(attribute='name') | list }}"
 
     - name: Dump databases
       ansible.builtin.include_tasks:
         file: database_dumps.yaml
 
-    - name: Create temporary directory on controller
-      ansible.builtin.tempfile:
-        state: directory
-        suffix: foremanctl-backup
-      delegate_to: localhost
-      become: false
-      register: backup_temp_dir
-
     - name: Archive foremanctl state on controller
       community.general.archive:
         path: "{{ obsah_state_path }}"
-        dest: "{{ backup_temp_dir.path }}/foremanctl-state.tar.gz"
+        dest: "{{ obsah_state_path }}/foremanctl-state.tar.gz"
         format: gz
+        mode: '0644'
       delegate_to: localhost
       become: false
 
     - name: Copy foremanctl state archive to target
       ansible.builtin.copy:
-        src: "{{ backup_temp_dir.path }}/foremanctl-state.tar.gz"
+        src: "{{ obsah_state_path }}/foremanctl-state.tar.gz"
         dest: "{{ backup_dir_full }}/foremanctl-state.tar.gz"
         mode: '0644'
 
-    - name: Clean up temporary directory on controller
+    - name: Clean up foremanctl state archive on controller
       ansible.builtin.file:
-        path: "{{ backup_temp_dir.path }}"
+        path: "{{ obsah_state_path }}/foremanctl-state.tar.gz"
         state: absent
       delegate_to: localhost
       become: false
--- a/src/roles/backup/tasks/main.yaml
+++ b/src/roles/backup/tasks/main.yaml
@@ -1,7 +1,7 @@
 ---
 - name: Set backup timestamp
   ansible.builtin.set_fact:
-    backup_timestamp: "{{ ansible_date_time.iso8601_basic_short }}"
+    backup_timestamp: "{{ ansible_facts['date_time']['iso8601_basic_short'] }}"
 
 - name: Set full backup directory path
   ansible.builtin.set_fact:
@@ -49,7 +49,7 @@
       ansible.builtin.systemd:
         name: postgresql.service
       register: backup_postgres_status
-      until: backup_postgres_status.status.ActiveState == 'inactive'
+      until: backup_postgres_status.status.ActiveState in ['inactive', 'failed']
       retries: "{{ backup_postgresql_stop_retries }}"
       delay: "{{ backup_postgresql_stop_delay }}"
       when: backup_database_mode == 'internal'
--- a/src/roles/backup/tasks/main.yaml
+++ b/src/roles/backup/tasks/main.yaml
@@ -92,12 +92,34 @@
       ansible.builtin.include_tasks:
         file: database_dumps.yaml
 
-    - name: Backup foremanctl state directory
+    - name: Create temporary directory on controller
+      ansible.builtin.tempfile:
+        state: directory
+        suffix: foremanctl-backup
+      delegate_to: localhost
+      become: false
+      register: backup_temp_dir
+
+    - name: Archive foremanctl state on controller
       community.general.archive:
         path: "{{ obsah_state_path }}"
-        dest: "{{ backup_dir_full }}/foremanctl-state.tar.gz"
+        dest: "{{ backup_temp_dir.path }}/foremanctl-state.tar.gz"
         format: gz
+      delegate_to: localhost
+      become: false
+
+    - name: Copy foremanctl state archive to target
+      ansible.builtin.copy:
+        src: "{{ backup_temp_dir.path }}/foremanctl-state.tar.gz"
+        dest: "{{ backup_dir_full }}/foremanctl-state.tar.gz"
         mode: '0644'
+
+    - name: Clean up temporary directory on controller
+      ansible.builtin.file:
+        path: "{{ backup_temp_dir.path }}"
+        state: absent
+      delegate_to: localhost
+      become: false
 
     - name: Backup pulp content
       ansible.builtin.include_tasks:
--- a/tests/backup_test.py
+++ b/tests/backup_test.py
@@ -1,4 +1,5 @@
 import re
+import subprocess
 
 import pytest
 import yaml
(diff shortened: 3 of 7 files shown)


src/roles/backup/tasks/main.yaml
```diff
@@ -49,7 +49,7 @@
       ansible.builtin.systemd:
         name: postgresql.service
       register: backup_postgres_status
-      until: backup_postgres_status.status.ActiveState == 'inactive'
+      until: backup_postgres_status.status.ActiveState in ['inactive', 'failed']
```
sjha4, line 52: Changed this to include `failed` as an acceptable state to continue with backup. This was happening in the CI test because postgresql.service was getting a SIGTERM and also makes sense for actual backup scenario where we may want to take a backup if state is failed.

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+
+    return {
+        'result': result,
+        'backup_dir': full_backup_path,
+        'backup_dir_name': backup_dir_name,
+    }
+
+
+def test_foreman_service_running(server):
+    """Test that Foreman is deployed and running before backup tests"""
```
ehelms: This seems unnecessary, if this is a worry then the test should ensure this during setup.

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+    foreman = server.service("foreman.target")
+    assert foreman.is_running, "Foreman must be running to test backup"
+
+
+def test_postgresql_running(server, database_mode):
+    """Test that PostgreSQL is running before backup tests"""
```
ehelms: This seems unnecessary, if this is a worry then the test should ensure this during setup.

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+
+    result = subprocess.run(
+        backup_cmd,
+        cwd=foremanctl_dir,
+        capture_output=True,
+        text=True
+    )
+
+    find_result = server.run(f"ls -1 {BACKUP_DIR}")
+    assert find_result.rc == 0, f"Backup directory should exist on VM. ls output: {find_result.stderr}"
```
ehelms, line 56: This looks like it overlaps with `test_backup_directory_created`

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+    assert result.rc == 0, f"Failed to create backup directory on VM: {result.stderr}"
+
+    foremanctl_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    backup_cmd = [
+        './foremanctl',
+        'backup',
+        BACKUP_DIR
+    ]
+
+    result = subprocess.run(
```
ehelms, line 49:
```suggestion
    result = server.run(
```

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+        assert file_check.is_file, f"{dump_file} should be a file"
+        assert file_check.size > 0, f"{dump_file} should not be empty"
+
+
+@pytest.mark.feature("iop")
+def test_iop_database_dumps_created(server, backup_result):
+    backup_dir = backup_result['backup_dir']
+
+    expected_iop_dumps = [
+        'iop_advisor.dump',
```
ehelms, line 133: Are these the right names?

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+
+    for dump_file in expected_iop_dumps:
+        dump_path = f"{backup_dir}/{dump_file}"
+        file_check = server.file(dump_path)
+        assert file_check.exists, f"IOP database dump {dump_file} should exist at {dump_path}"
+        assert file_check.is_file, f"{dump_file} should be a file"
+        assert file_check.size > 0, f"{dump_file} should not be empty"
+
+
+def test_database_dumps_valid_format(server, backup_result):
```
ehelms: You could just add this to the assertions above and combine the tests.

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+        assert result.rc == 0
+        assert result.stdout.startswith('PGDMP'), \
+            f"{dump_file} should be a valid pg_dump custom format file (should start with PGDMP)"
+
+
+def test_foremanctl_state_archived(server, backup_result):
+    """
+    Test that foremanctl state directory is archived.
+
+    NOTE: This test is conditional because foremanctl-state.tar.gz is only
```
ehelms: This test should not be conditional, that tells me there is a design flaw that should be fixed.

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+    pulp_archive = f"{backup_dir}/pulp-content.tar.gz"
+
+    archive_contents = server.run(f"tar -tzf {pulp_archive}").stdout
+
+    assert 'media/exports' not in archive_contents, "Archive should exclude media/exports"
+    assert 'media/imports' not in archive_contents, "Archive should exclude media/imports"
+    assert 'media/sync_imports' not in archive_contents, "Archive should exclude media/sync_imports"
+
+
+def test_metadata_file_created(server, backup_result):
```
ehelms:
```suggestion
def test_metadata_file_exists(server, backup_result):
```

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+
+
+def test_metadata_structure(server, backup_result):
+    backup_dir = backup_result['backup_dir']
+    metadata_file = f"{backup_dir}/metadata.yml"
+
+    result = server.run(f"cat {metadata_file}")
+    assert result.rc == 0, "Should be able to read metadata.yml"
+
+    metadata = yaml.safe_load(result.stdout)
```
ehelms: Could you use `server.file(metadata_file).content_string` instead?

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+
+    required_fields = ['hostname', 'os_version', 'type', 'timestamp', 'databases', 'database_mode']
+    for field in required_fields:
+        assert field in metadata, f"Metadata should contain '{field}' field"
+
+    assert metadata['type'] == 'offline', "Backup type should be 'offline'"
+    assert metadata['incremental'] is False, "Backup should not be incremental"
+    assert metadata['database_mode'] in ['internal', 'external'], "Database mode should be 'internal' or 'external'"
+
+    # Core databases should always be present
```
ehelms: This will end up being wrong in the long run due to flavors.

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+    core_databases = {'foreman', 'candlepin', 'pulp'}
+    actual_databases = set(metadata['databases'])
+    assert core_databases.issubset(actual_databases), \
+        f"Core databases {core_databases} should be present, got {actual_databases}"
+
+    # If IOP is enabled, check for IOP databases
```
ehelms: I'd pull this out to rely on markers like you did before in this file.

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+        assert iop_databases.issubset(actual_databases), \
+            f"IOP databases {iop_databases} should be present when IOP is enabled, got {actual_databases}"
+
+
+def test_metadata_backed_up_components(server, backup_result):
+    backup_dir = backup_result['backup_dir']
+    metadata_file = f"{backup_dir}/metadata.yml"
+
+    result = server.run(f"cat {metadata_file}")
+    metadata = yaml.safe_load(result.stdout)
```
ehelms: See previous comment. And also, given how many times this happens, should it be a fixture?

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+    dir_timestamp = backup_dir_name.replace('foreman-backup-', '')
+
+    result = server.run(f"cat {metadata_file}")
+    metadata = yaml.safe_load(result.stdout)
+
+    assert metadata['timestamp'] == dir_timestamp, \
+        "Metadata timestamp should match backup directory timestamp"
+
+
+def test_services_running_after_backup(server, backup_result, database_mode):
```
ehelms: Could we rely on the health check command to do this instead?

tests/backup_test.py
```diff
@@ -0,0 +1,367 @@
...
+    # Only check PostgreSQL service in internal database mode
+    if database_mode == 'internal':
+        key_services.insert(0, "postgresql")
+
+    for service_name in key_services:
+        service = server.service(service_name)
+        assert service.is_running, f"{service_name} should be running after backup"
+
+
+def test_backup_files_readable(server, backup_result):
```
ehelms: Since these files are all read in previous tests, should the permissions just be added into each one?

tests/backup_test.py
sjha4: Updated..

tests/backup_test.py
sjha4: I have updated backup to handle localhost and quadlet based deployments. Have kept it as a separate commit to isolate the change there..

tests/backup_test.py
sjha4: We base these based on name : https://github.com/theforeman/foremanctl/blob/e5e86e29b95d3aaba8c4651203401e2d74ebaac3/src/vars/database.yml#L104, not database..Hence the names like iop_advisor.dump..

tests/backup_test.py
sjha4: The server.run runs it inside quadlet for quadlet based deploys like CI..We want the command to run locally hence the subprocess.run ..

src/roles/backup/tasks/main.yaml
```diff
@@ -92,12 +92,39 @@
       ansible.builtin.include_tasks:
         file: database_dumps.yaml
 
-    - name: Backup foremanctl state directory
+    - name: Check if foremanctl state directory exists locally on target
+      ansible.builtin.stat:
+        path: "{{ obsah_state_path }}"
+      register: backup_state_path_stat
+
+    - name: Copy foremanctl state from controller to target (remote deployment)
```
ehelms: I think you could consolidate remote vs. local by just always performing this action.

sjha4:
@ehelms Greenery.. 🥗

src/playbooks/backup/metadata.obsah.yaml
```diff
@@ -9,6 +9,11 @@ variables:
     type: AbsolutePath
     persist: false
 
+  target_host:
+    help: Target host to backup (defaults to quadlet)
+    default: quadlet
```
sjha4: This doesn't actually do anything..I'll remove this..

src/roles/backup/tasks/main.yaml
```diff
@@ -86,18 +86,41 @@
...
-    - name: Backup foremanctl state directory
+    - name: Create temporary directory on controller
+      ansible.builtin.tempfile:
+        state: directory
+        suffix: foremanctl-backup
+      delegate_to: localhost
+      become: false
+      register: backup_temp_dir
+
+    - name: Archive foremanctl state on controller
```
ehelms, line 95: Could you do this within the `obsah_state_path`?

1. generate archive inside obsah_state_path
2. transfer to target location

Then no temporary directory or cleanup is needed?

sjha4: Updated to archive /var/lib/foremanctl/ directory on target into the backup directory on target..No cleanup/temp files..`/var/lib/foremanctl` works on localhost, quadlet and proxy where the obsah_state_path will be mounted.. 👍🏼

sjha4: As we discussed, archiving state on target doesn't work since it doesn't have all the relevant files. Added a test to check for a file I figured will be on all setups proxy, katello flavors.

ehelms (approved):

sjha4:
Rebased + push to retrigger tests..One test failure was unrelated..Want to see this green..

ehelms (approved):
This looks good based on current design, merging so we have some backup tests running regularly and allowing restore work to continue.

PR #591, opened 2026-06-23
generate random rails enc key and attach as secret

bochi:
Currently, the rails encryption key is baked into the foreman container image (it is created by the post install script during RPM installation and never deleted). I checked upstream, the 3.18 and 3.19 even have different keys, so currently, an upgrade would break encrypted values in the DB.

* Create a rails encryption key during deployment and write to obsah state dir
* Attach as secret to the foreman and dynflow containers as env var inside the container. 

Note: This works because the env var takes precedence over the file. Yet maybe the encryption key created during installation should be removed from the image after build.

Steps to reproduce:

* Deploy and encrypt a secret
* Redeploy with a newer image, see that secrets can still be decrypted

#### Checklist

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -29,6 +29,15 @@
     - Restart foreman
     - Restart dynflow-sidekiq@
 
+- name: Create secret for ENCRYPTION_KEY
+  containers.podman.podman_secret:
+    state: present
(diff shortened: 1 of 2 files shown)


evgeni (approved):

evgeni:
https://github.com/theforeman/foreman-oci-images/pull/51 is removing the baked in key, just to be sure

PR #600, opened 2026-06-25
Install the skopeo utility

stejskalleos:
Skopeo is a utility that performs various operations on container images and image repositories
This is a prerequisite for disconnected installations.

See https://github.com/podman-container-tools/skopeo

### DIFF
--- a/src/roles/pre_install/tasks/main.yaml
+++ b/src/roles/pre_install/tasks/main.yaml
@@ -3,10 +3,11 @@
   ansible.builtin.include_role:
     name: debug_tools
 
-- name: Install podman
+- name: Install podman and utilities
   ansible.builtin.package:
     name:
       - podman
+      - skopeo
 
 - name: Install other dependencies
   ansible.builtin.package:


src/roles/pre_install/tasks/main.yaml
```diff
@@ -3,10 +3,11 @@
   ansible.builtin.include_role:
     name: debug_tools
 
-- name: Install podman
+- name: Install packages
```
evgeni: the next task is "other dependencies", which makes it rather non-descriptive.
How about "Install podman and utilities" here?

evgeni (changes requested):
naming nitpick and remember to update packaging please

stejskalleos:
Task renamed and packaging PR is here: https://github.com/theforeman/foreman-packaging/pull/13669

evgeni (approved):

src/roles/pre_install/tasks/main.yaml
Gauravtalreja1: like installing `container-tools` package you mean?

evgeni: Do we need that one too?

PR #602, opened 2026-06-26
add support for providing an own ca

### LINKED ISSUE #297: Certificate workflow support
## Summary

The current certificate implementation lacks support for CNAME (Subject Alternative Names) configuration, direct user input for custom certificates, and providing complete certificate stacks. The first two are gaps from the existing `foreman-installer` supported workflows and the latter is a feature that has long been requested by users.

## Current State

**Certificate Generation:**
- Only supports single hostname via `subjectAltName = DNS:{{ certificates_hostname }}`
- Certificate generation limited to `default` source (self-generated)

**Custom Certificate Support:**
- `installer` certificate source exists but targets legacy paths
- No direct way for users to provide custom certificate files as deploy parameters
- Missing CNAME/SAN support for multi-domain certificates
- No streamlined approach for complete certificate bundles

### 1. CNAME Support

Add direct parameter support for multiple DNS names:

**New Parameter:**
```bash
foremanctl deploy --certificate-cnames="foreman.example.com,satellite.example.com,api.example.com"
```

### 2. Providing custom certificates for server components

This scenario follows the current method that allows users to provide their own certificates for the server certificates.
This is needed for continuity of upgrades.
Add direct certificate file parameters for replacing just the server certificates, and not the client/Candlepin CA:

```bash
foremanctl deploy
  --certificate-server-cert=</path/to/server.crt>
  --certificate-server-key=</path/to/server.key>
  --certificate-server-ca-cert=</path/to/ca.crt>
```

### 3. Providing the entire certificate stack

This scenario aims to provide new functionality heavily requested by users to provide the entire certificate stack.

**Server Certificate Parameters (Co-Required):**
```bash
foremanctl deploy
  --certificate-server-cert=</path/to/server.crt>
  --certificate-server-key=</path/to/server.key>
  --certificate-server-ca-cert=</path/to/ca.crt>
  --certificate-localhost-cert=</path/to/localhost.crt>
  --certificate-localhost-key=</path/to/localhost.key>
  --certificate-client-cert=</path/to/client.crt>
  --certificate-client-key=</path/to/client.key>
  --certificate-client-ca-cert=</path/to/ca.crt>
  --certificate-ca-cert=</path/to/ca.crt>
  --certificate-ca-key=</path/to/ca.key>
```

bochi:
Foremanctl should be able to handle user supplied CAs for signing certificates

partly fixes #297 

* New certificate source `custom_ca`: imports a supplied CA and uses it to sign certificates
* New command line parameters `--certificate-ca-certificate`, `--certificate-ca-key`,                                                                                                                                                                               
    `--certificate-ca-key-password` 
* In case the CA is unencrypted it will be reencrypted so it looks like a generated one to the system, some tools like candlepin seem to want an encrypted CA and doing it differently would introduce many non-trivial changes

Steps to reproduce:

* Create a CA
* Deploy with the command line options above

#### Checklist
* [X] Documentation updated (if applicable)

### DIFF
--- a/docs/user/certificates.md
+++ b/docs/user/certificates.md
@@ -6,7 +6,7 @@ This document describes how certificate generation and management works in forem
 
 ### Certificate Sources
 
-foremanctl supports two certificate sources that determine how certificates are obtained:
+foremanctl supports three certificate sources that determine how certificates are obtained:
 
 **Default Source (`certificate_source: default`)**
 - Automatically generates self-signed certificates during deployment
@@ -18,6 +18,14 @@ foremanctl supports two certificate sources that determine how certificates are 
 - Automatically generates an internal CA for client certificates and localhost
 - Server certificate, key, and CA bundle are copied to `/var/lib/foremanctl/certs/`
 - Certificate source persists across deployments; original files only needed on first deploy or when updating certificates
+
+**Custom CA (`certificate_source: custom_ca`)**
+- Uses a CA provided certificate and key that issues all certs
+- The provided CA key is stored unencrypted, the same way the managed internal CA key is stored
+- The CA must have `basicConstraints CA:TRUE`
+- **On an existing deployment, replacing the CA invalidates all certs**
+  `custom_ca` refuses to overwrite an existing CA unless you pass `--replace-ca`.
+  A fresh install, or re-applying the same CA, does not need `--replace-ca`
 
 ### Usage
 
@@ -52,6 +60,19 @@ foremanctl deploy \
 
 # Switch back to auto-generated certificates
 foremanctl deploy --certificate-source=default
+```
+
+#### Using a Custom CA
+
+```bash
+# First deployment
+foremanctl deploy \
+  --certificate-source=custom_ca \
+  --certificate-ca-certificate /path/to/ca.crt \
+  --certificate-ca-key /path/to/ca.key
+
+# With encrypted CA, add:
+  --certificate-ca-key-password 'secret'
 ```
 
 #### Migrating from foreman-installer
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,6 +82,10 @@ jobs:
             security: none
             database: internal
             box: centos/stream9
+          - certificate_source: custom_ca
+            security: none
+            database: internal
+            box: centos/stream9
     runs-on: ubuntu-24.04
     env:
       FOREMANCTL_BASE_BOX: ${{ matrix.box }}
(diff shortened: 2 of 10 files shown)


stejskalleos:
@coderabbitai review

evgeni:
I'd like to see some tests for this, probably very similar how we test `custom_server` today (also can probably re-use the CA-generation we implemented there).
Very curious to see how well Candlepin will take this.

Also, may I assume, that *switching* to a custom_ca will invalidate all consumer (client) certs? Sounds like lots of fun :)

bochi:
TBH haven't thought about switching, and yes, you are right. 
Candlepin seems to work fine but only after I started re-encrypting an potentially unencrypted CA

I will write some tests and convert this to draft, I agree that it may take some more work :)

docs/user/certificates.md
```diff
@@ -19,6 +19,11 @@ foremanctl supports two certificate sources that determine how certificates are
 - Server certificate, key, and CA bundle are copied to `/var/lib/foremanctl/certs/`
 - Certificate source persists across deployments; original files only needed on first deploy or when updating certificates
 
+**Custom CA (`certificate_source: custom_ca`)**
+- Uses a CA provided certificate and key that issues all certs
+- The provided CA key is re-encrypted with the managed CA password so it behaves like a generated one
```
ehelms, line 24: What does re-encryption provide?

bochi: well so far foremanctl expects a managed ca to be encrypted with a password _everywhere_ - so I figured keeping that would be the best solution. (ie. candlepin: `openssl pkey -passin 
  file:ca.pwd`, but there are others)

ehelms: This is a bit of a hold-over from the `foreman-installer` design that is worth re-thinking. As we store the CA key with the password, and we also hand an unencrypted version of it to Candlepin, the "encryption" of the CA key feels rather worthless and just adds management overhead.

bochi: ok I will look into removing the ca encryption entirely

bochi:
> I'd like to see some tests for this, probably very similar how we test `custom_server` today (also can probably re-use the CA-generation we implemented there). Very curious to see how well Candlepin will take this.
> 
> Also, may I assume, that _switching_ to a custom_ca will invalidate all consumer (client) certs? Sounds like lots of fun :)

Now that removal of CA encryption is merged I went back to this PR and thought about your comment.

Well of course it will invalidate all consumer certs when switching to a custom CA. Yet I would think that somebody using that feature knows that they are doing and is aware of the implications? I mean, we could implement a check for existing consumer certs and print a big fat warning or something, but is it actually worth the effort?

The other option would be not having the feature :-) At least from my POV. If you guys have and other/better ideas I am glad to hear them

evgeni:
Oh I absolutely think that having this feature is great. We just need to politely point out what all will go up in flames if you try this on an existing setup.

bochi:
I added some more tests as requested by @evgeni  plus implemented a gate in case there is already an existing  CA. 

If it differs, it refuses to do anything unless `--replace-ca` is specified and throws an error message stating why this might be a bad idea.

PR #603, opened 2026-06-29
use podman_quadlet to remove redis

evgeni:
podman_quadlet takes care of reloading systemd too

it's better than file

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/roles/valkey/tasks/cleanup_redis.yaml
+++ b/src/roles/valkey/tasks/cleanup_redis.yaml
@@ -13,24 +13,14 @@
     name: redis
     state: absent
 
-- name: Remove Redis image
-  ansible.builtin.file:
-    path: /etc/containers/systemd/redis.image
+- name: Remove Redis image and container definitions
+  containers.podman.podman_quadlet:
     state: absent
+    name:
+      - redis.image
+      - redis.container
 
 - name: Remove Redis image drop-in directory
   ansible.builtin.file:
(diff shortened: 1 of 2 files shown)


evgeni:
```
Failed to list quadlets: Error: querying systemd for unit status: Unit name dynflow-sidekiq@.service is missing the instance name.
```

Well, that's a nice failure, which I will not debug this week.

ShimShtein:
>`Failed to list quadlets: Error: querying systemd for unit status: Unit name dynflow-sidekiq@.service is missing the instance name.`
Well, that's a nice failure, which I will not debug this week.

It is a known bug in podman, https://github.com/podman-container-tools/podman/issues/26960. We will have to wait until it gets into our streams I suppose.

src/roles/valkey/tasks/cleanup_redis.yaml
```diff
@@ -13,24 +13,14 @@
...
     state: absent
 
-- name: Remove Redis image
-  ansible.builtin.file:
-    path: /etc/containers/systemd/redis.image
+- name: Remove Redis image and container definitions
+  containers.podman.podman_quadlet:
     state: absent
+    name:
+      - redis.image
```
ShimShtein, line 20: Opened https://github.com/containers/ansible-podman-collections/issues/1044 to handle all related suffixes instead of enumerating them one by one.

PR #593, opened 2026-06-23
Use a unix socket between httpd and Foreman's Puma

ekohl:
#### Why are you introducing these changes? (Problem description, related links)

https://github.com/theforeman/foremanctl/pull/118 included this previously, but splitting the focus.

In foreman-installer we saw using a unix socket had reliability benefits.

#### What are the changes introduced in this pull request?

* https://github.com/theforeman/foremanctl/pull/118
* Use a unix socket for httpd and Puma instead of a TCP socket

#### How to test this pull request

Steps to reproduce:

* The Foreman UI loads
* Test suite passes

#### Checklist
* [x] Tests added/updated (if applicable)
* [ ] Documentation updated (if applicable)

PR #471, opened 2026-04-23
configure hosts for ActionDispatch::HostAuthorization middleware

evgeni:
#### Why are you introducing these changes? (Problem description, related links)

moar secure

#### What are the changes introduced in this pull request?

configure `hosts` so that Rails can reject HTTP requests with invalid `Host` headrs

#### How to test this pull request

test in the patch

#### Checklist
* [x] Tests added/updated (if applicable)
* [x] Documentation updated (if applicable)

PR #608, opened 2026-06-30
Use unix sockets for httpd -> Foreman and Pulp communication

Gauravtalreja1:
On hosts without IPv4 (IPv6-only), Apache resolves `localhost` to `::1` but Puma/Gunicorn only listen on `127.0.0.1`, causing connection refused errors:
 ```
AH00957: http: attempt to connect to [::1]:3000 (localhost:3000) failed
```
This affects all three httpd backends: Foreman (port 3000), Pulp API (port 24817), and Pulp Content (port 24816).
Builds on the work started by @ekohl in #118 and #593. In foreman-installer, using a unix socket had reliability benefits, this brings the same approach to foremanctl.

**Why unix sockets instead of binding to `127.0.0.1` explicitly:**

  * **Parity with foreman-installer:** puppet-foreman and puppet-pulpcore already use unix sockets
  * **No TCP port conflicts:**  doesn't occupy TCP ports 3000, 24816, 24817
  * **Filesystem-level access control:**  socket permissions via `SocketUser`/`SocketMode`
  * **Eliminates TCP loopback overhead**
  * **Systemd socket activation:** clean dependency ordering via `Requires=`

  Foreman (cherry-picked from #593, authored by @ekohl):
  * Systemd socket activation for Foreman
  * Unix socket between httpd and Puma at `/run/httpd.foreman.sock`

  SELinux fixes (on top of #593):
  * Add missing `when` guard for `daemons_enable_cluster_mode` boolean
  * Remove `httpd_can_network_connect`,  no longer needed with all backends on unix sockets

  Pulp:
  * Unix sockets for Pulp API (`/run/httpd.pulp-api.sock`) and Pulp Content (`/run/httpd.pulp-content.sock`)
  * Systemd socket units for both services
  * Container command overrides to bind Gunicorn to unix sockets

  * Deploy using foremanctl on an IPv6-only or IPv4-only host
  * Verify no TCP listeners on ports 3000, 24816, 24817: `ss -tlnp | grep -E '(3000|24816|24817)'`
  * Verify unix sockets exist: `ls -laZ /run/httpd.*.sock`
  * The Foreman UI loads
  * Pulp API responds: `curl -k https://$(hostname -f)/pulp/api/v3/status/`
  * Test suite passes

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/roles/foreman/templates/foreman.socket.j2
+++ b/src/roles/foreman/templates/foreman.socket.j2
@@ -1,0 +1,8 @@
+[Unit]
+Description=Foreman socket
+
+[Socket]
+ListenStream={{ foreman_listen_stream }}
+
+[Install]
+WantedBy=sockets.target
--- a/src/roles/foreman/templates/foreman.socket.j2
+++ b/src/roles/foreman/templates/foreman.socket.j2
@@ -3,6 +3,12 @@ Description=Foreman socket
 
 [Socket]
 ListenStream={{ foreman_listen_stream }}
+SocketUser=apache
+SocketMode=0600
+
+NoDelay=false
+ReusePort=true
+Backlog=1024
 
 [Install]
 WantedBy=sockets.target
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -92,16 +92,37 @@
     - Restart pulp-content
     - Restart pulp-worker
 
+- name: Deploy Pulp API socket
+  ansible.builtin.template:
+    src: pulp-api.socket.j2
+    dest: /etc/systemd/system/pulp-api.socket
+    mode: '0644'
+
+# TODO: The --bind argument is passed here to override pulpcore-api's default
+# bind address (0.0.0.0:24817) with a unix socket. This duplicates arguments
+# from the container's pulp-api entry point script, which is fragile if the
+# image changes. The proper fix is to add a PULP_API_BIND environment variable
+# to the pulp-oci-images container (like PULP_API_WORKERS already works),
+# then set it via env: instead of overriding the command.
+# See: https://github.com/theforeman/pulp-oci-images
 - name: Deploy Pulp API Container
   containers.podman.podman_container:
     name: "{{ pulp_api_container_name }}"
     image: pulp.image
     state: quadlet
     sdnotify: true
-    command: pulp-api
+    command: >-
+      /usr/bin/pulpcore-api
+      --bind unix:{{ pulp_api_listen_stream }}
+      --timeout 90
+      --workers {{ pulp_api_service_worker_count }}
+      --max-requests 800
+      --max-requests-jitter 100
+      --access-logfile -
+      --access-logformat 'pulp [%({correlation-id}o)s]: %(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"'
     network: host
     hostname: "pulp-api.{{ ansible_facts['hostname'] }}.local"
-    volumes: "{{ pulp_volumes }}"
+    volumes: "{{ pulp_volumes + ['/run:/run:rw'] }}"
     security_opt:
       - "label=disable"
     secrets:
@@ -112,6 +133,8 @@
     env: "{{ pulp_settings_env }}"
     quadlet_options:
       - |
+        [Unit]
+        Requires=pulp-api.socket
         [Install]
         WantedBy=default.target foreman.target
         [Unit]
@@ -123,16 +146,29 @@
         RestartSec=3
   notify: Restart pulp-api
 
+- name: Deploy Pulp Content socket
+  ansible.builtin.template:
+    src: pulp-content.socket.j2
+    dest: /etc/systemd/system/pulp-content.socket
+    mode: '0644'
+
+# TODO: Same as pulp-api above — add PULP_CONTENT_BIND to pulp-oci-images
+# and use env: instead of overriding the command.
 - name: Deploy Pulp Content Container
   containers.podman.podman_container:
     name: "{{ pulp_content_container_name }}"
     image: pulp.image
     state: quadlet
     sdnotify: true
-    command: pulp-content
+    command: >-
+      /usr/bin/pulpcore-content
+      --bind unix:{{ pulp_content_listen_stream }}
+      --timeout 90
+      --workers {{ pulp_content_service_worker_count }}
+      --access-logfile -
     network: host
     hostname: "pulp-content.{{ ansible_facts['hostname'] }}.local"
-    volumes: "{{ pulp_volumes }}"
+    volumes: "{{ pulp_volumes + ['/run:/run:rw'] }}"
     security_opt:
       - "label=disable"
     secrets:
@@ -143,6 +179,8 @@
     env: "{{ pulp_settings_env }}"
     quadlet_options:
       - |
+        [Unit]
+        Requires=pulp-content.socket
         [Install]
         WantedBy=default.target foreman.target
         [Unit]
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -97,6 +97,7 @@
     src: pulp-api.socket.j2
     dest: /etc/systemd/system/pulp-api.socket
     mode: '0644'
+  notify: Restart pulp-api socket
 
 - name: Deploy Pulp API Container
   containers.podman.podman_container:
@@ -136,6 +137,7 @@
     src: pulp-content.socket.j2
     dest: /etc/systemd/system/pulp-content.socket
     mode: '0644'
+  notify: Restart pulp-content socket
 
 - name: Deploy Pulp Content Container
   containers.podman.podman_container:
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -98,31 +98,16 @@
     dest: /etc/systemd/system/pulp-api.socket
     mode: '0644'
 
-# TODO: The --bind argument is passed here to override pulpcore-api's default
-# bind address (0.0.0.0:24817) with a unix socket. This duplicates arguments
-# from the container's pulp-api entry point script, which is fragile if the
-# image changes. The proper fix is to add a PULP_API_BIND environment variable
-# to the pulp-oci-images container (like PULP_API_WORKERS already works),
-# then set it via env: instead of overriding the command.
-# See: https://github.com/theforeman/pulp-oci-images
 - name: Deploy Pulp API Container
   containers.podman.podman_container:
     name: "{{ pulp_api_container_name }}"
     image: pulp.image
     state: quadlet
     sdnotify: true
-    command: >-
-      /usr/bin/pulpcore-api
-      --bind unix:{{ pulp_api_listen_stream }}
-      --timeout 90
-      --workers {{ pulp_api_service_worker_count }}
-      --max-requests 800
-      --max-requests-jitter 100
-      --access-logfile -
-      --access-logformat 'pulp [%({correlation-id}o)s]: %(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"'
+    command: pulp-api
     network: host
     hostname: "pulp-api.{{ ansible_facts['hostname'] }}.local"
-    volumes: "{{ pulp_volumes + ['/run:/run:rw'] }}"
+    volumes: "{{ pulp_volumes }}"
     security_opt:
       - "label=disable"
     secrets:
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: src/roles/pulp/tasks/main.yaml [hunk 1]
```diff
@@ -90,16 +90,37 @@
...
+      --timeout 90
+      --workers {{ pulp_api_service_worker_count }}
+      --max-requests 800
+      --max-requests-jitter 100
+      --access-logfile -
+      --access-logformat 'pulp [%({correlation-id}o)s]: %(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"'
     network: host
     hostname: "pulp-api.{{ ansible_facts['hostname'] }}.local"
-    volumes: "{{ pulp_volumes }}"
+    volumes: "{{ pulp_volumes + ['/run:/run:rw'] }}"
```
ekohl: I don't think you should mount `/run`. systemd socket activation works by passing a file descriptor to the process (in this case a container) so it shouldn't need the actual file within the container.

src/roles/pulp/tasks/main.yaml
```diff
@@ -90,16 +90,37 @@
...
 - name: Deploy Pulp API Container
   containers.podman.podman_container:
     name: "{{ pulp_api_container_name }}"
     image: pulp.image
     state: quadlet
     sdnotify: true
-    command: pulp-api
+    command: >-
+      /usr/bin/pulpcore-api
+      --bind unix:{{ pulp_api_listen_stream }}
```
ekohl: This shouldn't be needed: systemd socket activation should pass a file descriptor. Note we don't pass it in for the existing code either: https://github.com/theforeman/puppet-pulpcore/blob/4e9d4db4bb753817ca23b049aae3272a913aa9d9/templates/pulpcore-api.service.erb#L15-L23.

See https://gunicorn.org/deploy/#systemd for more.

src/roles/pulp/tasks/main.yaml
```diff
@@ -121,16 +144,29 @@
...
 - name: Deploy Pulp Content Container
   containers.podman.podman_container:
     name: "{{ pulp_content_container_name }}"
     image: pulp.image
     state: quadlet
     sdnotify: true
-    command: pulp-content
+    command: >-
+      /usr/bin/pulpcore-content
+      --bind unix:{{ pulp_content_listen_stream }}
```
ekohl: Similar notes as above: https://github.com/theforeman/puppet-pulpcore/blob/4e9d4db4bb753817ca23b049aae3272a913aa9d9/templates/pulpcore-content.service.erb#L15-L20 doesn't pass it in either.

ekohl (commented):
https://github.com/theforeman/puppet-pulpcore/commit/04b9f5f4fbcaafd87c6ea056b9b499197739c7c4 may be an interesting reference.

src/roles/foreman/templates/foreman.socket.j2
```diff
@@ -0,0 +1,14 @@
...
+Description=Foreman socket
+
+[Socket]
+ListenStream={{ foreman_listen_stream }}
+SocketUser=apache
+SocketMode=0600
+
+NoDelay=false
+ReusePort=true
+Backlog=1024
```
pablomh: Why 1024 instead of 4294967295 (upstream value)?

https://github.com/theforeman/foreman/blob/e0522f25b34b7217e7c7484f4373e758eae2caa6/extras/systemd/foreman.socket#L11

evgeni:
I've rebased this now that #634 is merged, to be able to debug what's happening in the external proxy test case.

ekohl:
Looks like Apache sends the literal `%{http_host}` header, not interpolated. Now I wonder if that's also a problem in our regular install.
```
ul 08 07:41:50 proxy.example.com pulp-api[45255]: pulp [9d33dfa3a8194a68ad9b224b1fc1de0c]: django.security.DisallowedHost:ERROR: Invalid HTTP_HOST header: '%{http_host}'. The domain name provided is not valid according to RFC 1034/1035.
Jul 08 07:41:50 proxy.example.com pulp-api[45255]: Traceback (most recent call last):
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:   File "/usr/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:     response = get_response(request)
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:                ^^^^^^^^^^^^^^^^^^^^^
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:   File "/usr/lib/python3.12/site-packages/django/utils/deprecation.py", line 133, in __call__
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:     response = self.process_request(request)
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:   File "/usr/lib/python3.12/site-packages/django/middleware/common.py", line 48, in process_request
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:     host = request.get_host()
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:            ^^^^^^^^^^^^^^^^^^
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:   File "/usr/lib/python3.12/site-packages/django/http/request.py", line 151, in get_host
Jul 08 07:41:50 proxy.example.com pulp-api[45255]:     raise DisallowedHost(msg)
Jul 08 07:41:50 proxy.example.com pulp-api[45255]: django.core.exceptions.DisallowedHost: Invalid HTTP_HOST header: '%{http_host}'. The domain name provided is not valid according to RFC 1034/1035.
```

src/vars/base.yaml
```diff
@@ -23,6 +23,14 @@ foreman_plugins: "{{ enabled_features | features_to_foreman_plugins }}"
 foreman_name: "{{ ansible_facts['fqdn'] }}"
 foreman_url: "https://{{ foreman_name }}"
 
+foreman_listen_stream: /run/httpd.foreman.sock
+httpd_foreman_backend: "unix://{{ foreman_listen_stream }}|http://%{HTTP_HOST}"
+
+pulp_api_listen_stream: /run/httpd.pulp-api.sock
+pulp_content_listen_stream: /run/httpd.pulp-content.sock
+httpd_pulp_api_backend: "unix://{{ pulp_api_listen_stream }}|http://%{HTTP_HOST}"
+httpd_pulp_content_backend: "unix://{{ pulp_content_listen_stream }}|http://%{HTTP_HOST}"
```
evgeni: In puppet-pulpcore, we use a static name, not `HTTP_HOST`:
https://github.com/theforeman/puppet-pulpcore/blob/4e9d4db4bb753817ca23b049aae3272a913aa9d9/manifests/apache.pp#L16-L20

src/vars/base.yaml
```diff
@@ -23,6 +23,14 @@ foreman_plugins: "{{ enabled_features | features_to_foreman_plugins }}"
 foreman_name: "{{ ansible_facts['fqdn'] }}"
 foreman_url: "https://{{ foreman_name }}"
 
+foreman_listen_stream: /run/httpd.foreman.sock
+httpd_foreman_backend: "unix://{{ foreman_listen_stream }}|http://%{HTTP_HOST}"
```
evgeni: In puppet-foreman, we use a static name, not `HTTP_HOST`:
https://github.com/theforeman/puppet-foreman/blob/da521ac9ec4cd6acd55247a6325918ef244948e4/manifests/config/apache.pp#L144-L145

ekohl: When I experimented with it, my intent was to send what the client actually sent. But perhaps those are already set with `X-Forwarded-Host` and we trust those. https://github.com/theforeman/foremanctl/pull/471 should help to verify that it actually works as intended.

evgeni: Yeah, I think the x-forwarded-for is what we should aim at on app level

ekohl:
@Gauravtalreja1 https://github.com/theforeman/foremanctl/pull/608 was merged. Can you please rebase it to resolve the merge conflicts?

evgeni:
#471 was merged, so if we rebase here and fix the hostnames, we should see if things work properly

tests/foreman_test.py
```diff
@@ -16,8 +16,8 @@
 
 
 @pytest.fixture(scope="module")
-def foreman_status_curl(server, server_fqdn):
```
ekohl, line 19: This is now reverting the changes. Please keep `server_fqdn` in.

tests/foreman_test.py
```diff
@@ -5,7 +5,7 @@
 pytestmark = pytest.mark.feature('foreman')
 
 FOREMAN_HOST = 'localhost'
```
ekohl: This should be unused now.

tests/foreman_test.py
Gauravtalreja1: Oops, I messed up, I'll update this

tests/pulp_test.py
```diff
@@ -25,7 +25,7 @@ def load_pulp_paths_from_parameters():
 
 @pytest.fixture(scope="module")
 def pulp_status_curl(server):
-    return server.run(f"curl -k -s -w '%{{stderr}}%{{http_code}}' http://{PULP_HOST}:{PULP_API_PORT}/pulp/api/v3/status/")
+    return server.run(f"curl -k -s -w '%{{stderr}}%{{http_code}}' --unix-socket {PULP_API_SOCKET} http://{PULP_HOST}/pulp/api/v3/status/")
```
evgeni: We should use `server_fqdn` fixture here, then we can drop `PULP_HOST` constant too.

src/roles/pulp/templates/pulp-api.socket.j2
```diff
@@ -0,0 +1,16 @@
...
+ListenStream={{ pulp_api_listen_stream }}
+SocketUser=apache
+SocketMode=0600
+
+NoDelay=false
+ReusePort=true
+# Default on EL10 (v254+), needed explicitly for EL9 (v252)
+Backlog=4294967295
+
+
```
evgeni:
```suggestion
```

one empty line is enough ;)

ehelms:
> On hosts without IPv4 (IPv6-only), Apache resolves `localhost` to `::1` but Puma/Gunicorn only listen on `127.0.0.1`, causing connection refused errors:
> 
> ```
> AH00957: http: attempt to connect to [::1]:3000 (localhost:3000) failed
> ```

As I understand it, there are other ways to solve this issue. Can you expand on why we want to go with the unix socket solution? Are there other things that this choice solves or provides? I am not opposed to the change, but I would find it helpful to capture more about this choice in the pull request description and in our documentation given this is an explicit design choice we want to have / enforce.

Gauravtalreja1:
> > On hosts without IPv4 (IPv6-only), Apache resolves `localhost` to `::1` but Puma/Gunicorn only listen on `127.0.0.1`, causing connection refused errors:
> > ```
> > AH00957: http: attempt to connect to [::1]:3000 (localhost:3000) failed
> > ```
> 
> As I understand it, there are other ways to solve this issue. Can you expand on why we want to go with the unix socket solution? Are there other things that this choice solves or provides? I am not opposed to the change, but I would find it helpful to capture more about this choice in the pull request description and in our documentation given this is an explicit design choice we want to have / enforce.

@ehelms Good point, I've updated the PR description with the design rationale (with what I understand) for choosing unix sockets over binding to 127.0.0.1/localhost. 
For broader context on the original design decision, @ekohl drove this approach in #118 & #593 and can provide more insight into the long-term vision behind it.

src/vars/base.yaml
Gauravtalreja1: It worked, thanks!

evgeni:
`test_https_rhsm_proxy` is failing, which is the only test that uses the "httpd-on-proxy is a reverse proxy to httpd-on-server which in turn is a reverse proxy to rails" chain.

I bet this is some cross play with #471 - but that test was green there, so it's something about the sockets :see_no_evil: 

logs on the proxy say:
```
[Thu Jul 09 15:14:00.089578 2026] [proxy:error] [pid 44340:tid 44420] (13)Permission denied: AH00957: https: attempt to connect to 192.168.124.97:443 (quadlet.example.com:443) failed
[Thu Jul 09 15:14:00.089633 2026] [proxy_http:error] [pid 44340:tid 44420] [client 192.168.124.97:59898] AH01114: HTTP: failed to make connection to backend: quadlet.example.com

192.168.124.97 - - [09/Jul/2026:15:14:00 +0000] "GET /rhsm HTTP/1.1" 503 299 "-" "curl/7.76.1"
```

I see no requests reaching the quadlet box.

I think you need to put back that httpd network selinux boolean

Gauravtalreja1:
@coderabbitai full review

src/roles/pulp/tasks/main.yaml
```diff
@@ -92,6 +92,12 @@
     - Restart pulp-content
     - Restart pulp-worker
 
+- name: Deploy Pulp API socket
+  ansible.builtin.template:
+    src: pulp-api.socket.j2
+    dest: /etc/systemd/system/pulp-api.socket
+    mode: '0644'
+
```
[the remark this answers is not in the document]
Gauravtalreja1: That makes sense to me. I'll update it in a new commit

Satellite-QE:
**PRT Result**
```
Build Number: 16084
Build Status: UNSTABLE
PRT Comment: pytest tests/foreman/installer/test_install_foremanctl.py --external-logging
Test Result: ================= 34 warnings, 20 errors in 4107.43s (1:08:27) =================
```

evgeni (approved):

Satellite-QE:
**PRT Result**
```
Build Number: 16087
Build Status: UNSTABLE
PRT Comment: pytest tests/foreman/installer/test_install_foremanctl.py --external-logging
Test Result: ================= 34 warnings, 20 errors in 3812.58s (1:03:32) =================
```

evgeni:
PRT fails with `Unable to start service postgresql: A dependency job for postgresql.service failed.` which makes no sense :/

Edit: oooh, it used a v6 machine, but that probably can't reach quay.io…

Satellite-QE:
**PRT Result**
```
Build Number: 16089
Build Status: UNSTABLE
PRT Comment: pytest tests/foreman/installer/test_install_foremanctl.py --external-logging
Test Result: ================= 34 warnings, 20 errors in 3905.19s (1:05:05) =================
```

ehelms (approved):

ekohl (approved):
Thanks for taking this across the finish line.

stejskalleos:
Thanks everyone!

PR #609, opened 2026-06-30
Add incremental backup and restore

Chyenne8:
# Add incremental backup and restore

This PR implements incremental backup and restore functionality using GNU tar's `--listed-incremental` mechanism with `snar` snapshot files.

## Overview

Incremental backups contain only files that have changed since a previous backup, reducing backup size and time for frequent backup schedules. The implementation tracks file state using `snar` (snapshot) files that GNU tar uses to determine what has changed between backups.

### Full Backup
Creates a complete backup with initial `snar` snapshot files:

```bash
foremanctl backup /var/backup
```

This creates a timestamped directory like `/var/backup/foreman-backup-20260629T120000/` containing:
- Database dumps (`.dump` files)
- Foremanctl state archive (`foremanctl-state.tar.gz`)
- Pulp content archive (`pulp-content.tar.gz`)
- Snapshot files (`config.snar`, `pulp.snar`)
- Metadata (`metadata.yml`)

### Incremental Backup
Creates a differential backup containing only changed files:

```bash
foremanctl backup /var/backup --base-backup /var/backup/foreman-backup-20260629T120000
```

The `--base-backup` parameter specifies the previous backup to use as a baseline. The command validates that the base backup contains required `snar` files before proceeding.

### Full Backup Restore
```bash
foremanctl restore /var/backup/foreman-backup-20260629T120000
```

### Incremental Backup Restore
Incremental backups must be restored sequentially, starting with the full backup that began the chain:

```bash
# Restore full backup first
foremanctl restore /var/backup/foreman-backup-20260629T120000

# Restore first incremental
foremanctl restore /var/backup/foreman-backup-20260630T080000 --force

# Restore second incremental
foremanctl restore /var/backup/foreman-backup-20260701T080000 --force
```

The `--force` flag is required for subsequent restores since the system is already deployed after the first restore.

## Backup Chains

Incremental backups can be chained together. For example:
- Sunday: Full backup
- Monday: Incremental based on Sunday
- Tuesday: Incremental based on Monday
- Wednesday: Incremental based on Tuesday

To restore Wednesday's state, all four backups must be restored in order: Sunday → Monday → Tuesday → Wednesday.

### Backup Validation
When creating an incremental backup, the system validates:
- Base backup directory exists
- Required files present: `metadata.yml`, `config.snar`, `pulp.snar`
- Files are readable

Validation occurs before creating any backup directories or stopping services. Invalid base backups fail immediately with a clear error message.

### Restore Validation
When restoring an incremental backup, the system validates:
- Incremental backup metadata references a base backup
- Base backup directory exists
- Base backup metadata matches expected timestamp

Missing or mismatched base backups produce clear error messages explaining which backup is required.

## Metadata Structure

The `metadata.yml` file includes incremental backup information:

```yaml
incremental:
  is_incremental: true              # false for full backups
  base_backup_dir: /path/to/base    # empty string for full backups
  base_backup_timestamp: 20260629T120000  # empty string for full backups
```

This allows the restore process to detect incremental backups and validate the chain.

### Backup Process
1. Validate `--base-backup` parameter if provided (check for `snar` files)
2. Load base backup metadata to capture timestamp
3. Copy `snar` files from base backup
4. Execute backup with `tar --listed-incremental=<snar-file>`
5. Generate metadata with incremental information

### Restore Process
1. Read backup metadata
2. Detect if backup is incremental
3. If incremental, validate base backup exists and timestamp matches
4. Extract archives using `tar --listed-incremental=/dev/null` for incrementals
5. Record restore timestamp for chain validation

### File Handling
- `config.snar`: Tracks foremanctl state directory changes
- `pulp.snar`: Tracks Pulp content directory changes
- Both files are created during full backups and copied/updated during incremental backups

## Requirements

- GNU tar with `--listed-incremental` support (standard on RHEL/CentOS)
- Base backup must be created by foremanctl (contains `snar` files)
- All backups in a chain must be available for restore

## Testing

Tested scenarios:
- Full backup creation
- Incremental backup creation with valid base backup
- Incremental backup rejection when base backup missing `snar` files
- Metadata structure correctness
- Restore validation for incremental backups
- Variable scoping in backup and restore roles
- Shell quoting in tar and PostgreSQL commands

Test suite:
- `tests/backup_test.py` validates metadata structure
- Manual testing of full incremental backup and restore flow
- Robottelo integration tests for backup/restore workflows

## Typical Usage Pattern

Weekly full backup with daily incrementals:

```bash
# Sunday - Create full backup
foremanctl backup /var/backup

# Monday - Incremental from Sunday
foremanctl backup /var/backup --base-backup /var/backup/foreman-backup-20260629T120000

# Tuesday - Incremental from Monday  
foremanctl backup /var/backup --base-backup /var/backup/foreman-backup-20260630T080000

# Continue pattern through Saturday
```

Each incremental references the previous day's backup. To restore Saturday's state, restore all seven backups in order.

## Documentation

User documentation in `docs/user/backup-restore.md` covers:
- Incremental backup creation
- Restore procedures for incremental backups
- Backup chain concepts
- Validation behavior
- Common backup strategies


sjha4:
I like the idea here: `Restore with auto-chain (recommended):` We however do not have this in the foreman-maintain implementation..I can test it out and we can take a call with the team if it makes sense to have this.

sjha4:
One note about: `./foremanctl backup /backups --incremental /backups/foreman-backup-*`

There is some difference between command structure with foreman-maintain for incremental backups here: https://docs.theforeman.org/nightly/Administering_Project/index-katello.html#Performing_an_Incremental_Backup_admin

We'd like to stay close to the foreman-maintain command.

sjha4:
@coderabbitai review

sjha4:
My suggestion here is to remove --chain functionality to stick to feature parity and keep PR simple..

Chyenne8:
Removed the --restore-chain functionality. Restoring incrementals now follows the standard procedure: restore full backup first, then each incremental in order with
  --force.

Chyenne8:
Updated backup tests to use database_mapping from metadata to resolve dump filenames. Initially we changed expected_databases to use actual database names (advisor_db, etc.) to fix the dump file tests, but that broke test_metadata_structure and test_metadata_includes_iop_databases since metadata.databases stores logical names
 (iop_advisor, etc.). Instead, expected_databases now returns logical names to stay consistent with metadata, and the dump tests look up the actual database name via backup_metadata['database_mapping'] — mirroring how the backup and restore roles resolve names.

src/playbooks/backup/metadata.obsah.yaml
```diff
@@ -15,11 +15,20 @@ variables:
     persist: false
 
   skip_pulp_content:
-    help: Skip Pulp content directory backup
+    help: Skip content directory backup
```
sjha4: Specifying Pulp is fine here..Content has so many meanings in katello, it can be confusing.

src/roles/backup/tasks/main.yaml
```diff
@@ -92,25 +92,181 @@
...
+                msg: |
+                  Incremental backup requested, but previous backup is missing .snar files.
+                  Previous backup: {{ incremental }}
+                  Missing files: {{ backup_snar_files_check.results | selectattr('stat.exists', 'equalto', false) | map(attribute='item') | list }}
+
+                  This usually means the previous backup was not created with incremental support.
+                  The base backup must contain .snar files generated by foremanctl.
+
+                  Available backups with incremental support:
+                  {% for suggestion in backup_suggestions_formatted | default([]) | sort(attribute='timestamp', reverse=true) %}
```
sjha4: Not sure we need the whole `alternatives` workflow..We can fail backup if the user-provided backup dir is incorrect..Ensuring correctness of the suggested alternatives will be an issue plus this is not something we have in foreman-maintain so it is not something users expect AFAIK..

src/playbooks/backup/metadata.obsah.yaml
Chyenne8: Specified back to "pulp content" instead of "content"

src/roles/backup/tasks/main.yaml
Chyenne8: Replaced the entire alternatives workflow with a simple fail task.

src/roles/restore/tasks/restore_foremanctl_state.yaml
```diff
@@ -0,0 +1,48 @@
...
+    src: "{{ backup_dir }}/foremanctl-state.tar.gz"
+    dest: "{{ obsah_state_path | dirname }}"
+    remote_src: true
+  when: not restore_config_snar_check.stat.exists
+
+- name: Build list of expected state files
+  ansible.builtin.set_fact:
+    restore_expected_state_files: >-
+      {{
+        ['parameters.yaml', 'foreman-oauth-consumer-key', 'foreman-oauth-consumer-secret', 'postgresql-admin-password'] +
```
vsedmik: The "Verify critical files were restored" task checks for `foreman-oauth-consumer-key` and `foreman-oauth-consumer-secret` directly under `{{ obsah_state_path }}`, but those files actually live in the `oauth/` subdirectory (`{{ obsah_state_path}}/oauth/`), as defined in `src/vars/base.yaml:6`. The check passes for the other files because they do sit at the root of the state directory.

```suggestion
        ['parameters.yaml', 'oauth/foreman-oauth-consumer-key', 'oauth/foreman-oauth-consumer-secret', 'postgresql-admin-password'] +
```

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -0,0 +1,107 @@
...
+- name: Build database restore configuration
+  ansible.builtin.set_fact:
+    restore_databases_config: "{{ restore_databases_config | default([]) + [db_entry] }}"
+  vars:
+    db_entry:
+      name: "{{ item.name }}"
+      dump_file: "{{ restore_backup_metadata.database_mapping[item.name] }}.dump"
+      database: "{{ item.database }}"
+      user: "{{ item.user }}"
+  loop: "{{ databases }}"
```
vsedmik: The "Build database restore configuration" task loops over `{{ databases }}`, which is the full list of all databases (including IOP). For each entry, it eagerly evaluates `restore_backup_metadata.database_mapping[item.name]` in the `vars:` block. When IOP is not enabled, those keys don't exist in the backup metadata, and the task fails with `'dict object' has no attribute 'iop_advisor'`.

The subsequent "Filter databases that exist in backup" task would have excluded the IOP databases, but the error fires before the filter runs - Ansible evaluates `vars:` eagerly for every loop iteration.

Fix: filter the loop itself so it only iterates over databases present in the backup:

```suggestion
  loop: "{{ databases | selectattr('name', 'in', restore_backup_metadata.databases) | list }}"
```

This makes the separate "Filter databases that exist in backup" task redundant, so it could be removed or kept as a no-op safety net.

vsedmik (changes requested):
Tested downstream with an IOP-disabled SAT I bumped into two failures.

Proposing fixes that helped me get over them

src/roles/restore/tasks/restore_databases.yaml
Chyenne8: Fixed. The loop now only iterates over databases present in the backup metadata, so the database_mapping[item.name] lookup is always safe.

src/roles/restore/tasks/restore_foremanctl_state.yaml
Chyenne8: fixed and changed the path.  ({{ obsah_state_path}}/oauth/)

sjha4:
@Chyenne8 Can you force push here to rerun tests?

src/roles/backup/tasks/main.yaml
```diff
@@ -7,6 +7,30 @@
...
+      ansible.builtin.fail:
+        msg: |
+          Previous backup directory does not exist or is not a valid foremanctl backup: {{ incremental }}
+          Missing files: {{ backup_previous_files_check.results | selectattr('stat.exists', 'equalto', false) | map(attribute='item') | list }}
+
+          The previous backup must contain metadata.yml and .snar files generated by foremanctl.
+          Create a new full backup first:
+            foremanctl backup {{ backup_dir }}
+      when: backup_previous_files_check.results | selectattr('stat.exists', 'equalto', false) | list | length > 0
+
```
vsedmik, line 33: During testing of incremental backup/restore, restoring an incremental backup fails with:

```
  E         TASK [restore : Validate base backup timestamp matches] ************************                                                                                                                                                 
  E         fatal: [localhost]: FAILED! =>                                                                                                                                                                                                   
  E             changed: false                                                                                                                                                                                                               
  E             msg: |-                                                                                                                                                                                                                      
  E                 BACKUP CHAIN MISMATCH:                                                                                                                                                                                                   
  E                                                                                                                                                                                                                                          
  E                 The incremental backup references a different base backup than provided.                                                                                                                                                 
  E                                                                                                                                                                                                                                          
  E                 Incremental backup: /tmp/backup-orHsvwfvkD/foreman-backup-20260819T035326                                                                                                                                                
  E                 Expected base timestamp:                                                                                                                                                                                                 
  E                 Actual base timestamp: 20260819T035145                                                                                                                                                                                   
  E                                                                                                                                                                                                                                          
  E                 This incremental backup is not based on the backup at /tmp/backup-orHsvwfvkD/foreman-backup-20260819T035145.                                                                                                             
  E                 Verify you are using the correct base backup directory.
```

**Cause:** The incremental backup's `metadata.yml` has an empty `base_backup_timestamp`:

```yaml
incremental:
    base_backup_dir: /tmp/backup-xyz/foreman-backup-20260819T035145
    base_backup_timestamp: ''   # <-- EMPTY!
    is_incremental: true
```

This happens because `src/roles/backup/tasks/metadata.yaml` references `backup_previous_metadata`:

```yaml
base_backup_timestamp: "{{ (backup_previous_metadata | default({})).timestamp | default('') }}"
```

But there's no task that actually reads the previous backup's `metadata.yml` to populate this variable. The validation tasks only use `ansible.builtin.stat` to check file existence, but never load the file contents.

**Fix suggestion:** Add tasks to read and parse the previous backup's metadata after validation in `src/roles/backup/tasks/main.yaml`:

```suggestion
      when: backup_previous_files_check.results | selectattr('stat.exists', 'equalto', false) | list | length > 0

    - name: Read previous backup metadata
      ansible.builtin.slurp:
        src: "{{ incremental }}/metadata.yml"
      register: previous_metadata_content

    - name: Parse previous backup metadata
      ansible.builtin.set_fact:
        backup_previous_metadata: "{{ previous_metadata_content.content | b64decode | from_yaml }}"
```

With this fix, the incremental backup's metadata correctly stores:
```yaml
base_backup_timestamp: 20260819T054714
```

And the restore pass.

vsedmik (changes requested):
Yet one thing needs a change

src/roles/backup/tasks/main.yaml
Chyenne8: You're correct! When I simplified the validation, I removed the metadata read/parse tasks that metadata.yaml depends on and didn't add it back after moving the validation.

vsedmik (approved):
Giving ACK since all tests from [#22383](https://github.com/SatelliteQE/robottelo/pull/22383) passed locally against the packit box with the latest changes.
```
=== 8 passed, 61 warnings in 1871.14s (0:31:11) ===
```

docs/user/backup-restore.md
```diff
@@ -269,6 +302,35 @@ foremanctl restore /var/backup/foreman-backup-20260617T104115 --validate
 
 This validates the backup and checks system requirements before proceeding.
 
+### Restore Incremental Backups
+
+To restore from an incremental backup, restore each backup in the chain sequentially, starting with the full backup:
+
+```bash
+# Step 1: Restore the full (base) backup
```
ehelms, line 311: The most recent full backup or do you have to start with the first ever full backup?

docs/user/backup-restore.md
```diff
@@ -69,6 +70,38 @@ Allow in-progress tasks to complete before starting backup:
...
+```
+
+**How it works:**
+- The incremental backup contains only files changed since the previous backup
+- `.snar` snapshot files track which files were in the previous backup
+- Both `.config.snar` (foremanctl state) and `.pulp.snar` (Pulp content) are copied from the previous backup
+- Incremental backups are typically much smaller and faster than full backups
+
+**Typical backup strategy:**
+- Weekly full backup (Sunday)
```
ehelms, line 97: Does performing a full backup wipe out the previous incremental backups?

src/playbooks/backup/metadata.obsah.yaml
```diff
@@ -23,3 +23,12 @@ variables:
     help: Wait for running tasks to complete instead of failing immediately
     action: store_true
     persist: false
+
+  incremental:
+    parameter: --incremental
+    help: |
+      Path to previous backup directory for incremental backup.
```
ehelms: Can the backup directory be an incremental backup directory? or a does it need to be a full backup directory?

Do we expect users to manage multiple backups of the same installation? I am wondering if we can make this easier for users where it defaults to the latest but provides a way to override which backup to take as the base. 

`--incremental` sounds more like a *switch* than an option that expects me to specify a path

src/roles/backup/tasks/database_dumps.yaml
```diff
@@ -7,8 +7,8 @@
       --port={{ item.port }}
       --username={{ item.user }}
       --format=custom
-      --file={{ backup_dir_full }}/{{ item.database }}.dump
-      {{ item.database }}
+      --file='{{ backup_dir_full }}/{{ item.database }}.dump'
+      '{{ item.database }}'
```
ehelms: Why do these need to be in quotes? Was there a bug if they weren't?

ehelms: And if quotes are required, can we use `""` (double quotes) which is the Ansible convention I see the most?

src/roles/backup/tasks/main.yaml
```diff
@@ -7,6 +7,35 @@
...
+  when: incremental is defined and incremental | length > 0
+  block:
+    - name: Check previous backup files exist
+      ansible.builtin.stat:
+        path: "{{ incremental }}/{{ item }}"
+      register: backup_previous_files_check
+      loop:
+        - "metadata.yml"
+        - ".config.snar"
+        - ".pulp.snar"
```
ehelms: Will the `snar` files exist if an incremental has never been taken before?

src/roles/backup/tasks/metadata.yaml
```diff
@@ -40,7 +40,10 @@
       os_version: "{{ ansible_facts['distribution'] }} {{ ansible_facts['distribution_version'] }}"
       foremanctl_version: "{{ ansible_facts.packages['foremanctl'][0].version | default('unknown') if 'foremanctl' in ansible_facts.packages else 'unknown' }}"
       type: offline
-      incremental: false
+      incremental:
+        is_incremental: "{{ (incremental is defined and incremental | length > 0) | bool }}"
```
ehelms: This should be `backup_incremental` for role scoping everywhere it's used. Which, to my previous comment, is where it can be harder to read as this represents a path but it's not obvious from the variable name.

src/roles/backup/tasks/pulp_content.yaml
```diff
@@ -5,8 +5,9 @@
     - name: Backup pulp content directory with encryption keys  # noqa: command-instead-of-module
       ansible.builtin.command:
         cmd: >
-          tar -czf {{ backup_dir_full }}/pulp-content.tar.gz
-          --directory={{ backup_pulp_storage_path }}
+          tar -czf '{{ backup_dir_full }}/pulp-content.tar.gz'
```
ehelms: See previous comment about the type of quotes used

src/roles/restore/tasks/main.yaml
```diff
@@ -6,26 +6,35 @@
...
+      ansible.builtin.copy:
+        content: "{{ restore_backup_metadata.timestamp }}"
+        dest: /var/lib/foremanctl/.last_restore_timestamp
+        mode: '0644'
+
+    - name: Display restore completion
+      ansible.builtin.debug:
+        msg: |
+          Restore completed successfully!
+          Backup: {{ backup_dir }}
```
ehelms: `backup_dir` is not scoped to `restore` role.

src/roles/restore/tasks/validate.yaml
```diff
@@ -53,6 +53,128 @@
   ansible.builtin.set_fact:
     restore_backup_metadata: "{{ restore_metadata_content['content'] | b64decode | from_yaml }}"
 
+- name: Detect if backup is incremental
+  ansible.builtin.set_fact:
+    restore_is_incremental: "{{ restore_backup_metadata.incremental.is_incremental | default(restore_backup_metadata.incremental | default(false)) | bool }}"
```
ehelms: Are the defaults needed? Will the metadata not include a clear true / false for this?

docs/user/backup-restore.md
```diff
@@ -269,6 +302,35 @@ foremanctl restore /var/backup/foreman-backup-20260617T104115 --validate
 
 This validates the backup and checks system requirements before proceeding.
 
+### Restore Incremental Backups
+
+To restore from an incremental backup, restore each backup in the chain sequentially, starting with the full backup:
```
vijaysawant:
```suggestion
To restore from an incremental backup, restore each backup in the chain sequentially, starting with the full (base) backup:
```

vijaysawant (changes requested):

docs/user/backup-restore.md
Chyenne8: This would be the first full backup in the incremental chain (the most recent full backup) and the chain will continue with each incremental backup until you run a full backup starting a new chain.

docs/user/backup-restore.md
Chyenne8: No, performing a full backup doesn't delete anything. All backups (full or incremental) are just timestamped directories that coexist independently. You'd need to manually clean up old backups when you no longer need them.

src/playbooks/backup/metadata.obsah.yaml
Chyenne8: Yes, the backup directory be an incremental backup directory the code only checks for .snar files, not whether the previous backup is full or incremental. You can chain: full -> inc1 -> inc2 -> inc3 where each references the previous backup.

I've thought about auto backup option for users but my thoughts are how we would handle multiple independent backup sets in the same directory?

I agree --incremental sounds like a yes/ no flag.  I will change it to --base-backup ,so that it better indicates a path should follow the flag. let me know your thoughts on this name instead

src/roles/backup/tasks/database_dumps.yaml
Chyenne8: the quotes used here are defensive programming for shell safety, prevents issues if variables contain spaces or special characters. I will changes the instances of single quotes to double

src/roles/backup/tasks/main.yaml
Chyenne8: The .snar files are created even on the first full backup (when no --incremental flag is used) 
However, the .snar files will not exist if:
  - The backup was created before incremental support was added
  - The backup was created by a different tool
  - Someone manually created a backup directory

src/roles/backup/tasks/metadata.yaml
Chyenne8: updated this variable name to backup_incremental

src/roles/backup/tasks/pulp_content.yaml
Chyenne8: updated to use double quotes

src/roles/restore/tasks/main.yaml
Chyenne8: noted, backup_dir should be restore_backup_dir for role scoping.

src/roles/restore/tasks/validate.yaml
Chyenne8: Removed the defaults. If someone tries to restore from a very old backup (pre-incremental), they'll get a clear error that the metadata structure is incompatible. For all backups created with this code, the field is guaranteed to exist.

docs/user/backup-restore.md
Chyenne8: changed it to base backup

sjha4:
@Chyenne8 Can you update PR description to reflect the latest state of things. Ex: Removing --chain option etc.

docs/user/backup-restore.md
```diff
@@ -69,6 +70,38 @@ Allow in-progress tasks to complete before starting backup:
...
+
+```bash
+foremanctl backup /var/backup --base-backup /var/backup/foreman-backup-20260629T120000
+# Creates: /var/backup/foreman-backup-20260630T080000/ (incremental)
+```
+
+**How it works:**
+- The incremental backup contains only files changed since the previous backup
+- `.snar` snapshot files track which files were in the previous backup
+- Both `.config.snar` (foremanctl state) and `.pulp.snar` (Pulp content) are copied from the previous backup
```
sjha4: Why are these .pulp.snar and .config.snar vs pulp.snar and config.snar? I don't know if we need those hidden in the backup directory..

sjha4:
Tested this and it works well..Able to go through all incremental backups and restore in order..Got claude to stress test it a little and it looked all good. One question above about the hidden snar files.

docs/user/backup-restore.md
Chyenne8: I agree this should not be hidden from users. I updated this and other related sections to make it visible

sjha4 (approved):
Looking good to me..Works well in testing.

docs/user/backup-restore.md
ehelms: Should we then word it as such? `Restore the most recent full backup` ?

ehelms (approved):
Looks good. I think we should consider a future iteration that makes incremental more programmatic. Given it must start with a full backup, and increments are supposed to be generated sequentially, I think we could do something here to connect them together and make it a one-shot restore for users. That is, where the way we store the increments can tell the restore command:

1. This is the base backup
2. These are the incremental backups in order
3. Restore them in the correct order in one go for the user

sjha4:
We'll need to take a look at the coderabit reviews. @vsedmik Can you let us know if automation is good on this one to be merged?

src/roles/restore/tasks/restore_databases.yaml
```diff
@@ -85,13 +85,14 @@
...
   environment:
     PGPASSWORD: "{{ postgresql_admin_password }}"
   loop: "{{ restore_databases_to_restore }}"
   loop_control:
-    label: "{{ item.dump_file }} → {{ item.database }}"
+    label: "{{ item.dump_file }} -> {{ item.database }}"
   changed_when: true
-  failed_when: false
+  register: restore_pg_restore_result
+  failed_when: restore_pg_restore_result.rc != 0 and 'WARNING' not in restore_pg_restore_result.stderr
```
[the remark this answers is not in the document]
sjha4: This is a fair comment..We can burn some tokens on this to make sure we are not ignoring errors because of cleared warnings and report success.

src/roles/restore/tasks/restore_foremanctl_state.yaml
```diff
@@ -1,11 +1,48 @@
...
+- name: Check for config.snar file (incremental backup indicator)
+  ansible.builtin.stat:
+    path: "{{ restore_backup_dir }}/config.snar"
+  register: restore_config_snar_check
+
+- name: Remove existing foremanctl state directory for full restore
   ansible.builtin.file:
     path: "{{ obsah_state_path }}"
     state: absent
+  when: not restore_config_snar_check.stat.exists
```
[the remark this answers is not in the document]
sjha4: This is also a correct catch.

src/roles/restore/tasks/validate.yaml
```diff
@@ -0,0 +1,256 @@
...
+      - All configuration
+
+      There is NO UNDO operation.
+
+      If you are certain you want to proceed, re-run with --force:
+        foremanctl restore {{ backup_dir }} --force
+  when:
+    - restore_existing_deployment.stat.exists
+    - not force | default(false)
+    - not validate | default(false)
```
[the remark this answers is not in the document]
sjha4: We do not have restore_chain anymore. Right @Chyenne8 ?

vsedmik:
> We'll need to take a look at the coderabit reviews. @vsedmik Can you let us know if automation is good on this one to be merged?

After these two changes all 8 tests we have today pass:
- `--incremental` -> `--base-backup`
- `['.config.snar', '.pulp.snar']` -> `['config.snar', 'pulp.snar']`

src/roles/restore/tasks/validate.yaml
```diff
@@ -110,9 +232,8 @@
 
 - name: Get Pulp content archive size
   ansible.builtin.stat:
-    path: "{{ backup_dir }}/pulp-content.tar.gz"
+    path: "{{ restore_backup_dir }}/pulp-content.tar.gz"
   register: restore_pulp_archive
```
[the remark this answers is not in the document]
sjha4: Seems fair.Fail when stat doesn't exist on pulp content archive.

src/roles/restore/tasks/validate.yaml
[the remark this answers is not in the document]
Chyenne8: correct restore chain was removed

PR #611, opened 2026-07-01
Container Gateway support

ianballou:
The smart_proxy_container_gateway plugin is required for users to consume container content from smart proxies with Pulp.

* Adds support for the container gateway plugin to be configured with `foreman-proxy`. 

Steps to reproduce:

* Use quay.io/iballou/foreman-proxy-container-gateway:latest until https://github.com/theforeman/foreman-oci-images/pull/53 is merged
* Deploy a foreman proxy normally and see that the container gateway plugin is enabled.

#### Checklist


ehelms:
You can rebase this now that `deploy-proxy` work has been merged.

ianballou:
Linking up https://github.com/theforeman/foreman-oci-images/pull/53

ianballou:
Katello doesn't support the container gateway correctly yet. In its code, it expects the Pulp smart proxy to have the container gateway feature, but instead, the container gateway feature is on the non-Pulp smart proxy (since there are two per smart proxy host).

ianballou:
From my last comment - if this merges, containerized Katello + Container Gateway won't be any more broken than it is today, since Katello simply won't find the container gateway capsule and will skip the sync.

src/roles/httpd/defaults/main.yml
```diff
@@ -2,6 +2,7 @@ httpd_ssl_dir: /etc/pki/httpd
 httpd_pulp_api_backend: http://localhost:24817
 httpd_pulp_content_backend: http://localhost:24816
 httpd_foreman_backend: http://localhost:3000
+httpd_foreman_proxy_url: "https://{{ ansible_facts['fqdn'] }}:8443"
```
ianballou: This is the replacement for `$real_registered_proxy_url` from puppet-foreman_proxy.

https://github.com/theforeman/puppet-foreman_proxy/blob/725c3b518b7ea3a21e7db26ee74bbd622a4e9f83/manifests/init.pp#L431

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -127,6 +127,14 @@
   ProxyPass / {{ httpd_foreman_backend }}/ retry=0 timeout=900 upgrade=websocket
   ProxyPassReverse / {{ httpd_foreman_backend }}/
 {% else %}
+  ## Container Gateway
```
ehelms, line 134: The additions of these should technically be gated on enabled features. I think this is a gap across this whole file though.

src/features.yaml
```diff
@@ -62,6 +62,11 @@ iop:
   description: iop services
   dependencies:
     - rh-cloud
+container-gateway:
```
ehelms, line 76: These feels like it should be part of the "content/container" feature or a dependency of it... but only in the proxy use case.

src/roles/httpd/defaults/main.yml
ehelms: This is a bit confusing of a name since `foreman_proxy` has a whole nother meaning.

src/roles/foreman_proxy/templates/settings.d/container_gateway.yml.j2
```diff
@@ -0,0 +1,4 @@
+---
+:enabled: {{ feature_enabled }}
+:pulp_endpoint: {{ foreman_proxy_container_gateway_pulp_endpoint }}
+:db_connection_string: postgresql://{{ container_gateway_database_user }}:{{ container_gateway_database_password }}@{{ container_gateway_database_host }}:{{ container_gateway_database_port }}/{{ container_gateway_database_name }}
```
ehelms: Given these are inside the `foreman_proxy` role, they should be namespaced.

src/features.yaml
ianballou: ~~Sure, that makes sense to me.~~

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
ianballou: +1 to that

src/features.yaml
ianballou: Actually, I gave this some thought. What is the "proxy" use case we're talking about here? It's foreman-proxy-content specifically. That is the only time we want to install the container gateway. I thought about having a dependency relationship with a conditional on the use of foreman-proxy-content, but that seems like a roundabout way to achieve the same result as just having the container gateway be a part of the foreman-proxy-content flavor.

ianballou: So, while there isn't a direct relationship between the container gateway and `content/container`, there is a connection between them created by the `foreman-proxy-content` flavor.

src/roles/foreman_proxy/templates/settings.d/container_gateway.yml.j2
```diff
@@ -0,0 +1,4 @@
+---
+:enabled: {{ feature_enabled }}
+:pulp_endpoint: {{ foreman_proxy_container_gateway_pulp_endpoint }}
+:db_connection_string: {{ foreman_proxy_container_gateway_db_connection_string }}
```
ianballou, line 4: Is it an issue that the password ends up in plaintext in this config file? To get around this we could potentially implement environment variable support for the password.

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
ianballou: done!

src/roles/foreman_proxy/templates/settings.d/container_gateway.yml.j2
ianballou: Previously we relied on socket auth, so the password wasn't even there for a typical setup.

ehelms: Yes, this is fine. This is not the only place we do this.

src/vars/base.yaml
```diff
@@ -55,6 +55,12 @@ hammer_plugins: "{{ enabled_features | features_to_hammer_plugins }}"
 hammer_foreman_server_url: "{{ foreman_url }}"
 
 foreman_proxy_plugins: "{{ enabled_features | features_to_foreman_proxy_plugins }}"
+foreman_proxy_container_gateway_db_connection_string: >-
```
ehelms: As this is a construct, I feel like it should be defined in the role itself? Another way to look at it is, what is the public interface? The variables that are in the connection string or the connection string itself.

ianballou: It's debatable, but I think publicly we'd want the variables that make up the connection string to be the interface. That's how it worked before with the Puppet module.

+1 to moving the variable to the role, that way it'll be scoped to foreman-proxy.

src/features.yaml
ehelms: If there is no Pulp or there is no Pulp with container content present, does this feature work? What happens if a container-gateway is deployed without Pulp or Pulp with container content on the same server?

We can go this route for now, if we need to, to allow the feature to work, but I would ask that we track doing the dependency relationship properly across foremanctl and Katello.

ianballou:
@ehelms I reorganized the DB connection string variables so an Ansible role user can set the individual components or the entire string manually if they wish.

I also added an integration test along with a temporary container image to see it working while we wait for https://github.com/theforeman/foreman-oci-images/pull/53.

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -18,6 +18,20 @@ foreman_proxy_disabled_features: "{{ foreman_proxy_available_features | differen
 
 foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
 
+# Container gateway settings
+foreman_proxy_container_gateway_pulp_endpoint: "https://{{ ansible_facts['fqdn'] }}"
+foreman_proxy_container_gateway_db_host: "{{ container_gateway_database_host }}"
```
ehelms: These aren't really defaults if they are defined by variables from somewhere else. I would expect default values as strings here, and for `database.yml` to then define the left hand side.

src/vars/database.yml
```diff
@@ -17,6 +17,13 @@ pulp_database_user: pulp
 pulp_database_password_file: "{{ obsah_state_path }}/pulp-db-password"
 pulp_database_password: "{{ lookup('ansible.builtin.password', pulp_database_password_file, chars=['ascii_letters', 'digits']) }}"
 
+container_gateway_database_name: container_gateway
+container_gateway_database_user: container_gateway
+container_gateway_database_password_file: "{{ obsah_state_path }}/container-gateway-db-password"
+container_gateway_database_password: "{{ lookup('ansible.builtin.password', container_gateway_database_password_file, chars=['ascii_letters', 'digits']) }}"
+container_gateway_database_host: "{{ database_host }}"
+container_gateway_database_port: "{{ database_port }}"
```
ehelms: I would expect these to all be the `foreman_proxy_` scoped variable names -- https://github.com/theforeman/foremanctl/pull/611/changes#r3639114753

sjha4:
Ran a test for enabling container_gateway with foremanctl deploy-proxy on this branch on an existing proxy deployment and it worked well for me..Was able to run a backup as a follow up test and it correctly identified container_gateway and backed up the pulp and container_gateway DBs. 👍🏼

Update: Was able to test the proxy backup/restore on top of this and it worked well. https://github.com/theforeman/foremanctl/pull/684#issuecomment-5061216977

src/features.yaml
ianballou: The container gateway does require Pulp, so from the standpoint of looking into deploying a smart proxy manually via features it would make sense for there to be some dependency there. I'm trying to weigh the complexity for the short term of using the foreman-proxy-content flavor as the glue between Pulp and container gateway since we seem to be putting a lot of emphasis into it being "the" smart proxy deployment at the moment.

But I totally agree that the dependency should be linked more generally for users constructing their own capsules without using the foreman proxy content flavor. I can think up an issue for that  that covers what we'd need to achieve the dependency with a conditional on the foreman-proxy context.

ianballou:
@ehelms I'm wrapping my head around the proper organization of the variables your asking about, but I made a push in response to your comments.

I need to look into the failing tests, probably some expectations in other tests that I did not update.

src/features.yaml
ianballou: I'm cooking something up, I'll push a commit soon.

ianballou: Okay, so while I am going to push something that links these dependencies - I currently see no way for an admin to build a foreman-proxy-content without using the flavor, which means the dependency link is purely for sanity rather than user benefit. At least for now.

The Pulp feature and its content types (besides ostree) are all internal, so I don't see how a user can use the foreman-proxy feature to build their own foreman-proxy-content. I suppose this may get smoothed out in the future.

ianballou:
I pushed a change that makes the container gateway a dependency of content/container for all cases not matching "Foreman", which I assume here would mean smart proxy.

I did this because foreman-proxy can be on Foreman or not on Foreman. Thus, the only marker of a "foreman-proxy-content" machine is the "foreman-proxy-content" flavor. The flavor alone is not enough to say there is a dependency between content/container and the container gateway, so my next best idea is to use the lack of Foreman in a deployment to mean that we are talking about a "foreman proxy content".

This change means that we have a new concept of "unless" for features that are incompatible with each other. To accomplish this, I turned feature dependencies into dictionaries that can have "unless" as a list of incompatible plugins.

ehelms:
There are similar conversations happening in https://github.com/theforeman/foremanctl/pull/519#discussion_r3637571695 

CC @ekohl @jeremylenz @shubhamsg199 

The tension here is we have a single definition of features and then multiple scenarios that can apply:

- Features that only apply to Foreman server
- Features that apply to Foreman server and external proxy
- Features that apply to only external proxy

ianballou:
I had difficulty writing this new feature in a way that didn't assume "no Forman == Forman proxy content" because the only real indicator of this that I could find is the use of the flavor. On top of that, the dependency linking feels unhelpful with current foremanctl because of this huge reliance on the flavor. 

I think the fact that flavors include internal only features is part of the problem. If we want flavor use to be optional, then they should not include internal features that cannot be enabled outside of the flavors. If we don't want flavors to be optional, then I think it should be fine for them to be a first class part of the feature dependency chain.

ehelms:
> I think the fact that flavors include internal only features is part of the problem. If we want flavor use to be optional, then they should not include internal features that cannot be enabled outside of the flavors. If we don't want flavors to be optional, then I think it should be fine for them to be a first class part of the feature dependency chain.

I would agree -- flavors should not specify internal features at least based on my understanding of internal features.

The current design is that *a* flavor is required as it defines the baseline, unchangeable properties of the deployment. However, I do not see why features need to know about flavors.

Should features know about "server" and "proxy" -- I think so. I still strongly contend there are two foundational entities: server and proxy. And it matters which one is the target. That's why right now we have two distinct deployment commands. 

So from my point of view:

- Flavors should explicitly apply to either `deploy` or `deploy-proxy` but not both.
- Features should be marked as applying to one of the the three scenarios I mentioned previously.

ianballou:
I was considering if we could simply add boolean values to each feature to declare what deployment scenario(s) each support.

```
katello:
  description: Content and Subscription Management plugin for Foreman
  foreman:
    plugin_name: katello
  hammer: katello
  dependencies:
    - tasks
    - pulp
    - candlepin
  deploy_on_foreman: true
  deploy_on_foreman_proxy: false

container-gateway:  
  description: Container registry gateway for smart proxy
  internal: true
  foreman_proxy:
    plugin_name: container_gateway
  deploy_on_foreman: false
  deploy_on_foreman_proxy: true
```

Are there other features that can only be installed on external foreman proxies?

jeremylenz:
> Are there other features that can only be installed on external foreman proxies?

templates and registration are two! I mean, I guess you _could_ install them on your foreman server, but they're not really relevant since they just proxy api endpoints.

ianballou:
Before I make any changes, I'd like opinions on adding the feature incompatibility feature I have here (`unless`) versus making deployment methods a first-class indicator of which features should go on foreman/proxy/both.

I like using deployment methods as an indicator since checking for the existence of `foreman` feels like a workaround.

jeremylenz:
> I was considering if we could simply add boolean values to each feature to declare what deployment scenario(s) each support.

I would go beyond booleans here, and have a list of supported scenarios. They could perhaps mirror the foremanctl commands. Something like

```
relevant_to_scenarios: 
  - deploy
  - deploy-proxy
```

ianballou:
I split out the idea of having features applicable to only some deployment methods into a separate issue. That will unblock the container gateway. For now I'll have it just be a part of the foreman-proxy-content flavor.

https://github.com/theforeman/foremanctl/issues/723

src/roles/foreman_proxy/defaults/main.yaml
```diff
@@ -21,6 +21,19 @@ foreman_proxy_foreman_server_url: "https://{{ ansible_facts['fqdn'] }}"
 # Templates settings
 foreman_proxy_templates_url: "http://{{ foreman_proxy_name }}:8000"
 
+# Container gateway settings
+foreman_proxy_container_gateway_pulp_endpoint: "https://{{ ansible_facts['fqdn'] }}"
```
evgeni: Is this the pulp on the proxy or on the main server?

ianballou: This is the Pulp on the smart proxy, the container gateway doesn't know about the Pulp on the main server

src/features.yaml
evgeni: What if we add the proxy part to `content/container`, so you get:
```
content/container:
  description: Container content type for Pulp
  internal: true
  dependencies:
    - pulp
  foreman_proxy:
    plugin_name: container_gateway
```

but the real enablement in `src/roles/foreman_proxy/templates/settings.d/container_gateway.yml.j2` and everywhere else is guarded by a second boolean that is only true on external proxy deployments?

ekohl:
> katello:
>   description: Content and Subscription Management plugin for Foreman
>   foreman:
>     plugin_name: katello
>   hammer: katello
>   dependencies:
>     - tasks
>     - pulp
>     - candlepin
>   deploy_on_foreman: true
>   deploy_on_foreman_proxy: false

I'm not too keen on this. You can already make it depend on `foreman`. I'd rather make sure you can't have the `foreman` feature in the `deploy-proxy` case and then do dependency resolution.

Perhaps the flavor support could gain support for something like:
```yaml
mask_features:
  - foreman
```

> container-gateway:  
>   description: Container registry gateway for smart proxy
>   internal: true
>   foreman_proxy:
>     plugin_name: container_gateway
>   deploy_on_foreman: false
>   deploy_on_foreman_proxy: true

This isn't going to work for a plain Foreman Proxy case (https://github.com/theforeman/foremanctl/pull/527).

How about:
```yaml
container-gateway:  
  description: Container registry gateway for smart proxy
  foreman_proxy:
    plugin_name: container_gateway
  dependencies:
    - foreman-proxy
    - pulp
  conflicts:
    - foreman
```

src/roles/foreman_proxy/defaults/main.yaml
evgeni: Then I'd use `foreman_proxy_name` here:

```suggestion
foreman_proxy_container_gateway_pulp_endpoint: "https://{{ foreman_proxy_name }}"
```

the fewer we use fqdn, the better :)

ianballou:
> How about:
> 
> ```yaml
> container-gateway:  
>   description: Container registry gateway for smart proxy
>   foreman_proxy:
>     plugin_name: container_gateway
>   dependencies:
>     - foreman-proxy
>     - pulp
>   conflicts:
>     - foreman
> ```

This is similar to the "unless" thing I had before, it works for me. This is now going down the route of making the deployment methods / flavors less strong I suppose. So a user could craft their own proxy from features, theoretically.

ekohl:
> So a user could craft their own proxy from features, theoretically.

And that's exactly the original intent we had with features. You can compose it all to build your own deployment. Flavor is just your starting point.

ianballou:
> > So a user could craft their own proxy from features, theoretically.
> 
> And that's exactly the original intent we had with features. You can compose it all to build your own deployment. Flavor is just your starting point.

I like this strategy too, the flavors are help but not a mandate.

It appears I cannot mark container gateway as a requirement of content/container for it to be pulled in automagically when it is paired with foreman-proxy (due to the _resolved_ container-gateway dependency not being checked by the new conflict checker), but I think that's okay. Users crafting their own environments without flavors will at least find that they are not allowed to put the container gateway on a Foreman machine.

ianballou:
Oh - and since the container gateway is not pulled in automatically with container-content for the foreman-proxy feature, I'm going to make it no longer an internal one.

ianballou:
I caught @evgeni 's comment that was in-thread a bit late, it would theoretically allow the container gateway to actually be auto-included with content/container:

> What if we add the proxy part to `content/container`, so you get:
> 
> ```
> content/container:
>   description: Container content type for Pulp
>   internal: true
>   dependencies:
>     - pulp
>   foreman_proxy:
>     plugin_name: container_gateway
> ```
> 
> but the real enablement in `src/roles/foreman_proxy/templates/settings.d/container_gateway.yml.j2` and everywhere else is guarded by a second boolean that is only true on external proxy deployments?

I can add it if it sounds interesting to others too. It would be an add-on to my latest commit. 

The part that gives me some pause is `a second boolean that is only true on external proxy deployments` since it sounds like it could end up being another discussion piece. Does this boolean have meaning for foreman-proxy-content vs foreman-proxy, does it introduce complexity to our deployment method fundamentals, things like that.

src/vars/images.yml
```diff
@@ -6,8 +6,8 @@ candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "foreman-{{ container_tag_stream }}"
 foreman_container_image: quay.io/foreman/foreman
 foreman_container_tag: "{{ container_tag_stream }}"
-foreman_proxy_container_image: "quay.io/foreman/foreman-proxy"
-foreman_proxy_container_tag: "{{ container_tag_stream }}"
+foreman_proxy_container_image: "quay.io/iballou/foreman-proxy-container-gateway"
```
ehelms: Noting this needs to be updated before merge.

src/features.yaml
ehelms: Thinking on it more, I do think this is a broader design question that affects multiple features. I asked @ianballou to file https://github.com/theforeman/foremanctl/issues/723 to discuss and solve it broadly and allow this PR to progress without blocking on this needed bigger solution.

src/vars/images.yml
ianballou: Once someone soft ACKs the PR, I can remove this commit.

tests/flavor/foreman-proxy-content/container_gateway_test.py
ekohl, line 1: I'd love to avoid the `tests/flavor` directory when possible. Is it an idea to use `test/feature` to test it when the feature is enabled?

tests/flavor/foreman-proxy-content/container_gateway_test.py
```diff
@@ -0,0 +1,17 @@
+from urllib.parse import urlparse
+
+import yaml
+
+
+def test_container_gateway_db_connection_string(server):
+    result = server.run("podman exec foreman-proxy cat /etc/foreman-proxy/settings.d/container_gateway.yml")
```
ekohl: Instead of doing a lot of internal things, is there a way to observe it works externally? In other places we look at `/features` to see if it's really presented as enabled (and thus loaded up correctly), or `/v2/features`. https://github.com/theforeman/foremanctl/blob/master/tests/feature/foreman-proxy/base_test.py does that.

And for more end-to-end functional testing I still think https://github.com/theforeman/smart-proxy/pull/847 is the best approach. Then you can even extend Foreman's ping command with a deep verification to debug issues at runtime.

ekohl (commented):
#726 will create some merge conflicts. Perhaps wait for that to be merged before rebasing.

tests/flavor/foreman-proxy-content/container_gateway_test.py
ianballou: Sure, I can take a look at implementing it more from a user's standpoint.

ekohl:
I'd say the overall direction is good and now that we merged #726 you can rebase this.

ianballou:
Let's see if these new tests pass. I added a podman login integration test. In the future I'd like to have a full container sync to podman pull test, but that'll require a bit more change than I'd like to add here.

ianballou:
Ah the failure is due to the container gateway being too old.

ianballou:
Alrighty folks, tests are passing, this is ready for a (final?) look.

src/vars/images.yml
```diff
@@ -6,8 +6,8 @@ candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "foreman-{{ container_tag_stream }}"
 foreman_container_image: quay.io/foreman/foreman
 foreman_container_tag: "{{ container_tag_stream }}"
-foreman_proxy_container_image: "quay.io/foreman/foreman-proxy"
-foreman_proxy_container_tag: "{{ container_tag_stream }}"
+foreman_proxy_container_image: "quay.io/iballou/foreman-proxy"
```
ehelms: This still needs updating

src/features.yaml
```diff
@@ -72,6 +74,15 @@ iop:
...
+  description: Container registry gateway for smart proxy
+  foreman_proxy:
+    plugin_name: container_gateway
+  dependencies:
+    - foreman-proxy
+    - pulp
+  conflicts:
```
ehelms, line 84: Does this do anything at present?

ekohl: https://github.com/theforeman/foremanctl/blob/master/docs/developer/feature-metadata.md describes `conflicts` and there is code for it:
https://github.com/theforeman/foremanctl/blob/3492e079a90d6b4151b2b330783d82fa47e71eeb/src/filter_plugins/foremanctl.py#L104-L111

ehelms: What silly person added that. I wasn't thinking of it's use in this use case :)

src/vars/images.yml
ianballou: Will update once someone says "I'd ack this if that hack wasn't there!"

tests/feature/container-gateway/base_test.py
```diff
@@ -0,0 +1,34 @@
...
+    assert cmd.succeeded, f"Failed to query /v2/features: {cmd.stderr}"
+    v2_features = json.loads(cmd.stdout)
+    assert 'container_gateway' in v2_features
+    assert v2_features['container_gateway'].get('state') == 'running'
+
+
+@pytest.fixture
+def podman_login(server, server_fqdn):
+    """Login to the container registry and logout after the test."""
+    result = server.run(f"podman login {server_fqdn} --username admin --password changeme --tls-verify=false")
```
ekohl: You're supposed to use interpolation because that also performs escaping. See https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.host.Host.run for more:
```suggestion
    result = server.run("podman login %s --username admin --password changeme --tls-verify=false", server_fqdn)
```

tests/feature/container-gateway/base_test.py
```diff
@@ -0,0 +1,34 @@
...
+    assert cmd.succeeded
+    features = json.loads(cmd.stdout)
+    assert "container_gateway" in features
+
+
+def test_container_gateway_running(curl_request, proxy_base_url):
+    """Verify container_gateway is loaded and running via /v2/features."""
+    cmd = curl_request("v2/features", base_url=proxy_base_url, return_body=True)
+    assert cmd.succeeded, f"Failed to query /v2/features: {cmd.stderr}"
+    v2_features = json.loads(cmd.stdout)
```
ekohl: This is repeated in a few places. Perhaps it should be a fixture in `conftest.py` by now?

tests/feature/container-gateway/base_test.py
```diff
@@ -0,0 +1,34 @@
...
+    assert 'container_gateway' in v2_features
+    assert v2_features['container_gateway'].get('state') == 'running'
```
ekohl: Any particular reason you don't just assert it straight away, counting on a `KeyError` to be raised if it fails? I'd assume pytest is smart enough to provide sufficient context when it fails.
```suggestion
    assert v2_features['container_gateway']['state'] == 'running'
```

tests/feature/container-gateway/base_test.py
```diff
@@ -0,0 +1,34 @@
...
+
+import pytest
+
+
+def test_container_gateway_in_proxy_features(curl_request, proxy_base_url):
+    """Verify container_gateway appears in the proxy's feature list."""
+    cmd = curl_request("features", base_url=proxy_base_url, return_body=True)
+    assert cmd.succeeded
+    features = json.loads(cmd.stdout)
+    assert "container_gateway" in features
```
ekohl: With the `test_container_gateway_running` below I think this test may be redundant.

ekohl:
And overall I'd say this is :+1: so you can drop the hack.

src/vars/images.yml
ehelms: Which hack?

ekohl: I'm guessing using an image built from a PR.

sjha4:
I was able to test through the current state of this and it continues to work. 🟢 

Side question: Since there are multiple proxies now, do we need to ensure that all proxies running on a server are attached to the same orgs/locations. For pulp and container_gateway in particular, having those 2 in different orgs/locations seems a little odd.

src/vars/images.yml
ianballou: Yeah the use of my personal image is the hack.

ekohl:
> Side question: Since there are multiple proxies now, do we need to ensure that all proxies running on a server are attached to the same orgs/locations. For pulp and container_gateway in particular, having those 2 in different orgs/locations seems a little odd.

That's a good point and I don't think we track that today. Feels like something that would easily break things in subtle ways.

ianballou:
> Side question: Since there are multiple proxies now, do we need to ensure that all proxies running on a server are attached to the same orgs/locations. For pulp and container_gateway in particular, having those 2 in different orgs/locations seems a little odd.

Frankly at the moment I've been assuming that admins will need to manage this, at least at first. That is to say, I think we should help with this, but we need to be careful not to over prescribe what taxonomies the smart proxies are connected to.

tests/feature/container-gateway/base_test.py
ianballou: Works for me, I'll create a fixture.

tests/feature/container-gateway/base_test.py
ianballou: Eh, just some safe navigation I threw in,  but it's really not necessary in the test. In fact I suppose it may even hide regressions. I'll remove it.

ekohl:
You're right that admins should, but we can help point them in the right direction in our docs. We're thinking about enablement docs that point out "this is different in a containerized install", which we could point out.

ianballou:
> You're right that admins should, but we can help point them in the right direction in our docs. We're thinking about enablement docs that point out "this is different in a containerized install", which we could point out.

I agree that this would be great for an enablement doc, while this smart proxy situation isn't something too new from a technical standpoint, users have built up so much "muscle memory" around having one "internal smart proxy" that this sort of doc should go a long way.

tests/feature/container-gateway/base_test.py
ianballou: Actually, I can reuse `proxy_v2_features`

ianballou:
I updated the tests. Will remove the custom container image once I see them pass here.

ianballou:
Ah, right, tests are dead on CP 5 :)

jeremylenz:
@ianballou Should be green after a rebase now!

src/vars/images.yml
ehelms: You can drop this now.

ianballou:
Tests are green! Removing my custom image.

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -130,6 +130,16 @@
...
   ProxyPassReverse / {{ httpd_foreman_backend }}/
 {% else %}
+{% if httpd_with_container_gateway %}
+  ## Container Gateway
+  ProxyPass /v1/ {{ foreman_proxy_url }}/container_gateway/v1/
+  ProxyPassReverse /v1/ {{ foreman_proxy_url }}/container_gateway/v1/
+  ProxyPass /v2/ {{ foreman_proxy_url }}/container_gateway/v2/
+  ProxyPassReverse /v2/ {{ foreman_proxy_url }}/container_gateway/v2/
+  ProxyPass /index/ {{ foreman_proxy_url }}/container_gateway/index/
+  ProxyPassReverse /index/ {{ foreman_proxy_url }}/container_gateway/index/
```
lzap, line 140: So `/v{1,2}/` is basically claimed by container gateway now and nothing else can use this location prefix? It worked until no because everything else was our APIs we could probably easily customize, but container registry is I guess fixed and we can't do much about it.

I guess it is good enough for now, we can always add more statements here for each and every endpoint in the worst case. I am thinking if virtual hosts would be a cleaner approach. We do not manage DNS tho, starting asking our users to create entries like `cw.satellite.example.com` could be a pain.

ehelms: > We do not manage DNS tho, starting asking our users to create entries like cw.satellite.example.com could be a pain.

We have discussed this idea for a variety of things in the past and while it's a great idea, we never felt we could enforce this on users and at best it could be a best practice. Since these routes only apply to a content proxy (aka Capsule) there's not been any real risk of clashing.

PR #612, opened 2026-07-01
Enable parameter persistence for forge

jeremylenz:

`foremanctl` already persists parameters between runs via `OBSAH_PERSIST_PARAMS`, but `forge` does not. This means arguments like `--add-feature` passed to `forge deploy-dev` are forgotten on subsequent runs, requiring developers to re-specify them every time.

* Set `OBSAH_PERSIST_PARAMS=true` in the `forge` wrapper script, matching `foremanctl`'s behavior. This ensures that deploy-dev arguments (features, plugins, etc.) are remembered across runs.

Steps to reproduce:

* Run `forge deploy-dev --add-feature iop` (or any other arguments)
* Run `forge deploy-dev` again without specifying the feature
* Verify the previously added feature is still enabled

#### Checklist


ehelms:
What do you think about setting `OBSAH_STATE=${OBSAH_BASE}/.var/lib/foremanctl` to `OBSAH_STATE=${OBSAH_BASE}/.var/lib/foremanctl-dev` or `OBSAH_STATE=${OBSAH_BASE}/.var/lib/forge` in order to prevent clashing parameters between prod and dev?

jeremylenz:
I think that's a great idea! Updated

ehelms:
I think this leads to a variety of assumptions that our code makes that won't work. I guess we keep it where `foremanctl` points for now.

ehelms (approved):

Issue #565, opened 2026-06-12
Decouple service name from hostname


ekohl:
Today foreman-installer is unintentionally quite tied to the system's hostname. This necessitates tools like `katello-change-hostname`. In foremanctl I'd like to decouple it fully, which aligns with a containerized model.

A use cases that I'd like to see work: install on 1 hostname, then change it and keep serving the application on the old.
```console
# hostname
foreman.example.com
# foremanctl deploy
# hostnamectl hostname host12345.example.com
# foremanctl deploy
# curl https://foreman.example.com
valid output here
```

Another is to install straight away on a different service name:
```console
# hostname
host12345.example.com
# foremanctl deploy --service-name foreman.example.com
# curl https://foreman.example.com
valid output here
```

The exact parameter name is up for debate.

Today we use the FQDN fact too often and that's IMHO an anti-pattern. For example:
https://github.com/theforeman/foremanctl/blob/099a9feea9725c82be7c5e9771db9fdc94671a27/src/vars/base.yaml#L3
https://github.com/theforeman/foremanctl/blob/099a9feea9725c82be7c5e9771db9fdc94671a27/src/vars/base.yaml#L34-L35

ekohl:
Duplicates https://github.com/theforeman/foremanctl/issues/73.

PR #618, opened 2026-07-03
Only set pulp_rhsm_url for content proxies

pablomh:
## Problem

`pulp_rhsm_url` is currently set for all Pulp deployments.

On Foreman server deployments, that makes foremanctl export
`PULP_SMART_PROXY_RHSM_URL` even though Katello can already derive the
server RHSM URL on its own when the setting is absent.

On content proxies, an explicit RHSM URL is still needed so clients use the
proxy's `/rhsm` endpoint rather than the Foreman server directly.

## Solution

- Stop exporting `PULP_SMART_PROXY_RHSM_URL` for Foreman server deployments
- Set `pulp_rhsm_url` only for the `foreman-proxy-content` flavor
- Keep the explicit content proxy RHSM URL as `https://<proxy-fqdn>/rhsm`

## Testing

- Added coverage that server deployments do not export
  `PULP_SMART_PROXY_RHSM_URL`, while Katello still derives the server
  `/rhsm` URL for the server Pulp smart proxy
- Added coverage that `foreman-proxy-content` deployments do export it and
  point it to the proxy host

### DIFF
--- a/src/roles/pulp/defaults/main.yaml
+++ b/src/roles/pulp/defaults/main.yaml
@@ -29,7 +29,6 @@ pulp_pulp_url: "http://{{ ansible_facts['fqdn'] }}:24817"
 pulp_enable_analytics: false
 pulp_mirror: false
 pulp_register_foreman_proxy: true
-pulp_rhsm_url: "https://{{ ansible_facts['fqdn'] }}/rhsm"
 
 pulp_default_import_paths: "{{ pulp_mirror | ternary([], ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']) }}"
 pulp_default_export_paths: "{{ pulp_mirror | ternary([], ['/var/lib/pulp/exports']) }}"
@@ -80,7 +79,7 @@ pulp_settings_other_env:
     ['rest_framework.authentication.SessionAuthentication', 'pulpcore.app.authentication.PulpRemoteUserAuthentication']
   PULP_SMART_PROXY_PULP_URL: "{{ pulp_pulp_url }}"
   PULP_SMART_PROXY_MIRROR: "{{ pulp_mirror }}"
-  PULP_SMART_PROXY_RHSM_URL: "{{ pulp_rhsm_url }}"
+  PULP_SMART_PROXY_RHSM_URL: "{{ pulp_rhsm_url | default('') }}"
   PULP_API_WORKERS: "{{ pulp_api_service_worker_count }}"
   PULP_CONTENT_WORKERS: "{{ pulp_content_service_worker_count }}"
   PULP_TOKEN_AUTH_DISABLED: "true"
--- a/src/vars/base.yaml
+++ b/src/vars/base.yaml
@@ -44,7 +44,6 @@ pulp_storage_path: /var/lib/pulp
 pulp_content_origin: "https://{{ ansible_facts['fqdn'] }}"
 pulp_pulp_url: "https://{{ ansible_facts['fqdn'] }}"
 pulp_plugins: "{{ enabled_features | select('contains', 'content/') | map('replace', 'content/', 'pulp_') | list }}"
-pulp_rhsm_url: "{{ foreman_proxy_foreman_server_url }}/rhsm"
 pulp_foreman_url: "{{ foreman_proxy_foreman_server_url }}"
 pulp_foreman_ca_certificate: "{{ foreman_ca_certificate }}"
 pulp_foreman_oauth_consumer_key: "{{ foreman_proxy_oauth_consumer_key }}"
--- a/src/vars/flavors/foreman-proxy-content.yml
+++ b/src/vars/flavors/foreman-proxy-content.yml
@@ -7,6 +7,7 @@ flavor_features:
   - content/python
 
 pulp_mirror: true
+pulp_rhsm_url: "https://{{ ansible_facts['fqdn'] }}/rhsm"
 
 checks_to_execute:
   - check_features
--- a/tests/flavor/foreman-proxy-content/pulp_test.py
+++ b/tests/flavor/foreman-proxy-content/pulp_test.py
@@ -15,11 +15,16 @@ def pulp_settings(server):
     py = (
         'from django.conf import settings; import json; '
         'print(json.dumps({"import": list(settings.ALLOWED_IMPORT_PATHS), '
-        '"export": list(settings.ALLOWED_EXPORT_PATHS)}))'
+        '"export": list(settings.ALLOWED_EXPORT_PATHS), '
+        '"rhsm_url": settings.SMART_PROXY_RHSM_URL}))'
     )
     result = server.run(f"podman exec pulp-api pulpcore-manager shell -c '{py}'")
     assert result.succeeded, f"Failed to read Pulp settings: {result.stderr}"
     return json.loads(result.stdout)
+
+
+def test_pulp_rhsm_url_set_on_content_proxy(pulp_settings, server_fqdn):
+    assert pulp_settings['rhsm_url'] == f'https://{server_fqdn}/rhsm'
 
 
 def test_import_paths_restricted(pulp_settings):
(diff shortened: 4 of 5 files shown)


ekohl (commented):
Not looking at the details, but more about the context. When submitting PRs upstream, please avoid using downstream terminology. Satellite and Capsule do no exist in upstream, nor should upstream care. Please refer to Foreman server and content proxies instead.

src/vars/base.yaml
```diff
@@ -36,7 +36,7 @@ pulp_storage_path: /var/lib/pulp
 pulp_content_origin: "https://{{ ansible_facts['fqdn'] }}"
 pulp_pulp_url: "https://{{ ansible_facts['fqdn'] }}"
 pulp_plugins: "{{ enabled_features | select('contains', 'content/') | map('replace', 'content/', 'pulp_') | list }}"
-pulp_rhsm_url: "{{ foreman_proxy_foreman_server_url }}/rhsm"
```
arvind4501, line 47: Could we just remove `pulp_rhsm_url` from base and fallback to defaults https://github.com/theforeman/foremanctl/blame/master/src/roles/pulp/defaults/main.yaml#L29 which will always be hosts fqdn

ekohl: What it should be:
* On Foreman server it should be empty
* On a content proxy it should be the service name that Apache binds to

arvind4501: But i see that in puppet based installer, we did kept it default to fqdn https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/plugin/pulp/params.pp#L6 :confused:

pablomh: I hadn't thought such simple solution. Doing that I could remove the addition of foreman_proxy_name as well.

Allow me some time to test it.

ekohl: > But i see that in puppet based installer, we did kept it default to fqdn https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/plugin/pulp/params.pp#L6 😕

That predates https://github.com/Katello/katello/commit/46abdb24234df46c94424cf376b6119e386c8c70. 55fa1d2c4f93abb02acb8eda2ebba289a4873d93 was the first to make use of it.

So to be clear: for content proxies that have deployed an RHSM proxy it is needed, but when I wrote 55fa1d2c4f93abb02acb8eda2ebba289a4873d93 we didn't have those.

arvind4501: Thanks for details, so does that mean we will only set rhsm url for content proxy only? sorry if i missed something

i think make sense now, if no pulp_rhsm_url given fallback to foreman_url otherwise used configured pulp_rhsm_url

ekohl: > Thanks for details, so does that mean we will only set rhsm url for content proxy only? sorry if i missed something

Yes.

> i think make sense now, if no pulp_rhsm_url given fallback to foreman_url otherwise used configured pulp_rhsm_url

Pretty much.

The idea behind https://github.com/Katello/katello/commit/46abdb24234df46c94424cf376b6119e386c8c70 is that you can decouple Pulp and the RHSM part. For example, you could deploy a scalable Pulp deployment in OpenShift without the RHSM component. In that case, Katello would already know the Foreman URL and point clients there.

We're not there yet, but on a Foreman server (where there is no RHSM proxy) there's no point in setting the URL: Katello already knows it.

pablomh:
Now it should be good to go.

arvind4501:
@pablomh , i think this still sets `pulp_rhsm_url` in both(server and proxy)

src/roles/pulp/defaults/main.yaml
```diff
@@ -26,7 +26,6 @@ pulp_pulp_url: "http://{{ ansible_facts['fqdn'] }}:24817"
 pulp_enable_analytics: false
 pulp_mirror: false
 pulp_register_foreman_proxy: true
-pulp_rhsm_url: "https://{{ ansible_facts['fqdn'] }}/rhsm"
```
ekohl, line 32: I think it handles an empty string just fine: https://github.com/theforeman/pulp_smart_proxy/blob/53894170bd37b8f012dbee9aa09bf82137020e7b/pulp_smart_proxy/app/views.py#L61-L62

I'd prefer to keep the code simple. and avoid the conditionals.
```suggestion
pulp_rhsm_url: ""
```

tests/flavor/foreman-proxy-content/pulp_test.py
```diff
@@ -13,15 +13,20 @@ def pulp_smart_proxy_features(curl_request):
 @pytest.fixture(scope="module")
 def pulp_settings(server):
     py = (
-        'from django.conf import settings; import json; '
+        'from django.conf import settings; import json, os; '
         'print(json.dumps({"import": list(settings.ALLOWED_IMPORT_PATHS), '
-        '"export": list(settings.ALLOWED_EXPORT_PATHS)}))'
+        '"export": list(settings.ALLOWED_EXPORT_PATHS), '
+        '"rhsm_url": os.environ.get("PULP_SMART_PROXY_RHSM_URL")}))'
```
ekohl: Please don't use the env var. This is a setting
```suggestion
        '"rhsm_url": settings.SMART_PROXY_RHSM_URL}))'
```

src/vars/flavors/foreman-proxy-content.yml
```diff
@@ -7,6 +7,7 @@ flavor_features:
   - content/python
 
 pulp_mirror: true
+pulp_rhsm_url: "https://{{ ansible_facts['fqdn'] }}/rhsm"
```
ekohl, line 10: We shouldn't use FQDN facts in any vars, other than in defaults. See https://github.com/theforeman/foremanctl/issues/565 for the background.

I now see that httpd also uses the fqdn fact so perhaps we should ignore that for now.

src/roles/pulp/defaults/main.yaml
pablomh: Thanks for the code hint. I looked for "empty"vars in the repo and didn't find any, so I decided to use the alternative implementation.

src/vars/flavors/foreman-proxy-content.yml
pablomh: Could we use `foreman_proxy_name` as an alternative here?

ekohl: Since httpd is configured with the FQDN, I think it's also fine to use that here for now.

ekohl (commented):
Conceptually this looks OK to me, but the tests are failing. Is that related?

pablomh:
They seem to be related to backups and the proxy-specific RHSM test passed. I've just re-run the CI and it fully passed. I'm going to rebase and squahs and hope for it to succeed again.

pablomh:
Finally a clean CI :)

src/roles/pulp/defaults/main.yaml
```diff
@@ -29,7 +29,7 @@ pulp_pulp_url: "http://{{ ansible_facts['fqdn'] }}:24817"
...
+pulp_rhsm_url: "" # noqa: no-empty-defaults
```
ehelms: I think the right way to do this is to not include it here, and instead set:

```
    PULP_SMART_PROXY_RHSM_URL: "{{ pulp_rhsm_url | default('') }}"
```

ehelms:
A small nitpick, then I think this is good to merge based on the current design. Overall, I think the `rhsm_url` should be removed from the Pulp smart-proxy and done instead as part of something like the `registration` feature or a http-proxy type feature given with the way we are doing the pulp smart-proxy now, it's not tied to the existence of Apache.

ekohl:
That's the direction I've been wanting for a long time and I've left some breadcrumbs along the way to get there though. In https://github.com/theforeman/foreman/pull/8563#discussion_r643051682 I pushed for it being abstracted into a helper to get the RHSM URL. https://github.com/Katello/katello/pull/9417 implemented that. In https://github.com/Katello/katello/pull/11328 I changed it to make it optional.

Now we're getting rid of the old helpers that hardcoded things (the registration RPM) and slowly we're we're ready to fully decouple Pulp and RHSM. At least, make it optional.

pablomh:
Ping after implementing @ehelms suggestion.

ehelms (approved):

PR #619, opened 2026-07-03
Add copy-bundle forge command for proxy deployments

ehelms:

Deploying a proxy requires manually fetching the certificate bundle, copying it to the proxy VM, and then constructing the `deploy-proxy` command with the correct bundle path, FQDN, and oauth credentials. This is error-prone and requires looking up values from multiple sources.

* Extend `forge fetch-bundle` to copy the certificate bundle to the proxy VM at `/root/<hostname>.tar.gz` when a proxy is in the inventory
* Print the full `deploy-proxy` command with the correct `--certificate-bundle`, `--foreman-fqdn`, `--oauth-consumer-key`, and `--oauth-consumer-secret` arguments
* When no proxy is in the inventory, the command behaves as before (fetch only)
* Update the certificates `extract.yml` to use `remote_src: true` on the `unarchive` task and remove `delegate_to: localhost` on the stat check, since the bundle now resides on the target host rather than the Ansible controller

Steps to reproduce:

* Deploy Foreman on quadlet: `./foremanctl deploy --initial-admin-password=changeme --tuning development --add-feature foreman-proxy`
* Generate a certificate bundle: `./foremanctl certificate-bundle proxy.example.com`
* Fetch and copy the bundle: `./forge fetch-bundle proxy.example.com`
* Verify the output shows the `deploy-proxy` command with all required arguments
* Run the printed `deploy-proxy` command and verify the proxy deploys successfully
* Verify `./forge fetch-bundle` still works without a proxy VM in the inventory

#### Checklist

### DIFF
--- a/development/playbooks/fetch-bundle/fetch-bundle.yaml
+++ b/development/playbooks/fetch-bundle/fetch-bundle.yaml
@@ -6,9 +6,36 @@
   vars_files:
     - "../../../src/vars/base.yaml"
     - "../../../src/vars/certificates.yml"
+    - "../../../src/vars/foreman.yml"
   tasks:
     - name: Fetch bundle
       ansible.builtin.fetch:
         src: "{{ certificates_ca_directory }}/bundles/{{ hostname }}.tar.gz"
         dest: "{{ obsah_state_path }}/{{ hostname }}.tar.gz"
         flat: true
+
+    - name: Store oauth credentials
+      ansible.builtin.set_fact:
+        oauth_consumer_key: "{{ foreman_oauth_consumer_key }}"
+        oauth_consumer_secret: "{{ foreman_oauth_consumer_secret }}"
+
+- name: Copy bundle to proxy
+  hosts:
+    - proxy
+  become: true
+  tasks:
+    - name: Copy bundle
+      ansible.builtin.copy:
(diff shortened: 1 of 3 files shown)


development/playbooks/copy-bundle/copy-bundle.yaml
```diff
@@ -0,0 +1,40 @@
...
+- name: Fetch generated bundle from server
+  hosts:
+    - quadlet
+  become: true
+  vars_files:
+    - "../../../src/vars/base.yaml"
+    - "../../../src/vars/certificates.yml"
+    - "../../../src/vars/foreman.yml"
+  tasks:
+    - name: Fetch bundle
```
arvind4501: We already have https://github.com/theforeman/foremanctl/blob/master/development/playbooks/fetch-bundle/fetch-bundle.yaml

arvind4501:
I tried similar thing in CI earlier https://github.com/theforeman/foremanctl/pull/571/changes/959a00186d7c253c931bb0c0912aa8ac40ef912e#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R418-R419 , also i like what @evgeni mentioned about the approch of keeping things on system and then use it to deploy to any no of nodes

PR #616, opened 2026-07-02
Fixes #615 - Add documentation for update procedure

qcjames53:
#### Why are you introducing these changes? (Problem description, related links)
https://github.com/theforeman/foremanctl/issues/615

#### What are the changes introduced in this pull request?
* Introduce user documentation to cover the manual update process for connected RPM, disconnected RPM, and from-source installs.

Note: this pull request has a few TODOs which are open questions that need to be answered by the foremanctl team. Please feel free to push to this branch to address these issues (no AI fluff please). I'll try to keep an eye on this during my paternity leave if I have any time :crossed_fingers: 

#### How to test this pull request
* Attempt the upgrade procedure from source and the portions of the RPM procedures which can be currently attempted.
* Validate the document for accuracy. Ensure it is concise, useful, and well organized.

#### Checklist
* [X] Tests added/updated (if applicable)
* [X] Documentation updated (if applicable)

PR #621, opened 2026-07-06
Fixes #615 - Add documentation for upgrade procedure

### LINKED ISSUE #615: Create documentation for manual update process
Create documentation for users to manually update foremanctl from one z-stream to the next for connected and disconnected installations. This documentation will inform how to automate the update procedure and also act as a backup in case updates cannot be automated by the next release.

aidenfine:
#615 

* Introduce user documentation to cover the manual update process for connected RPM, disconnected RPM, and from-source installs.

Note: this pull request has a few TODOs which are open questions that need to be answered by the foremanctl team. Please feel free to push to this branch to address these issues (no AI fluff please). This PR was originally opened https://github.com/theforeman/foremanctl/pull/616. The author (@qcjames53) is on paternity leave. This PR will replace the existing PR.

- Attempt the upgrade procedure from source and the portions of the RPM procedures which can be currently attempted.
- Validate the document for accuracy. Ensure it is concise, useful, and well organized.

#### Checklist

### DIFF
--- a/docs/developer/development-environment.md
+++ b/docs/developer/development-environment.md
@@ -143,6 +143,8 @@ After deployment, the environment includes:
 
 ### Upgrading from source
 
+This section is intended for users with foremanctl source cloned onto the same machine where foreman is being deployed on.
+
 In a source install, git branches control versioning instead of RPM packages. Stable branches (e.g. `2.y-stable`) track a specific Foreman Y-stream, while the `master` branch tracks nightly changes (most recent).
 
 Run all the steps as the root user in the foremanctl source directory unless otherwise specified.
--- a/docs/developer/development-environment.md
+++ b/docs/developer/development-environment.md
@@ -141,6 +141,22 @@ After deployment, the environment includes:
   - `smart-proxy` and its plugins
   - the development smart proxy registered into Foreman
 
+### Upgrading from source
+
+In a source install, git branches control versioning instead of RPM packages. Stable branches (e.g. `2.y-stable`) track a specific Foreman Y-stream, while the `master` branch tracks nightly changes (most recent).
+
+Run all the steps as the root user in the foremanctl source directory unless otherwise specified.
+
+1. Pull the latest changes using `git pull`.
+2. Begin the upgrade by stopping all Foreman processes by running `systemctl stop foreman.target`.
+3. Switch to the preferred target branch:
+    - Switch to a stable Y-stream: `git fetch origin && git checkout origin/X.y-stable`
+    - Switch to nightly: `git fetch origin && git checkout origin/master`
+4. Run upgrade tasks by re-deploying foremanctl with your customized deploy command: `foremanctl deploy [...]`. Please see [Parameters](parameters.md) for available deploy options.
+
+This final deploy command will pull new images and run all upgrade jobs required by Foreman, its dependencies, and your configured plugins. Expect this deploy to take longer than typical deploys.
+
+
 ## Architecture
 
 ### Service Integration
--- a/docs/user/upgrade.md
+++ b/docs/user/upgrade.md
@@ -2,6 +2,8 @@
 
 Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
 
+Upgrades are only supported to the next Z or Y version.
+
 For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM and disconnected RPM).
 
 ## Upgrading foremanctl from RPM install
--- a/docs/user/upgrade.md
+++ b/docs/user/upgrade.md
@@ -1,50 +1,59 @@
 # Upgrading foremanctl
 
-Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+Foremanctl releases are locked to specific Foreman images with version-compatible dependencies and plugins. Your system's installed `foreman-release` repository configuration RPM locks your system to the correct foremanctl version, which in turn will pull the correct Foreman images. For nearly all install situations, upgrading your Foreman server should be approached through upgrading `foreman-release` and `foremanctl`.
 
-Upgrades are only supported to the next Z or Y version.
+Foreman MUST be upgraded one release at a time (e.g. 3.19 -> 3.20).
 
-For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM and disconnected RPM).
+Scroll to the correct procedure below which matches your installation environment type (RPM and disconnected RPM).
 
 ## Upgrading foremanctl from RPM install
 
 All steps must be run as root user.
 
-1. Run a `foremanctl health` check to ensure a smooth upgrade.
-2. Update the Foreman repository to the next X or Y release:
+1. Run `foremanctl health` to ensure your existing Foreman server is healthy. Correct any issues before continuing.
+2. Consider backing up your Foreman environment before upgrade. See [Backup](backup.md).
+3. Manually update your Foreman Release repository to the next Foreman Y release:
     - `dnf upgrade https://yum.theforeman.org/releases/<next-version>/el9/x86_64/foreman-release.rpm`
-3. Upgrade the foremanctl package:
+    - Example: Foreman 3.19 -> 3.20 upgrades would use `dnf upgrade https://yum.theforeman.org/releases/3.20/el9/x86_64/foreman-release.rpm`.
+4. Upgrade the foremanctl package:
     - `dnf upgrade foremanctl`
-4. Run upgrade tasks by re-deploying: `foremanctl deploy`. Please see [Parameters](parameters.md) for additional deploy options.
-
-This final deploy command will pull new images and run all upgrade jobs required by Foreman, its dependencies, and your configured plugins. Expect this deploy to take longer than typical deploys.
+5. Run upgrade tasks by re-deploying your Foreman environment: 
+    - `foremanctl deploy` 
+    - Please see [Parameters](parameters.md) for additional deployment options.
+    - This deploy command will pull new images and run all upgrade jobs required by Foreman, its dependencies, and your configured plugins. Expect this deploy to take longer than typical deploys.
 
 ## Upgrading foremanctl from disconnected RPM install
 
-All below steps must be run as root user. We also recommend that a `foremanctl health` check is run before these steps.
+All steps must be run as root user.
 
-Disconnected users should also create a local repository mirror of foreman, an example of this from Red Hat is linked here https://access.redhat.com/solutions/7019225.
-
-1. Run a `foremanctl health` check to ensure a smooth upgrade.
-2. Stage the foremanctl RPM package
-    - The Foreman repository is needed for dependencies related to the foremanctl RPM.
-    - The foremanctl RPM must be available in a repository accessible to your disconnected Foreman server. Please transfer the RPM to your disconnected system via an available transport mechanism (USB drive, rsync over a bastion, etc.).
-    - The foremanctl RPM can be downloaded from `https://yum.theforeman.org`
-    - Once staged, `dnf info foremanctl` will resolve as in a connected environment.
-3. Stage the required container images
-    - On a connected machine, pull all required images with `foremanctl pull-images`.
-    - Confirm the correct images were downloaded by running `podman images` on both the connected and disconnected machines. All images from your previous-version disconnected environment should be present on the connected environment. If images are missing, ensure parameters are identical between machines.
-    - On the connected environment, run `podman save $(podman images --format "{{.Repository}}:{{.Tag}}" | tr '\n' ' ') -o <filename>.tar` to export all downloaded images as a tarball.
-    - Transfer the tar file to your disconnected environment via an available transport mechanism.
-    - Run `podman load -i <filename>.tar` to stage the required images.
-4. Complete the [Upgrading foremanctl from RPM install](#upgrading-foremanctl-from-rpm-install) section above starting from step 3 to install from locally staged packages and images.
-
-## Recovering from a failed upgrade
-
-In the event of a failed upgrade, don't panic! A failed deploy will typically reveal the details of what went wrong and can give hints regarding the nature of your issue. Here are some troubleshooting steps:
-
-#### (RPM install) `dnf upgrade foremanctl` had "Nothing to do"
-Run `dnf versionlock list` to see if your system is configured to allow X or Y version upgrades. Update the versionlock using the steps above.
-
-#### `foremanctl deploy` could not pull images from remote
-Ensure that https://quay.io is unblocked on your network. You can manually open https://quay.io/foreman/foreman in a browser to view available images.
+1. On your disconnected environment, run `foremanctl health` to ensure your existing Foreman server is healthy. Correct any issues before continuing.
+2. Consider backing up your Foreman environment before upgrade. See [Backup](backup.md).
+3. On a connected machine, install the Foreman Release repository for the next Foreman Y release:
+    - `dnf install https://yum.theforeman.org/releases/<next-version>/el9/x86_64/foreman-release.rpm`
+    - Example: Foreman 3.19 -> 3.20 upgrades would use `dnf install https://yum.theforeman.org/releases/3.20/el9/x86_64/foreman-release.rpm`.
+4. On a connected machine, create a local mirror of the foreman repository:
+    - `reposync -n -p /path/to/mirror --download-metadata --repoid=foreman`
+5. On a connected machine, install foremanctl and configure it identially to your disconnected environment:
+    - `dnf install foremanctl`
+    - Note the version of foremanctl which installed.
+    - Configure your connected foremanctl to use the same features as your disconnected environment.
+6. On a connected machine, pull required images and prepare them for transfer:
+    - `foremanctl pull-images`
+    - Confirm the correct images were downloaded by running `podman images`. All images from your previous-version disconnected environment should be present on the connected environment. If images are missing, ensure foremanctl features parameters are identical between machines.
+    - Run `podman save $(podman images --format "{{.Repository}}:{{.Tag}}" | tr '\n' ' ') -o <filename>.tar` to export all downloaded images as a tarball.
+7. Using an available transport mechanism, move the following to your disconnected environment:
+    - The foreman repo mirror and contents (contains foremanctl).
+    - The foremanctl container image tarball.
+8. On the disconnected environment, set up the repository mirrors:
+    - Copy the mirrored directory to a stable location (e.g., `/var/repos/foreman`).
+    - Redirect the existing repository configuration to use your local mirror:
+      - `dnf config-manager --setopt=foreman.baseurl=file:///var/repos/foreman --save`
+    - Verify the mirror is serving the correct package version with `dnf info foremanctl`. This version should match step 5.
+9. On the disconnected environment, stage the required container images:
+    - `podman load -i <filename>.tar`
+10. On the disconnected environment, upgrade the foremanctl package:
+    - `dnf upgrade foremanctl`
+11. Run upgrade tasks by re-deploying your Foreman environment: 
+    - `foremanctl deploy` 
+    - Please see [Parameters](parameters.md) for additional deployment options.
+    - This deploy command will pull new images and run all upgrade jobs required by Foreman, its dependencies, and your configured plugins. Expect this deploy to take longer than typical deploys.
--- a/docs/user/upgrade.md
+++ b/docs/user/upgrade.md
@@ -1,0 +1,48 @@
+# Upgrading foremanctl
+
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+
+For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM and disconnected RPM).
+
+## Upgrading foremanctl from RPM install
+
+All steps must be run as root user.
+
+1. Run a `foremanctl health` check to ensure a smooth upgrade.
+2. Update the Foreman repository to the next X or Y release:
+    - `dnf upgrade https://yum.theforeman.org/releases/<next-version>/el9/x86_64/foreman-release.rpm`
+3. Upgrade the foremanctl package:
+    - `dnf upgrade foremanctl`
+4. Run upgrade tasks by re-deploying: `foremanctl deploy`. Please see [Parameters](parameters.md) for additional deploy options.
+
+This final deploy command will pull new images and run all upgrade jobs required by Foreman, its dependencies, and your configured plugins. Expect this deploy to take longer than typical deploys.
+
+## Upgrading foremanctl from disconnected RPM install
+
+All below steps must be run as root user. We also recommend that a `foremanctl health` check is run before these steps.
+
+Disconnected users should also create a local repository mirror of foreman, an example of this from Red Hat is linked here https://access.redhat.com/solutions/7019225.
+
+1. Run a `foremanctl health` check to ensure a smooth upgrade.
+2. Stage the foremanctl RPM package
+    - The Foreman repository is needed for dependencies related to the foremanctl RPM.
+    - The foremanctl RPM must be available in a repository accessible to your disconnected Foreman server. Please transfer the RPM to your disconnected system via an available transport mechanism (USB drive, rsync over a bastion, etc.).
+    - The foremanctl RPM can be downloaded from `https://yum.theforeman.org`
+    - Once staged, `dnf info foremanctl` will resolve as in a connected environment.
+3. Stage the required container images
+    - On a connected machine, pull all required images with `foremanctl pull-images`.
+    - Confirm the correct images were downloaded by running `podman images` on both the connected and disconnected machines. All images from your previous-version disconnected environment should be present on the connected environment. If images are missing, ensure parameters are identical between machines.
+    - On the connected environment, run `podman save $(podman images --format "{{.Repository}}:{{.Tag}}" | tr '\n' ' ') -o <filename>.tar` to export all downloaded images as a tarball.
+    - Transfer the tar file to your disconnected environment via an available transport mechanism.
+    - Run `podman load -i <filename>.tar` to stage the required images.
+4. Complete the [Upgrading foremanctl from RPM install](#upgrading-foremanctl-from-rpm-install) section above starting from step 3 to install from locally staged packages and images.
+
+## Recovering from a failed upgrade
+
+In the event of a failed upgrade, don't panic! A failed deploy will typically reveal the details of what went wrong and can give hints regarding the nature of your issue. Here are some troubleshooting steps:
+
+#### (RPM install) `dnf upgrade foremanctl` had "Nothing to do"
+Run `dnf versionlock list` to see if your system is configured to allow X or Y version upgrades. Update the versionlock using the steps above.
+
+#### `foremanctl deploy` could not pull images from remote
+Ensure that https://quay.io is unblocked on your network. You can manually open https://quay.io/foreman/foreman in a browser to view available images.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,3 +46,23 @@ AI agent specifications (rules, skills, agent personas) live under `.agents/`.
 - [Check Roles](docs/developer/checks.md) — check role catalog and integration patterns
 - [Testing](docs/developer/testing.md) — test infrastructure, fixtures, patterns
 - [Parameters](docs/user/parameters.md) — installation parameter map; update when adding parameters
+Developer docs:
+- [Check roles](docs/developer/checks.md) - How to integrate check roles; update as checks are created/modified
+- [Container Image Builds](docs/developer/container-image-builds.md) - Info on image naming, registries
+- [Deployment Architecture](docs/developer/deployment.md)
+- [Development Environment](docs/developer/development-environment.md) - Dev environment setup with Foreman from source
+- [How to Add a Feature](docs/developer/how-to-add-a-feature.md) - End-to-end feature development
+- [Playbooks and Roles](docs/developer/playbooks-and-roles.md) - Playbook structure, naming, metadata
+- [Testing](docs/developer/testing.md) - Additional info on test infrastructure, fixtures, patterns
+
+User docs:
+- [Backup](docs/user/backup.md) - How to back up your data
+- [Certificates](docs/user/certificates.md) - Overview of certificate sources
+- [Parameters](docs/user/parameters.md) - Map of Foreman installation parameters; update as parameters are created/modified
+- [Upgrade](docs/user/upgrade.md) - How to upgrade your Foreman server through foremanctl
+
+- [CONTRIBUTING](CONTRIBUTING.md) - How to contribute
+- [Development](DEVELOPMENT.md) - Foremanctl development overview
+- [IOP](docs/iop.md) - Overview of insights on premise
+- [Migration Guide](docs/migration-guide.md) - Migrating from foreman-installer to foremanctl
+- [Release](RELEASE.md) - Info on Foremanctl releases


docs/user/upgrade.md
```diff
@@ -0,0 +1,65 @@
+# Upgrading foremanctl
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins. Specific versions of foremanctl always map to the same version of Foreman, regardless of parameter configuration:
+
+- 1.2.z+ -> Foreman 3.17
+- 2.0.z -> Foreman 3.18
+- 2.2.z+ -> Foreman 3.19
+
+TODO: This table needs an explanation for the Y stream schema. Ask Evgeni or Ewoud about this. Add a resource where users can match a Foreman version to a foremanctl version.
```
aidenfine: @ehelms on the previous PR you mentioned "This assumes we intend users to use it from git for production rather than the RPM. That's a design question we haven't tried to answer."

Does it make sense to leave this part out or maybe move this into the developer docs since the rest of this doc assumes installing/upgrading from RPM.

aidenfine:
Based on previous comments on https://github.com/theforeman/foremanctl/pull/616. I added the health check mention in `docs/user/upgrade.md` and moved **Upgrading from source** section into `docs/developer/development-environment.md`

docs/developer/development-environment.md
```diff
@@ -141,6 +122,23 @@ After deployment, the environment includes:
   - `smart-proxy` and its plugins
   - the development smart proxy registered into Foreman
 
+### Upgrading from source
+
+In a source install, git branches control versioning instead of RPM packages. Stable branches (e.g. `2.y-stable`) track a specific Foreman Y-stream, while the `master` branch tracks nightly changes (most recent).
+
+All below steps must be run as root user.
```
ehelms:
```suggestion
All steps must be run as the root user.
```

docs/developer/development-environment.md
```diff
@@ -141,6 +122,23 @@ After deployment, the environment includes:
...
+
+1. Please take a moment to determine your preferred Foreman version / git branch. Cross-reference with the Foreman version table above.
```
ehelms: This makes it sound like the user can upgrade from any version to any other version, which we do not test.

docs/developer/development-environment.md
```diff
@@ -141,6 +122,23 @@ After deployment, the environment includes:
   - `smart-proxy` and its plugins
   - the development smart proxy registered into Foreman
 
+### Upgrading from source
+
+In a source install, git branches control versioning instead of RPM packages. Stable branches (e.g. `2.y-stable`) track a specific Foreman Y-stream, while the `master` branch tracks nightly changes (most recent).
```
ehelms, line 146: This procedure needs to be specific about where foremanctl is being run from since from a git checkout you can run it from your Laptop pointed at a VM. Or you can run it from a git checkout on the VM.

docs/developer/development-environment.md
aidenfine: We support y+1 and y+2 correct? Ex: 2.1 -> 2.2 or 2.3.

ehelms: We support sequential updates only AFAIK.

docs/user/upgrade.md
ehelms: I think the latter makes sense.

docs/developer/development-environment.md
aidenfine: @ehelms I added "These steps assume that foremanctl is being ran from the same machine that the source lives on." on 137.

docs/developer/development-environment.md
```diff
@@ -141,6 +122,31 @@ After deployment, the environment includes:
...
+
+Specific versions of foremanctl always map to the same version of Foreman, regardless of parameter configuration:
+
+- 1.2.z+ -> Foreman 3.17
+- 2.0.z+ -> Foreman 3.18
+- 2.2.z+ -> Foreman 3.19
+
+TODO: This table needs an explanation for the Y stream schema. Ask Evgeni or Ewoud about this. Add a resource where users can match a Foreman version to a foremanctl version.
```
ehelms: Noting this TODO here still.

docs/user/upgrade.md
```diff
@@ -0,0 +1,59 @@
+# Upgrading foremanctl
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+
+For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM, disconnected RPM, source).
+
+
+
```
ehelms, line 6: A couple extra empty lines here.

docs/user/upgrade.md
```diff
@@ -0,0 +1,59 @@
+# Upgrading foremanctl
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
```
ehelms, line 3:
```suggestion

Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
```

docs/user/upgrade.md
```diff
@@ -0,0 +1,59 @@
...
+1. We recommend setting up a dnf versionlock for foremanctl if control over Foreman versioning is critical to your use case. If applicable, please take a moment to review your current versionlock configuration:
+    - `dnf versionlock list` - View the current versionlock configuration
+    - `dnf versionlock add 'foremanctl-X.*'` - Lock upgrades to a certain X-stream (replace with exact X version).
+    - `dnf versionlock add 'foremanctl-X.Y.*'` - Lock upgrades to a certain Y-stream (replace with exact X.Y version).
+    - `dnf versionlock delete foremanctl` - Remove any foremanctl versionlock.
+2. We recommend a full foremanctl backup before all upgrade operations. Run `foremanctl backup <filepath for backup>`.  Please see [Backup](backup.md) for more information on this process.
+3. Begin the upgrade by stopping all Foreman processes with `systemctl stop foreman.target`.
+4. Upgrade foremanctl using dnf:
+    - Upgrade to latest version (or latest allowed by versionlock): `dnf upgrade foremanctl`
+    - Upgrade to specific version: `dnf upgrade foremanctl-X.Y.Z`
```
ehelms: I vote we drop this line. This should be a special case, not something we encourage users to do as we won't be testing this scenario.

docs/user/upgrade.md
```diff
@@ -0,0 +1,59 @@
...
+5. Run upgrade tasks by re-deploying foremanctl with your customized deploy command: `foremanctl deploy [...]`. Please see [Parameters](parameters.md) for available deploy options.
```
ehelms: The `customized deploy command` part I think can lead to confusion or is an advanced upgrade path that I don't think we should encourage. We generally want user upgrade path to be just run deploy, not run it with extra options.

docs/user/upgrade.md
```diff
@@ -0,0 +1,59 @@
+# Upgrading foremanctl
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+
+For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM, disconnected RPM, source).
+
+
+
+## Upgrading foremanctl from RPM install
+
+All below steps must be run as root user. We also recommend that a `foremanctl health` check is run before these steps.
```
ehelms:
```suggestion
All steps must be run as root user. We also recommend that a `foremanctl health` check is run before these steps.
```

docs/user/upgrade.md
```diff
@@ -0,0 +1,59 @@
...
+    - Upgrade to specific version: `dnf upgrade foremanctl-X.Y.Z`
+5. Run upgrade tasks by re-deploying foremanctl with your customized deploy command: `foremanctl deploy [...]`. Please see [Parameters](parameters.md) for available deploy options.
+
+This final deploy command will pull new images and run all upgrade jobs required by Foreman, its dependencies, and your configured plugins. Expect this deploy to take longer than typical deploys.
+
+
+
+## Upgrading foremanctl from disconnected RPM install
+
+All below steps must be run as root user.
```
ehelms:
```suggestion
All steps must be run as root user.
```

ehelms: Mention health check here also?

docs/user/upgrade.md
```diff
@@ -0,0 +1,59 @@
...
+
+1. Stage the foremanctl RPM package
+    - The foremanctl RPM must be available in a repository accessible to your disconnected Foreman server. Please transfer the RPM to your disconnected system via an available transport mechanism (USB drive, rsync over a bastion, etc.).
```
ehelms, line 27: Is the RPM enough? Or will they need the Foreman repository for dependencies?

ehelms:
Minor wording mismatch -- the PR title and commits mention `update` which is different than `upgrade` in our nomenclature. I think we should try to be clear about which one we are talking about. Hopefully the procedure for the two becomes the same, but for now I like the differentiation:

1. upgrade: X.Y -> X.Y+1 
2. update: X.Y.Z -> X.Y.Z+1

`upgrade` requires things like switching repositories, different version tags on the images and possibly "bigger" upgrade actions happening. Where as "update" should feel smaller, and more direct to achieve for the user.

docs/user/upgrade.md
```diff
@@ -0,0 +1,59 @@
...
+4. Upgrade foremanctl using dnf:
+    - Upgrade to latest version (or latest allowed by versionlock): `dnf upgrade foremanctl`
+    - Upgrade to specific version: `dnf upgrade foremanctl-X.Y.Z`
+5. Run upgrade tasks by re-deploying foremanctl with your customized deploy command: `foremanctl deploy [...]`. Please see [Parameters](parameters.md) for available deploy options.
+
+This final deploy command will pull new images and run all upgrade jobs required by Foreman, its dependencies, and your configured plugins. Expect this deploy to take longer than typical deploys.
+
+
+
+## Upgrading foremanctl from disconnected RPM install
```
ianballou, line 19: "RPM install" makes me think un-containerized Foreman and "upgrade" typically means a y-stream upgrade in our historic verbiage.

ianballou (commented):
Just one note so far, thanks for putting all this together:

docs/user/upgrade.md
aidenfine: If I recall you should also need the repo for the deps, let me test right now.

docs/user/upgrade.md
aidenfine: What do you think of the way it is worded now?

aidenfine:
> Minor wording mismatch -- the PR title and commits mention `update` which is different than `upgrade` in our nomenclature. I think we should try to be clear about which one we are talking about. Hopefully the procedure for the two becomes the same, but for now I like the differentiation:
> 
> 1. upgrade: X.Y -> X.Y+1
> 2. update: X.Y.Z -> X.Y.Z+1
> 
> `upgrade` requires things like switching repositories, different version tags on the images and possibly "bigger" upgrade actions happening. Where as "update" should feel smaller, and more direct to achieve for the user.

Updated commits and PR title.

docs/user/upgrade.md
aidenfine: Do you have a suggestion for what this should be?

docs/user/upgrade.md
aidenfine: @ehelms Take a look I added an extra bullet point what do you think?

docs/developer/development-environment.md
```diff
@@ -141,6 +122,29 @@ After deployment, the environment includes:
...
+
+In a source install, git branches control versioning instead of RPM packages. Stable branches (e.g. `2.y-stable`) track a specific Foreman Y-stream, while the `master` branch tracks nightly changes (most recent).
+
+Specific versions of foremanctl always map to the same version of Foreman, regardless of parameter configuration:
+
+- 1.2.z+ -> Foreman 3.17
+- 2.0.z+ -> Foreman 3.18
+- 2.2.z+ -> Foreman 3.19
+
+All steps must be run as the root user. These steps assume that foremanctl is being ran from the same machine that the source lives on.
```
ehelms:
```suggestion
All steps must be run as the root user. These steps assume that foremanctl is being ran from the same machine that Foreman was deployed on.
```

docs/user/upgrade.md
```diff
@@ -0,0 +1,51 @@
+# Upgrading foremanctl
+
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+
+For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM, disconnected RPM, source).
+
+## Upgrading foremanctl from RPM install
```
ehelms, line 7: This procedure doesn't explain to the user where to get the new version of `foremanctl`. I would expect it to include managing the Foreman repository. I think the title here is also confusing as it talks about "upgrading foremanctl". But upgrading foremanctl inherently means also updating or upgrading Foreman. Right now, we do not support one without the other. Therefore, I think the focus needs to be on "How do I upgrade my installation from X.Y to X.Y+1"

docs/user/upgrade.md
ehelms: See https://github.com/theforeman/foremanctl/pull/621/changes#r3639168322 

I think this procedure is a little better, but we shouldn't think about foremanctl by itself -- it has dependencies. The repositories matter.

ianballou: For the naming of this section, I might suggest something like 

```suggestion
## Updating the foremanctl RPM
```

docs/user/upgrade.md
ianballou: @aidenfine it might help to run through the upgrading and updating guides in the existing documentation to understand how we handle defining repositories. I _think_ the general strategy for how we manage the repo will stay the same.

docs/user/upgrade.md
```diff
@@ -0,0 +1,51 @@
...
+In the event of a failed upgrade, don't panic! A failed deploy will typically reveal the details of what went wrong and can give hints regarding the nature of your issue. Here are some troubleshooting steps:
+
+#### (RPM install) `dnf upgrade foremanctl` had "Nothing to do"
+Run `dnf versionlock list` to see if your system is configured to allow X or Y version upgrades. Update the versionlock using the steps above.
+
+#### `foremanctl deploy` could not pull images from remote
+Ensure that https://quay.io is unblocked on your network. You can manually open https://quay.io/foreman/foreman in a browser to view available images.
+
+#### I'd like to roll back to a known working version (RPM install)
+`dnf downgrade foremanctl-X.Y.Z` will roll back your foremanctl install to the requested version. Please follow instructions in our [Restore Guide](restore.md) to restore your system to a working state.
```
ianballou: I'm not certain that this is safe, an older version of foremanctl might clash with a newer versions context files. Also, Foreman has never supported downgrades in the past with `foreman-installer`.

docs/user/upgrade.md
```diff
@@ -0,0 +1,51 @@
...
+    - Once staged, `dnf info foremanctl` will resolve as in a connected environment.
+2. Stage the required container images
+    - On a connected machine, pull all required images with `foremanctl pull-images`.
+    - Confirm the correct images were downloaded by running `podman images` on both the connected and disconnected machines. All images from your previous-version disconnected environment should be present on the connected environment. If images are missing, ensure parameters are identical between machines.
+    - On the connected environment, run `podman save $(podman images --format "{{.Repository}}:{{.Tag}}" | tr '\n' ' ') -o <filename>.tar` to export all downloaded images as a tarball.
+    - Transfer the tar file to your disconnected environment via an available transport mechanism.
+    - Run `podman load -i <filename>.tar` to stage the required images.
+3. Complete the [Upgrading foremanctl from RPM install](#upgrading-foremanctl-from-rpm-install) section above to install from locally staged packages and images.
+
+## Recovering from a failed upgrade
```
ianballou, line 38: The troubleshooting section should probably have a section about varying Ansible errors, after which a user should probably open up an issue on GitHub or a post at community.theforeman.org.

docs/developer/development-environment.md
```diff
@@ -141,6 +122,29 @@ After deployment, the environment includes:
...
+
+Specific versions of foremanctl always map to the same version of Foreman, regardless of parameter configuration:
+
+- 1.2.z+ -> Foreman 3.17
+- 2.0.z+ -> Foreman 3.18
+- 2.2.z+ -> Foreman 3.19
+
+All steps must be run as the root user. These steps assume that foremanctl is being ran from the same machine that Foreman was deployed on.
+
+1. Please take a moment to determine your preferred Foreman version / git branch, we only support sequential updates.
```
ianballou: We don't support anything on development environments, so people can `git pull` at will.

docs/developer/development-environment.md
```diff
@@ -141,6 +122,29 @@ After deployment, the environment includes:
...
+2. We recommend a full foremanctl backup before all upgrade operations. Run `foremanctl backup <filepath for backup>`. Please see [Backup](backup.md) for more information on this process.
```
ianballou: While it's a good practice, I'm not sure how many people will be taking backups of their development environment. Actually, it makes me wonder if they're somehow incompatible with backup/restore.

docs/developer/development-environment.md
```diff
@@ -141,6 +122,29 @@ After deployment, the environment includes:
...
 
+### Upgrading from source
+
+In a source install, git branches control versioning instead of RPM packages. Stable branches (e.g. `2.y-stable`) track a specific Foreman Y-stream, while the `master` branch tracks nightly changes (most recent).
+
+Specific versions of foremanctl always map to the same version of Foreman, regardless of parameter configuration:
+
+- 1.2.z+ -> Foreman 3.17
+- 2.0.z+ -> Foreman 3.18
+- 2.2.z+ -> Foreman 3.19
```
ianballou: Maintaining this mapping might grow into a maintenance headache.

docs/user/upgrade.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Upgrading foremanctl from RPM install
+
+All steps must be run as root user. We also recommend that a `foremanctl health` check is run before these steps.
+
+1. We recommend setting up a dnf versionlock for foremanctl if control over Foreman versioning is critical to your use case. If applicable, please take a moment to review your current versionlock configuration:
+    - `dnf versionlock list` - View the current versionlock configuration
+    - `dnf versionlock add 'foremanctl-X.*'` - Lock upgrades to a certain X-stream (replace with exact X version).
+    - `dnf versionlock add 'foremanctl-X.Y.*'` - Lock upgrades to a certain Y-stream (replace with exact X.Y version).
+    - `dnf versionlock delete foremanctl` - Remove any foremanctl versionlock.
```
ianballou: Upstream we "lock" users into specific versions of packages by the repository. That way, a user can `dnf upgrade` without fear of pulling in something strange. It's one reason why we don't support installations with EPEL enabled on them.

If we continue to guide users to enable only the correct Foreman repository, they should not have to worry about `versionlock`. Plus - we don't really want to "recommend" specific upgrade setups. If users are doing different things, it can make supporting them difficult.

docs/user/upgrade.md
```diff
@@ -0,0 +1,50 @@
...
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+
+For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM, disconnected RPM, source).
+
+## Upgrading foremanctl from RPM install
+
+All steps must be run as root user. We also recommend that a `foremanctl health` check is run before these steps.
+
+1. Update the Foreman repository to the target version:
+    - `dnf install https://yum.theforeman.org/releases/<target-version>/el9/x86_64/foreman-release.rpm`
```
ianballou:
```suggestion
    - `dnf upgrade https://yum.theforeman.org/releases/<target-version>/el9/x86_64/foreman-release.rpm`
```

Just to be more consistent with the existing docs.

docs/user/upgrade.md
```diff
@@ -0,0 +1,50 @@
...
+
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+
+For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM, disconnected RPM, source).
+
+## Upgrading foremanctl from RPM install
+
+All steps must be run as root user. We also recommend that a `foremanctl health` check is run before these steps.
+
+1. Update the Foreman repository to the target version:
```
ianballou:
```suggestion
1. Update the Foreman repository to the target version for X or Y stream upgrades:
```

This alerts users that they do not need to update the repo for patch-level updates.

docs/user/upgrade.md
ianballou: I would recommend telling users to stage the entire repository locally. `python3.12-obsah` is a dependency of foremanctl for example, and users may want to update it even if the update isn't a hard requirement. One reason could be for a CVE, for example.

docs/user/upgrade.md
aidenfine: Is there any link to common Ansible errors that may occur?

aidenfine:
@ianballou Are you talking about the Foreman repository?

ianballou:
> @ianballou Are you talking about the Foreman repository?

Yep! Like https://yum.theforeman.org/releases/3.19/el9/x86_64/. Where users will find the foremanctl RPM. The repository contains the dependencies that foremanctl needs.

Check this out:

```
# rpm -q --requires foremanctl 
/usr/bin/bash
/usr/bin/env
/usr/bin/python3
config(foremanctl) = 2.3.0-1.el9sat
python3.12-obsah >= 1.8.1
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsZstd) <= 5.4.18-1
```

The basic deps likely come from the main OS repos. But then there's python3.12-obsah, which we package up. But anyway - when a user is installing Foreman we want them to have the whole repo. For disconnected users in Satellite, the full setup typically looks like having a connected Satellite host the repositories and the disconnected Satellite consuming from it in some fashion - possibly via import/export. The foremanctl docs don't need to fully define this whole setup though, it could be as basic as saying the user needs to bundle up the repository in some manner and present it to the host.

It might be acceptable if we identify all the dependencies and tell them to copy them over, but it's technically safer to use the entire repo where `foremanctl` lives.

I think the tricky thing here is that "disconnected" installation is truly a bit complicated and I don't think we've really documented it before for regular Foreman or Katello, just Satellite.

So - with all this said, my recommendation would be to suggest that the disconnected user creates a local repository mirror of the foreman distribution of their choice, but leave the complexities of how to their choosing. See https://access.redhat.com/solutions/7019225 for an example of how Red Hat suggests users make local mirrors of RHEL repos.

aidenfine:
@ianballou I added the suggestion let me know what you think.

qcjames53:
Thank you guys for all the movement on this :) 

Something I think is worth bringing up is that downgrading rpm version troubleshooting section was intended for people with a full backup. If you upgrade from x.1 to x.2 and something breaks, there's a path to restore your working x.1 backup. It could use better wording.

Another thought: is there a way to force upgrade jobs to run sequentially if a user goes from foremanctl x.1 to x.10 in one dnf command? Eric mentioned we require version bumping foreman one release at a time so if we can't automate that in packaging there should be clear instructions to the user here.

ianballou:
Hi @qcjames53 !

> Thank you guys for all the movement on this :)
> 
> Something I think is worth bringing up is that downgrading rpm version troubleshooting section was intended for people with a full backup. If you upgrade from x.1 to x.2 and something breaks, there's a path to restore your working x.1 backup. It could use better wording.

I think we need to test  that the downgrade is possible, with containers it might be easier to support. One question that comes to mind, for example, is what happens if the new deployment pull some newer container images. If you downgrade and run `foremanctl restore`, will the older container images be downgraded too?

> Another thought: is there a way to force upgrade jobs to run sequentially if a user goes from foremanctl x.1 to x.10 in one dnf command? Eric mentioned we require version bumping foreman one release at a time so if we can't automate that in packaging there should be clear instructions to the user here.

+1 to this, we need a note saying that users must upgrade to the next Z or Y version and it's the only upgrade path we support.

docs/user/upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+
+For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM, disconnected RPM, source).
+
+## Upgrading foremanctl from RPM install
+
+All steps must be run as root user. We also recommend that a `foremanctl health` check is run before these steps.
+
+1. Update the Foreman repository to the target version for the X or Y stream upgrades:
```
ianballou:
```suggestion
1. Update the Foreman repository to the next X or Y release:
```

docs/user/upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+    - `dnf upgrade https://yum.theforeman.org/releases/<target-version>/el9/x86_64/foreman-release.rpm`
```
ianballou:
```suggestion
    - `dnf upgrade https://yum.theforeman.org/releases/<next-version>/el9/x86_64/foreman-release.rpm`
    - Sequential upgrades of X (3.19->5.0) or Y versions (3.18->3.19) of Foreman are required.
```

docs/user/upgrade.md
```diff
@@ -0,0 +1,46 @@
+# Upgrading foremanctl
+
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+
+For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM, disconnected RPM, source).
```
aneta-petrova: Upgrading from source is in a different file and in deverloper-facing documentation. Should it be listed here?

docs/developer/development-environment.md
```diff
@@ -141,6 +141,22 @@ After deployment, the environment includes:
   - `smart-proxy` and its plugins
   - the development smart proxy registered into Foreman
 
+### Upgrading from source
+
+In a source install, git branches control versioning instead of RPM packages. Stable branches (e.g. `2.y-stable`) track a specific Foreman Y-stream, while the `master` branch tracks nightly changes (most recent).
+
+All steps must be run as the root user. These steps assume that foremanctl is being ran from the same machine that Foreman was deployed on.
```
qcjames53, line 148:
```suggestion
Run all steps as the root user in the foremanctl source directory unless otherwise specified.
```

docs/developer/development-environment.md
```diff
@@ -141,6 +141,22 @@ After deployment, the environment includes:
...
+
+1. Pull the latest changes using `git pull`.
+2. Begin the upgrade by stopping all Foreman processes with `systemctl stop foreman.target`.
```
qcjames53, line 151:
```suggestion
2. Begin the upgrade by stopping all Foreman processes by running `systemctl stop foreman.target` on your deployment environment.
```

docs/user/upgrade.md
```diff
@@ -0,0 +1,46 @@
+# Upgrading foremanctl
+
+Foremanctl releases bundle specific versions of Foreman with version-compatible dependencies and plugins.
+
+For nearly all install situations, upgrading your Foreman server should be approached through upgrading foremanctl. Scroll to the correct procedure below which matches your installation environment type (RPM and disconnected RPM).
```
qcjames53, line 5:
```suggestion
Foremanctl releases bundle-specific versions of Foreman with version-compatible dependencies and plugins. Upgrading your containerized Foreman server should be approached through upgrading foremanctl. Scroll to the matching installation environment type procedure below (internet-connected RPM install or disconnected RPM install).
```

qcjames53: @evgeni  @ehelms Sorry to pull you guys in. I'm not entirely sure this is the correct language and I'd like to ensure I'm understanding everything correctly before suggesting more changes to the procedure below.

Please let me know if this is an incorrect understanding of our current versioning strategy:

- Foremanctl's version controls the image that is pulled for Foreman's components. On foremanctl deploy, a container image will be pulled from quay and migrations will run to bump the old N-2 or N-1 install to N. This all works today.
- We control foremanctl's version through the currently installed foreman-release.rpm that users manually install to bump versions? Not a subscribed repository or anything but the actual foreman rpm? 
    - Is there a dependency setup on foremanctl such that the latest version supporting foreman-release-X.Y is what ends up on the machine? Or do users need to manually coordinate foremanctl version still?
    - I'm definitely misunderstanding this but why are users installing the foreman RPM then downloading a quay image that contains the same foreman rpm?

qcjames53: Sorry if this was addressed in a prior conversation and I missed it. I've been out for a while.

docs/user/upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Upgrading foremanctl from RPM install
+
+All steps must be run as root user. We also recommend that a `foremanctl health` check is run before these steps.
```
qcjames53, line 9: I suggest moving `foremanctl health` to step 1 below instead of having it here.

docs/user/upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+
+1. Update the Foreman repository to the next X or Y release:
+    - `dnf upgrade https://yum.theforeman.org/releases/<next-version>/el9/x86_64/foreman-release.rpm`
```
qcjames53, line 12: See https://github.com/theforeman/foremanctl/pull/621/changes#r3738946826 - I don't know enough about this to know how to fix the steps yet. This feels backwards to me though.

docs/user/upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+    - Once staged, `dnf info foremanctl` will resolve as in a connected environment.
+2. Stage the required container images
+    - On a connected machine, pull all required images with `foremanctl pull-images`.
+    - Confirm the correct images were downloaded by running `podman images` on both the connected and disconnected machines. All images from your previous-version disconnected environment should be present on the connected environment. If images are missing, ensure parameters are identical between machines.
+    - On the connected environment, run `podman save $(podman images --format "{{.Repository}}:{{.Tag}}" | tr '\n' ' ') -o <filename>.tar` to export all downloaded images as a tarball.
+    - Transfer the tar file to your disconnected environment via an available transport mechanism.
+    - Run `podman load -i <filename>.tar` to stage the required images.
+3. Complete the [Upgrading foremanctl from RPM install](#upgrading-foremanctl-from-rpm-install) section above starting from step 3 to install from locally staged packages and images.
+
+## Recovering from a failed upgrade
```
qcjames53, line 38: Now that backup / deploy are merged I would love if we could add a section on restoring to N-1 like we talked about above. @aidenfine would you like me to work on this at some point?

qcjames53 (changes requested):
Aiden, thanks so much for working so hard on this. I have way too many open questions right now but I'm confident we can get this through in due time.

docs/user/upgrade.md
aidenfine: Yeah is this something we want to include in this PR or a separate PR? If you want to include it here, I can give you permissions on this branch.

qcjames53: Sure I can take care of adding that, no worries.

aidenfine:
@ehelms Can you please take a look at the updates.

docs/user/upgrade.md
qcjames53: Update: I think I tracked down answers to most of these questions. I had a flawed understanding of foreman-release. Will push changes shortly!

qcjames53:
I cleaned up the steps a little and rolled all the disconnected steps from the Red Hat docs Ian mentioned into this file itself. It should read a lot simpler now.

docs/user/upgrade.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. On your disconnected environment, run `foremanctl health` to ensure your existing Foreman server is healthy. Correct any issues before continuing.
+2. Consider backing up your Foreman environment before upgrade. See [Backup](backup.md).
+3. On a connected machine, install the Foreman Release repository for the next Foreman Y release:
+    - `dnf install https://yum.theforeman.org/releases/<next-version>/el9/x86_64/foreman-release.rpm`
+    - Example: Foreman 3.19 -> 3.20 upgrades would use `dnf install https://yum.theforeman.org/releases/3.20/el9/x86_64/foreman-release.rpm`.
+4. On a connected machine, create local mirrors of the foreman and foreman-plugins repositories:
+    - `reposync -n -p /path/to/mirror --download-metadata --repoid=foreman`
+    - `reposync -n -p /path/to/mirror --download-metadata --repoid=foreman-plugins`
+5. On a connected machine, install foremanctl and configure it identially to your disconnected environment:
```
ianballou, line 36:
```suggestion
5. On a connected machine, install foremanctl and configure it identically to your disconnected environment:
```

docs/user/upgrade.md
```diff
@@ -0,0 +1,61 @@
...
+
+Foremanctl releases are locked to specific Foreman images with version-compatible dependencies and plugins. Your system's installed `foreman-release` repository configuration RPM locks your system to the correct foremanctl version, which in turn will pull the correct Foreman images. For nearly all install situations, upgrading your Foreman server should be approached through upgrading `foreman-release` and `foremanctl`.
+
+Foreman MUST be upgraded one release at a time (e.g. 3.19 -> 3.20).
+
+Scroll to the correct procedure below which matches your installation environment type (RPM and disconnected RPM).
+
+## Upgrading foremanctl from RPM install
+
+Please run all steps as root user.
```
aidenfine:
```suggestion
All steps must be run as root user.
```

docs/user/upgrade.md
```diff
@@ -0,0 +1,61 @@
...
+4. Upgrade the foremanctl package:
+    - `dnf upgrade foremanctl`
+5. Run upgrade tasks by re-deploying your Foreman environment: 
+    - `foremanctl deploy` 
+    - Please see [Parameters](parameters.md) for additional deployment options.
+    - This deploy command will pull new images and run all upgrade jobs required by Foreman, its dependencies, and your configured plugins. Expect this deploy to take longer than typical deploys.
+
+## Upgrading foremanctl from disconnected RPM install
+
+Please run all steps as root user.
```
aidenfine: Could we change to "All steps must be run as root user."

docs/user/upgrade.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. On your disconnected environment, run `foremanctl health` to ensure your existing Foreman server is healthy. Correct any issues before continuing.
+2. Consider backing up your Foreman environment before upgrade. See [Backup](backup.md).
+3. On a connected machine, install the Foreman Release repository for the next Foreman Y release:
+    - `dnf install https://yum.theforeman.org/releases/<next-version>/el9/x86_64/foreman-release.rpm`
+    - Example: Foreman 3.19 -> 3.20 upgrades would use `dnf install https://yum.theforeman.org/releases/3.20/el9/x86_64/foreman-release.rpm`.
+4. On a connected machine, create local mirrors of the foreman and foreman-plugins repositories:
```
ianballou: Why do `foremanctl` users need the `foreman-plugins` repo?

qcjames53:
Thanks, Aiden! I changed those two lines.

docs/user/upgrade.md
qcjames53: You know, good point. It's not. It was one of the repos in foreman-release so I included it. Will remove.

qcjames53: This is corrected now :+1:

ianballou:
I think we may want a review from @ShimShtein or @stejskalleos to ensure this meshes with the installation process they're working on. It's fine if it's different here because Katello isn't relied on, but the methods should be similar so users aren't confused.

The steps here are what should end up in docs.theforeman.org for containerized Foreman.

docs/developer/development-environment.md
```diff
@@ -176,6 +141,22 @@ After deployment, the environment includes:
   - `smart-proxy` and its plugins
   - the development smart proxy registered into Foreman
 
+### Upgrading from source
+
+In a source install, git branches control versioning instead of RPM packages. Stable branches (e.g. `2.y-stable`) track a specific Foreman Y-stream, while the `master` branch tracks nightly changes (most recent).
+
+Run all the steps as the root user in the foremanctl source directory unless otherwise specified.
```
ehelms, line 150: This needs to note that these steps assume foremanctl is cloned on the same machine foreman is deployed on.

ehelms:
Other than my nitpick, I think this reads good as a start.

ehelms (approved):

qcjames53:
Pushed with reword to re-run automation

qcjames53 (approved):

ehelms:
Merging as it's only docs changes which does not affect the current Candlepin failures

PR #622, opened 2026-07-06
Fix Candlepin service failure with Tomcat 9

jeremylenz:

The candlepin service fails to start on deployments using the nightly candlepin container image. The failure occurs because `server.xml` references `org.apache.catalina.core.OpenSSLLifecycleListener`, which is a Tomcat 10.1+ class, but the current nightly candlepin image ships Tomcat 9.0.87.

* Added tasks to query the Tomcat version from the container image at deploy time
* Made the OpenSSLLifecycleListener conditional on Tomcat major version >= 10
* Updated connector comment to be version-agnostic

Steps to reproduce:

* Deploy with the nightly candlepin image (which has Tomcat 9)
* Verify candlepin service starts successfully without errors
* Check that server.xml does not include OpenSSLLifecycleListener

#### Checklist

### DIFF
--- a/src/roles/candlepin/tasks/main.yml
+++ b/src/roles/candlepin/tasks/main.yml
@@ -28,6 +28,19 @@
   notify:
     - Restart candlepin
 
+- name: Query Tomcat version from container image
+  ansible.builtin.command:
+    cmd: >-
+      podman run --rm --entrypoint rpm
+      {{ candlepin_container_image }}:{{ candlepin_container_tag }}
+      -q --queryformat '%{VERSION}' tomcat
+  register: _candlepin_tomcat_version
+  changed_when: false
+
+- name: Set Tomcat major version fact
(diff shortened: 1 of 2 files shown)


ehelms:
> The candlepin service fails to start on deployments using the nightly candlepin container image. The failure occurs because `server.xml` references `org.apache.catalina.core.OpenSSLLifecycleListener`, which is a Tomcat 10.1+ class, but the current nightly candlepin image ships Tomcat 9.0.87.

I think your container image might just be old. I show `tomcat-9.0.117-1.el9.noarch` which has the FFM support. Also, we haven't seen any CI tests failing.

jeremylenz:
That was it, I just had to pull a new image.

This led me to discover that there is only `foremanctl pull-images`, and no `forge pull-images`. raised a PR: https://github.com/theforeman/foremanctl/pull/631

PR #624, opened 2026-07-07
Send httpd vhost logs to journald via systemd-cat

Alleny244:

foremanctl deploys httpd as a host service, but the Foreman vhosts currently write access and error logs to files under `/var/log/httpd/`. This does not align with the containerized logging model, where logs should be written to `stdout`/`stderr` and collected by `journald`.

On EL9, logging directly to `/dev/stdout` was unreliable under systemd. Using `systemd-cat` with dedicated journal tags provides reliable integration with `journald`.

---

- Update `foreman-vhost.conf.j2` and `foreman-ssl-vhost.conf.j2` to pipe access and error logs through `systemd-cat` instead of writing to `/var/log/httpd/`.
- Add integration tests to verify:
  - vhost configuration uses `systemd-cat`
  - HTTPS access logs are written to the system journal

---

## How to test this pull request

1. Deploy foremanctl from this branch.
2. Verify the generated Apache configuration uses `systemd-cat`:
   ```bash
   grep -E 'ErrorLog|CustomLog' /etc/httpd/conf.d/foreman-ssl.conf
   ```
3. Generate an HTTPS request:
   ```bash
   curl -sk https://localhost/api/v2/ping
   ```
4. Confirm the request appears in the journal:
   ```bash
   sudo journalctl -t httpd-ssl-access --since "1 min ago" --no-pager
   ```
5. Verify the old log files under `/var/log/httpd/` are no longer being updated.
6. Run the new integration tests.

---

## Checklist

- [x] Tests added/updated

### DIFF
--- a/src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
+++ b/src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
@@ -5,9 +5,9 @@
   IncludeOptional "/etc/httpd/conf.d/05-foreman-ssl.d/*.conf"
 
   ## Logging
-  ErrorLog "/var/log/httpd/{{ 'foreman-ssl_error_ssl' if httpd_with_foreman else 'pulpcore_error' }}.log"
+  ErrorLog "|/usr/bin/systemd-cat -t {{ 'httpd-ssl-error' if httpd_with_foreman else 'pulpcore-ssl-error' }} -p err"
   ServerSignature Off
-  CustomLog "/var/log/httpd/{{ 'foreman-ssl_access_ssl' if httpd_with_foreman else 'pulpcore_access' }}.log" combined
+  CustomLog "|/usr/bin/systemd-cat -t {{ 'httpd-ssl-access' if httpd_with_foreman else 'pulpcore-ssl-access' }}" combined
 
   ## Request header rules
   ## as per http://httpd.apache.org/docs/2.4/mod/mod_headers.html#requestheader
--- a/src/roles/httpd/templates/foreman-vhost.conf.j2
+++ b/src/roles/httpd/templates/foreman-vhost.conf.j2
@@ -5,9 +5,9 @@
   IncludeOptional "/etc/httpd/conf.d/05-foreman.d/*.conf"
 
   ## Logging
-  ErrorLog "/var/log/httpd/{{ 'foreman_error' if httpd_with_foreman else 'mirror_error' }}.log"
+  ErrorLog "|/usr/bin/systemd-cat -t {{ 'httpd-error' if httpd_with_foreman else 'mirror-error' }} -p err"
   ServerSignature Off
-  CustomLog "/var/log/httpd/{{ 'foreman_access' if httpd_with_foreman else 'mirror_access' }}.log" combined
+  CustomLog "|/usr/bin/systemd-cat -t {{ 'httpd-access' if httpd_with_foreman else 'mirror-access' }}" combined
 
   ## Request header rules
   ## as per http://httpd.apache.org/docs/2.4/mod/mod_headers.html#requestheader
--- a/src/roles/httpd/README.md
+++ b/src/roles/httpd/README.md
@@ -1,0 +1,32 @@
+httpd Role
+==========
+
+Deploys Apache httpd as the reverse proxy for Foreman and Pulp.
+
+Journal logging
+---------------
+
+Apache access and error logs are sent to journald via `systemd-cat`.
+Each vhost uses its own syslog identifier (filter with `journalctl -t`).
+
+| Identifier | VHost | Log type | Typical content |
+|------------|-------|----------|-------------------|
+| `httpd-access` | HTTP (80) | Access | All requests, including 404 responses |
+| `httpd-error` | HTTP (80) | Error | Server/proxy errors (backend unreachable, malformed requests) |
+| `httpd-ssl-access` | HTTPS (443) | Access | All requests, including 404 responses |
+| `httpd-ssl-error` | HTTPS (443) | Error | Server/proxy errors |
+
+Pulp-only or mirror deployments use `pulpcore-ssl-access`, `pulpcore-ssl-error`,
+`mirror-access`, and `mirror-error` instead.
+
+Examples:
+
+```bash
+# Access log (404s, normal traffic)
+journalctl -t httpd-ssl-access --since "10 min ago"
+journalctl -t httpd-access | grep "/pub/"
+
+# Error log (proxy failures, malformed requests)
+journalctl -t httpd-ssl-error --since "1 hour ago"
+```
+
--- a/tests/httpd_test.py
+++ b/tests/httpd_test.py
@@ -147,6 +147,42 @@ def test_httpd_headers_use_dashes(server):
     assert cmd.stdout.strip() == '', f"HTTP header names should use dashes, not underscores:\n{cmd.stdout}"
 
 
+@pytest.mark.feature('foreman')
+def test_httpd_vhost_custom_logs_in_journal(server, server_fqdn, curl_request):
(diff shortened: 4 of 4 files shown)


src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -5,9 +5,9 @@
   IncludeOptional "/etc/httpd/conf.d/05-foreman-ssl.d/*.conf"
 
   ## Logging
-  ErrorLog "/var/log/httpd/{{ 'foreman-ssl_error_ssl' if httpd_with_foreman else 'pulpcore_error' }}.log"
+  ErrorLog "|/usr/bin/systemd-cat -t {{ 'httpd-ssl-error' if httpd_with_foreman else 'pulpcore-ssl-error' }} -p err"
   ServerSignature Off
-  CustomLog "/var/log/httpd/{{ 'foreman-ssl_access_ssl' if httpd_with_foreman else 'pulpcore_access' }}.log" combined
+  CustomLog "|/usr/bin/systemd-cat -t {{ 'httpd-ssl-access' if httpd_with_foreman else 'pulpcore-ssl-access' }}" combined
```
ehelms, line 10: Did you try ` CustomLog "journald:info" combined` ?

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -5,9 +5,9 @@
   IncludeOptional "/etc/httpd/conf.d/05-foreman-ssl.d/*.conf"
 
   ## Logging
-  ErrorLog "/var/log/httpd/{{ 'foreman-ssl_error_ssl' if httpd_with_foreman else 'pulpcore_error' }}.log"
+  ErrorLog "|/usr/bin/systemd-cat -t {{ 'httpd-ssl-error' if httpd_with_foreman else 'pulpcore-ssl-error' }} -p err"
```
ehelms, line 8: I saw mention of `ErrorLog syslog` as a native method to send to journald. However, it wasn't clear if that would respect "tagging" to differentiate virtual hosts.

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
Alleny244: I didn't try` journald:info` initially since I had switched to` systemd-cat `after hitting the /dev/stdout issue on EL9. I just tested it now: `CustomLog "journald:info" `combined works and access logs appear in `journalctl -u httpd,` but `ErrorLog "journald:err" `still prevents httpd from reloading.

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
Alleny244: ErrorLog syslog reaches journald, but its tag is global, so it can't provide separate tags per vhost. Since systemd-cat -t gives us distinct tags for :80 and :443 (and  tests verify this), I thought of the current approach.

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
ekohl: Looks like Apache 2.5 has [mod_journald](https://httpd.apache.org/docs/trunk/mod/mod_journald.html) but that doesn't exist in Apache 2.4 (which is what everyone uses).

Edit: also has:
> Performance warning
>
> Currently, systemd-journald is not designed for high-throughput logging and logging access_log to systemd-journald could decrease the performance a lot.

I think we'll have to keep an eye on this in the future.

evgeni: You were faster than me. Was about to post the same warning.

`Customlog journald:…` seems to be an EL specific patch: https://gitlab.com/redhat/centos-stream/rpms/httpd/-/blob/c9s/httpd-2.4.43-logjournal.patch?ref_type=heads

And for ErrorLog, I think it should be `journald:error`, not `err`, as that's what the Apache priority is (see https://github.com/apache/httpd/blob/736bb657405eb73fd68a64772c3a908807bdb887/server/log.c#L152 - `err` is systemd/syslog notation)

Alleny244: Thanks for the pointers! I tested this on` Apache 2.4.62 (CentOS Stream)`.  "`journald:error`"  failed, Apache treats them as file paths and reload/start fails. `CustomLog "journald:info" `does work, but all access logs end up under a single SYSLOG_IDENTIFIER=httpd, so we lose the per-vhost separation.

src/roles/httpd/templates/foreman-vhost.conf.j2
```diff
@@ -5,9 +5,9 @@
   IncludeOptional "/etc/httpd/conf.d/05-foreman.d/*.conf"
 
   ## Logging
-  ErrorLog "/var/log/httpd/{{ 'foreman_error' if httpd_with_foreman else 'mirror_error' }}.log"
+  ErrorLog "|/usr/bin/systemd-cat -t {{ 'httpd-error' if httpd_with_foreman else 'mirror-error' }} -p err"
   ServerSignature Off
-  CustomLog "/var/log/httpd/{{ 'foreman_access' if httpd_with_foreman else 'mirror_access' }}.log" combined
+  CustomLog "|/usr/bin/systemd-cat -t {{ 'httpd-access' if httpd_with_foreman else 'mirror-access' }}" combined
```
stejskalleos, line 10:
```
curl -s http://localhost/pub/nonexistent-path

journalctl -t httpd-error --since "1 min ago" --no-pager | grep nonexistent-path

=> empty
```

Looks like errors from the `/pub` dir aren't logged.

src/roles/httpd/templates/foreman-vhost.conf.j2
```diff
@@ -5,9 +5,9 @@
   IncludeOptional "/etc/httpd/conf.d/05-foreman.d/*.conf"
 
   ## Logging
-  ErrorLog "/var/log/httpd/{{ 'foreman_error' if httpd_with_foreman else 'mirror_error' }}.log"
+  ErrorLog "|/usr/bin/systemd-cat -t {{ 'httpd-error' if httpd_with_foreman else 'mirror-error' }} -p err"
```
stejskalleos, line 8: I think we should document the specified syslog identifiers (`httpd-error`, `httpd-access`); it might be useful for debugging purposes for our users.

stejskalleos (changes requested):

src/roles/httpd/templates/foreman-vhost.conf.j2
Alleny244: 404s belong in httpd-access, not httpd-error, right?. ErrorLog is for server/proxy errors; missing files under /pub are logged as normal requests with status 404. Try journalctl -t httpd-access for that curl test

evgeni: That's correct.

src/roles/httpd/templates/foreman-vhost.conf.j2
Alleny244: Yeah, I have added a README.md file under httpd role. Feel free to take a look and let me know if you think anything else should be added.

src/roles/httpd/templates/foreman-vhost.conf.j2
stejskalleos: > 404s belong in httpd-access, not httpd-error

Yeah that make sense.

stejskalleos (approved):
* Code LGTM
* CI is 🍏
* Tests are 🍏 

Unless there are any further objections from eee (@ehelms @ekohl @evgeni), I will merge it today.

ehelms (approved):

PR #625, opened 2026-07-07
Route Candlepin logs to journald via stdout

Alleny244:

Candlepin runs as a Podman container but currently writes application and Tomcat logs to host-mounted directories under `/var/log/candlepin` and `/var/log/tomcat`. This does not align with the containerized logging model, where services should log to `stdout`/`stderr` and rely on `journald`.

This change switches Candlepin, Tomcat, and HTTP access logging to stdout so all logs are collected by `journald`.

Related: Logging audit for journal-only deployments.

---

- Add `logback.xml` template using `ConsoleAppender` for Candlepin application logs.
- Add `logging.properties` template using `ConsoleHandler` for Tomcat internal logs.
- Mount both configuration files as Podman secrets.
- Remove host log volume mounts (`/var/log/candlepin` and `/var/log/tomcat`) and the task creating those directories.
- Add a Tomcat `AccessLogValve` to write HTTP access logs to stdout.
- Add integration tests to verify:
  - Candlepin logs are written to the journal.
  - Log volume mounts are no longer present.

---

## How to test this pull request

1. Deploy foremanctl from this branch.
2. Verify Candlepin is running:
   ```bash
   curl -sk https://localhost:23443/candlepin/status
   systemctl is-active candlepin
   ```
3. Generate a request and verify logs appear in the journal:
   ```bash
   curl -sk https://localhost:23443/candlepin/status
   sudo journalctl -u candlepin --since "1 min ago" --no-pager
   ```
4. Confirm the container no longer mounts `/var/log/candlepin` or `/var/log/tomcat`:
   ```bash
   sudo podman inspect candlepin --format '{{range .Mounts}}{{.Destination}} {{end}}'
   ```
5. Verify the old log files are no longer updated.
6. Run the new integration tests.

---

## Checklist

- [x] Tests added/updated

### DIFF
--- a/tests/candlepin_test.py
+++ b/tests/candlepin_test.py
@@ -25,6 +25,28 @@ def test_candlepin_status(server, certificates):
     assert status.stdout == '200'
 
 
+def test_candlepin_logs_in_journal(server, certificates):
+    server.run(
+        f"curl --cacert {certificates['ca_certificate']} --silent --output /dev/null "
+        f"https://localhost:23443/candlepin/status"
+    )
+
+    journal = server.run("journalctl -u candlepin --since '2 min ago' --no-pager").stdout
+    assert 'candlepin/status' in journal
+    assert 'LoggingFilter' in journal
+
+
+def test_candlepin_tomcat_logs_in_journal(server, certificates):
+    server.run(
+        f"curl --cacert {certificates['ca_certificate']} --silent --output /dev/null "
+        f"https://localhost:23443/candlepin/status"
+    )
+
+    journal = server.run("journalctl -u candlepin --no-pager").stdout
+    assert '"GET /candlepin/status HTTP/1.1"' in journal
+    assert 'org.apache.catalina' in journal
+
+
 def test_tls(server):
     result = server.run('nmap --script +ssl-enum-ciphers localhost -p 23443')
     result = result.stdout
--- a/src/roles/candlepin/tasks/main.yml
+++ b/src/roles/candlepin/tasks/main.yml
@@ -2,17 +2,6 @@
 - name: Deploy candlepin image
   ansible.builtin.include_tasks: image.yaml
 
-- name: Create log directories
-  ansible.builtin.file:
-    path: "{{ item }}"
-    state: directory
-    owner: root
-    group: root
-    mode: '0755'
-  loop:
-    - /var/log/candlepin
-    - /var/log/tomcat
-
 - name: Configure certificates
   ansible.builtin.include_tasks:
     file: certs.yml
@@ -50,6 +39,28 @@
   notify:
     - Restart candlepin
 
+- name: Create Candlepin logback configuration
+  containers.podman.podman_secret:
+    state: present
+    name: candlepin-logback-xml
+    data: "{{ lookup('ansible.builtin.template', 'logback.xml') }}"
+    labels:
+      filename: logback.xml
+      app: candlepin
+  notify:
+    - Restart candlepin
+
+- name: Create Tomcat logging configuration
+  containers.podman.podman_secret:
+    state: present
+    name: candlepin-tomcat-logging-properties
+    data: "{{ lookup('ansible.builtin.template', 'logging.properties') }}"
+    labels:
+      filename: logging.properties
+      app: tomcat
+  notify:
+    - Restart candlepin
+
 - name: Create DB SSL cert
   containers.podman.podman_secret:
     state: present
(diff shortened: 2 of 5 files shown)


tests/candlepin_test.py
```diff
@@ -25,6 +27,27 @@ def test_candlepin_status(server, certificates):
...
+    )
+    assert request.succeeded
+
+    journal = server.run("journalctl -u candlepin --since '2 min ago' --no-pager")
+    assert journal.succeeded
+    assert 'candlepin/status' in journal.stdout
+    assert 'LoggingFilter' in journal.stdout
+
+
+def test_candlepin_no_log_volume_mounts(server):
```
ehelms: I don't think this test is necessary.

nikosmoum (approved):

tests/candlepin_test.py
```diff
@@ -25,6 +27,27 @@ def test_candlepin_status(server, certificates):
     assert status.stdout == '200'
 
 
+def test_candlepin_logs_in_journal(server, certificates):
```
ehelms, line 28: Can you also test that tomcat logs are in the journal?

ehelms:
@Alleny244 please also rebase, that should get tests going

tests/candlepin_test.py
Alleny244: Sure, have removed that test

tests/candlepin_test.py
Alleny244: Done

tests/candlepin_test.py
```diff
@@ -25,6 +25,41 @@ def test_candlepin_status(server, certificates):
...
+    assert 'candlepin/status' in journal.stdout
+    assert 'LoggingFilter' in journal.stdout
+
+
+def test_candlepin_tomcat_logs_in_journal(server, certificates):
+    request = server.run(
+        f"curl --cacert {certificates['ca_certificate']} --silent --output /dev/null "
+        f"https://localhost:23443/candlepin/status"
+    )
+    assert request.succeeded
```
ehelms: I'm not a fan of asserts like this since we aren't testing that curl against Candlepin works in this test, this test is about the logs.

tests/candlepin_test.py
```diff
@@ -25,6 +25,41 @@ def test_candlepin_status(server, certificates):
...
+
+    journal = server.run("journalctl -u candlepin --since '2 min ago' --no-pager")
+    assert journal.succeeded
```
ehelms: Similar comment -- this test isn't about whether `journalctl` itself works, we just want to focus on the log itself.

tests/candlepin_test.py
```diff
@@ -25,6 +25,41 @@ def test_candlepin_status(server, certificates):
...
+    assert '"GET /candlepin/status HTTP/1.1"' in journal.stdout
+
+    catalina = server.run("journalctl -u candlepin --no-pager -g 'org.apache.catalina'")
+    assert catalina.succeeded
+    assert catalina.stdout
+
+    tomcat_logs = server.run(
```
ehelms: What are these three lines testing?

tests/candlepin_test.py
```diff
@@ -25,6 +25,41 @@ def test_candlepin_status(server, certificates):
     assert status.stdout == '200'
 
 
+def test_candlepin_logs_in_journal(server, certificates):
+    request = server.run(
+        f"curl --cacert {certificates['ca_certificate']} --silent --output /dev/null "
+        f"https://localhost:23443/candlepin/status"
+    )
+    assert request.succeeded
```
ehelms: I'm not a fan of asserts like this since we aren't testing that curl works in this test.

tests/candlepin_test.py
```diff
@@ -25,6 +25,41 @@ def test_candlepin_status(server, certificates):
...
+
+    journal = server.run("journalctl -u candlepin --since '2 min ago' --no-pager")
+    assert journal.succeeded
```
ehelms: Similar comment

tests/candlepin_test.py
Alleny244: Yeah , got it, I ll remove this part

tests/candlepin_test.py
Alleny244: yup, sure !!

tests/candlepin_test.py
Alleny244: The catalina check verifies Tomcat internal logs (e.g. org.apache.catalina.startup.Catalina.start) land in journald via logging.properties → ConsoleHandler.. But this can be condensed as a single  assert ,as mentioned earlier

tests/candlepin_test.py
Alleny244: Sure, will be updated

tests/candlepin_test.py
Alleny244: yeah , removed!!

ehelms (approved):

PR #628, opened 2026-07-07
Add satellite flavor for downstream deployments

zjhuntin:
## Summary
- Adds a `satellite` flavor (`src/vars/flavors/satellite.yml`) with the feature set for Satellite deployments
- Enables: foreman, katello, content/{ansible,container,deb,python,rpm}, remote-execution, rh-cloud, theme-satellite, webhooks
- Compared to `katello` flavor: adds features that Satellite ships by default (remote-execution, rh-cloud, theme-satellite, webhooks)

## Details

The `theme-satellite` feature requires a corresponding entry in `features.yaml` (or `features.d/`) to take effect. Without one, it is silently ignored. It is marked as `internal` so users cannot enable it manually.

Usage:
```bash
foremanctl deploy --flavor satellite --initial-admin-password=changeme
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -465,6 +465,76 @@ jobs:
           limit-access-to-actor: true
           wait-timeout-minutes: 5
 
+  satellite-tests:
+    runs-on: ubuntu-24.04
+    name: "Tests Satellite Deployment"
+    steps:
+      - uses: actions/checkout@v7
+      - name: Set up Python
+        uses: actions/setup-python@v7
+        with:
+          python-version: '3.12'
+      - name: Setup libvirt for Vagrant
+        uses: voxpupuli/setup-vagrant@v0
+        with:
+          configure_dns: true
+      - name: Install Ansible
+        run: pip install --upgrade ansible-core
+      - name: Setup environment
+        run: ./setup-environment
+      - name: Start VMs
+        run: |
+          ./forge vms start --vms "quadlet client"
+      - name: Configure repositories
+        run: |
+          ./forge setup-repositories
+      - name: Run image pull
+        run: |
+          ./foremanctl pull-images
+      - name: Run deployment
+        run: |
+          ./foremanctl deploy \
+            --flavor satellite \
+            --initial-admin-password=changeme \
+            --initial-organization "Foreman CI" \
+            --initial-location "Internet" \
+            --tuning development
+      - name: Deploy features
+        run: |
+          ./foremanctl deploy \
+            --add-feature hammer \
+            --add-feature foreman-proxy \
+            --add-feature azure-rm \
+            --add-feature google \
+            --add-feature bmc
+      - name: Run tests
+        run: |
+          ./forge test
+      - name: Run smoker
+        run: |
+          ./forge smoker
+      - name: Archive smoker report
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: smoker-satellite
+          path: "/home/runner/smoker/report/"
+      - name: Generate sos reports
+        if: ${{ always() }}
+        run: ./forge sos
+      - name: Archive sos reports
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: sosreport-satellite
+          path: sos/
+      - name: Setup upterm session
+        if: ${{ failure() }}
+        uses: owenthereal/action-upterm@v1
+        with:
+          limit-access-to-actor: true
+          wait-timeout-minutes: 5
+
   # A dummy job that you can mark as a required check instead of each individual test
   test-suite:
     if: always()
@@ -474,6 +544,7 @@ jobs:
       - upgrade
       - migration
       - foreman-proxy-content-tests
+      - satellite-tests
       - ansible-lint
       - python-lint
     runs-on: ubuntu-latest
--- a/src/features.yaml
+++ b/src/features.yaml
@@ -92,5 +92,6 @@ registration:
     - templates
 theme-satellite:
   description: Satellite theme for Foreman
+  internal: true
   foreman:
     plugin_name: foreman_theme_satellite
--- a/src/features.yaml
+++ b/src/features.yaml
@@ -90,3 +90,7 @@ registration:
     plugin_name: registration
   dependencies:
     - templates
+theme-satellite:
+  description: Satellite theme for Foreman
+  foreman:
+    plugin_name: foreman_theme_satellite
--- a/src/playbooks/deploy/metadata.obsah.yaml
+++ b/src/playbooks/deploy/metadata.obsah.yaml
@@ -7,6 +7,7 @@ variables:
     help: Base flavor to use in this deployment.
     choices:
       - katello
+      - satellite
   external_authentication:
     help: External authentication method to use
     choices:
--- a/src/vars/flavors/satellite.yml
+++ b/src/vars/flavors/satellite.yml
@@ -10,3 +10,20 @@ flavor_features:
   - remote-execution
   - rh-cloud
   - theme-satellite
+  - webhooks
+
+checks_to_execute:
+  - check_features
+  - check_hostname
+  - check_database_connection
+  - check_system_requirements
+  - check_podman_network_backend
+
+health_checks_to_execute:
+  - check_hostname
+  - check_database_connection
+  - check_services
+  - check_foreman_api
+  - check_foreman_tasks
+  - check_host_facts_count
+  - check_duplicate_permissions
--- a/src/vars/flavors/satellite.yml
+++ b/src/vars/flavors/satellite.yml
@@ -1,0 +1,12 @@
+---
+flavor_features:
+  - foreman
+  - katello
+  - content/ansible
+  - content/container
+  - content/deb
+  - content/python
+  - content/rpm
+  - remote-execution
+  - rh-cloud
+  - theme-satellite
--- a/src/roles/post_install/tasks/message.yaml
+++ b/src/roles/post_install/tasks/message.yaml
@@ -1,6 +1,6 @@
 ---
 - name: Post install message
-  when: flavor in ['katello']
+  when: flavor in ['katello', 'satellite']
   ansible.builtin.debug:
     msg:
       - "{{ _post_install_url_msg }}"
--- a/tests/backup_test.py
+++ b/tests/backup_test.py
@@ -18,8 +18,8 @@ def expected_databases(enabled_features, flavor):
     """
     databases = []
 
-    # Katello flavor has foreman, candlepin, and pulp
-    if flavor == 'katello':
+    # Katello and Satellite flavors have foreman, candlepin, and pulp
+    if flavor in ('katello', 'satellite'):
         databases = ['foreman', 'candlepin', 'pulp']
 
     # Foreman-proxy-content flavor only has pulp
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -284,16 +284,25 @@ def pytest_configure(config):
     config.user_parameters = UserParameters(config)
 
 
+FLAVOR_TEST_DIRS = {
+    'katello': ['katello', 'satellite'],
+    'foreman-proxy-content': ['foreman-proxy-content', 'capsule'],
+}
+
+
 def pytest_collection_modifyitems(config, items):
     active_flavor = config.user_parameters.flavor
-    active_flavor_dir = FLAVOR_TESTS_DIR / active_flavor
+    allowed_dirs = set()
+    for test_dir, flavors in FLAVOR_TEST_DIRS.items():
+        if active_flavor in flavors:
+            allowed_dirs.add(test_dir)
 
     deselected = []
     selected = []
     for item in items:
         test_path = py.path.local(item.fspath)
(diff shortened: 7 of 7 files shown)


src/vars/flavors/satellite.yml
```diff
@@ -0,0 +1,12 @@
+---
+flavor_features:
+  - foreman
+  - katello
+  - content/ansible
+  - content/container
+  - content/rpm
+  - hammer
```
evgeni: Do we really want to enable hammer by default?

While I know this is something most users will want to have to keep the "old" "all in one" deployment style, I would prefer if they have to explicitly opt in to that.

src/vars/flavors/satellite.yml
```diff
@@ -0,0 +1,12 @@
+---
+flavor_features:
+  - foreman
+  - katello
```
Gauravtalreja1, line 4: foreman-proxy feature can be added to this list

src/vars/flavors/satellite.yml
```diff
@@ -0,0 +1,12 @@
...
+  - content/ansible
+  - content/container
+  - content/rpm
+  - hammer
+  - remote-execution
+  - rh-cloud
+  - iop
```
Gauravtalreja1: iop feature can be optional

evgeni: I don't think IOP should be enabled by default.

src/vars/flavors/satellite.yml
zjhuntin: ack, I was only assuming it'd be there, happy to remove. In a lot of ways I think that makes sense too cause they need to enable other repos for this to work.

evgeni: hammer will come from the same repos as foremanctl? :)

src/vars/flavors/satellite.yml
evgeni: as with hammer, I think that is something people should be actively opt in for?
content distribution to external proxies works without the feature, so in many architectures it won't be needed

src/vars/flavors/satellite.yml
zjhuntin: lol fair point :smile: only a problem with the way i'm testing locally.

src/vars/flavors/satellite.yml
zjhuntin: Done

zjhuntin: Done

src/vars/flavors/satellite.yml
evgeni: yepp! today, with the way upstream repos are laid out, it's a bit of a pain.

zjhuntin:
tested the most recent version and that worked as well :smile:

src/vars/flavors/satellite.yml
zjhuntin: for now I do not have foreman-proxy enabled by default.

src/playbooks/_flavor_features/metadata.obsah.yaml
```diff
@@ -4,6 +4,7 @@ variables:
     help: Base flavor to use in this deployment.
     choices:
       - katello
+      - satellite
```
evgeni: IMHO a Satellite user should not be able to select Katello, and a Katello user not Satellite 

But we don't need to solve this right now

src/vars/flavors/satellite.yml
Gauravtalreja1: I think it should be identical to what we've currently with rpm based installations and we should have similar features pre-enabled on satellite deploy for existing users, where foreman-proxy and hammer are enabled by default
@ekohl @vijay8451 wdyt?

ekohl: Yes, I do think that.

ekohl:
How should we approach testing this? In 2198602d41c83e92c6e8568fc0bea642771a62d1 @arvind4501 introduced flavor based tests. Are we going to duplicate all katello tests for satellite? Should we also deploy this in CI to verify it works?

src/vars/flavors/satellite.yml
ehelms: One of our goals with foremanctl and containers was to encourage users to deploy external proxy/capsule/thing when they need functionality that is not available within the core Foreman(+Katello) stack rather than getting a local proxy by default. So that they have to make the choice whether to have it locally or not actively.

Upgrades will result in the same functionality, but the baseline feature set doesn't necessarily have to replicate what we had before.

For example, as a best practice, `hammer` should be installed on and used from a separate node. We should discourage admins from sshing in as root and using a CLI on the same machine that the service exists on. That's just a good security posture for us to encourage users to have. Should we block this? No. But it shouldn't be the default behaviour IMO.

ekohl: > One of our goals with foremanctl and containers was to encourage users to deploy external proxy/capsule/thing when they need functionality that is not available within the core Foreman(+Katello) stack rather than getting a local proxy by default. So that they have to make the choice whether to have it locally or not actively.

I think this is a good discussion topic. It will mean that by default you won't have a working remote execution setup. Is that what we expect?

> Upgrades will result in the same functionality, but the baseline feature set doesn't necessarily have to replicate what we had before.

Agreed

> For example, as a best practice, `hammer` should be installed on and used from a separate node. We should discourage admins from sshing in as root and using a CLI on the same machine that the service exists on. That's just a good security posture for us to encourage users to have. Should we block this? No. But it shouldn't be the default behaviour IMO.

This is IMHO a good thing, but I wonder about the documentation impact. Today we often document the Hammer procedure. Should we modify those to all to refer to a chapter to install Hammer (somewhere)?

zjhuntin:
@ekohl I looked into this with Claude's help.

I don't think duplicating the `tests/flavor/katello/` tests into a `tests/flavor/satellite/` directory makes much sense here. The two tests there — `images_test.py` and `postgresql_test.py` — verify that the foreman, pulp, and candlepin container images, databases, and users are present. Those are all guaranteed by the `katello` feature's dependency chain in `src/features.yaml`, and both flavors enable it. The assertions would be identical.

I think the directory-based flavor tests work well for `foreman-proxy-content`, where the deployment topology is fundamentally different and you need different assertions for the same kinds of checks. But when two flavors share the same Katello stack and only differ in which optional features are layered on top, tying tests to the flavor name means every new flavor that deploys Katello needs its own copy of the same tests.

The feature marker approach seems like a better fit here. `@pytest.mark.feature('katello')` ties the tests to what they actually depend on — the Katello stack being deployed — rather than which flavor was selected. This is already how `candlepin_test.py`, `katello_api_test.py`, and `client_test.py` work.

I'd propose:
1. Move the two `tests/flavor/katello/` files into the shared test directory with `@pytest.mark.feature('katello')` decorators
2. Keep `tests/flavor/` for `foreman-proxy-content` where the structural differences justify it
3. Add a `satellite-tests` CI job (single centos/stream9 run with `--flavor satellite`)

zjhuntin:
I went ahead and pushed a commit with what I think this would look like so we can see if CI passes.

zjhuntin:
Realized, that won't work as we don't serve the package. I'm not 100% sure how we can test this totally here. Is it okay without foreman_theme_satellite here?

arvind4501:
For the testing part, if we wil have exactly similar tests for `katello` and `satellite` flavor, then i would make change to https://github.com/theforeman/foremanctl/blob/master/tests/conftest.py#L267 to such that it runs katello specific tests when any of katello and satellite flavor is deployed. and similarly i would do for capsule flavor as well. i think this would require some mapping in same something like 
```
FLAVOR_TEST_DIRS = {
    'katello': ['katello', 'satellite'],
    'foreman-proxy-content': ['foreman-proxy-content', 'capsule'],
}
```

src/playbooks/_flavor_features/metadata.obsah.yaml
evgeni: https://github.com/theforeman/foremanctl/pull/643

zjhuntin:
Switched to @arvind4501's approach — added a FLAVOR_TEST_DIRS mapping in conftest.py so the satellite flavor runs the katello test directory. Also cleaned up the commit history.

src/vars/flavors/satellite.yml
jeremylenz: +1 to everything Ewoud said here. I think users will expect the same plugins and features to be enabled on a containerized install that they used to get with the classic installer. If that's not going to be the case, our documentation should be _loud_ about that.

zjhuntin:
Circling back on the two open threads here:

**1. Feature set defaults (hammer, foreman-proxy)**

Eric's point about encouraging users to deploy proxy externally and install hammer on a separate node makes sense from a security posture perspective. But Ewoud and Jeremy raise a valid concern — users migrating from foreman-installer will expect the same features to be available out of the box. If we change the defaults, we need to be deliberate about it and make sure documentation reflects what's different and why.

What should the actual outcome be here? Should the satellite flavor ship with hammer and foreman-proxy enabled by default to match existing user expectations, or should we take this as an opportunity to set new defaults and document the change prominently?

**2. Testing theme-satellite**

The `theme-satellite` feature is in the satellite flavor, but `rubygem-foreman_theme_satellite` isn't available in the repos that the upstream container images are built from. The nightly foreman image at `quay.io/foreman/foreman` only includes plugins listed in the `FOREMAN_PLUGINS` build arg in [foreman-oci-images](https://github.com/theforeman/foreman-oci-images), and `foreman_theme_satellite` isn't one of them.

This means CI can't actually deploy with `--flavor satellite` as-is — the theme plugin won't be found. The best thing I can come up with is: Accept that upstream CI tests the satellite flavor without the theme (maybe skip/ignore theme-satellite if the plugin isn't available).

But I can also understand not liking that. Is there another alternative option?

zjhuntin:
Rebased on master to pick up the post_install refactor from #641. Also added the satellite flavor to the post-install message condition so it shows the admin credentials after deploy, same as katello.

.github/workflows/test.yml
```diff
@@ -455,6 +455,81 @@ jobs:
...
+            --initial-location "Internet" \
+            --tuning development
+      - name: Deploy features
+        run: |
+          ./foremanctl deploy \
+            --add-feature hammer \
+            --add-feature foreman-proxy \
+            --add-feature azure-rm \
+            --add-feature google \
+            --add-feature bmc
```
adamruzicka, line 509: Shouldn't those be part of the flavor or am I misinterpreting why those would get enabled?

src/features.yaml
```diff
@@ -66,3 +66,7 @@ bmc:
   description: Power management for bare metal hosts (IPMI, Redfish)
   foreman_proxy:
     plugin_name: bmc
+theme-satellite:
```
ekohl, line 93: I think this should be at least an `internal` feature.

evgeni: Yepp, nothing and nobody should ever enable this manually

src/playbooks/deploy/metadata.obsah.yaml
```diff
@@ -7,6 +7,7 @@ variables:
     help: Base flavor to use in this deployment.
     choices:
       - katello
+      - satellite
```
evgeni, line 10: We should not allow users to select between katello and satellite.

In https://github.com/theforeman/foremanctl/pull/643 I started with a PoC how this *could* work

Feel free to  implement this part in a follow up tho.

evgeni:
The test failures in the Satellite test seem related. Holler if you need help deciphering them.

.github/workflows/test.yml
```diff
@@ -463,6 +463,81 @@ jobs:
...
+      - name: Setup environment
+        run: ./setup-environment
+      - name: Start VMs
+        run: |
+          ./forge vms start --vms "quadlet client"
+      - name: Downgrade openssl
+        run: |
+          for vm in quadlet client; do
+            vagrant ssh "$vm" -- sudo dnf downgrade -y openssl openssl-libs
+          done
```
evgeni: this is not needed anymore, it was removed in #659

.github/workflows/test.yml
evgeni: right now they are optional, because it's a bit undecided which features will get enabled by default on new installs. this PR gives enough building blocks for future adjustment :)

src/vars/flavors/satellite.yml
```diff
@@ -0,0 +1,10 @@
+---
+flavor_features:
+  - foreman
+  - katello
+  - content/ansible
+  - content/container
+  - content/rpm
+  - remote-execution
+  - rh-cloud
+  - theme-satellite
```
adamruzicka, line 12: since #694 is already in
```suggestion
  - theme-satellite
  - webhooks
```

evgeni:
@zjhuntin do you have time to further work on this? Or do you need any help?
I don't think this needs to be in perfect final form, but having a `satellite` flavor defined would allow us to continue working on the details in follow ups as those details emerge.

src/vars/flavors/satellite.yml
```diff
@@ -0,0 +1,27 @@
+---
+flavor_features:
+  - foreman
+  - katello
+  - content/ansible
+  - content/container
+  - content/rpm
```
adamruzicka, line 9: Since #444 is already in
```suggestion
  - content/rpm
  - ansible
```

PR #631, opened 2026-07-07
Add pull-images playbook for forge

jeremylenz:

The `pull-images` playbook was only available in `src/playbooks/` for `foremanctl` production deployment. Development environments using `forge` couldn't pull container images because `forge` uses `development/playbooks/` for playbook discovery.

* Added `development/playbooks/pull-images/metadata.obsah.yaml` with obsah metadata, help text, and includes
* Added `development/playbooks/pull-images/pull-images.yaml` that targets `quadlet` host, loads production vars, and reuses the existing `images` role

Steps to reproduce:

* Deploy a development environment with `forge deploy`
* Run `forge pull-images` to verify the command is discovered and executes successfully
* Verify container images are pulled to the development environment

#### Checklist

### DIFF
--- a/development/playbooks/pull-images/metadata.obsah.yaml
+++ b/development/playbooks/pull-images/metadata.obsah.yaml
@@ -1,0 +1,7 @@
+---
+help: |
+  Pull necessary container images
+
+include:
+  - _database_mode
+  - _flavor_features
--- a/development/playbooks/pull-images/pull-images.yaml
+++ b/development/playbooks/pull-images/pull-images.yaml
@@ -1,0 +1,72 @@
+---
+- name: Pull images
+  hosts: "{{ target_host if target_host is defined and target_host != '' else 'quadlet' }}"
+  vars_files:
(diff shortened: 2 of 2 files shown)


ehelms:
I would think that `foremanctl` commands should work as long as the target host is the one you are aiming at. Both `deploy-dev` and `pull-images` target `quadlet` by default. That should be the same entity in inventory. 

Is this rather a case of one supports custom target hosts and the other does not?

jeremylenz:
I feel it's essential to allow non-quadlet target hosts, since about half of us have development environments set up that way (using forge deploy-dev with `--target-host localhost`). If there's a more correct way to do that, of course I'm open to that too.

ehelms:
I don't think we should repeat functionality so either in these scenarios we should:

- inherit the playbook
- add the `target_host` paradigm to playbooks in `src/`

The second option likely makes the most sense given we need this for things like `deploy-proxy` or `backup`.

Issue #630, opened 2026-07-07
State management redesign for multi-host control


ehelms:
## User Stories

**US1: Deploy a server and add proxies over time**
A sysadmin deploys a Foreman server from their workstation, then days or weeks later adds 2-5 content proxies one at a time. Each `deploy-proxy` invocation must not inherit or overwrite the server's persisted parameters, and vice versa. Ongoing operations (upgrades, config changes, feature enablement, certificate renewal, backup) target one host at a time without affecting other hosts' state.

**US2: Certificate bundle flow from a single control node**
When adding a proxy, the sysadmin generates a certificate bundle -- either on the server or directly on the control node -- and `deploy-proxy` picks it up automatically from the control node's state without manual file transfer between machines.

**US3: Rebuild a failed server or proxy**
When a server dies, the sysadmin restores it from backup on the control node and reconnects existing proxies. When a proxy dies, the sysadmin redeploys just that proxy from the control node using its stored state (secrets, parameters, certificates) without affecting the server or other proxies.

**US4: Manage identical but independent Foreman servers**
A sysadmin maintains multiple independent Foreman servers (e.g., dev/staging/prod) with the same features, tuning, and configuration but different hostnames and separate data. They may define the config once and apply it to each server, or deploy one and replicate its settings to others -- all from a single control node or from separate control nodes sharing a configuration template.

**US5: Development environment alongside production**
A developer uses `forge` for a dev environment and `foremanctl` for a test deployment on the same machine. Running one must not contaminate the other's persisted parameters or secrets. This includes the case where the developer runs a dev Foreman server via `forge` and then deploys a proxy to connect to it.

**US6: Multiple development environments**
A developer runs multiple `forge` environments simultaneously (e.g., two VMs with different configurations, or a dev server and a dev proxy). Each environment's state must be isolated.

## Problem

foremanctl currently uses a single flat state directory (`OBSAH_STATE`, defaulting to `.var/lib/foremanctl/`) on the control node for two distinct categories of state:

1. **Parameter persistence** - obsah writes `parameters.yaml` to `OBSAH_STATE` after every successful run, and loads it as defaults on the next run
2. **Secret/credential files** - Ansible `password` lookups generate and cache secrets (DB passwords, OAuth keys, CA passwords, encryption keys) in `obsah_state_path`

[...]

PR #633, opened 2026-07-07
Add --target-host parameter for multi-host management

jeremylenz:
## Summary

Adds a `--target-host` argument to all 8 production `foremanctl` commands (deploy, features, health, migrate, backup, pull-images, certificate-bundle, checks), allowing users to specify which host to run against instead of the hardcoded `quadlet` default.

This is a first step toward multi-host management — running commands against different targets such as external proxies or remote servers. Full multi-host workflows will additionally require per-host state isolation (#630), since today all commands share a single `parameters.yaml`.

The parameter defaults to `quadlet` when not specified, preserving existing behavior. The value is persisted between runs and can be cleared with `--reset-target-host`.

For remote hosts, the specified host must be present in the Ansible inventory.

## Changes

- New shared metadata fragment `_target_host` included by all 8 commands
- Playbook `hosts:` directives use `{{ target_host | default('quadlet') }}`
- Updated `docs/user/parameters.md` with a Global Parameters section

## Test plan

- [x] `foremanctl <command> --help` shows `--target-host` for all 8 commands
- [x] ansible-lint passes
- [x] All playbook tests pass

### DIFF
--- a/docs/user/parameters.md
+++ b/docs/user/parameters.md
@@ -7,6 +7,12 @@ This document covers installation parameters as presented to users, and includes
 * Using a single parameter when there are multiple variables that should have the same value
 * Aim for minimal configuration values for the user
 
+## Global Parameters
+
+| Parameter | Description | Applies to |
+| ----------| ----------- | ---------- |
+| `--target-host` | Target hostname or IP address for deployment. Defaults to `quadlet`. This value is persisted between runs. For remote hosts, the specified host must be present in the Ansible inventory. | All commands except `deploy-proxy` |
+
 ## Mapping
 
 Parameters are split into either mapped or unmapped groupings per category.
--- a/src/playbooks/_target_host/metadata.obsah.yaml
+++ b/src/playbooks/_target_host/metadata.obsah.yaml
@@ -1,0 +1,5 @@
+---
+variables:
+  target_host:
+    help: Target hostname or IP address for deployment
+    action: store
--- a/src/playbooks/backup/metadata.obsah.yaml
+++ b/src/playbooks/backup/metadata.obsah.yaml
@@ -23,3 +23,6 @@ variables:
     help: Wait for running tasks to complete instead of failing immediately
     action: store_true
     persist: false
+
+include:
+  - _target_host
--- a/src/playbooks/certificate-bundle/certificate-bundle.yaml
+++ b/src/playbooks/certificate-bundle/certificate-bundle.yaml
@@ -1,7 +1,6 @@
 ---
 - name: Generate a certificate bundle for a hostname
-  hosts:
-    - quadlet
+  hosts: "{{ target_host | default('quadlet') }}"
   become: true
   vars_files:
     - "../../vars/base.yaml"
--- a/src/playbooks/certificate-bundle/metadata.obsah.yaml
+++ b/src/playbooks/certificate-bundle/metadata.obsah.yaml
@@ -31,4 +31,5 @@ constraints:
     - ['certificates_source', 'default', ['certificates_custom_server_certificate', 'certificates_custom_server_key']]
 
 include:
+  - _target_host
   - _certificate_source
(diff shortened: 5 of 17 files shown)


ianballou:
I think you might also need to ensure the host is in the Ansible inventory. If so, perhaps it's worth a small docs line.

ehelms:
> This brings feature parity between the `forge deploy-dev` command and the 8 production `foremanctl` commands. Previously, only `forge deploy-dev` supported the `--target-host` parameter, which allows targeting a host other than the default 'quadlet'.

For me this change is not about parity, it's about the need to run these commands against multiple hosts in some cases such as deploying multiple external proxies or executing a backup of a server vs. a proxy. That's my way of saying can you update the commit message and description to speak more to this idea?

Additionally, while this will unlock the ability to run commands against more targets, this will be limited by https://github.com/theforeman/foremanctl/issues/630

@evgeni what do you think? I'm generally in favor of this `target-host` idea. Is there any more "ansible" way to do it that fits in with our CLI design?

jeremylenz:
@ehelms Updated the description and commit message. :+1:

ianballou:
What is the intended end state for where we recommend users run foremanctl, if there is one? Do we want users to be able to do every foremanctl action against any reachable host? For example, if setting up smart proxies, will it be totally legit for users to deploy the Foreman and the proxy from some exterior controller machine? 

I just want to understand if this `target-host` option fits in with the end game plan, or if it's a helpful intermediate workaround for the current "quadlet" assumptions.

jeremylenz:
I will defer to @ehelms to answer this

ianballou:
> I will defer to @ehelms to answer this

Whatever the answer ends up being, I like this :+1:

docs/user/parameters.md
```diff
@@ -7,6 +7,12 @@ This document covers installation parameters as presented to users, and includes
 * Using a single parameter when there are multiple variables that should have the same value
 * Aim for minimal configuration values for the user
 
+## Global Parameters
+
+| Parameter | Description | Applies to |
+| ----------| ----------- | ---------- |
+| `--target-host` | Target hostname or IP address for deployment. Defaults to `quadlet`. This value is persisted between runs. For remote hosts, the specified host must be present in the Ansible inventory. | All commands except `deploy-proxy` |
```
evgeni, line 14: Why except `deploy-proxy`? (This is not a "I think you're wrong", more a "I want to understand your trail of thought")

evgeni:
I like this, but I also think it's dangerous in the current state, so not sure I'd like to expose it to users just yet.

Right now this allows setting up multiple Foreman instances, with the same set of features/configs. I don't think that's terribly useful in most cases? Unless you have a prod and a test box that should be identical and you can manage them from the same controller.

The PR description says that can be used to deploy multiple proxies, but it's not used in `deploy-proxy`. (And even if we'd expose it there, they would end up having the same config all over, maybe even the same as the main server if you run it from there)

Overall I think I'd like to understand more which user workflows this enables now, without #630 being solved.

jeremylenz:
> Overall I think I'd like to understand more which user workflows this enables now, without #630 being solved.

The one workflow that started it for me is that I was unable to run `foremanctl pull-images` in a development environment, to update my images.

jeremylenz:
> The PR description says that can be used to deploy multiple proxies, but it's not used in deploy-proxy.

Good catch — `deploy-proxy` was omitted because it uses `hosts: proxy` rather than `hosts: quadlet`, so it wasn't part of the "replace hardcoded quadlet" scope. But if we want to support multi-proxy workflows, adding `--target-host` there would be the same pattern, just defaulting to `proxy` instead of `quadlet`. Happy to add that if we think it's worth including.

> Overall I think I'd like to understand more which user workflows this enables now, without #630 being solved.

Agreed that without #630, this is limited to single-target workflows where you're only ever pointing at one host at a time — just not one called `quadlet`. The immediate motivation was being unable to run commands like `foremanctl pull-images` in environments where the target host has a different name in the inventory.

PR #636, opened 2026-07-08
Make sure Podman uses  netavark as network backend

arvind4501:
Foremanctl requires netavark as the Podman network backend. This adds a preflight check using containers.podman.podman_system_info that fails with a clear error if the system is configured to use a different backend such as CNI. The check runs for both katello and foreman-proxy-content flavors.

* New `check_podman_network_backend` role that uses `containers.podman.podman_system_info` to retrieve the active Podman network backend and fails with a clear error message if it is
   not `netavark`
 * Role added to `checks_to_execute` for both `katello` and `foreman-proxy-content` flavors so it runs as a preflight check before both deployments
* `docs/developer/checks.md` updated with description, fail state, and rationale for the new check

Steps to reproduce:

* Deploy foremanctl against a host where `podman info --format '{{.Host.NetworkBackend}}'` returns `netavark` — the preflight check should pass
* Deploy foremanctl against a host configured to use CNI — the preflight check should fail with a message indicating the network backend must be `netavark`

#### Checklist

### DIFF
--- a/src/roles/check_podman_network_backend/tasks/main.yaml
+++ b/src/roles/check_podman_network_backend/tasks/main.yaml
@@ -1,0 +1,14 @@
+---
+- name: Gather Podman system info
+  containers.podman.podman_system_info:
+  register: check_podman_network_backend_info
+
+- name: Assert Podman network backend is netavark
+  ansible.builtin.assert:
+    that:
+      - check_podman_network_backend_value == 'netavark'
+    fail_msg: >-
+      Podman network backend is '{{ check_podman_network_backend_value }}'
+      but must be 'netavark'. See the installation prerequisites for more information.
+  vars:
+    check_podman_network_backend_value: "{{ check_podman_network_backend_info.podman_system_info.host.networkBackendInfo.backend }}"
--- a/docs/developer/checks.md
+++ b/docs/developer/checks.md
@@ -40,6 +40,11 @@ Please update this file as check usage evolves.
 - **Fail state**: Fails if FQDN conditions are not met.
 - **Rationale**: Foreman requires a properly-configured server FQDN. This check ensures server hostname was modified from the default and approximates a valid FQDN.
 
+### check_podman_network_backend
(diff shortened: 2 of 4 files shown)


src/roles/check_podman_network_backend/tasks/main.yaml
```diff
@@ -0,0 +1,16 @@
...
+  ansible.builtin.set_fact:
+    check_podman_network_backend_value: "{{ check_podman_network_backend_info.podman_system_info.host.networkBackendInfo.backend }}"
+
+- name: Assert Podman network backend is netavark
+  ansible.builtin.assert:
+    that:
+      - check_podman_network_backend_value == 'netavark'
+    fail_msg: >-
+      Podman network backend is '{{ check_podman_network_backend_value }}'
+      but must be 'netavark'. See the installation prerequisites for more information.
```
evgeni, line 12:
```suggestion
  ansible.builtin.assert:
    that:
      - check_podman_network_backend_value == 'netavark'
    fail_msg: >-
      Podman network backend is '{{ check_podman_network_backend_value }}'
      but must be 'netavark'. See the installation prerequisites for more information.
  vars:
    check_podman_network_backend_value: "{{ check_podman_network_backend_info.podman_system_info.host.networkBackendInfo.backend }}"
```

Then you don't need the `set_fact` above but still have the result in a speakable variabl

evgeni (approved):

src/roles/check_podman_network_backend/tasks/main.yaml
ehelms: What does a user do if this check fails? Are we expecting our installation prerequisites to discuss why netavark is needed and how to ensure that it is present?

arvind4501: I think we definately need to document it as requirement in the installation prerequisites, and expect user to configure it to netavark(which is default for podman https://docs.podman.io/en/stable/markdown/podman-network.1.html#description)

PR #640, opened 2026-07-09
Foremanct - Log levels for Foreman & Proxy

stejskalleos:

To give users the option to change the log level of the services.

### What changes are introduced in this pull request?

Three new `deploy` parameters:

* `--log-level` Default log level for all services.
* `--foreman-log-level` Log level for `foreman.service`
* `--foreman-proxy-log-level` Log level for `foreman-proxy.service`

### How to test this pull request

```
foremanctl deploy --initial-admin-password=changeme \
--add-feature foreman-proxy \
--tuning development \
--log-level=debug
```

or

```
foremanctl deploy --foreman-log-level=error --foreman-proxy-log-level=error
```

and then:

```
journalctl -f -u foreman.service
journalctl -f -u foreman-proxy.service
```

### Notes

Resetting params won't work

```shell
./foremanctl deploy --reset-foreman-log-level --reset-foreman-proxy-log-level
```
Fixed in https://github.com/theforeman/obsah/pull/127/changes

#### Checklist

### DIFF
--- a/docs/user/parameters.md
+++ b/docs/user/parameters.md
@@ -72,6 +72,21 @@ There are multiple use cases from the users perspective that dictate what parame
 | foreman-installer Parameter | Description | Reason |
 | --------------------------- | ----------- | ------ |
 
+#### Logging
+
+##### New
+
+| Parameter | Description |
+| ----------| ----------- |
+| `--log-level` | Default log level for all services. Accepted values: `debug`, `info`, `warn`, `error`, `fatal`. Defaults to `info`. |
+
+##### Mapped
+
+| Parameter | Description | foreman-installer Parameters |
+| ----------| ----------- | ---------------------------- |
+| `--foreman-log-level` | Log level for Foreman. Overrides `--log-level` for Foreman only. | `--foreman-logging-level` |
+| `--foreman-proxy-log-level` | Log level for Foreman Proxy. Overrides `--log-level` for Foreman Proxy only. | `--foreman-proxy-log-level` |
+
 #### Undetermined
 
 | foreman-installer Parameter | Description | Module | Puppet Parameter | Keep |
@@ -95,7 +110,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-keycloak-app-name` | | |
 | `--foreman-keycloak-realm` | | |
 | `--foreman-logging-layout` | | |
-| `--foreman-logging-level` | | |
 | `--foreman-logging-type` | | |
 | `--foreman-oauth-map-users` | | |
 | `--foreman-plugin-remote-execution-cockpit-ensure` | | |
@@ -165,7 +179,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-proxy-content-enable-ostree` | | | |
 | `--foreman-proxy-http` | | | |
 | `--foreman-proxy-log` | | | |
-| `--foreman-proxy-log-level` | | | |
 | `--foreman-proxy-plugin-ansible-working-dir` | | | |
 | `--foreman-proxy-plugin-dhcp-infoblox-dns-view` | | | |
 | `--foreman-proxy-plugin-dhcp-infoblox-network-view` | | | |
--- a/src/playbooks/_logging/metadata.obsah.yaml
+++ b/src/playbooks/_logging/metadata.obsah.yaml
@@ -1,0 +1,27 @@
+---
+variables:
+  log_level:
+    help: Default log level for all services.
+    parameter: --log-level
+    choices:
+      - debug
+      - info
+      - warn
+      - error
+      - fatal
+  foreman_log_level:
+    help: Log level for Foreman. Defaults to the global log level.
+    choices:
+      - debug
+      - info
+      - warn
+      - error
+      - fatal
+  foreman_proxy_log_level:
+    help: Log level for Foreman Proxy. Defaults to the global log level.
+    choices:
+      - debug
+      - info
+      - warn
+      - error
+      - fatal
--- a/src/playbooks/_logging/metadata.obsah.yaml
+++ b/src/playbooks/_logging/metadata.obsah.yaml
@@ -9,14 +9,6 @@ variables:
       - warn
       - error
       - fatal
-  foreman_log_level:
-    help: Log level for Foreman. Defaults to the global log level.
-    choices:
-      - debug
-      - info
-      - warn
-      - error
-      - fatal
   foreman_proxy_log_level:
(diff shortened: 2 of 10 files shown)


src/playbooks/_logging/metadata.obsah.yaml
```diff
@@ -0,0 +1,27 @@
...
+  log_level:
+    help: Default log level for all services.
+    parameter: --log-level
+    choices:
+      - debug
+      - info
+      - warn
+      - error
+      - fatal
+  foreman_log_level:
```
arvind4501: Since `_logging` is included in both flavors (katello, and proxy), i think `foreman_log_level` will be added to `deploy-proxy --help` and proxy will never have foreman so i think we could strip this and add directly to deploy/metadata.obsah.yaml?

src/roles/foreman/tasks/main.yaml
```diff
@@ -128,6 +128,7 @@
       FOREMAN_PUMA_THREADS_MAX: "{{ foreman_puma_threads_max }}"
       FOREMAN_PUMA_WORKERS: "{{ foreman_puma_workers }}"
       FOREMAN_ENABLED_PLUGINS: "{{ foreman_plugins | join(' ') }}"
+      FOREMAN_LOGGING_LEVEL: "{{ foreman_log_level }}"
```
arvind4501, line 131: where os `FOREMAN_LOGGING_LEVEL` configured, i am not sure if this env will configure foreman log level, i think if we see puppet implementation(https://github.com/theforeman/puppet-foreman/blob/10dafb00096cfa8c162161c00e766a18ca079bec/templates/settings.yaml.erb#L30) we can update settings.yaml for foreman as well, wdyt?

stejskalleos: https://github.com/theforeman/foreman/blob/develop/app/services/foreman/env_settings_loader.rb#L39

arvind4501: ah.. okay, make sense, thanks

src/playbooks/_logging/metadata.obsah.yaml
```diff
@@ -0,0 +1,27 @@
...
+    help: Log level for Foreman. Defaults to the global log level.
+    choices:
+      - debug
+      - info
+      - warn
+      - error
+      - fatal
+  foreman_proxy_log_level:
```
arvind4501, line 12: I am also thinking how would this behave when we don't have foreman-proxy feature enabled on server

src/vars/logging.yml
```diff
@@ -0,0 +1,12 @@
+---
+log_level: info
+
+foreman_log_level: "{{ log_level }}"
+foreman_proxy_log_level: "{{ log_level }}"
+
+_foreman_proxy_log_level_map:
```
evgeni: do we need this map, or can we just use `{{ log_level | upper }}`?

src/roles/foreman/tasks/main.yaml
```diff
@@ -222,6 +224,7 @@
     env:
       SEED_ORGANIZATION: "{{ foreman_initial_organization }}"
       SEED_LOCATION: "{{ foreman_initial_location }}"
+      FOREMAN_LOGGING_LEVEL: "{{ foreman_log_level }}"
```
evgeni, line 227: not directly related to the PR here, but we have a set of env vars we pass to *all* foreman containers. time to consolidate them into a variable so we don't have to edit them in multuple places? same for secrets, I guess?

the `pulp` role has some prior work on this if you look for inspiration

evgeni:
I like this!

evgeni:
Memo to self about Valkey: need to pass `--loglevel X` on the CLI.
Possible values documented in https://github.com/valkey-io/valkey/blob/3886bee1ca595f0fafa576c409fc77871a1226c4/valkey.conf#L342-L349

evgeni:
pulp (according to https://pulpproject.org/pulpcore/docs/admin/reference/settings/#logging and https://docs.djangoproject.com/en/4.2/topics/logging/#loggers) will need some translation in the settings too

src/playbooks/_logging/metadata.obsah.yaml
stejskalleos: From my testing:
```
./foremanctl deploy --tuning development --log-level debug --foreman-proxy-log-level info
```

works fine

src/roles/foreman/tasks/main.yaml
stejskalleos: https://github.com/theforeman/foremanctl/issues/648

arvind4501 (approved):

evgeni (approved):

docs/user/parameters.md
```diff
@@ -95,7 +110,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-keycloak-app-name` | | |
 | `--foreman-keycloak-realm` | | |
 | `--foreman-logging-layout` | | |
-| `--foreman-logging-level` | | |
 | `--foreman-logging-type` | | |
```
ekohl, line 113: Not a blocker, but I think I think we can say that layout and type will likely not be supported.

docs/user/parameters.md
```diff
@@ -165,7 +179,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-proxy-content-enable-ostree` | | | |
 | `--foreman-proxy-http` | | | |
 | `--foreman-proxy-log` | | | |
```
ekohl, line 181: Similar to above. This controls where the proxy logs to (stdout, syslog, journal). I think in a containerized world we won't make that configurable so move this to the Unmapped section?

docs/user/parameters.md
evgeni: Given the author of this PR won't get to fix it quickly, I propose you track this as a separate work item

shubhamsg199:
🍏 Tested with foremanctl-3.0.0.develop-2.20260713052924310495.pr640.79.g9b74dc0.el9.noarch

Scenarios tested:

1. Set a log level for foreman service and verify its correctly set in the parameters file. Also verify the output in foreman journal
2. Set a log level for foreman service and verify its correctly set in the parameters file. Also verify the output in foreman journal
3. Set log level for foreman-proxy without enabling it and verify deployment succeeds
4. Reset the log levels(tested with https://github.com/theforeman/obsah/pull/127/changes) and verify they are not present in parameters file

Automation PR: https://github.com/SatelliteQE/robottelo/pull/22148

shubhamsg199 (approved):

PR #641, opened 2026-07-09
Refactor post_install role

ehelms:

The `Warn about unrecognized Smart Proxy features` task looped over all registered
smart proxies and emitted a verbose skip line for each one when there were no
unrecognized features. Each skip line dumped the full proxy data structure, producing
a large noisy block at the end of every run even when everything was fine.

Additionally, `main.yaml` dispatched to flavor-specific task files (`{{ flavor }}.yaml`)
via a dynamic include, which required a separate file per flavor (`katello.yaml`,
`foreman-proxy-content.yaml`). These files duplicated the "mark installation as complete"
task and made the structure harder to follow.

* Pre-filter smart proxies into a fact before looping, so the warn task only iterates
  over proxies that actually have unrecognized features — eliminating per-item skip output
  in the common (no-problem) case
* Add `loop_control: label` to the warn task so any loop output shows only the proxy name,
  not the full data dict
* Collapse flavor-specific task dispatch in `main.yaml` into explicit, unconditional
  includes for `smart_proxy_features_check.yaml` and `message.yaml`, with the
  flavor-specific debug guarded inline by `when: flavor == 'katello'`
* Rename `katello.yaml` → `message.yaml` to reflect its new flavor-agnostic role
* Delete the now-redundant `foreman-proxy-content.yaml`

Steps to reproduce:

* Deploy with `./foremanctl deploy --foreman-initial-admin-password=changeme --tuning development`
* Verify the post-install output no longer contains a large `skipping` block for the
  smart proxy features check
* Verify the post-install admin credentials message still appears at the end of a katello deployment
* Run with a proxy that has an unrecognized feature and confirm the warning is still emitted,
  showing only the proxy name (not the full dict) in the task label

#### Checklist

### DIFF
--- a/src/roles/post_install/tasks/message.yaml
+++ b/src/roles/post_install/tasks/message.yaml
@@ -1,5 +1,6 @@
 ---
-- name: Admin credentials
+- name: Post install message
+  when: flavor in ['katello']
   ansible.builtin.debug:
     msg:
       - "{{ _post_install_url_msg }}"
@@ -7,10 +8,3 @@
   vars:
     _post_install_url_msg: "Foreman is running at {{ post_install_foreman_server_url }}"
     _post_install_cred_msg: "Admin credentials: {{ post_install_foreman_initial_admin_username }}:{{ post_install_foreman_initial_admin_password }}"
-- name: Mark installation as complete
-  ansible.builtin.copy:
-    dest: "{{ post_install_done_flag }}"
-    content: ''
-    mode: '0640'
-  delegate_to: localhost
-  become: false
--- a/src/roles/post_install/tasks/smart_proxy_features_check.yaml
+++ b/src/roles/post_install/tasks/smart_proxy_features_check.yaml
@@ -14,6 +14,11 @@
     msg: >-
       Proxy {{ item.name }} reports unrecognized features: {{ item.unrecognized_features | join(', ') }}.
       Enable matching features in foremanctl or disable them on the proxy.
-  loop: "{{ _post_install_smart_proxies.resources }}"
-  when:
-    - item.unrecognized_features | default([]) | length > 0
+  loop: "{{ _post_install_proxies_with_unrecognized_features }}"
+  loop_control:
+    label: "{{ item.name }}"
+  vars:
+    _post_install_proxies_with_unrecognized_features: >-
+      {{ _post_install_smart_proxies.resources |
+         selectattr('unrecognized_features', 'defined') |
+         selectattr('unrecognized_features', '!=', []) | list }}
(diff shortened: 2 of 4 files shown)


src/roles/post_install/tasks/message.yaml
```diff
@@ -1,16 +1,10 @@
 ---
-- name: Admin credentials
+- name: Post install message
+  when: flavor == 'katello'
```
evgeni: This will break once we add other Foreman-based flavors (like Satellite), while `ansible.builtin.include_tasks: "{{ flavor }}.yaml"` would still work.

You could do:
```yaml
- name: Include additional tasks for {{ flavor }}
  ansible.builtin.include_tasks: '{{ flavor_file }}'
  when:
    - flavor_file is not none
    - flavor_file != ""
  vars:
    tasks_file: "{{ lookup('ansible.builtin.first_found', [flavor + '.yaml'], errors='ignore') }}"
```

src/roles/post_install/tasks/smart_proxy_features_check.yaml
```diff
@@ -9,11 +9,18 @@
     full_details: true
   register: _post_install_smart_proxies
 
+- name: Set list of Smart Proxies with unrecognized features
+  ansible.builtin.set_fact:
```
evgeni: this can be a `vars` on the below `debug` statement

src/roles/post_install/tasks/message.yaml
ehelms: I could do a `in []` type of check? I'm not sure I see a good reason to have a per-flavor file as there is nothing unique yet.

evgeni: And the list could come from a vendor override? Yeah, that'd work

ehelms: Also, technically `flavor` is a non-role variable :/ Should we consider having some (carefully curated) list of a global variables that we support?

ehelms: > And the list could come from a vendor override? Yeah, that'd work

Why or how do you expect that to work inside role code?

evgeni: Well, I was thinking that we should allow vendors to define own flavors.

OTOH, #628 happens in this repo, so the list of flavors is known… :shrug:

src/roles/post_install/tasks/message.yaml
```diff
@@ -1,16 +1,10 @@
 ---
-- name: Admin credentials
+- name: Post install message
+  when: when: "'{{ flavor }}' in ['katello']"
```
ehelms: I bungled this, will update after finalizing discussion.

src/roles/post_install/tasks/message.yaml
evgeni: Yeah, let's do `flavor in list` and we figure out how the list is conceived later.

ehelms:
@evgeni Ready for re-review

evgeni (approved):

PR #642, opened 2026-07-10
Don't become when checking the existence of the bundle

evgeni:

This allows running without root/sudo on the controller node

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/roles/certificates/tasks/extract.yml
+++ b/src/roles/certificates/tasks/extract.yml
@@ -4,6 +4,7 @@
     path: "{{ certificates_bundle }}"
   register: _certificates_tar_file_path
   delegate_to: localhost
+  become: false
 
 - name: Fail if path to certificate tar file does not exist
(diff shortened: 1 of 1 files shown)


arvind4501:
I wonder It would work fine until the whole foremanctl deployment runs as root, i am curious if we have plans to setup foremanctl as non-root ever. then it could give permission issues, thoughts?

arvind4501 (approved):

evgeni:
> I wonder It would work fine until the whole foremanctl deployment runs as root, i am curious if we have plans to setup foremanctl as non-root ever. then it could give permission issues, thoughts?

If it runs as root (as we do in prod), `become: true` is a noop already.
In CI it works as the CI user can do `sudo` without any password etc.
On my laptop it failed as my user can't do `sudo` without password.

We use the same pattern in the post-install tasks too, so if it ever blows up, then in multiple places ;)

PR #643, opened 2026-07-10
allow overriding of deploy params by a vendor

evgeni:

Vendors might want to override our CLI definition

* Make it so

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/playbooks/_vendor_override/metadata.obsah.yaml
+++ b/src/playbooks/_vendor_override/metadata.obsah.yaml
@@ -1,0 +1,5 @@
+---
+variables:
+  flavor:
+    choices:
+      - satellite
--- a/docs/developer/vendor-overrides.md
+++ b/docs/developer/vendor-overrides.md
@@ -1,0 +1,66 @@
+# Vendor overrides
+
+As a downstream consumer of `foremanctl` you might want to diverge from defaults `foremanctl` offers.
+Here's how.
+
+## Parameters
+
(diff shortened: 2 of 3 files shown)


src/playbooks/_vendor_override/metadata.obsah.yaml
```diff
@@ -0,0 +1,6 @@
+---
+variables:
+  flavor:
+    help: Base flavor to use in this deployment.
```
evgeni: We can drop the `help` key once https://github.com/theforeman/obsah/pull/129 lands

src/playbooks/_vendor_override/metadata.obsah.yaml
```diff
@@ -0,0 +1,6 @@
...
+    choices:
+      - satellite
```
evgeni, line 5: Obviously, this file should not be shipped like this in `foremanctl`, but be provided by a vendor-branding-package

ehelms:
I was thinking this should get merged (the parts that make sense to) and then CI for https://github.com/theforeman/foremanctl/pull/726 should create the vendor file. This would then reflect the model we expect and test this override method?

evgeni:
updated based on latest learnings and experiments from #726

ehelms:
@evgeni I think it's safe to take this out of draft now that we have an implementation.

evgeni:
@ehelms it is!

arvind4501 (approved):

PR #644, opened 2026-07-10
prepare proxy before deployment

arvind4501:
Deploying a smart proxy requires a certificate bundle and OAuth credentials that live on the Foreman server. Previously there was no dedicated step to transfer these to the proxy host before running deploy-proxy, leaving users to handle that manually.

* Add foremanctl prepare-proxy  <fqdn> command — validates that a certificate bundle exists on the Foreman server, then copies it along with the OAuth consumer key and secret to the proxy host over SSH
* Add src/playbooks/prepare-proxy/prepare-proxy.yaml — two-play playbook: first play runs on the Foreman host to validate and fetch the certificate bundle; second play connects to the proxy over SSH to transfer the bundle and OAuth credentials
* Add src/playbooks/prepare-proxy/metadata.obsah.yaml — exposes the command via obsah with a required --hostname parameter
* Move OAuth credential reading into src/roles/pre_install/tasks/foreman-proxy-content.yaml and include it conditionally for the foreman-proxy-content flavor

Steps to reproduce:

* Deploy a Foreman server with foremanctl deploy
* Generate a certificate bundle for the proxy: foremanctl certificate-bundle  <proxy-fqdn>
* Ensure SSH key-based access from the Foreman server to the proxy host is configured
* Run foremanctl prepare-proxy <proxy-fqdn> and verify it completes without errors
* Run foremanctl deploy-proxy  --flavor foreman-proxy-content --certificate-bundle /root/proxy.example.com.tar.gz --foreman-fqdn quadlet.example.com and verify the OAuth credentials are picked up correctly from the transferred files

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -449,9 +449,9 @@ jobs:
           ./foremanctl certificate-bundle proxy.example.com \
             --certificate-source=${{ matrix.certificate_source }} \
             ${{ matrix.certificate_source == 'custom_server' && '--certificate-server-certificate /root/custom-certificates/certs/proxy.example.com.crt --certificate-server-key /root/custom-certificates/private/proxy.example.com.key' || '' }}
-      - name: Fetch certificates bundle from quadlet
+      - name: Prepare proxy host
         run: |
-          ./forge fetch-bundle proxy.example.com
+          ./foremanctl prepare-proxy proxy.example.com
       - name: Deploy content proxy
         run: |
           ./foremanctl deploy-proxy \
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -449,9 +449,9 @@ jobs:
           ./foremanctl certificate-bundle proxy.example.com \
             --certificate-source=${{ matrix.certificate_source }} \
             ${{ matrix.certificate_source == 'custom_server' && '--certificate-server-certificate /root/custom-certificates/certs/proxy.example.com.crt --certificate-server-key /root/custom-certificates/private/proxy.example.com.key' || '' }}
-      - name: Prepare proxy host
+      - name: Fetch certificates bundle from quadlet
         run: |
-          ./foremanctl prepare-proxy proxy.example.com
+          ./forge fetch-bundle proxy.example.com
       - name: Deploy content proxy
         run: |
           ./foremanctl deploy-proxy \
--- a/src/playbooks/prepare-proxy/prepare-proxy.yaml
+++ b/src/playbooks/prepare-proxy/prepare-proxy.yaml
@@ -1,0 +1,67 @@
+---
+- name: Validate and fetch proxy prerequisites
+  hosts: quadlet
+  become: true
+  vars_files:
+    - "../../vars/certificates.yml"
+    - "../../vars/foreman.yml"
+  tasks:
+    - name: Check certificate bundle exists
+      ansible.builtin.stat:
+        path: "{{ certificates_ca_directory }}/bundles/{{ hostname }}.tar.gz"
+      register: _bundle_stat
+
+    - name: Fail if certificate bundle does not exist
+      ansible.builtin.fail:
+        msg: >-
+          Certificate bundle for {{ hostname }} not found at
+          {{ certificates_ca_directory }}/bundles/{{ hostname }}.tar.gz.
+          Run 'foremanctl certificate-bundle --hostname {{ hostname }}' first.
+      when: not _bundle_stat.stat.exists
+
+    - name: Fetch certificate bundle to controller
+      ansible.builtin.fetch:
+        src: "{{ certificates_ca_directory }}/bundles/{{ hostname }}.tar.gz"
+        dest: "{{ obsah_state_path }}/{{ hostname }}.tar.gz"
+        flat: true
+
+    - name: Add proxy host to inventory
+      ansible.builtin.add_host:
+        name: "{{ hostname }}"
+        groups: proxy
+        ansible_connection: ssh
+        ansible_host: "{{ hostname }}"
+        inventory_dir: "{{ inventory_dir }}"
+
+- name: Transfer proxy certificate bundle and oauth credentials to proxy host
+  hosts: proxy
+  become: true
+  vars_files:
+    - "../../vars/foreman.yml"
+  tasks:
+    - name: Check SSH connectivity and authentication to proxy
+      ansible.builtin.ping:
+
+    - name: Copy certificate bundle to proxy
+      ansible.builtin.copy:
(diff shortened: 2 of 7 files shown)


arvind4501:
The prerequisites for `prepare-proxy` is , user can do key based ssh to proxy server from foreman server

evgeni:
What's the envisioned user workflow here, given that the current way of storing parameters limits us pretty hard on having "single foremanctl install manages multiple servers and proxies" working?

Something like:
* on server: `foremanctl deploy`
* on server: `foremanctl certificate-bundle proxy.example.com`
* on server: `foremanctl prepare-proxy --target proxy.example.com`
* on proxy: `foremanctl deploy-proxy …`

arvind4501:
> What's the envisioned user workflow here, given that the current way of storing parameters limits us pretty hard on having "single foremanctl install manages multiple servers and proxies" working?
> 
> Something like:
> 
> * on server: `foremanctl deploy`
> * on server: `foremanctl certificate-bundle proxy.example.com`
> * on server: `foremanctl prepare-proxy --target proxy.example.com`
> * on proxy: `foremanctl deploy-proxy …`

yes, the goal was to get rid of manually copying certificate bundle and passing oauth creds with deploy-proxy

src/playbooks/prepare-proxy/prepare-proxy.yaml
```diff
@@ -0,0 +1,68 @@
...
+- name: Transfer proxy certificate bundle and oauth credentials to proxy host
+  hosts: proxy
+  become: true
+  vars_files:
+    - "../../vars/foreman.yml"
+  tasks:
+    - name: Check SSH connectivity and authentication to proxy
+      ansible.builtin.ping:
+
+    - name: Copy certificate bundle to proxy
```
arvind4501, line 45: this particular step does behave differently for dev and prod setup, idea was to copy the bundle to proxy machine, but that contradicts with dev workflow where we fetch bundle locally on controller and use it

arvind4501: on dev and CI , fetching  bundle would work but in production we need to copy bundle over to proxy machine

src/playbooks/prepare-proxy/prepare-proxy.yaml
```diff
@@ -0,0 +1,67 @@
...
+        src: "{{ foreman_oauth_consumer_key_file }}"
+        dest: "/root/foreman-oauth-consumer-key"
+        mode: "0600"
+        owner: root
+        group: root
+
+    - name: Copy OAuth consumer secret to proxy
+      ansible.builtin.copy:
+        src: "{{ foreman_oauth_consumer_secret_file }}"
+        dest: "/root/foreman-oauth-consumer-secret"
```
ehelms: I don't love this. Copying and storing this value at `/root`. I think these need to go into the state directory or include them in the bundle.

src/playbooks/prepare-proxy/prepare-proxy.yaml
ehelms: Try not to think of this as "dev" and "prod". Instead, think of this as `control node` and `target node`. See https://github.com/theforeman/foremanctl/issues/630

src/playbooks/prepare-proxy/prepare-proxy.yaml
arvind4501: This causes CI failure, because(in context of controller and target), we have single `controller`  managing both `quadlet` and `proxy` targets in Dev and CI. but thats not the case on production, there we have separate proxy machine which itself acts as `controller` so we need to  copy oauth credentials and certificates bundle from foreman machine.

.github/workflows/test.yml
```diff
@@ -449,9 +449,9 @@ jobs:
           ./foremanctl certificate-bundle proxy.example.com \
             --certificate-source=${{ matrix.certificate_source }} \
             ${{ matrix.certificate_source == 'custom_server' && '--certificate-server-certificate /root/custom-certificates/certs/proxy.example.com.crt --certificate-server-key /root/custom-certificates/private/proxy.example.com.key' || '' }}
-      - name: Fetch certificates bundle from quadlet
+      - name: Prepare proxy host
         run: |
-          ./forge fetch-bundle proxy.example.com
+          ./foremanctl prepare-proxy proxy.example.com
```
Gauravtalreja1: Instead of introducing a new subcommand just for preparation, could we leverage the existing one — something like `foremanctl deploy-proxy --prepare`, I'd like to understand the reasoning behind this design, what led us to prefer a new subcommand over extending the existing one?
Adding a separate subcommand for each step feels like unnecessary from the user's perspective.

src/playbooks/prepare-proxy/prepare-proxy.yaml
```diff
@@ -0,0 +1,67 @@
...
+        owner: root
+        group: root
+
+    - name: Copy OAuth consumer secret to proxy
+      ansible.builtin.copy:
+        src: "{{ foreman_oauth_consumer_secret_file }}"
+        dest: "{{ obsah_state_path }}/foreman-oauth-consumer-secret"
+        mode: "0600"
+        owner: root
+        group: root
```
Gauravtalreja1, line 67: Once the certificate bundle and OAuth keys are copied, could we also print a ready-to-run `foremanctl deploy-proxy..` command to the console? This would mirror the experience users are familiar with from
`capsule-certs-generate`, which outputs the exact install command with all the necessary parameters filled in.

.github/workflows/test.yml
arvind4501: my reasoning to pick a separate `prepare-proxy` command  is that it keeps two distinct tasks seprate as below:-

- preparing proxy machine for deployment
- actually deploying proxy

> Adding a separate subcommand for each step feels like unnecessary from the user's perspective.

we are not adding subcommand for each task here rather `prepare-proxy` is the bundle of tasks(copying oauth credentials and certificate bundle) that helps to take care of the whole prepration we need to deploy the proxy.

src/playbooks/prepare-proxy/prepare-proxy.yaml
arvind4501: we could, yes, but i think after copying we eliminate the need of passing oauth credentianls(directly slurp from copied path) so essentially user only need to pass certificate bundle and foreman fqdn

ehelms:
@arvind4501 Could you add docs updates? That would help understand the user flow.

arvind4501:
Closing in favor of https://github.com/theforeman/foremanctl/pull/698

evgeni:
Thanks for the experiment!

PR #646, opened 2026-07-10
don't include the certs/ directory in the bundle

evgeni:

this makes it easer to unarchive it on the target

*

Steps to reproduce:

*

#### Checklist


evgeni:
This will require an update to https://github.com/SatelliteQE/robottelo/pull/21325 once merged

arvind4501:
Should the https://github.com/theforeman/foremanctl/blob/master/docs/user/certificates.md?plain=1#L168 be updated and removed `/certs` from bundle path too?

evgeni:
Good catch!

PR #647, opened 2026-07-12
Add foreman-rake wrapper

ShimShtein:

Fixes https://redhat.atlassian.net/browse/SAT-47063: We want to make sure the `foreman-rake` option still exists in containerized deployments.

* Add a `foreman-rake` wrapper
* Make sure to reuse the foreman container name

Steps to reproduce:

* Install the foreman
* run `foreman-rake <your command>` from the host.

#### Checklist


ShimShtein:
Converted to draft, since I want to explore the possibility of adding the wrapper in the foremanctl rpm package.

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -0,0 +1,17 @@
...
+if [ "$(id -u)" -ne 0 ]; then
+    exec sudo "$0" "$@"
+fi
+
+cat >&2 <<'EOF'
+NOTICE: Files created by rake tasks are stored inside the foreman container.
+        Copy them to the host with: podman cp {{ foreman_container_name }}:<container-path> <host-path>
+EOF
+
+TTY_ARGS=()
```
evgeni, line 55: Why is this an array? Do you expect to ever have to add anything to it?

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+if [ "$(id -u)" -ne 0 ]; then
+    exec sudo "$0" "$@"
+fi
```
evgeni, line 50: Given the file lives in sbin, I'd argue we can assume the caller is root?

ShimShtein: I have tried it as is on the vagrant machine, and I was able to call it with vagrant user, so switching to root to find the container is needed anyway.

src/roles/foreman/templates/foreman-rake.j2
ShimShtein: There is a caveat around `-t` flag and running the script with or without terminal. For now I didn't see the need for it, but another AI suggested adding the `-t` flag conditionally, so I left the option to do it more easily

src/roles/foreman/templates/foreman-rake.j2
evgeni: why the F is `/usr/sbin` in the `PATH` of a regular user, weird

evgeni: TIL: Fedora and EL do that on purpose: https://src.fedoraproject.org/rpms/setup/blob/rawhide/f/profile#_45

src/roles/foreman/templates/foreman-rake.j2
evgeni: That's what the `if` in the next line is about. I am fine with that.

My question is why we don't do
```shell
TTY_ARGS=""
if [ -t 0 ]; then
    TTY_ARGS="-it"
fi
exec podman exec "${TTY_ARGS}" {{ foreman_container_name }} foreman-rake "$@"
```

evgeni: Or did you meant always doing `-i` but only sometimes `-it`?

ShimShtein:
Switched to RPM to deliver the script here: https://github.com/theforeman/foreman-packaging/pull/13717.

src/roles/foreman/templates/foreman-rake.j2
ekohl: I'd prefer erroring out if the user isn't root.

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -0,0 +1,17 @@
...
+NOTICE: Files created by rake tasks are stored inside the foreman container.
+        Copy them to the host with: podman cp {{ foreman_container_name }}:<container-path> <host-path>
+EOF
+
+TTY_ARGS=()
+if [ -t 0 ]; then
+    TTY_ARGS=(-it)
+fi
+
+exec podman exec "${TTY_ARGS[@]}" {{ foreman_container_name }} foreman-rake "$@"
```
ehelms, line 60: Inception...foreman-rake calling foreman-rake that are the same-same but different. That may cause confusion for developers, but I suspect it shouldn't affect users.

ehelms:
This introduces a secondary script / tool instead of embracing `foremanctl` as our interface. Are we OK with that design because this is a low-level tool that "users" shouldn't ever use? As in, this should only be used by developers / users doing deep debugging?

ekohl:
> This introduces a secondary script / tool instead of embracing `foremanctl` as our interface. Are we OK with that design because this is a low-level tool that "users" shouldn't ever use? As in, this should only be used by developers / users doing deep debugging?

In a private conversation I've said I'm firmly in the "you shouldn't use this unless you know what you're doing" camp.

Ideally we'd expose all supported procedures in `foremanctl` or the appropriate tool, but having this tool is a "hack" to allow us to release sooner. The tool should reflect that.

What I'd propose is a whitelist of "supported" commands. That list should be taken from the docs:
```console
$ rg '# foreman-rake' | awk '{ print $3 }' | sort -u
audits:anonymize
audits:expire
audits:list_attributes
console
facts:clean
foreman_openscap:bulk_upload:default
foreman_tasks:cleanup
interfaces:clean
katello:delete_orphaned_content
katello:migrate_deb_content_attributes_to_pulp3
katello:upgrade_check
ldap:refresh_usergroups
orchestration:dhcp:add_missing
orchestration:dhcp:remove_offending
permissions:reset
reports:expire
reports:_My_Frequency_
rh_cloud_inventory:report:generate
telemetry:metrics
templates:cleanup
```

Then in pseudo-Python code:
```python
if sys.argv[0] not in SUPPORTED_TASKS:
    print("Use of foreman-rake is not supported. Use with caution", file=sys.stderr)
```

src/roles/foreman/templates/foreman-rake.j2
ShimShtein: Actually it makes sense: the same command (let's take for example `foreman-rake console`: no matter where you call i from, it will do the same thing - it will show you a rails console to your server.

ShimShtein:
Addressed the issues, should be ready for another round

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -0,0 +1,57 @@
...
+        "interfaces:clean"
+        "katello:delete_orphaned_content"
+        "katello:migrate_deb_content_attributes_to_pulp3"
+        "katello:upgrade_check"
+        "ldap:refresh_usergroups"
+        "orchestration:dhcp:add_missing"
+        "orchestration:dhcp:remove_offending"
+        "permissions:reset"
+        "reports:expire"
+        "reports:_My_Frequency_"
```
ekohl: The documentation convention is to use underscores for what the user inputs. That won't work in this script and you'll need to expand it to all the options we have. https://docs.theforeman.org/nightly/Administering_Project/index-katello.html#Testing_Email_Notifications_admin lists those.

Today it's also in a procedure we don't publish today. Long term we should build something into the application itself, if we don't already have that. But that goes for most of these tasks.

tests/feature/foreman/foreman_rake_test.py
ekohl, line 1: Since b61642c6630109eb2e06b1cad00d7c091020b631 we have feature directories. Please move it into `tests/feature/foreman` and drop `pytestmark`.

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    local joined_pattern
+    joined_pattern=$(IFS="|"; echo "${allowed_actions[*]}")
+
+    local pattern="(^|[[:space:]])($joined_pattern)([[:space:]]|$)"
+
+    if [[ ! "$*" =~ $pattern ]]; then
+        cat >&2 <<'EOF'
+NOTICE: Unrecognized action detected, please use with caution.
```
ekohl: Please upgrade this to warning and explicitly state it's unsupported.

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -0,0 +1,57 @@
...
+NOTICE: Files created by rake tasks are stored inside the foreman container.
+        Copy them to the host with: podman cp {{ foreman_container_name }}:<container-path> <host-path>
```
ekohl: I don't think we should mention this. Please make sure the supported actions don't do any file operations and then consider the rest unsupported.
```suggestion
```

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+validate_actions() {
+    # This list is generated by running:
+    # rg '# foreman-rake' | awk '{ print $3 }' | sort -u
+    # against the documentation.
```
ehelms: `the documentation` is rather ambiguous, can this be more concrete? Like including the repository link.

src/roles/foreman/tasks/main.yaml
```diff
@@ -109,7 +109,7 @@
 
 - name: Deploy Foreman Container
   containers.podman.podman_container:
-    name: "foreman"
+    name: "{{ foreman_container_name }}"
```
ehelms, line 112: This change seems unrelated and unnecesary.

ShimShtein: I'm using it in the script: https://github.com/theforeman/foremanctl/pull/647/changes#diff-e88ccd8fc4cf101614b443c40e5d117b52c4d21970af9f1c7fdd511c9722d700R54

I want to run the command against a container with the same name. I think we should use it as a parameter (a const to be more precise) rather than hard coding it in two different places.

src/roles/foreman/defaults/main.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 foreman_container_image: "quay.io/foreman/foreman"
 foreman_container_tag: "nightly"
+foreman_container_name: foreman
```
ehelms, line 4: This reminds me that we should consider moving our roles to having `defaults` and `vars`. Where defaults represent the "public API" for the role, and `vars` represent the "constants" for the role. I'll file an issue for further discussion around this design principle.

tests/feature/foreman/foreman_rake_test.py
```diff
@@ -0,0 +1,10 @@
+def test_foreman_rake_wrapper_exists(server):
+    wrapper = server.file('/usr/sbin/foreman-rake')
+    assert wrapper.exists
+    assert wrapper.mode == 0o755
+
+
+def test_foreman_rake_list_tasks(server):
+    result = server.run('foreman-rake -T')
```
ehelms: I am surprised `-T` is allowed based on the code. Could you add a negative test that ensures that something *not* in the list is properly blocked?

tests/feature/foreman/foreman_rake_test.py
```diff
@@ -0,0 +1,10 @@
...
+    assert result.succeeded
+    assert 'rake cron:daily' in result.stdout
```
ehelms, line 11: This kinda shows a goofy element of this -- users can list the set of rake tasks but then are limited by the script on what can actually be ran. I'd rather we block `-T` for example, or override it to only list the actions that are permitted.

tests/feature/foreman/foreman_rake_test.py
ShimShtein: It will not be blocked, it will just show a warning. I would advise against hard-block.

tests/feature/foreman/foreman_rake_test.py
ShimShtein: That's an option, but I am not sure it's worth investing in it for the long run, since we will want to replace the `foreman-rake` entirely.

tests/feature/foreman/foreman_rake_test.py
ehelms: Oh it just says a warning? I don't forsee that doing much, especially if the rake command scrolls the warning off-screen. @ekohl This was from your original suggestion, and you did indicate warning there. Should we just go with "you can only do what's in the allowed list"  rather than a warning?

"Power users" can always exec into the container to run whatever they want anyway.

ShimShtein: Another option is to fail unless an `--i-know-what-i-am-doing` flag is set. But again, how much time do we want to spend on it, if in the end we want to remove this whole command altogether.

ehelms: Do we have a plan or design or timeline written out for removing it? Otherwise "in the end" might be years. :grin:

ShimShtein:
Waiting for @ekohl's response about hard failing for unsupported tasks.

tests/feature/foreman/foreman_rake_test.py
ekohl: I like making it exit hard instead of a warning. Didn't think of that. Perhaps use an environment variable to bypass it instead of a command line argument? I think manipulating those is a pain.

> Do we have a plan or design or timeline written out for removing it? Otherwise "in the end" might be years. 😁

I wonder if we should create a feature to track the removal. Each command that we whitelist could be an epic to solve it properly. Which also makes me think, should we deprecate it immediately in our containerized release notes?

tests/feature/foreman/foreman_rake_test.py
ekohl: I like the idea. If we want to limit it, really prevent users from using it.

src/roles/foreman/templates/foreman-rake.j2
ShimShtein: Apparently when we do not use an array (as you suggested) it will fail to execute the command without tty. Git an error with this code and the AI suggested to use an array notation. Added a comment to describe this behavior.

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -0,0 +1,58 @@
...
+        "templates:cleanup"
+    )
+
+    local joined_pattern
+    joined_pattern=$(IFS="|"; echo "${allowed_actions[*]}")
+
+    local pattern="(^|[[:space:]])($joined_pattern)([[:space:]]|$)"
+
+    if [[ ! "$*" =~ $pattern ]]; then
+        echo "WARNING: Unsupported action detected. Please use with caution." >&2
```
ekohl: Please tune this message so it's a warning when `ALLOW_UNSUPPORTED` is true and otherwise an error.

ShimShtein:
Ready for another round

ehelms (approved):
Looks good from the users perspective for me.

ekohl (approved):

ekohl:
I should mention my approval was based on reading the code. I'll leave it open so someone can actually test this. Though it also has tests that should cover most already.

ehelms:
Given we have a test that does run one allowed command and tests that it executes, I am going to opt to merge this. This then should allow further automated testing to pick up the existence and check for pass / failure across those workflows.

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+validate_actions() {
+    # This list is generated by running:
+    # rg '# foreman-rake' | awk '{ print $3 }' | sort -u
+    # against the foreman documentation repository (https://github.com/theforeman/foreman-documentation).
+    # The intention is to have a list of tasks that were mentioned explicitly in the documentation.
+    local allowed_actions=(
```
evgeni, line 8: I am late to the party, but how does the user learn about this list? Neither `foreman-rake -h` nor `foreman-rake -T` work anymore.

ShimShtein: The idea is that the user looks up in the documentation for a command (and those are allowed). Everything else is rejected. We can add a comment along the lines "if you know better, use the `ALLOW_UNSUPPORTED` flag", but I am not sure we want to encourage people using it.

PR #651, opened 2026-07-13
Support non-root Candlepin by running as tomcat with owned secrets

Alleny244:
## Why are you introducing these changes? (Problem description, related links)

Candlepin is moving to a non-root container user (`tomcat`). By default, Podman mounts secrets as `root:root` with mode `0440`, which prevents the `tomcat` user from reading the mounted certificates and configuration files unless it is added to the `root` group inside the container.

This PR updates `foremanctl` to run Candlepin as `tomcat` and sets the ownership of secret mounts to the same user, enabling the companion `candlepin-oci-images` change (`USER tomcat`, no `root` group membership).

- Add `candlepin_container_user: tomcat` as the default.
- Configure the Candlepin Quadlet to run with `User=tomcat`.
- Add `user.yml` to determine the `tomcat` UID/GID from the container image at deploy time (`podman run ... id -u/-g`).
- Set the `uid`/`gid` on all Candlepin secret mounts so the mounted files are owned by `tomcat` (mode `0440`).
- Add `test_candlepin_runs_as_tomcat` to verify Candlepin runs as `tomcat` and can read the mounted secrets.

## Related
- candlepin-oci-images: (https://github.com/theforeman/candlepin-oci-images/pull/52) (SHOULD  BE MERGED FIRST)

### Deploy

Use a Candlepin image built with the companion `candlepin-oci-images` change (`USER tomcat`):

```bash
./foremanctl deploy --foreman-initial-admin-password=changeme --tuning development
```

### Verify

```bash
podman exec candlepin id -un
# tomcat

podman exec candlepin ls -l /etc/candlepin/candlepin.conf

podman exec candlepin ls -l /etc/candlepin/certs/

podman exec candlepin test -r /etc/candlepin/certs/tomcat.key

curl -k https://localhost:23443/candlepin/status
# HTTP 200
```

### Run automated tests

```bash
./forge test --pytest-args="tests/candlepin_test.py -vv"
```

> **Note:** The non-root behavior is fully validated only with the companion `candlepin-oci-images` PR. Deployments continue to work with the current image, but the container will still run as the existing image user.

## Checklist

- [x] Tests added/updated (if applicable)
- [ ] Documentation updated (if applicable)

### DIFF
--- a/src/roles/candlepin/defaults/main.yml
+++ b/src/roles/candlepin/defaults/main.yml
@@ -11,6 +11,7 @@ candlepin_ciphers:
   - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
 candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "4.4.14"
+candlepin_secret_mount_opts: "mode=0440,uid=0,gid=53,type=mount"
 
 candlepin_database_host: localhost
 candlepin_database_port: 5432
--- a/src/roles/candlepin/tasks/main.yml
+++ b/src/roles/candlepin/tasks/main.yml
@@ -77,16 +77,16 @@
     network: host
     hostname: "{{ ansible_facts['hostname'] }}.local"
     secrets:
-      - 'candlepin-ca-cert,target=/etc/candlepin/certs/candlepin-ca.crt,mode=0440,type=mount'
-      - 'candlepin-ca-key,target=/etc/candlepin/certs/candlepin-ca.key,mode=0440,type=mount'
-      - 'candlepin-tomcat-cert,target=/etc/candlepin/certs/tomcat.crt,mode=0440,type=mount'
-      - 'candlepin-tomcat-key,target=/etc/candlepin/certs/tomcat.key,mode=0440,type=mount'
-      - 'candlepin-candlepin-conf,target=/etc/candlepin/candlepin.conf,mode=0440,type=mount'
-      - 'candlepin-tomcat-server-xml,target=/etc/tomcat/server.xml,mode=0440,type=mount'
-      - 'candlepin-tomcat-conf,target=/etc/tomcat/tomcat.conf,mode=0440,type=mount'
-      - 'candlepin-tomcat-logging-properties,target=/etc/tomcat/logging.properties,mode=0440,type=mount'
-      - 'candlepin-logback-xml,target=/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/logback.xml,mode=0440,type=mount'
-      - 'candlepin-db-ca,target={{ candlepin_database_ssl_ca_path }},mode=0440,type=mount'
+      - 'candlepin-ca-cert,target=/etc/candlepin/certs/candlepin-ca.crt,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-ca-key,target=/etc/candlepin/certs/candlepin-ca.key,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-tomcat-cert,target=/etc/candlepin/certs/tomcat.crt,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-tomcat-key,target=/etc/candlepin/certs/tomcat.key,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-candlepin-conf,target=/etc/candlepin/candlepin.conf,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-tomcat-server-xml,target=/etc/tomcat/server.xml,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-tomcat-conf,target=/etc/tomcat/tomcat.conf,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-tomcat-logging-properties,target=/etc/tomcat/logging.properties,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-logback-xml,target=/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/logback.xml,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-db-ca,target={{ candlepin_database_ssl_ca_path }},{{ candlepin_secret_mount_opts }}'
     quadlet_options:
       - |
         [Install]
--- a/tests/feature/katello/candlepin_test.py
+++ b/tests/feature/katello/candlepin_test.py
@@ -9,6 +9,23 @@ def test_candlepin_service(server):
     assert candlepin.is_running
 
 
+def test_candlepin_runs_as_tomcat(server):
+    assert server.run("podman exec candlepin id -un").stdout.strip() == 'tomcat'
+    assert server.run("podman exec candlepin id -u").stdout.strip() != '0'
+
+    groups = server.run("podman exec candlepin id -Gn").stdout.split()
+    assert 'tomcat' in groups
+    assert 'root' not in groups
+
+    assert server.run("podman exec candlepin test -r /etc/candlepin/certs/tomcat.key").succeeded
+    assert server.run("podman exec candlepin test -r /etc/tomcat/tomcat.conf").succeeded
+
+    secret_ownership = server.run(
+        "podman exec candlepin stat -c '%U:%G %a' /etc/candlepin/certs/tomcat.key"
+    ).stdout.strip()
+    assert secret_ownership == 'root:tomcat 440'
+
+
 def test_candlepin_port(server):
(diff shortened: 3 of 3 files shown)


src/roles/candlepin/tasks/user.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Resolve candlepin container UID from image
```
ehelms: Can we just hardcode this as a known value since we control building the container and can either know it or set it and avoid this look up?

Alleny244: yeah sure

tests/feature/katello/candlepin_test.py
```diff
@@ -14,6 +14,12 @@ def test_candlepin_service(server):
     assert candlepin.is_running
 
 
+def test_candlepin_runs_as_tomcat(server):
```
archanaserver, line 12: Could we also assert that tomcat has a non-zero UID and is not a member of GID 0?

archanaserver (approved):

src/roles/candlepin/tasks/main.yml
```diff
@@ -77,17 +77,20 @@
...
+      - 'candlepin-candlepin-conf,target=/etc/candlepin/candlepin.conf,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-tomcat-server-xml,target=/etc/tomcat/server.xml,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-tomcat-conf,target=/etc/tomcat/tomcat.conf,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-tomcat-logging-properties,target=/etc/tomcat/logging.properties,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-logback-xml,target=/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/logback.xml,{{ candlepin_secret_mount_opts }}'
+      - 'candlepin-db-ca,target={{ candlepin_database_ssl_ca_path }},{{ candlepin_secret_mount_opts }}'
     quadlet_options:
+      - |
+        [Container]
+        User={{ candlepin_container_user }}
```
evgeni: Should we set this in https://github.com/theforeman/candlepin-oci-images/blob/master/images/candlepin/Containerfile, so all users of the container benefit from it?

(Not blocking, we can change it here, see it works, change it in the container and then remove the two lines here again)

evgeni: Hah, just realized you already have https://github.com/theforeman/candlepin-oci-images/pull/52 open. Nice!

tests/feature/katello/candlepin_test.py
Alleny244: yeah, added

archanaserver (approved):
LGTM

src/roles/candlepin/defaults/main.yml
```diff
@@ -11,6 +11,8 @@ candlepin_ciphers:
   - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
 candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "4.4.14"
+candlepin_container_user: tomcat
+candlepin_secret_mount_opts: "mode=0440,uid=53,gid=53,type=mount"
```
ekohl: Would it be safer to use `uid=0` so root owns the file and the `candlepin` user can't change the permissions? It can still read it then via the group.

Alleny244: I hadn't thought about that, but that's a good point. Having the secrets owned by root with uid=0,gid=53,mode=0440 is a better approach since tomcat can still read them through the group without being able to change their permissions. I'll update the PR.

evgeni:
"migration" test failure unrelated, explanation and possible fix in https://github.com/theforeman/foremanctl/pull/661

src/roles/candlepin/tasks/main.yml
ekohl: If the image change is merged, this should be dropped, right?

evgeni: Correct.

Alleny244: okay, got it.  Can I drop it then, assuming the other pr would be merged first?

ekohl: You'll need to update `candlepin_container_tag` as well, unless that tag gets updated with the non-root container. Though I think that's a bad idea because it will break existing deployments in Foreman 3.19.

evgeni: The changes in this PR will never reach 3.19 deployments.
And we're using nightly here already: https://github.com/theforeman/foremanctl/blob/688d3da8e4bc7a74136b8260e2133ada7a43ab7a/src/vars/images.yml#L4-L6

I've just merged the oci PR, once Konflux has built it to https://quay.io/repository/foreman/candlepin?tab=tags&tag=latest you can drop this line and CI will verify it worked

evgeni: huh, that got built, but the `foreman-nightly` tag is not updated, only `nightly`

@Odilhao when is the `foreman-nightly` tag updated on those?

ekohl: > And we're using nightly here already:

Oh yes, I was looking at the role defaults:
https://github.com/theforeman/foremanctl/blob/688d3da8e4bc7a74136b8260e2133ada7a43ab7a/src/roles/candlepin/defaults/main.yml#L13

Odilhao: > huh, that got built, but the `foreman-nightly` tag is not updated, only `nightly`
> 
> @Odilhao when is the `foreman-nightly` tag updated on those?

It's already updated, we have `quay.io/foreman/candlepin:nightly` pushed 3 hours ago. The container version tag will only be update if someone update the Containerfile at https://github.com/theforeman/candlepin-oci-images/blob/foreman-3.19/images/candlepin/Containerfile . We don't have `branched` versions on nightly atm, the candlepin version/tag is tied to the foreman version.

evgeni: But we're using `foreman-nightly`, not `nightly` in https://github.com/theforeman/foremanctl/blob/688d3da8e4bc7a74136b8260e2133ada7a43ab7a/src/vars/images.yml#L4-L6

What updates the `foreman-nightly` tag?

Odilhao: Any merge that you do on `foreman-nightly` the image is built and published by Konflux right after.

This is the tag that we are using now https://gitlab.com/fedora/infrastructure/konflux/tenants-config/-/blob/main/clusters/kflux-fedora-01/tenants/theforeman-org-tenant/candlepin/releaseplans/develop/kustomization.yaml?ref_type=heads#L30

Should we also add `foreman-nightly` alongside `nightly` as tag?

Odilhao: Just added here https://gitlab.com/fedora/infrastructure/konflux/tenants-config/-/merge_requests/212, I'll make sure it's there in one hour after everything syncs

Odilhao: You can now use `quay.io/foreman/candlepin:foreman-nightly` 

```
podman pull quay.io/foreman/candlepin:foreman-nightly
Trying to pull quay.io/foreman/candlepin:foreman-nightly...
Getting image source signatures
Copying blob 0a3d34aa24dc done   | 
Copying blob 452eff874a29 done   | 
Copying config f7a84f58b6 done   | 
Writing manifest to image destination
f7a84f58b60c2c0a002523d70594e708264d17d5b3c652d9357d49028202592d
```

evgeni: I don't necessarily say this needs to be updated at the same time you tag a new candlepin (or pulp!) nightly. More when there is a new candlepin (or pulp) that should go *with* foreman nightly (e.g. after a pipeline run that verifies those are compatible, or whatever).

src/roles/candlepin/defaults/main.yml
```diff
@@ -11,6 +11,7 @@ candlepin_ciphers:
   - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
 candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "4.4.14"
```
Odilhao, line 13: Right now our container on nightly is built with `4.8.2`, do we need to update it here?

evgeni: No

ekohl:
https://github.com/theforeman/foremanctl/pull/657 was merged and introduced a merge conflict. My apologies.

PR #652, opened 2026-07-13
Add ostree content type support

ianballou:

* Adds support for enabling the ostree content type, which Katello supports.
  * The ostree type is optional.

There is some mention in the docs about the ostree plugin not migrating properly - I have not touched the migration mapping logic, but this should help unblock that.

Steps to reproduce:

* Deploy a machine and check that you can create ostree repos.

#### Checklist

### DIFF
--- a/docs/user/parameters.md
+++ b/docs/user/parameters.md
@@ -110,6 +110,7 @@ There are multiple use cases from the users perspective that dictate what parame
 | Parameter | Description | foreman-installer Parameters |
 | --------- | ----------- | ---------------------------- |
 | `--add-feature bmc` | Enable BMC feature | `--foreman-proxy-bmc` |
+| `--add-feature content/ostree` | Enable OSTree content type | `--foreman-proxy-content-enable-ostree` |
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
 
@@ -162,7 +163,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-proxy-plugin-remote-execution-script-mode` | | foreman_proxy::plugin::remote_execution_script | mode |
 | `--foreman-proxy-plugin-openscap-ansible-module` | | foreman_proxy::plugin::openscap | ansible_module |
 | `--foreman-proxy-plugin-openscap-puppet-module` | | foreman_proxy::plugin::openscap | puppet_module |
-| `--foreman-proxy-content-enable-ostree` | | | |
 | `--foreman-proxy-http` | | | |
 | `--foreman-proxy-log` | | | |
 | `--foreman-proxy-log-level` | | | |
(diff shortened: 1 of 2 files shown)


src/features.d/content.yaml
```diff
@@ -24,3 +24,7 @@ content/python:
   internal: true
   dependencies:
     - pulp
+content/ostree:
```
ianballou, line 27: This name matches the other content types, but is it strange to keep the slash in there if it's user configurable?

src/features.d/content.yaml
```diff
@@ -24,3 +24,7 @@ content/python:
...
+  description: OSTree content type for Pulp
```
vijaysawant, line 28: Does Katello have defalut support for OS tree content type? any specific reason for omitting it ?

docs/user/parameters.md
```diff
@@ -110,6 +110,7 @@ There are multiple use cases from the users perspective that dictate what parame
 | Parameter | Description | foreman-installer Parameters |
 | --------- | ----------- | ---------------------------- |
 | `--add-feature bmc` | Enable BMC feature | `--foreman-proxy-bmc` |
+| `--add-feature content/ostree` | Enable OSTree content type | `--foreman-proxy-content-enable-ostree` |
```
sjha4, line 113: Is this functional?

src/features.d/content.yaml
ianballou: Katello does not enable OSTree by default since it's one of the lesser known content types. You could almost say it's in a permanent state of tech preview.

docs/user/parameters.md
ianballou:
```
$ ./foremanctl features
FEATURE                   STATE        DESCRIPTION
foreman                   enabled      Base Foreman Server
...
content/ostree            enabled      OSTree content type for Pulp
...

```

```
# OBSAH_STATE=/var/lib/foremanctl/ ./foremanctl deploy --add-feature content/ostree
[DEPRECATION WARNING]: community.postgresql.postgresql_set has been deprecated. Use postgresql_alter_system instead. This feature will 
be removed from community.postgresql in version 5.0.0. Deprecation warnings can be disabled by setting deprecation_warnings=False in 
ansible.cfg.
[WARNING]: Collection ansible.posix does not support Ansible version 2.15.13

PLAY [Setup machine] ********************************************************************************************************************

TASK [Gathering Facts] ******************************************************************************************************************
ok: [localhost]

```

Yep! 

My quadlet machine has ostree on it now.

ianballou: I put this under the smart proxy section because the old code was enabled via "foreman proxy content".

src/features.d/content.yaml
vijaysawant: users need to explicitly enable it. Okay thanks.

vijaysawant (approved):

ianballou:
Tests seem to have failed because of a (transient?) issue with Galaxy.

sjha4 (approved):
Ack..The parameter works as expected..The workflow is basically default katello doesn't install ostree by default driven by src/vars/flavors/katello.yml and users can add ostree using --add-feature. Ack 👍🏼

ehelms (approved):

PR #654, opened 2026-07-14
Unset proxy environment variables before running Ansible

Gauravtalreja1:
When proxy environment variables (`http_proxy`/`HTTPS_PROXY`) are set on the host, Ansible tasks that make HTTP calls (`ansible.builtin.uri`,`theforeman.foreman` modules) route through the proxy and fail with `Tunnel connection failed: 503 Service Unavailable`. This affects health checks, smart proxy registration, and service readiness checks.
This is the same approach used by foreman-installer, which unsets proxy variables in `hooks/boot/05-environment.rb` before running Puppet.
- https://github.com/theforeman/foreman-installer/blob/develop/hooks/boot/05-environment.rb

Image pulls are unaffected, they go through Podman/systemd, which reads proxy settings from `/etc/containers/containers.conf`, not from the shell environment.

* Unset `http_proxy`, `HTTP_PROXY`, `https_proxy`, and `HTTPS_PROXY` in the`foremanctl` wrapper script before invoking Ansible, so no playbook inherits proxy settings from the host environment.

* Deploy on a host with `https_proxy` set in the environment
* Verify `foremanctl deploy` completes without proxy-related errors
* Verify container image pulls still work (they use `containers.conf`)
* Verify health checks pass: `foremanctl health`
 
Steps to reproduce:
* Set proxy on the host: `export https_proxy=http://proxy.example.com:3128`
* Run `foremanctl deploy --foreman-initial-admin-password=changeme`
* Observe failures at "Wait for Foreman service to be accessible" or "Configure Foreman Proxy" with `Tunnel connection failed: 503`

#### Checklist


src/playbooks/backup/backup.yaml
```diff
@@ -3,6 +3,11 @@
   hosts: "{{ target_host | default('quadlet') }}"
   become: true
   gather_facts: true
+  environment:
```
ehelms: Is there a way to make this block re-usable across playbooks?

Gauravtalreja1: Good point, I could move this to the foremanctl wrapper script instead, that would cover all playbooks with a single unset. The user's environment wouldn't be affected since the script runs as a child process. What do you think?

evgeni: I'd put it into `foremanctl` too. Easier to not forget again.

evgeni (approved):

Gauravtalreja1:
PRT passed on https://github.com/SatelliteQE/robottelo/pull/22125 (remaining failures are unrelated to this changes)
And, migration failure is probably due to https://github.com/theforeman/foremanctl/pull/661

PR #655, opened 2026-07-14
Mark slow tests to allow for quicker iteration

ekohl:

While working on https://github.com/theforeman/foremanctl/pull/511 I wanted to quickly run tests again and again. The full service restart is slow and not really something I was interested in.

You can now skip slow tests with:

    pytest -k "not slow"

Steps to reproduce:

* Watch CI and verify it still runs the slow tests
* Manually run pytest to skip the slow tests

#### Checklist

### DIFF
--- a/tests/target_lifecycle_test.py
+++ b/tests/target_lifecycle_test.py
@@ -1,8 +1,12 @@
 import time
 
+import pytest
+
 TARGET_ACTIVE_RETRIES = 90
 TARGET_ACTIVE_DELAY = 10
 CURL_CMD = "curl --silent --output /dev/null"
+
+pytestmark = pytest.mark.slow
 
 
 def _wait_for_target_active(server, target="foreman.target"):
--- a/docs/developer/testing.md
+++ b/docs/developer/testing.md
@@ -139,9 +139,7 @@ These fixtures create a resource before the test and delete it afterward:
 
 ### Service test
 
-Check that a systemd unit is running and its port is reachable:
-
-For example:
+Check that a systemd unit is running and its port is reachable using testinfra's [service](https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.service.Service) and [addr](https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.addr.Addr) modules:
 
 ```python
 def test_service_running(server):
--- a/docs/developer/testing.md
+++ b/docs/developer/testing.md
@@ -66,6 +66,12 @@ pytest tests/postgresql_test.py
 pytest tests/foreman_test.py::test_foreman_service
 ```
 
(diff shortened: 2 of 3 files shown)


tests/target_lifecycle_test.py
```diff
@@ -1,5 +1,7 @@
 import time
 
+import pytest
+
```
arvind4501, line 4:
```suggestion
pytestmark = pytest.mark.slow
```
we only have these two tests and both are marked slow so marking entire file slow make sense

arvind4501:
I wonder if these are the only tests which are slow, I was thinking if there is a more definative way to figure out slow tests and mark them accordingly? and seems like pytest has a way to figure that out https://docs.pytest.org/en/stable/how-to/usage.html#profiling-test-execution-duration

tests/target_lifecycle_test.py
ekohl: Works for me. Done.

ekohl:
We even have that enabled in our test playbook:
https://github.com/theforeman/foremanctl/blob/57966bd9d3fd9a5e796afaaad5855264d02ddcf5/development/playbooks/test/test.yaml#L46

arvind4501 (approved):

ekohl:
Migration is probably #661, the other is failing on:
```
RUNNING HANDLER [iop_kafka : Restart kafka] ************************************
Error: : Task failed: Failed to connect to the host via ssh: ssh: connect to host 192.168.121.253 port 22: No route to host
```

I don't see how that could be caused by my changes.

arvind4501:
> I don't see how that could be caused by my changes.

It  not i think, failure seems unrelated

ehelms:
A note in our developer documentation about our use of `slow` and when to apply it, or at least how to think about when and when not to apply it would be helpful.

ekohl:
I added some documentation, but wasn't quite sure about how to describe when to use the marker. I'm not sure if we need to provide that context now.

ehelms:
> I added some documentation, but wasn't quite sure about how to describe when to use the marker. I'm not sure if we need to provide that context now.

I was kinda hoping for that description. Ideally something like "tests longer than 1 minute can (or should) be marked as slow" for example. 

I can see it both as being a good heuristic to have a "time" based component, but also a way for us to grab the list of slow tests for analysis sometime in the future. 

This is not a hard requirement on my part, just thinking out loud.

ekohl:
Slow is always relative. I wanted to avoid that.

ehelms:
You'll need a rebase here to fix tests.

ekohl:
I tried to be a bit more creative now in the docs. Also couldn't resist adding links to the section above it.

ehelms (approved):

PR #657, opened 2026-07-14
Mark tests with the appropriate feature

ekohl:

As part of https://github.com/theforeman/foremanctl/pull/511 I've been looking at deploying Foreman without Katello. This properly marks tests so they're only executed when the katello feature is present.

* It adds the katello feature marker to tests that require Katello

Steps to reproduce:

* Deploy Foreman with Katello and verify the tests still run

Deployment without Katello isn't supported yet, so no need to test that yet.

#### Checklist

### DIFF
--- a/tests/feature/foreman/base_test.py
+++ b/tests/feature/foreman/base_test.py
@@ -45,6 +45,7 @@ def test_foreman_status_cache(foreman_status):
     assert foreman_status['results']['foreman']['cache']['servers'][0]['status'] == 'ok'
 
 
+@pytest.mark.feature('katello')
 @pytest.mark.parametrize("katello_service", ['candlepin', 'candlepin_auth', 'foreman_tasks', 'katello_events', 'pulp3', 'pulp3_content'])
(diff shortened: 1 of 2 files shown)


tests/httpd_test.py
```diff
@@ -37,36 +37,42 @@ def test_https_foreman_ping(server, certificates, server_fqdn):
     assert cmd.stdout == '200'
 
 
+@pytest.mark.feature('katello')
```
evgeni: those also apply to standalone proxies that have content

ekohl: Hmm yes, I wrote this before `foreman-proxy-content` was a flavor. It ties into some work I have in https://github.com/theforeman/foremanctl/pull/527, in particular https://github.com/theforeman/foremanctl/pull/527/changes/5a6cf5f3a807c2e91f06210dedcf2372c11aaa1b. I'll remove it from this PR and move it over there so I won't forget.

ehelms (approved):

PR #661, opened 2026-07-15
Run pre_install role in mock-installer

evgeni:

pre_install ensures things like the right PostgreSQL repos are available, which is required for the migration tests to succeed

otherwise the installation of `foreman-installer` pulls in PostgreSQL 13 clients, and backup tests fail as they require postgresql 16

See e.g. https://github.com/theforeman/foremanctl/actions/runs/29394054376/job/87283471606?pr=659

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/development/playbooks/mock-installer/mock-installer.yaml
+++ b/development/playbooks/mock-installer/mock-installer.yaml
@@ -4,4 +4,5 @@
     - quadlet
   become: true
   roles:
+    - pre_install
     - mock_foreman_installer


evgeni:
The alternative could be to enforce `postgresql >= 16` in https://github.com/theforeman/foremanctl/blob/57966bd9d3fd9a5e796afaaad5855264d02ddcf5/src/playbooks/backup/backup.yaml#L16-L19

Gauravtalreja1 (approved):

PR #663, opened 2026-07-15
test valkey UID is not directly mapped to same host UID

arvind4501:
To test permissions leak issue for rootful containers

*

Steps to reproduce:

*

#### Checklist


ehelms:
This reads like it should follow our check role paradigm and encapsulate it.

arvind4501:
yes, checks are a better places for validations. for now this was just a test to see how non-root containers UID leaks to random host user. valkey ships with id 1001 in container and that gives a user on host with UID 1001 full access to mounted directories. 
for now i will keep this in a draft state as it was an expirement with userns.

PR #664, opened 2026-07-15
Simplify test infrastructure

ehelms:

Simplify the `forge test` command and allow running non-slow marked tests.

* Move ssh-config generation from `forge test` into a session-scoped autouse
  pytest fixture (`generate_ssh_config` in `tests/conftest.py`). The fixture
  reads the static inventory (defaulting to `inventories/local_vagrant`,
  overridable via `OBSAH_INVENTORY`) and writes `.tmp/ssh-config` before any
  test runs, so pytest can be invoked directly without going through forge.

Steps to reproduce:

* Confirm `forge test` still works (now a thin wrapper around pytest):
  ```
  ./forge test
  ```

* Run fast tests:
  ```
  ./forge test --fast
  ```

#### Checklist

### DIFF
--- a/development/playbooks/test/metadata.obsah.yaml
+++ b/development/playbooks/test/metadata.obsah.yaml
@@ -5,3 +5,7 @@ help: |
 variables:
   pytest_args:
     help: Pass arguments to pytest.
+  fast:
+    help: Skip tests marked as slow.
+    parameter: --fast
+    action: store_true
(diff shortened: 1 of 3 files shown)


ehelms:
There are things in here I think we should do for sure, the open question is testing interface:

- `forge` for consistency with development actions
- `make` since it will stream the testing output real-time rather than dump it at the end
- `pytest` directly which has the benefits of direct interaction, but the downsides of having to "remember" the baseline set of options we run it with today: `--durations=10 -vv`

PR #666, opened 2026-07-15
Improve test suite performance via module-scoped fixtures and batched…

ehelms:

Test runs were spending a large fraction of their wall-clock time recreating
the same remote resources for every test. The ten Foreman/Katello resource
fixtures (organizations, products, repositories, lifecycle environments, content
views, activation keys) were all function-scoped by default, so each individual
test paid the full create-over-API cost. Additionally, two pulp-manager tests
each spawned a separate Django process inside the container, doubling that
startup cost, and inherently-slow tests (backup, target lifecycle) were mixed
with fast ones making it impossible to exclude them during development.

**Fixture call reduction** (measured with `--fixture-durations`):

| Fixture | Before calls | Before total | After calls | After total | Saved |
|---------|:---:|---:|:---:|---:|---:|
| Pulp manager Django startup¹ | 2 | 87.5s | 1 | 24.3s | 63.2s |
| `yum_repository` | 4 | 61.2s | 2 | 10.8s | 50.4s |
| `client_environment` | 2 | 35.2s | 1 | 13.6s | 21.6s |
| `organization` | 10 | 23.1s | 3 | 7.9s | 15.2s |
| `product` | 7 | 3.5s | 2 | 1.1s | 2.4s |
| **Total** | | **210.5s** | | **57.7s** | **152.8s** |

¹ Before, each test launched its own `pulpcore-manager shell` process (45.0s + 42.5s in test call time); after, both tests share one module-scoped fixture.

**Test phase overhead** (pytest `--durations` grand totals):

| Phase | Before | After | Saved |
|-------|---:|---:|---:|
| Test setup | 156s | 2s | 154s |
| Test teardown | 120s | <1s | ~120s |

**Skippable slow tests** (with `-m "not slow"`):

| Test module | Duration |
|-------------|---:|
| `backup_test.py` | ~2m 1s |
| `target_lifecycle_test.py` | ~2m 46s |
| **Skippable total** | **~4m 47s** |

* `tests/conftest.py` — Promote all remote resource fixtures to
  `scope="module"` so resources are created once per test module, not once per
  test. Expand `content_view` teardown to handle version cleanup (previously
  left to test bodies, which broke with shared fixtures). Simplify
  `client_environment` teardown to delegate version cleanup to `content_view`.
  Change `wait_for_metadata_generate` to filter with `state != stopped` instead
  of fetching all historical tasks. Register the `slow` marker in
  `pytest_configure`.
* `tests/pulp_test.py` — Add `_PULP_MANAGER_SCRIPT` constant and module-scoped
  `pulp_manager_info` fixture that runs one Django startup via
  `pulpcore-manager shell` (base64-encoded to avoid shell quoting issues).
  `test_pulp_manager_check` and `test_pulp_import_export_settings` now consume
  the shared result instead of each launching their own container process.
* `tests/feature/katello/api_test.py` — Remove per-test content view version
  cleanup loop from `test_foreman_content_view`; the `content_view` fixture
  teardown now handles this.
* `tests/backup_test.py` — Add `pytestmark = pytest.mark.slow`.
* `tests/target_lifecycle_test.py` — Add `pytestmark = pytest.mark.slow`.

Steps to reproduce:

* Deploy a foremanctl environment and run the full test suite:
  ```
  python -m pytest tests/ --durations=20 --fixture-durations=10 -v
  ```
* Verify resource fixtures appear once per module in the setup column, not once
  per test
* Deselect slow tests and confirm the run completes faster:
  ```
  python -m pytest tests/ -m "not slow"
  ```
* Confirm the `content_view` teardown cleans up published versions without
  error when the `katello` feature is enabled

#### Checklist

### DIFF
--- a/tests/pulp_test.py
+++ b/tests/pulp_test.py
@@ -1,9 +1,30 @@
 import json
+import textwrap
 
 import pytest
 
 PULP_API_SOCKET = '/run/httpd.pulp-api.sock'
 PULP_CONTENT_SOCKET = '/run/httpd.pulp-content.sock'
+
+# Run check --deploy and settings inspection in a single Django startup.
+# Passed via heredoc to avoid shell quoting issues with multi-line Python.
+_PULP_MANAGER_SCRIPT = textwrap.dedent("""\
+    import json
+    from django.conf import settings
+    from django.core.management import call_command
+    from django.core.management.base import SystemCheckError
+    check_ok = True
+    try:
+        call_command('check', '--deploy', verbosity=0)
+    except SystemCheckError:
+        check_ok = False
+    print(json.dumps({
+        'check_ok': check_ok,
+        'import': sorted(settings.ALLOWED_IMPORT_PATHS),
+        'export': sorted(settings.ALLOWED_EXPORT_PATHS),
+        'rhsm_url': settings.SMART_PROXY_RHSM_URL,
+    }))
+""")
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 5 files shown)


tests/pulp_test.py
```diff
@@ -1,12 +1,33 @@
...
 import os
+import textwrap
 
 import pytest
 import yaml
 
 PULP_API_SOCKET = '/run/httpd.pulp-api.sock'
 PULP_CONTENT_SOCKET = '/run/httpd.pulp-content.sock'
 
+# Run both check --deploy and settings inspection in a single Django startup.
```
ehelms, line 12: There's got to be a better / cleaner way to do this, consider this just a "hack" for now.

PR #668, opened 2026-07-16
Fix PostgreSQL 13→16 upgrade path in deploy-dev

jeremylenz:

Commit ca470b8 ("Switch to PostgreSQL 16 image") introduced issues that cause `forge deploy-dev` to fail when re-deploying over an existing PostgreSQL 13 database.

Three problems were identified:

1. The `upgrade.yaml` stop task assumes `foreman.target` exists, but deploy-dev never created it (the `systemd_target` role was not included)
2. Without `foreman.target`, services like `pulp-api` aren't stopped before re-deploy, causing handler conflicts
3. The host's `pg_dump` stays at version 13 after the containerized PG upgrades to 16, breaking `db:migrate` schema dumps

* Add `failed_when: false` to the "Stop foreman target" task in `upgrade.yaml` — on first deploy, `foreman.target` doesn't exist yet (it's created by `systemd_target` which runs after `postgresql`)
* Add `systemd_target` role to `deploy-dev.yaml` — creates `foreman.target` so all backing services can be managed as a group, matching the production deploy
* Add a deploy-dev pre-task to switch the postgresql dnf module to 16 if needed — handles dev re-deploys where the host still has PG 13 client tools

1. Have an existing dev environment deployed with PG 13
2. Run `forge deploy-dev` — should complete without errors
3. Verify `cat /var/lib/pgsql/data/userdata/PG_VERSION` shows `16`
4. Run `forge deploy-dev` again — idempotent re-run should succeed

### DIFF
--- a/development/playbooks/deploy-dev/deploy-dev.yaml
+++ b/development/playbooks/deploy-dev/deploy-dev.yaml
@@ -15,6 +15,12 @@
     - "../../../src/vars/foreman.yml"
     - "../../../src/vars/base.yaml"
   pre_tasks:
+    - name: Stop foreman target
+      ansible.builtin.systemd:
+        name: foreman.target
+        state: stopped
+      failed_when: false
+
     - name: Set development postgresql databases
       ansible.builtin.set_fact:
         postgresql_databases: >-
--- a/development/playbooks/deploy-dev/deploy-dev.yaml
+++ b/development/playbooks/deploy-dev/deploy-dev.yaml
@@ -15,6 +15,15 @@
     - "../../../src/vars/foreman.yml"
     - "../../../src/vars/base.yaml"
   pre_tasks:
+    - name: 'Switch to postgresql:16 module if needed'
+      ansible.builtin.command: dnf module switch-to -y postgresql:16
+      register: module_switch
+      changed_when: "'Nothing to do' not in module_switch.stdout"
+      failed_when: false
+      when:
+        - ansible_facts['os_family'] == 'RedHat'
+        - ansible_facts['distribution_major_version'] == '9'
+
     - name: Set development postgresql databases
       ansible.builtin.set_fact:
         postgresql_databases: >-
--- a/development/playbooks/deploy-dev/deploy-dev.yaml
+++ b/development/playbooks/deploy-dev/deploy-dev.yaml
@@ -15,15 +15,6 @@
     - "../../../src/vars/foreman.yml"
     - "../../../src/vars/base.yaml"
   pre_tasks:
-    - name: 'Switch to postgresql:16 module if needed'
-      ansible.builtin.command: dnf module switch-to -y postgresql:16
-      register: module_switch
-      changed_when: "'Nothing to do' not in module_switch.stdout"
-      failed_when: false
-      when:
-        - ansible_facts['os_family'] == 'RedHat'
-        - ansible_facts['distribution_major_version'] == '9'
-
     - name: Set development postgresql databases
       ansible.builtin.set_fact:
         postgresql_databases: >-
--- a/development/playbooks/deploy-dev/deploy-dev.yaml
+++ b/development/playbooks/deploy-dev/deploy-dev.yaml
@@ -15,12 +15,6 @@
     - "../../../src/vars/foreman.yml"
     - "../../../src/vars/base.yaml"
   pre_tasks:
-    - name: Stop foreman target
-      ansible.builtin.systemd:
-        name: foreman.target
-        state: stopped
-      failed_when: false
-
     - name: Set development postgresql databases
       ansible.builtin.set_fact:
         postgresql_databases: >-
--- a/development/playbooks/deploy-dev/deploy-dev.yaml
+++ b/development/playbooks/deploy-dev/deploy-dev.yaml
@@ -54,6 +54,7 @@
     - role: candlepin
     - role: httpd
     - role: pulp
+    - role: systemd_target
     - role: foreman_development
       vars:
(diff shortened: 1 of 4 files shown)


src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -39,7 +46,7 @@
         state: quadlet
         sdnotify: false
         volumes:
-          - "{{ postgresql_data_dir }}:/var/lib/pgsql/data:rw,Z"
+          - "{{ postgresql_data_dir }}:/var/lib/pgsql/data:rw,z"
```
evgeni: `:z` means "shared" - shared with whom? There should be no other containers running off that volume at the same time.

(and if there should be, `main.yml` would need an update too)

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -15,6 +15,12 @@
     - "../../../src/vars/foreman.yml"
     - "../../../src/vars/base.yaml"
   pre_tasks:
+    - name: Stop foreman target
```
evgeni: Why would you stop for a deployment?

src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -25,6 +25,13 @@
       ansible.builtin.systemd:
         name: foreman.target
         state: stopped
+      failed_when: false
+
+    - name: Stop postgresql service
```
evgeni: it's part of foreman.target, so why twice?

src/roles/pre_install/tasks/main.yaml
```diff
@@ -3,16 +3,32 @@
...
-  # https://github.com/ansible/ansible/issues/64852
-  args:
-    creates: /etc/dnf/modules.d/postgresql.module
+  changed_when: true
+  when:
+    - ansible_facts['os_family'] == 'RedHat'
+    - ansible_facts['distribution_major_version'] == '9'
+    - postgresql_module_file is failed or postgresql_module_file.content is not defined
+
+- name: 'Switch to postgresql:16 module'
```
evgeni: Switching means another module was already enabled, so we're not deploying on a clean system (which is against our support policy).
But more importantly, it means that there are potentially already postgresql packages installed, and they might not be updated if you just `switch`.

jeremylenz: > we're not deploying on a clean system (which is against our support policy).

How do I allow redeploying in a dev environment, then?

evgeni: I have no idea about dev :)

What I am saying: this should never happen in prod and thus should not be fixed in code that runs on prod.

jeremylenz: Okay, I moved it to deploy-dev, see if you're happier with that?

development/playbooks/deploy-dev/deploy-dev.yaml
jeremylenz: I was trying to make it easier to redeploy in a dev environment, when you probably already have things running.

src/roles/postgresql/tasks/upgrade.yaml
jeremylenz: I think most of my problem was because of SELinux denials, so I removed the :z part.

jeremylenz:
addressed comments and also updated the PR description to reflect the current state.

src/roles/postgresql/tasks/upgrade.yaml
```diff
@@ -25,6 +25,7 @@
       ansible.builtin.systemd:
         name: foreman.target
         state: stopped
+      failed_when: false
```
ehelms: Is this still needed?

jeremylenz: On my box `foreman.target` didn't exist for some reason, and so was failing at this step. is that SELinux too??

jeremylenz: Yes — this is needed as a transitional case. Since deploy-dev didn't include `systemd_target` before this PR, existing dev environments won't have `foreman.target`. On the first deploy-dev after this PR, `postgresql` runs before `systemd_target`, so the target won't exist yet when the upgrade task tries to stop it. After that first run, `systemd_target` creates it and subsequent re-deploys are fine.

ehelms: Can you move the `systemd_target` role before the postgresql role in `deploy-dev` and therefore avoid this?

jeremylenz: Good idea. Done

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -15,6 +15,15 @@
     - "../../../src/vars/foreman.yml"
     - "../../../src/vars/base.yaml"
   pre_tasks:
+    - name: 'Switch to postgresql:16 module if needed'
```
ehelms: Can you move this to https://github.com/theforeman/foremanctl/blob/master/development/roles/foreman_development/tasks/main.yaml#L2 and we'll be GTG.

I tested this upgrade cleanly and that worked and is cleaner design.

jeremylenz: done

ehelms (approved):

src/roles/pre_install/tasks/main.yaml
evgeni: Yepp, thanks!

PR #670, opened 2026-07-17
Rename name of Microsoft Azure plugin

maximiliankolb:
The docs call this plugin "Microsoft Azure". Only Foreman/Katello WebUI calls it, on Settings > About > Available Providers, "Azure Resource Manager".

Refs PR 5069 in foreman-documentation

A long time ago, we renamed it in the docs from "Microsoft Azure Resource Manager" to "Microsoft Azure", partly due to the unwieldy name.

* Rename the name (not ID etc) of the plugin

Steps to reproduce:

* ?

#### Checklist

* [ ] Tests added/updated (if applicable) -> No, no idea if necessary. grep indicated that there's no need in _this_ repo.
* [ ] Documentation updated (if applicable) -> does not apply IMO

### DIFF
--- a/src/features.yaml
+++ b/src/features.yaml
@@ -25,8 +25,8 @@ google:
   hammer: foreman_google
   foreman:
     plugin_name: foreman_google
-azure-rm:
-  description: Azure Resource Manager plugin for Foreman
+azure:
+  description: Microsoft Azure plugin for Foreman
   hammer: foreman_azure_rm
   foreman:
     plugin_name: foreman_azure_rm
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -148,7 +148,7 @@ jobs:
           ./foremanctl deploy \
             --add-feature hammer \
             --add-feature foreman-proxy \
-            --add-feature azure-rm \
+            --add-feature azure \
             --add-feature google \
             --add-feature remote-execution \
             --add-feature ansible \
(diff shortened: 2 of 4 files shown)


src/features.yaml
```diff
@@ -26,7 +26,7 @@ google:
   foreman:
     plugin_name: foreman_google
 azure-rm:
```
ekohl: This is a bigger change, but I think we should also change the feature name.
```suggestion
azure:
```

ehelms:
@maximiliankolb What is needed to move this out of draft?

maximiliankolb:
I have rebased my PR to HEAD of "master" & resolved the merge conflict. The upgrade test still fails. Is this because I need to implement some form of migration (X.Y: option called "azure-rm" -> X.Y+1: option called "azure") or is it because the GHA logic somehow tries to install an old version with new options from my feature branch?

I am only really interested in rewording line 29 in `src/features.yaml` & I do not yet have a local playground to test/investigate this further.

What do you think/suggest?

ekohl:
@evgeni we haven't done any thinking around migrations during upgrades, have we?

evgeni:
Nope, so far we didn't need them.

evgeni:
For this particular case: IMHO we can just tell users to do `--remove-feature azure-rm --add-feature azure` and update our CI accordingly.

PR #671, opened 2026-07-17
Test refresh of internal CA doesn't break the bundle

evgeni:

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -420,16 +420,20 @@ jobs:
             --initial-admin-password=changeme \
             --tuning development \
             --add-feature foreman-proxy
-      - name: Remove old parameters.yaml
-        run: rm -f .var/lib/foremanctl/parameters.yaml
       - name: Generate certificates bundle on quadlet
         run: |
           ./foremanctl certificate-bundle proxy.example.com \
-            --certificate-source=${{ matrix.certificate_source }} \
             ${{ matrix.certificate_source == 'custom_server' && '--certificate-server-certificate /root/custom-certificates/certs/proxy.example.com.crt --certificate-server-key /root/custom-certificates/private/proxy.example.com.key' || '' }}
+      - name: Refresh internal CA to ensure it doesn't break the bundle
+        if: matrix.certificate_source != 'custom_server'
+        run: |
+          ./foremanctl deploy --certificate-ca-renew
       - name: Fetch certificates bundle from quadlet
         run: |
           ./forge fetch-bundle proxy.example.com
+      - name: Remove old parameters, but keep certificates_source for tests
+        run: |
+          echo "certificates_source: ${{ matrix.certificate_source }}" > .var/lib/foremanctl/parameters.yaml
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -420,16 +420,20 @@ jobs:
             --initial-admin-password=changeme \
             --tuning development \
             --add-feature foreman-proxy
-      - name: Remove old parameters.yaml
-        run: rm -f .var/lib/foremanctl/parameters.yaml
       - name: Generate certificates bundle on quadlet
         run: |
           ./foremanctl certificate-bundle proxy.example.com \
-            --certificate-source=${{ matrix.certificate_source }} \
```
evgeni, line 428: This was what was making the tests pass before. Now we explicitly store only the source in L434.

.github/workflows/test.yml
```diff
@@ -420,16 +420,20 @@ jobs:
...
             ${{ matrix.certificate_source == 'custom_server' && '--certificate-server-certificate /root/custom-certificates/certs/proxy.example.com.crt --certificate-server-key /root/custom-certificates/private/proxy.example.com.key' || '' }}
+      - name: Refresh internal CA to ensure it doesn't break the bundle
+        if: matrix.certificate_source != 'custom_server'
+        run: |
+          ./foremanctl deploy --certificate-renew-ca
```
evgeni: #672 renames this

.github/workflows/test.yml
```diff
@@ -420,16 +420,20 @@ jobs:
...
             --tuning development \
             --add-feature foreman-proxy
-      - name: Remove old parameters.yaml
-        run: rm -f .var/lib/foremanctl/parameters.yaml
       - name: Generate certificates bundle on quadlet
         run: |
           ./foremanctl certificate-bundle proxy.example.com \
-            --certificate-source=${{ matrix.certificate_source }} \
             ${{ matrix.certificate_source == 'custom_server' && '--certificate-server-certificate /root/custom-certificates/certs/proxy.example.com.crt --certificate-server-key /root/custom-certificates/private/proxy.example.com.key' || '' }}
+      - name: Refresh internal CA to ensure it doesn't break the bundle
```
arvind4501, line 427: are we testing that `existing` (default ca) bundle does not change after `--certificate-ca-renew`?
Does `--certificate-ca-renew` itself changes all issued certificates, if yes then does existing bundle(include old certificates) should work?

evgeni: Yeah, the idea was to see that the following works:
- create a fresh deployment (CA and everything)
- create a bundle
- renew the certificate of the CA (which should update expiry dates, but not the key and not the certs signed by the CA)
- use the bundle created *before* the renewal to deploy a system and assert it still works

arvind4501 (approved):

.github/workflows/test.yml
arvind4501: thanks, make sense to me now!

PR #674, opened 2026-07-17
store the CA private key unencrypted

bochi:
The CA private key is stored encrypted, but that password is in ca.pwd next to the key, and Candlepin is handed a decrypted copy. So there is no real benefit. 

Followup to #602 

* Store key in an unencrypted way
* Remove ca_key_password and candlepin_ca_key_password variables
* migrate existing deployments in place

Steps to reproduce:

* on initial deployment: 
-> check that the ca.key is unencrypted
* on upgrade
-> deploy, see that key is NOW unencrypted

#### Checklist

### DIFF
--- a/src/roles/candlepin/tasks/certs.yml
+++ b/src/roles/candlepin/tasks/certs.yml
@@ -9,16 +9,11 @@
   notify:
     - Restart candlepin
 
-- name: Decrypt Candlepin CA key
-  ansible.builtin.command: openssl pkey -in "{{ candlepin_ca_key }}" -passin "file:{{ candlepin_ca_key_password }}"
-  register: _candlepin_ca_key
-  changed_when: false
-
 - name: Create the podman secret for Candlepin CA key
   containers.podman.podman_secret:
     state: present
     name: candlepin-ca-key
-    data: "{{ _candlepin_ca_key.stdout }}"
+    path: "{{ candlepin_ca_key }}"
     labels:
       app: candlepin
   notify:
--- a/src/roles/certificates/tasks/ca.yml
+++ b/src/roles/certificates/tasks/ca.yml
@@ -4,25 +4,45 @@
     path: "{{ certificates_ca_directory_certs }}/ca.crt"
   register: _certificates_ca_cert
 
+- name: 'Check for an existing CA key'
+  ansible.builtin.stat:
+    path: "{{ certificates_ca_directory_keys }}/ca.key"
+  register: _certificates_existing_ca_key
+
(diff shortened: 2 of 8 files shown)


src/roles/candlepin/tasks/certs.yml
```diff
@@ -9,8 +9,8 @@
   notify:
     - Restart candlepin
 
-- name: Decrypt Candlepin CA key
-  ansible.builtin.command: openssl pkey -in "{{ candlepin_ca_key }}" -passin "file:{{ candlepin_ca_key_password }}"
+- name: Export Candlepin CA key
+  ansible.builtin.command: openssl pkey -in "{{ candlepin_ca_key }}"
```
ehelms: You should be able to pass `candlepin_ca_key` directly to line 21 now just like the certificate.

tests/feature/katello/candlepin_test.py
```diff
@@ -37,6 +37,17 @@ def test_candlepin_status(server, certificates):
     assert status.stdout == '200'
 
 
+def test_candlepin_ca_key_secret_is_plain_ca_key(server, certificates):
```
ehelms: We don't test this for any other service, so I'd opt to not include it and make Candlepin a special case.

src/roles/candlepin/tasks/certs.yml
bochi: right, fixed

tests/feature/katello/candlepin_test.py
bochi: fair, thought it might make sense to actually check if the secret is really the unencrypted key. but your call, removed.

ehelms:
Valid test failure for backup tests:

```
FAILED tests/backup_test.py::test_foremanctl_state_archive_contents - AssertionError: Archive should contain certificates-ca-password (generated during deployment)
```

ehelms:
This generally looks good to me. @evgeni could you give it a second pair of eyes?

evgeni:
Test failures related, I think it tries to check `ca.key` on the proxy, but we do not copy that file there.

Otherwise LGTM

bochi:
> Test failures related, I think it tries to check `ca.key` on the proxy, but we do not copy that file there.
> 
> Otherwise LGTM

I fixed the test so it doesn't check for ca.key on the proxy. Thanks

ehelms:
Looks like you'll need a rebase.

bochi:
> Looks like you'll need a rebase.

Done

ehelms (approved):

PR #676, opened 2026-07-19
Fixes #39446 - Reject non-PEM content in foreman-certificate-check

Satellite-RedHat:
## Summary:

- Port `check-pem-content` from [foreman-installer #1062](https://github.com/theforeman/foreman-installer/pull/1062) into `foreman-certificate-check`
- Reject certificate and CA bundle files that contain content outside `BEGIN`/`END CERTIFICATE` blocks (for example PKCS#12 Bag Attributes or comment lines)

Reference discussion:

https://github.com/theforeman/foreman-installer/pull/1062#issuecomment-5008537407

## Why

`foreman-certificate-check` could previously accept cert/CA files with non-PEM text between blocks. Those files can pass OpenSSL checks but fail later in production. Installer #1062 adds the same guard to `katello-certs-check`; this change keeps the containerized install path aligned.

Related:
- https://projects.theforeman.org/issues/39446
- https://github.com/theforeman/foreman-installer/pull/1062

## Test cases  added
- test_fails_with_pkcs12_metadata_in_ca_bundle — CA bundle with PKCS#12 metadata (Bag Attributes, subject=, issuer=) between PEM blocks → exit 12
- test_fails_with_comment_lines_in_certificate_files — server cert and CA bundle with # RETRAITE AC Intermédiaire TLS before PEM blocks → exit 12


ehelms:
@Satellite-RedHat have you investigated a method using `openssl` to check for valid certificate file structure?

Satellite-RedHat:
@ehelms 

Thanks for the question.

Yes — this script already relies on `openssl` for most certificate validation (encoding, expiry, key/cert match, CA verification, SAN, SHA-1, etc.). The issue is that these checks can still pass when there is non-PEM content between certificate blocks, such as PKCS#12 `Bag Attributes` or comment lines. This is what triggered #39446.

I looked at whether `openssl` alone could reliably catch this, but it doesn't reject this class of input consistently. The additional check is therefore structural: the file must contain only `BEGIN`/`END CERTIFICATE` blocks and blank lines.

This is the same approach used in foreman-installer #1062, which I ported here to keep both installation paths aligned.

PR #677, opened 2026-07-19
Rename cname to server_alias

### LINKED ISSUE #639: Replace `--certificate-cname` with `--server-alias`
Today we have a `--certificate-cname` parameter that adds SANs to the generated (self-signed) certificate.

On the one hand, "cname" is the wrong term here as CNAMEs are DNS record, and nothing dictates how the alias is routed to the server.

On the other we need to know the (possible) aliases for more than just the generated certificate, as shown in #471 

We should replace that parameter with something more generic, like `--server-alias` and then feed the value of that to the various places that need to know the aliases of the server.

Will need updates to 
* documentation: https://github.com/theforeman/foreman-documentation/blob/117029d705336c179ba6ab7625e8ac809cb6378d/guides/common/modules/proc_configuring-project-with-an-alternate-cname.adoc?plain=1#L16
* robottelo: https://github.com/SatelliteQE/robottelo/blob/34d59f812478723609712f74307f1b1a1855d82f/tests/foreman/installer/test_install_foremanctl.py#L252

ShimShtein:

Implements https://redhat.atlassian.net/browse/SAT-47835

cname is a term from DNS records, which is not entirely correct in foremanctl context. Broader `server-alias` feels more natural.

* Renamed the user-facing `--certificate-cname` parameter to `--server-alias`
* Updated the internal documentation

Steps to reproduce:

* Run the installer
* notice `--certificate-cname` is rejected
* notice `--server-alias` is accepted

#### Checklist


Gauravtalreja1 (approved):
ACK, changes look good to me. PRT passed as well on https://github.com/SatelliteQE/robottelo/pull/22161

docs/developer/playbooks-and-roles.md
```diff
@@ -158,11 +158,11 @@ variables:
     help: Initial password for the admin user.
   foreman_puma_workers:
     help: Number of workers for Puma.
-  certificates_cnames:
+  certificates_server_aliases:
```
evgeni:
```suggestion
  server_aliases:
```

docs/developer/playbooks-and-roles.md
```diff
@@ -158,11 +158,11 @@ variables:
...
     help: Additional DNS name for Subject Alternative Names. Can be specified multiple times.
```
evgeni:
```suggestion
    help: Server alias. Used for additional Subject Alternative Names in generated certificates and accepted server names. Can be specified multiple times.
```

src/playbooks/_foreman/metadata.obsah.yaml
```diff
@@ -14,3 +14,8 @@ variables:
     parameter: --initial-location
   foreman_puma_workers:
     help: Number of workers for Puma.
+  server_aliases:
+    help: Additional DNS name to include in Subject Alternative Names for certificates. Can be specified multiple times.
```
evgeni:
```suggestion
    help: Server alias. Used for additional Subject Alternative Names in generated certificates and accepted server names. Can be specified multiple times.
```

evgeni (changes requested):
Doc nitpicks :)

ShimShtein:
Rebased and changed

evgeni (approved):

PR #679, opened 2026-07-20
Renew all the CAs

evgeni:

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -427,8 +427,7 @@ jobs:
         run: |
           ./foremanctl auth-bundle proxy.example.com \
             ${{ matrix.certificate_source == 'custom_server' && '--certificate-server-certificate /root/custom-certificates/certs/proxy.example.com.crt --certificate-server-key /root/custom-certificates/private/proxy.example.com.key' || '' }}
-      - name: Refresh internal CA to ensure it doesn't break the bundle
-        if: matrix.certificate_source != 'custom_server'
+      - name: Renew CA to ensure it doesn't break the bundle
         run: |
           ./foremanctl deploy --certificate-ca-renew
       - name: Fetch auth bundle from quadlet
--- a/src/roles/certificates/tasks/ca.yml
+++ b/src/roles/certificates/tasks/ca.yml
@@ -79,6 +79,8 @@
         remote_src: true
         force: "{{ certificates_ca_renew | bool }}"
         mode: '0444'
+      when:
+        - certificates_source == 'default'
 
     - name: 'Create CA bundle'
       ansible.builtin.copy:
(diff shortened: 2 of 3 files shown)


evgeni:
This includes https://github.com/theforeman/foremanctl/pull/671 for reasons to build ontop of the changes there

.github/workflows/test.yml
```diff
@@ -420,16 +420,19 @@ jobs:
...
             --tuning development \
             --add-feature foreman-proxy
-      - name: Remove old parameters.yaml
-        run: rm -f .var/lib/foremanctl/parameters.yaml
       - name: Generate certificates bundle on quadlet
         run: |
           ./foremanctl certificate-bundle proxy.example.com \
-            --certificate-source=${{ matrix.certificate_source }} \
             ${{ matrix.certificate_source == 'custom_server' && '--certificate-server-certificate /root/custom-certificates/certs/proxy.example.com.crt --certificate-server-key /root/custom-certificates/private/proxy.example.com.key' || '' }}
+      - name: Refresh CA to ensure it doesn't break the bundle
```
ehelms, line 427: I don't understand what this is testing.

evgeni: The naïve idea was:
- deploy foreman
- create bundle for a proxy
- renew ca
- see that the previously created bundle still works (assuming the certificate has new expiration dates, but otherwise was not changed)

ehelms: Does the bundle get re-created? Or just testing that the certs in the bundle before renewal continue to work?

evgeni: No recreation, "old bundle still works" is basically the test

ehelms: Should the old bundle work if the CA was refreshed?

evgeni: In the case where the private key didn't change, and only the CA expiration date did - yes.

Otherwise no

ehelms: Gotcha... perhaps we should define terms to clarify that? I read refresh thinking complete change of the CA but your intent was the certificate not the private key.

- refresh: changes to the CA certificate without modifying the CA private key
- renew/reissue: completely regenerate the CA private key and certificate

evgeni: We had a similar discussion in docs recently, and there I said (rephrasing):
- renew - update expiration or similar "minor" update (which doesn't invalidate old signatures)
- refresh - update the used certificate (or CA certificate) on a consumer (read: fetch the renewed foreman ca to a client host)

And now we'd add:
- reissue - completely replace the certificate/key/etc (invalidates old signatures)

evgeni: (and yes, this PR uses "refresh" wrongly, by that definition)

evgeni:
Updated commit message and step name to "renew"

ehelms (approved):

PR #680, opened 2026-07-20
Remove katello:migrate_deb_content_attributes_to_pulp3 from foreman-rake

ShimShtein:
Since https://github.com/theforeman/foreman-documentation/pull/5082 we no longer need to allow this action in foreman-rake.

* removed the command from `foreman-rake` allow list

Steps to reproduce:

* Try to run the command and see that it is rejected.

#### Checklist

### DIFF
--- a/src/roles/foreman/templates/foreman-rake.j2
+++ b/src/roles/foreman/templates/foreman-rake.j2
@@ -15,7 +15,6 @@ validate_actions() {
         "foreman_tasks:cleanup"
         "interfaces:clean"
         "katello:delete_orphaned_content"
-        "katello:migrate_deb_content_attributes_to_pulp3"
         "katello:upgrade_check"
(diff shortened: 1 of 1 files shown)


stejskalleos:
If it is not needed, should we remove it [from the Katello](https://github.com/Katello/katello/blob/master/lib/katello/tasks/pulp2to3_migrate_deb_attributes.rake#L3) repo as well?

ShimShtein:
> If it is not needed, should we remove it [from the Katello](https://github.com/Katello/katello/blob/master/lib/katello/tasks/pulp2to3_migrate_deb_attributes.rake#L3) repo as well?

I suppose that is a question for @ianballou and @quba42.

stejskalleos (approved):
Removal from Katello doesn't need to block this change; let's get this in!

PR #684, opened 2026-07-20
Support offline restore on proxy deployments

sjha4:
Support offline restore on proxies. Built on top of offline restore commit hence marking draft.

* Adds some proxy specific logic to offline restore to be able to run against proxy deployments.

1. Set up a proxy
2. Run foremanctl backup /tmp/backup --target-host proxy
3. Run foremanctl restore /tmp/backup/backup_dir --target-host proxy --force

#### Checklist

### DIFF
--- a/src/playbooks/restore/restore.yaml
+++ b/src/playbooks/restore/restore.yaml
@@ -1,6 +1,6 @@
 ---
 - name: Restore from a backup
-  hosts: quadlet
+  hosts: "{{ target_host | default('quadlet') }}"
   become: true
   gather_facts: true
   vars_files:
@@ -10,6 +10,49 @@
   roles:
     - restore
 
-- name: Deploy services after restore
+- name: Load restored parameters for deploy
+  hosts: "{{ target_host | default('quadlet') }}"
+  become: true
+  gather_facts: false
+  tasks:
+    - name: Read restored parameters
+      ansible.builtin.slurp:
+        path: "{{ obsah_state_path }}/parameters.yaml"
+      register: restore_params_raw
+      failed_when: false
+      when: not validate | default(false)
+
+    - name: Parse restored parameters
+      ansible.builtin.set_fact:
+        restore_params: "{{ restore_params_raw.content | b64decode | from_yaml }}"
+      when:
+        - not validate | default(false)
+        - restore_params_raw is succeeded
+
+    - name: Set restored features
+      ansible.builtin.set_fact:
+        restore_enabled_features: "{{ restore_backup_metadata.enabled_features | default([]) }}"
+      when: restore_backup_metadata is defined
+
+    - name: Set proxy deploy parameters
+      ansible.builtin.set_fact:
+        foreman_name: "{{ restore_params.foreman_name }}"
+        foreman_proxy_oauth_consumer_key: "{{ restore_params.foreman_proxy_oauth_consumer_key }}"
+        foreman_proxy_oauth_consumer_secret: "{{ restore_params.foreman_proxy_oauth_consumer_secret }}"
+      when:
+        - restore_params is defined
+        - restore_enabled_features | default([]) | has_feature('foreman-proxy')
+        - not (restore_enabled_features | default([]) | has_feature('foreman'))
+
+- name: Deploy services after restore (foreman-proxy)
+  import_playbook: ../deploy-proxy/deploy-proxy.yaml
+  when:
+    - not validate | default(false)
+    - restore_enabled_features | default([]) | has_feature('foreman-proxy')
+    - not (restore_enabled_features | default([]) | has_feature('foreman'))
+
+- name: Deploy services after restore (server)
   import_playbook: ../deploy/deploy.yaml
-  when: not validate | default(false)
+  when:
+    - not validate | default(false)
+    - restore_enabled_features | default([]) | has_feature('foreman')
--- a/src/playbooks/restore/metadata.obsah.yaml
+++ b/src/playbooks/restore/metadata.obsah.yaml
@@ -17,6 +17,10 @@ variables:
     action: store_true
     persist: false
 
+  target_host:
+    help: Target host to restore (defaults to quadlet)
+    persist: false
+
   force:
     help: Force restore on existing system (bypasses safety check)
     action: store_true
--- a/src/roles/restore/tasks/restore_databases.yaml
+++ b/src/roles/restore/tasks/restore_databases.yaml
@@ -27,7 +27,7 @@
       dump_file: "{{ restore_backup_metadata.database_mapping[item.name] }}.dump"
       database: "{{ item.database }}"
       user: "{{ item.user }}"
-  loop: "{{ databases }}"
+  loop: "{{ databases | selectattr('name', 'in', restore_backup_metadata.databases) | list }}"
   loop_control:
     label: "{{ item.name }}"
   when: item.name in restore_backup_metadata.database_mapping


sjha4:
In first read of the PR, this should work with https://github.com/theforeman/foremanctl/pull/611 with this so we can claim proxy backup-restore really brings proxy to backed up state. Otherwise we'll require a complete proxy sync after restore to handle container gateway. I'll test both PRs together and see if we need any updates.

sjha4:
I did some testing of this with Cheyenne's restore PR and Ian's container_gateway PR with this. Was able to backup container_gateway DB and restore the DB and redeploy proxy with correct features, content and container_gateway. Will keep this in draft till the restore PR goes in so the commits can be cleaned up.

Backup on proxy with pulp and container_gateway
```
TASK [backup : Display backup completion] ********************************************************************************************************
ok: [localhost] => 
    msg: |-
        Backup completed successfully.
        Location: /tmp/backup/foreman-backup-20260723T130214
        Databases: pulp, container_gateway

```
Restore validate:
```
TASK [restore : Verify database dump files exist] ************************************************************************************************
ok: [localhost] => (item=pulp)
ok: [localhost] => (item=container_gateway)
```
Restore:
```
TASK [restore : Build database restore configuration] ********************************************************************************************
ok: [localhost] => (item=pulp)
ok: [localhost] => (item=container_gateway)

TASK [restore : Verify dump files exist] *********************************************************************************************************
ok: [localhost] => (item=pulp.dump)
ok: [localhost] => (item=container_gateway.dump)

TASK [restore : Drop existing databases] *********************************************************************************************************
changed: [localhost] => (item=pulp)
changed: [localhost] => (item=container_gateway)

TASK [restore : Create empty databases] **********************************************************************************************************
changed: [localhost] => (item=pulp)
changed: [localhost] => (item=container_gateway)

TASK [restore : Enable postgres_fdw extension for IOP databases] *********************************************************************************
skipping: [localhost] => (item=pulp) 
skipping: [localhost] => (item=container_gateway) 
skipping: [localhost]

TASK [restore : Restore databases from dump files] ***********************************************************************************************
changed: [localhost] => (item=pulp.dump → pulp)
changed: [localhost] => (item=container_gateway.dump → container_gateway)

```

src/playbooks/restore/restore.yaml
```diff
@@ -10,6 +10,89 @@
...
+        foreman_proxy_oauth_consumer_secret: "{{ restore_params.foreman_proxy_oauth_consumer_secret }}"
+      when:
+        - restore_params is defined
+        - restore_flavor | default('') is match('foreman-proxy.*')
+
+- name: Deploy services after restore (foreman-proxy)
+  import_playbook: ../deploy-proxy/deploy-proxy.yaml
+  when:
+    - not validate | default(false)
+    - restore_flavor | default('') is match('foreman-proxy.*')
```
ehelms: This can work for now, but this will really need to change to be feature based or to have a separate command (e.g. `restore-proxy`).

src/playbooks/restore/restore.yaml
```diff
@@ -10,6 +10,89 @@
...
+
+- name: Deploy services after restore (server)
   import_playbook: ../deploy/deploy.yaml
-  when: not validate | default(false)
+  when:
+    - not validate | default(false)
+    - restore_flavor | default('') == 'katello'
+
+- name: Verify restore completion
```
ehelms: Can we rely on a subset of our existing checks here instead?

sjha4: We can remove the entire section post deploy cause deploy has these checks built in.

src/playbooks/restore/restore.yaml
```diff
@@ -10,6 +10,89 @@
...
+  hosts: "{{ target_host | default('quadlet') }}"
+  become: true
+  gather_facts: true
+  tasks:
+    - name: Verify services after restore
+      when: not validate | default(false)
+      block:
+        - name: Wait for services to stabilize
+          ansible.builtin.pause:
+            seconds: 30
```
qcjames53: I'm not so sure about this approach. It seems shaky to me. Ansible for sure allows a better option here that's a service-based wait or similar.

src/playbooks/restore/restore.yaml
qcjames53: +1 to that. If there are any additional checks needed you should add them as a role to ./checks.

src/playbooks/restore/restore.yaml
qcjames53: +1 for making this deploy and the one below feature-based instead of using flavor. Since enabled_features are informed by flavors and features I'd use them as a source of truth.

qcjames53 (changes requested):
I've been working on testing this playbook all day. Backup / restore seems to work great for empty proxies but I haven't yet had a content backup go through. It's some sort of pulpcore versioning issue I'm working through; hoping to get this addressed ASAP and approve the PR.

qcjames53 (changes requested):
WOW that took a long time to fully test. Two full workdays. You'll have to excuse a little weariness and mental fog in this review :)

I haven't been able to validate proxy restore works at all. I think it's my fault but I want to document what's failed for me:
- I set up foreman on quadlet + proxy on another vagrant VM using the `foreman-proxy` feature. Content sync worked. Backup went through but restore did not function due to some obsah_state issues encountered on restore setup?
- I set up foreman and the smart proxy on different RHEL 9 satlab hosts. Foreman was configured regularly and then I created an auth bundle and used `./foremanctl deploy-proxy --auth-bundle ...` to register the proxy. Backup would not go through with this proxy. According to Claude it was attempting to back up the foreman and candlepin dbs in addition to pulp and they were not being found on the proxy.

Based on this I have a couple of questions:
- Is restore supposed to work on a quadlet-based environment?
- Should that satlab proxy backup have gone through?
- I don't think `--force` does anything? It seems to always be enabled when restoring a proxy due to broken obsah handling on the local VM proxy I tried to restore. It might work on a remote proxy but I did not get that far.

Also a few requests:
- The postgreql service shutdown is really flakey in `src/roles/restore/tasks/prepare_system.yaml` and `src/roles/backup/tasks/main.yaml`. I had to do it manually on 9/10 runs across multiple installs. I know this isn't a part of this PR but could we address both before saying backup/restore is good to go? I'd really want to see this improved.
- `docs/user/backup-restore.md` severely needs reformatted. It is 400 lines of AI-generated filler and it is so hard to find useful info in that file. I think we could get away with like 40 lines of info in that file. Can you please give this an honest try? I'd personally do commands and common params, a BRIEF list of what is backed up, and supported configs like external DBs. We don't need to include anything about compression ratios or tell users they need to rotate the backups. Why does BACKUP_DIR, the most important param, not say anything about it being a remote directory? Please :pleading_face:

qcjames53 (approved):
Samir, thank you for meeting up yesterday morning to explain the details of proxy backup/restore with me. For future reference, the missing pieces were this:
- Foremanctl backup and restore currently requires a single-environment foremanctl client like foreman-maintain. We can't support multi-host foremanctl for backup/restore until obsah can support mutli-host cleanly.
- `--force` is single-environment only for now
- Documentation is a lot of carryover from various satellite docs mixed together.

I was able to verify backup/restore of a proxy hosted on a local VM functions great provided you have two foremanctl instances configured (one for the quadlet and one for the proxy VMs). `--force` works great for overriding an existing install. Restore onto a blank VM also functioned provided Foreman was pre-configured to use that VM as a proxy. I imagine configuring the proxy in Foreman itself would have worked as well. Existing content (+Pulp db) restores okay and new content can sync after restore.

I didn't have much luck getting https://github.com/theforeman/foremanctl/pull/611 to work in conjunction with my environment. Nothing related to this PR; I just couldn't get the container gateway feature enabled properly for some reason (It looks like the PR is still in flux, anyways). I'd be happy to revisit this if it's a hard requirement but I'm thinking if something _is_ broken we can fix it down the road.

qcjames53:
I'd love if you could add some of these critical details to the backup/restore docs but I'm happy to merge as-is.

ehelms (approved):
I'll wait to merge to see @sjha4 response to the docs request.

sjha4:
Had some discussions with @qcjames53 and opened a PR here: https://github.com/theforeman/foremanctl/pull/778 which has some documentation about using backup/restore in dev setups. The current backup-restore doc in user docs does not talk about non-localhost deployments which is correct since user installs always map quadlet and proxy to localhost. In dev setups, one has to consider some things I am trying to document here: https://github.com/theforeman/foremanctl/pull/778

qcjames53 (approved):

sjha4:
Thanks @qcjames53 @ehelms .. This is good to merge now.

PR #690, opened 2026-07-21
AI lint skill

stejskalleos:

I want AI agents to run Ansible-lint & Ruff checks every time they finish changes.

New skill `.agents/skills/lint/SKILL.md`

Check that the skill is picked up by agents and is executed.

#### Checklist

### DIFF
--- a/.agents/skills/lint/SKILL.md
+++ b/.agents/skills/lint/SKILL.md
@@ -17,24 +17,43 @@ Run project code checks.
 
 ## Workflow
 
-1. Set up virtual environment
+1. Set up virtual environment (builds Ansible collections)
 2. Ansible lint
 3. Ruff
+
+All commands below run from the **project root**.
 
 ### Set up virtual environment
 
 ```shell
-# (in project root dir)
 ./setup-environment
 source .venv/bin/activate
 ```
 
+`setup-environment` installs Python dependencies and runs `make build/collections/foremanctl build/collections/forge`. Ansible lint needs those collection trees (same paths as the `foremanctl` and `forge` wrappers).
+
 ### Ansible lint
 
+Run each tree in a subshell from the project root. Set `ANSIBLE_COLLECTIONS_PATH` and disable system collection scanning so lint resolves the same collections as deploy, without picking up unrelated system installs.
+
 ```shell
-# (in project root dir)
-cd src && ansible-lint
-cd development && ansible-lint
+# src (production playbooks and roles)
+ANSIBLE_COLLECTIONS_PATH="$PWD/build/collections/foremanctl" ANSIBLE_COLLECTIONS_SCAN_SYS_PATH=false \
+  bash -c '(cd src && ansible-lint)'
+
+# development (dev playbooks and roles; exclude vendored collections from lint targets)
+ANSIBLE_COLLECTIONS_PATH="$PWD/build/collections/forge" ANSIBLE_COLLECTIONS_SCAN_SYS_PATH=false \
+  bash -c '(cd development && ansible-lint --exclude ../build/collections)'
+```
+
+Use `$PWD` so collection paths stay correct after the subshell changes directory (a bare `build/collections/...` path would be resolved from inside `src/` or `development/` and fail).
+
+Do **not** chain `cd src && ansible-lint` and `cd development && ansible-lint` in one shell — after the first command the working directory is `src/`, so `cd development` fails.
+
+If ansible-lint reports missing modules or roles despite collections being built, clear stale caches and rerun:
+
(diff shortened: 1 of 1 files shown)


.agents/skills/lint/SKILL.md
```diff
@@ -0,0 +1,47 @@
...
+## Workflow
+
+- Ansible lint
+- Ruff
+
+### Run `ansible-lint`
+
+```shell
+cd src && ansible-lint
+cd development && ansible-lint
```
ekohl: Pretty sure this also needs the virtual environment.

.agents/skills/.gitkeep
ekohl, line 1: I think this file shouldn't be needed.

.agents/skills/lint/SKILL.md
```diff
@@ -0,0 +1,47 @@
+---
+name: lint
+description: >-
+  Run linters after finishing a change. Use when any code (except *.md files) has been changed.
```
ekohl: I think YAML files may not be considered code, but here they are relevant. Is that important to call out?

ehelms:
I've been thinking this kinda of action, and wanting it to be ran by agents is technically better handled by hooks (e.g. https://code.claude.com/docs/en/hooks) but the various tools do not support the same hook configuration making it hard to define it once and re-use it multiple places.

.agents/skills/.gitkeep
stejskalleos: Yeah, I removed it, but here it looks like I added it.

.agents/skills/lint/SKILL.md
stejskalleos: Ok, I will add it as a first step.

.agents/skills/lint/SKILL.md
stejskalleos: > Is that important to call out?

That's a good question. My approach was "better safe than sorry", but I can remove it.

ekohl: Perhaps this is good enough?
```suggestion
  Run linters after finishing a change. Use when any YAML and code files have been changed.
```

.agents/skills/.gitkeep
stejskalleos:
```shell
tree -a .agents/skills
.agents/skills
└── lint
    └── SKILL.md

2 directories, 1 file
```

.agents/skills/lint/SKILL.md
```diff
@@ -0,0 +1,46 @@
...
+# Lint
+
+Run project code checks.
+
+## Canonical docs
+
+- [AGENTS.md](../../../AGENTS.md) — common lint commands
+- [Playbooks and Roles](../../../docs/developer/playbooks-and-roles.md) — secrets, OAuth, naming
+- Custom rules: `.ansible-lint-rules/` (source of truth for enforceable constraints)
+- Config: `.ansible-lint`
```
ekohl: These are files relative to the root while the above are relative to this file. Feels a bit inconsistent to me.

.agents/skills/lint/SKILL.md
```diff
@@ -0,0 +1,46 @@
...
+```shell
+./setup-environment
+source .venv/bin/activate
+```
+
+### Ansible lint
+
+```shell
+cd src && ansible-lint
+cd development && ansible-lint
```
ekohl: Does this work as well?
```suggestion
ansible-lint src development
```

ehelms: No, because the ansible.cfg is within each of those respectively and has relative configuration inside it.

evgeni:
When I load this and ask it to lint, it reports:
```
  ansible-lint (src/) -- all failures are syntax-check[unknown-module] errors for missing collections (containers.podman, community.crypto, theforeman.foreman, ansible.posix). These aren't installed in the lint environment but are
  expected on deployment targets — not real code issues.

  ansible-lint (development/) -- same syntax-check[unknown-module] errors for missing collections (containers.podman, community.general), plus three syntax-check[specific] errors where roles from theforeman.forklift can't be found
  (etc_hosts, smoker, sos_report, resize_disk). The forklift collection is built during setup but the roles aren't resolving in the lint role search path.
```

Why is it not finding the collections?

stejskalleos:
> Why is it not finding the collections?

Huh, what? _It was working ..._

Anyway, I asked AI to fix it, and [this is the result](https://github.com/theforeman/foremanctl/pull/690/changes/2ef77588ef162c49bd0ec7abc15df6395efe6179). Now it's passing (in the cursor at least)

evgeni (approved):

PR #692, opened 2026-07-22
Fix SELinux context on httpd MPM event.conf

shubhamsg199:
After a foremanctl deploy, /etc/httpd/conf.modules.d/event.conf was labeled etc_t instead of httpd_config_t. 
RPM-shipped files in that directory were fine; only the Ansible-templated event.conf was wrong.

Fixes: SAT-47294

* Set setype: httpd_config_t on the MPM event template task so the file gets the correct label on create and on re-runs.
* Added a test to verify the selinux context of http configs

Steps to reproduce:

* foremanctl deploy 
* Check the event.conf selinux context under Apache modules config: `ls -1Z /etc/httpd/conf.modules.d/event.conf`

#### Checklist

### DIFF
--- a/src/roles/httpd/tasks/main.yml
+++ b/src/roles/httpd/tasks/main.yml
@@ -74,6 +74,7 @@
     src: event.conf.j2
     dest: /etc/httpd/conf.modules.d/event.conf
     mode: "0644"
+    setype: httpd_config_t
   notify:
     - Restart httpd
 
@@ -95,6 +96,7 @@
     src: foreman-vhost.conf.j2
     dest: /etc/httpd/conf.d/{{ 'foreman' if httpd_with_foreman else 'pulpcore' }}.conf
     mode: "0644"
+    setype: httpd_config_t
   notify:
     - Restart httpd
 
@@ -103,6 +105,7 @@
     src: foreman-ssl-vhost.conf.j2
     dest: /etc/httpd/conf.d/{{ 'foreman-ssl' if httpd_with_foreman else 'pulpcore-ssl' }}.conf
     mode: "0644"
+    setype: httpd_config_t
   notify:
     - Restart httpd
 
--- a/tests/httpd_test.py
+++ b/tests/httpd_test.py
@@ -136,6 +136,13 @@ def test_httpd_event_conf_contains_threads_per_child(server):
     assert event_conf.contains("ThreadsPerChild")
 
 
+def test_httpd_selinux_context(server):
+    cmd = server.run("ls -1Z /etc/httpd/*.d/*.conf")
+    assert cmd.succeeded
+    incorrect = [line for line in cmd.stdout.splitlines() if line and ":httpd_config_t:" not in line]
+    assert not incorrect, "Incorrect SELinux context (expected httpd_config_t):\n" + "\n".join(incorrect)
+
+
 def test_httpd_config_syntax(server):
     cmd = server.run("httpd -t")
     assert cmd.succeeded


src/roles/httpd/tasks/main.yml
```diff
@@ -74,6 +74,7 @@
     src: event.conf.j2
     dest: /etc/httpd/conf.modules.d/event.conf
     mode: "0644"
+    setype: httpd_config_t
```
evgeni, line 77: I wonder why this is needed.
There is a correct fcontext in place:
```
[root@quadlet ~]# semanage fcontext -l |grep etc.httpd
/etc/httpd(/.*)?                                   all files          system_u:object_r:httpd_config_t:s0 
```
And it correctly finds the misslabeling:
```
[root@quadlet ~]# restorecon -vn /etc/httpd/conf.modules.d/event.conf 
Would relabel /etc/httpd/conf.modules.d/event.conf from system_u:object_r:etc_t:s0 to system_u:object_r:httpd_config_t:s0
```

So why wasn't it labeled correctly to begin with?

And why did we notice it now, but not for the other files that exist much longer?
```
[root@quadlet ~]# ls -alhZ /etc/httpd/conf.d/foreman*
-rw-r--r--. 1 root root system_u:object_r:etc_t:s0 2.7K Jul 20 07:47 /etc/httpd/conf.d/foreman.conf
-rw-r--r--. 1 root root system_u:object_r:etc_t:s0 5.1K Jul 20 07:47 /etc/httpd/conf.d/foreman-ssl.conf
```

evgeni: Interesting, using `ansible.builtin.file` creates the file with the right type, but `.copy` and `.template` do `etc_t`.

evgeni: I had some fun: https://matrix.to/#/!DyJEpptWlrIexlAmAY:libera.chat/$52u85Mo0kU7D2bL-5qNwdERzo-PH0zQDaLZRpHrbTms?via=ansible.com&via=matrix.org&via=ansible.im

(Will write up a detailed bug report in a moment)

evgeni: https://github.com/ansible/ansible/issues/87285

shubhamsg199: > And why did we notice it now, but not for the other files that exist much longer?

Because we have robottelo coverage for this and was reported as the [test](https://github.com/SatelliteQE/robottelo/blob/600d95371fe1655c7caea3077705f670d6ec8204/tests/foreman/installer/test_installer.py#L518) was failing

shubhamsg199: @evgeni Should we switch to `ansible.builtin.file` or this solution is good enough?

evgeni: > > And why did we notice it now, but not for the other files that exist much longer?
> 
> Because we have robottelo coverage for this and was reported as the [test](https://github.com/SatelliteQE/robottelo/blob/600d95371fe1655c7caea3077705f670d6ec8204/tests/foreman/installer/test_installer.py#L518) was failing

Ah, but we should cover all apache files, not only modules.d.



> @evgeni Should we switch to `ansible.builtin.file` or this solution is good enough?

`file` can't manage the content, so… probably not.

This solution seems good enough for now, but should be applied to all files we create in `/etc/httpd`

tests/httpd_test.py
```diff
@@ -136,6 +136,13 @@ def test_httpd_event_conf_contains_threads_per_child(server):
     assert event_conf.contains("ThreadsPerChild")
 
 
+def test_httpd_modules_d_selinux_context(server):
+    cmd = server.run("ls -1Z /etc/httpd/conf.modules.d/*.conf")
```
evgeni:
```suggestion
    cmd = server.run("ls -1Z /etc/httpd/*.d/*.conf")
```
so we catch all of them

evgeni (approved):

tests/httpd_test.py
```diff
@@ -136,6 +136,14 @@ def test_httpd_event_conf_contains_threads_per_child(server):
     assert event_conf.contains("ThreadsPerChild")
 
 
+def test_httpd_modules_d_selinux_context(server):
```
evgeni: Can we then drop the test from robotello, given it's covered here now?

shubhamsg199: We can, but robottelo test also covers `Create a test file using Puppet without seltype parameter` scenario. Do you think that should also be removed?

evgeni: Given there is no Puppet in our installation path, I guess this is not needed for the foremanctl part of the world.

shubhamsg199: Raised a PR for removal: https://github.com/SatelliteQE/robottelo/pull/22236

evgeni (approved):

evgeni:
FTR, this was tested with foremanctl-3.0.0.develop-3.20260723095454638757.pr692.135.g2c2b010.el9.noarch

PR #693, opened 2026-07-22
Proof of concept to use --target within the wrapper script to allow m…

ehelms:
This is a proof of concept of the idea of using a wrapper script to handle multi-host management with the state management restrictions as described in https://github.com/theforeman/foremanctl/issues/630.

The general idea is put the `--target` into the wrapper script and use that to dynamically set the state directory based upon the target's value.

### DIFF
--- a/anvil
+++ b/anvil
@@ -1,0 +1,114 @@
+#!/bin/bash
+
+OBSAH_NAME=anvil
+OBSAH_BASE=$(dirname $(readlink -f $0))
+OBSAH_INVENTORY=${OBSAH_BASE}/inventories
+OBSAH_PERSIST_PARAMS=true
+
+ANSIBLE_COLLECTIONS_PATH=${OBSAH_BASE}/build/collections/foremanctl:${OBSAH_BASE}/build/collections/forge
+ANSIBLE_COLLECTIONS_SCAN_SYS_PATH=false
+
+obsah_help() {
+    OBSAH_NAME=anvil OBSAH_DATA="${OBSAH_BASE}/$1" OBSAH_INVENTORY="${OBSAH_INVENTORY}" \
+        OBSAH_STATE="${OBSAH_BASE}/.var/lib/anvil" obsah --help 2>&1
+}
+
+extract_commands() {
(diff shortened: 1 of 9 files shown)


anvil
```diff
@@ -0,0 +1,114 @@
...
+
+Development commands:
+$(obsah_help development | extract_commands)
+
+Run 'anvil <command> --help' for command-specific options.
+EOF
+}
+
+# Extract --target before obsah sees the arguments
+TARGET=""
```
ekohl, line 39: Why clear the value? If you keep it empty you can also use
```sh
export TARGET=myserver
```
And then keep using `anvil` without remembering the target the whole time.

```suggestion
```

anvil
```diff
@@ -0,0 +1,114 @@
...
+            TARGET="$2"
+            shift 2
+            ;;
+        *)
+            ARGS+=("$1")
+            shift
+            ;;
+    esac
+done
+set -- "${ARGS[@]}"
```
ekohl, line 61: Why is this needed? I don't think I've ever needed it. Isn't it safe to just drop that?
```suggestion
    esac
done
```

anvil
```diff
@@ -0,0 +1,114 @@
...
+        ANSIBLE_LOG_PATH=${OBSAH_STATE}/anvil.log
+        export ANSIBLE_LOG_PATH
+        unset http_proxy HTTP_PROXY https_proxy HTTPS_PROXY
+        ;;
+    deploy-dev|vms|test|smoker|custom-certs|fetch-bundle|security|sos|lock)
+        OBSAH_DATA=${OBSAH_BASE}/development
+        if [[ -n "$TARGET" ]]; then
+            OBSAH_STATE=${OBSAH_BASE}/.var/lib/anvil/dev/${TARGET}
+        else
+            OBSAH_STATE=${OBSAH_BASE}/.var/lib/anvil/dev
```
ekohl, line 96: Given the repetition, perhaps define `ANVIL_BASE=${OBSAH_BASE}/.var/lib/anvil` and reuse that?

PR #695, opened 2026-07-22
Add candlepin tuning

Imaanpreet:
Candlepin JVM heap is hardcoded in tomcat.conf. It should scale with the tuning profile like other components (httpd, PostgreSQL).

  - Templatized -Xms/-Xmx in tomcat.conf using candlepin_xms/candlepin_xmx variables
  - Role defaults: 1g/4g (matches puppet-candlepin (https://github.com/theforeman/puppet-candlepin/blob/072f08a/manifests/init.pp#L216))
  - Extra-large and extra-extra-large override candlepin_xmx: 8g (matches foreman-installer
  (https://github.com/theforeman/foreman-installer/blob/0448d5d/config/foreman.hiera/tuning/sizes/extra-extra-large.yaml#L5))

  1. foremanctl deploy --tuning default
  2. podman exec candlepin cat /proc/1/cmdline | tr '\0' ' ' → verify -Xms1g -Xmx4g
  3. foremanctl deploy --tuning extra-large (32+ cores, 128GB+) → verify -Xms1g -Xmx8g

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/roles/candlepin/defaults/main.yml
+++ b/src/roles/candlepin/defaults/main.yml
@@ -19,3 +19,6 @@ candlepin_database_ssl: false
 candlepin_database_ssl_mode: disable
 candlepin_database_ssl_ca: # noqa: no-empty-defaults
 candlepin_database_ssl_ca_path: /etc/candlepin/certs/db-ca.crt
+
+candlepin_java_opts_xms: 1g
+candlepin_java_opts_xmx: 4g
--- a/src/roles/candlepin/tasks/main.yml
+++ b/src/roles/candlepin/tasks/main.yml
@@ -32,7 +32,7 @@
   containers.podman.podman_secret:
     state: present
     name: candlepin-tomcat-conf
-    data: "{{ lookup('ansible.builtin.template', 'tomcat.conf') }}"
+    data: "{{ lookup('ansible.builtin.template', 'tomcat.conf.j2') }}"
     labels:
       filename: tomcat.conf
       app: tomcat
--- a/src/roles/candlepin/templates/tomcat.conf.j2
+++ b/src/roles/candlepin/templates/tomcat.conf.j2
@@ -28,7 +28,7 @@ CATALINA_HOME="/usr/share/tomcat"
 CATALINA_TMPDIR="/var/cache/tomcat/temp"
 
 # You can pass some parameters to java here if you wish to
-JAVA_OPTS="-Xms1024m -Xmx4096m -Dcom.redhat.fips=false --enable-native-access=ALL-UNNAMED"
+JAVA_OPTS="-Xms{{ candlepin_java_opts_xms }} -Xmx{{ candlepin_java_opts_xmx }} -Dcom.redhat.fips=false --enable-native-access=ALL-UNNAMED"
 
 # You can change your tomcat locale here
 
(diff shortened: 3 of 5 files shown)


pablomh:
+1

src/roles/candlepin/defaults/main.yml
```diff
@@ -19,3 +19,6 @@ candlepin_database_ssl: false
 candlepin_database_ssl_mode: disable
 candlepin_database_ssl_ca: # noqa: no-empty-defaults
 candlepin_database_ssl_ca_path: /etc/candlepin/certs/db-ca.crt
+
+candlepin_xms: 1g
+candlepin_xmx: 4g
```
ehelms: How about including `java_opts`?
```suggestion
candlepin_java_opts_xmx: 4g
```

evgeni:
In `puppet-candlepin`, we used to just have a single `java_opts` string, that you could fill with whatever you wanted: https://github.com/theforeman/puppet-candlepin/blob/072f08a2e3d9bb1f5dc15bcb768151ec0015476c/manifests/init.pp#L216
It's basically a tradeoff between "being able to selectively override defaults" (this PR) and "be free to pass any opt you want" (`puppet-candlepin`).

My Tomcat knowledge is super rusty, so maybe @Imaanpreet and @pablomh know better: what are the odds that you want to alter only ms or only mx values (looking at this PR and the tuning in foreman-installer: high?)? and do we ever recommend adding any other options? 

I only could find https://access.redhat.com/solutions/6999123 which adds `-XX:+UseParallelGC -XX:ParallelGCThreads=16` but looks rather old (Satellite 6.9 on EL7)

Imaanpreet:
> In `puppet-candlepin`, we used to just have a single `java_opts` string, that you could fill with whatever you wanted: https://github.com/theforeman/puppet-candlepin/blob/072f08a2e3d9bb1f5dc15bcb768151ec0015476c/manifests/init.pp#L216 It's basically a tradeoff between "being able to selectively override defaults" (this PR) and "be free to pass any opt you want" (`puppet-candlepin`).
> 
> My Tomcat knowledge is super rusty, so maybe @Imaanpreet and @pablomh know better: what are the odds that you want to alter only ms or only mx values (looking at this PR and the tuning in foreman-installer: high?)? and do we ever recommend adding any other options?
> 
> I only could find https://access.redhat.com/solutions/6999123 which adds `-XX:+UseParallelGC -XX:ParallelGCThreads=16` but looks rather old (Satellite 6.9 on EL7)

Thanks for raising this, @evgeni ! 

The separate variables work well for our use case since tuning profiles only need to adjust heap sizes — for example, XL and XXL just override xmx to 8g while keeping the default xms. The other JAVA_OPTS flags (-Dcom.redhat.fips=false, --enable-native-access=ALL-UNNAMED) are fixed in the template and don't change across profiles.

I don't think there's a current need for passing additional JVM options through tuning profiles. The GC tuning example you found looks like it was specific to Satellite 6.9/EL7. But if that need comes up in the future, we could always add something like candlepin_java_opts_extra without changing the existing setup.

Happy to hear what you and @pablomh think!

src/roles/candlepin/defaults/main.yml
Imaanpreet: Done! Renamed to candlepin_java_opts_xms / candlepin_java_opts_xmx. 
Thanks @ehelms

ehelms (approved):
Based on the discussion and conclusions, this change looks good.

PR #696, opened 2026-07-22
Fix httpd proxy backend precedence in deploy-dev playbook

jeremylenz:

The deploy-dev playbook sets `httpd_foreman_backend: "http://localhost:3000"` in the play-level `vars:` block, but `src/vars/base.yaml` (loaded via `vars_files:`) overrides it with the Unix socket path. In Ansible's variable precedence, `vars_files` beats `vars`, so httpd was always configured to proxy to a nonexistent Unix socket instead of TCP port 3000, resulting in 503 errors.

This was latent since `eedbe40` (March 2025) when Unix socket support was added to `base.yaml`, but went unnoticed because the `vars:` block originally appeared *below* `vars_files:` in the YAML, and the ordering happened to make the override work. [#623](https://github.com/theforeman/foremanctl/pull/623) ("Move flavor default into playbook vars", July 6 2026) moved `vars:` above `vars_files:`, which broke the accidental ordering and surfaced the bug.

* Add `development/vars/devel.yaml` with dev-specific variable overrides (currently `httpd_foreman_backend`)
* Load it last in `vars_files` so it overrides `base.yaml` at the same precedence level
* Remove the ineffective override from play-level `vars:`

* Deploy using `./forge deploy-dev --target-host localhost`
* Verify `/etc/httpd/conf.d/foreman.conf` contains `ProxyPass / http://localhost:3000/` (not a Unix socket path)
* Confirm no 503 errors when accessing Foreman through httpd

#### Checklist
* [x] Tests added/updated (if applicable) - No new tests needed; existing httpd functional tests cover this once [#715](https://github.com/theforeman/foremanctl/pull/715) enables running tests against dev deployments
* [ ] Documentation updated (if applicable) - No documentation changes needed

### DIFF
--- a/development/playbooks/deploy-dev/deploy-dev.yaml
+++ b/development/playbooks/deploy-dev/deploy-dev.yaml
@@ -4,7 +4,6 @@
   become: true
   vars:
     flavor: katello
-    httpd_foreman_backend: "http://localhost:3000"
     pulp_register_foreman_proxy: false
   vars_files:
     - "../../../src/vars/defaults.yml"
@@ -14,6 +13,7 @@
     - "../../../src/vars/database.yml"
     - "../../../src/vars/foreman.yml"
     - "../../../src/vars/base.yaml"
+    - "../../vars/devel.yaml"
   pre_tasks:
     - name: Set development postgresql databases
       ansible.builtin.set_fact:
--- a/development/vars/devel.yaml
+++ b/development/vars/devel.yaml
@@ -1,0 +1,2 @@
+---
+httpd_foreman_backend: "http://localhost:3000"


development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -48,6 +47,8 @@
     - role: valkey
     - role: candlepin
     - role: httpd
+      vars:
+        httpd_foreman_backend: "http://localhost:3000"
```
Gauravtalreja1: The `httpd` role already defines this variable in its defaults, so I don't think we need to set it explicitly in the dev playbook, the role default should take care of it, right?

jeremylenz: Good thought! The role default is indeed `http://localhost:3000`, but `base.yaml` overrides it to the Unix socket value — and since `vars_files` has higher precedence than role defaults, the socket value wins. That's actually what was causing the 503s. The role param here is what bumps us back above `vars_files` in Ansible's precedence order.

jeremylenz:
Updated the description with some additional info about how the bug was introduced and then surfaced.

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -42,12 +41,15 @@
...
-    - role: systemd_target
     - role: certificates
     - role: postgresql
     - role: valkey
     - role: candlepin
     - role: httpd
+      vars:
+        httpd_foreman_backend: "http://localhost:3000"
+        httpd_pulp_api_backend: "http://localhost:24817"
+        httpd_pulp_content_backend: "http://localhost:24816"
```
ehelms: Are the pulp ones needed? Given Pulp runs as a backend the same way in production and development, it should be able to continue to operate via socket.

jeremylenz: Before fixing this, I could not access content from a host:
```
# dnf install npm
Updating Subscription Management repositories.

The system clock is skewed. There is a time difference of 19.0 seconds with the entitlement server. Please check your clock settings to ensure access to all entitled content.

Red Hat Enterprise Linux 9 for x86_64 - BaseOS (RPMs)                                                                                                       5.1 kB/s | 299  B     00:00    
Errors during downloading metadata for repository 'rhel-9-for-x86_64-baseos-rpms':
  - Status code: 503 for https://<redacted>/pulp/content/Default_Organization/Library/content/dist/rhel9/9/x86_64/baseos/os/repodata/repomd.xml (IP: xxx)
Error: Failed to download metadata for repo 'rhel-9-for-x86_64-baseos-rpms': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried

```

The 503 was being served not by Rails, but by httpd.

jeremylenz: I am happy to make it work via socket if that is more correct, I would just need to learn how

ehelms: There are 503's on the test suite. Which points to a different issue. The use of sockets here shouldn't be affecting this.

ehelms: I dropped the two pulp lines and it worked locally for me.

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -42,12 +41,15 @@
...
   roles:
     - role: pre_install
-    - role: systemd_target
     - role: certificates
     - role: postgresql
     - role: valkey
     - role: candlepin
     - role: httpd
+      vars:
+        httpd_foreman_backend: "http://localhost:3000"
```
evgeni: I am debating whether these should be set in a `devel.yaml` that we load last in `vars_files`. Seems a bit cleaner to me, but certainly not blocking.

development/playbooks/deploy-dev/deploy-dev.yaml
jeremylenz: Seems CI agrees. resolving

development/playbooks/deploy-dev/deploy-dev.yaml
jeremylenz: I am beginning to be convinced. any other opinions?

evgeni: The other person with opinions is out, so we can merge quickly behind his back :p

evgeni (approved):

PR #689, opened 2026-07-21
New param --valkey-log-level for Valkye logs

stejskalleos:
#### Why are you introducing these changes? (Problem description, related links)

To allow users to configure log levels for the Valkey service.

#### What are the changes introduced in this pull request?

* New parameter `--valkye-log-level`

#### How to test this pull request

Steps to reproduce:

```
foremanctl deploy --valkey-log-level debug
# Accepted values: `debug`, `verbose`, `notice`, `warning`.

journalctl -f -u valkey.service
```

#### Checklist
* IMO tests are not needed
* `parameters.md` updated

PR #697, opened 2026-07-23
Add --log-level-pulp

stejskalleos:

New parameter for managing log levels for all pulp services.

* https://github.com/theforeman/pulp-oci-images/commit/8de94a4105b115370c2c63a36368f926e8360615
* `foremanctl deploy --pulp-log-level `

Steps to reproduce:

* `foremanctl deploy --pulp-log-level debug|info|warning|error|critical`
* `foremanctl deploy-proxy --pulp-log-level debug|info|warning|error|critical`

#### Checklist
* ~Tests added/updated!~

Similar PR for Valkey: https://github.com/theforeman/foremanctl/pull/689

### DIFF
--- a/src/playbooks/_logging/metadata.obsah.yaml
+++ b/src/playbooks/_logging/metadata.obsah.yaml
@@ -25,3 +25,12 @@ variables:
       - notice
       - warning
       - nothing
+  pulp_log_level:
+    help: Log level for Pulp. Defaults to the global log level.
+    parameter: --pulp-log-level
+    choices:
+      - debug
+      - info
+      - warning
+      - error
+      - critical
--- a/docs/user/parameters.md
+++ b/docs/user/parameters.md
@@ -88,6 +88,12 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--foreman-log-level` | Log level for Foreman. Overrides `--log-level` for Foreman only. | `--foreman-logging-level` |
 | `--foreman-proxy-log-level` | Log level for Foreman Proxy. Overrides `--log-level` for Foreman Proxy only. | `--foreman-proxy-log-level` |
 
+##### New
+
+| Parameter | Description |
+| ----------| ----------- |
+| `--pulp-log-level` | Log level for Pulp. Overrides `--log-level` for Pulp only. Accepted values: `debug`, `info`, `warning`, `error` & `critical` |
+
 #### Undetermined
 
 | foreman-installer Parameter | Description | Module | Puppet Parameter | Keep |
--- a/src/roles/pulp/README.md
+++ b/src/roles/pulp/README.md
@@ -28,5 +28,6 @@ Usage Inside foremanctl
 -----------------------
 When used as part of `foremanctl`, the variables are setup as the following
 - `pulp_worker_count`: `--pulp-worker-count`
+- `pulp_log_level`: `--pulp-log-level`
 - `pulp_import_paths`: `--content-import-path` (may be specified multiple times)
 - `pulp_export_paths`: `--content-export-path` (may be specified multiple times)
--- a/src/roles/pulp/defaults/main.yaml
+++ b/src/roles/pulp/defaults/main.yaml
@@ -82,6 +82,9 @@ pulp_settings_other_env:
   PULP_SMART_PROXY_RHSM_URL: "{{ pulp_rhsm_url | default('') }}"
   PULP_API_WORKERS: "{{ pulp_api_service_worker_count }}"
   PULP_CONTENT_WORKERS: "{{ pulp_content_service_worker_count }}"
+  PULP_LOG_LEVEL: "{{ pulp_log_level }}"
+  PULP_LOGGING__HANDLERS__CONSOLE__LEVEL: "{{ pulp_log_level | upper }}"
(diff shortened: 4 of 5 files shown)


src/playbooks/_logging/metadata.obsah.yaml
```diff
@@ -17,3 +17,12 @@ variables:
...
       - fatal
+  pulp_log_level:
+    help: Log level for Pulp. Defaults to the global log level.
+    parameter: --pulp-log-level
+    choices:
+      - debug
+      - info
+      - warning
+      - error
+      - critical
```
evgeni, line 36: Should we expose the names as Pulp expects them, or unify them to what we have for `--log-level` and translate later?

(I am not sure myself)

stejskalleos: We [do that](https://github.com/theforeman/foremanctl/pull/697/changes#diff-a533c2d5f9eec6c127521329da94ec169ed70a4fb8b0afcd306ecc7e2b2a88b0R7) for the initial setup when using `--log-level`, but when users configure log levels for the service itself with `--<service>-log-level`, I believe we should use the service's own names.

Alleny244:
Tested with `foremanctl-3.0.0.develop-3.20260723082058835462.pr697.136.gf6ad539.el9.noarch`.

`--pulp-log-level error` is only partially working. It sets `PULP_LOG_LEVEL=error`, and Gunicorn respects it, but Django logging for the API and workers still stays at `INFO`.

I think we also need to set the Django logging levels

```yaml
PULP_LOGGING__HANDLERS__CONSOLE__LEVEL: "{{ pulp_log_level | upper }}"
PULP_LOGGING__LOGGERS____LEVEL: "{{ pulp_log_level | upper }}"
```

src/playbooks/_logging/metadata.obsah.yaml
evgeni: I know, but for foreman-proxy, we did not: https://github.com/theforeman/foremanctl/pull/640 uses all lower case names, and only does upper when it does the deployment.

stejskalleos: Huh, in that case, I think we should fix the proxy logging so it behaves the same as valkey and pulp.

stejskalleos: On the other hand, would it not be better to have the same naming conventions for foreman-* services?

I always found `foreman-installer --foreman-logging-level "debug" --foreman-proxy-log-level "DEBUG"` confusing, never able to remember which one is supposed to be uppercase and which one lowercase.

evgeni: I do think the current behaviour for the proxy is fine.
I was merely pointing out the inconsistency to this PR.

stejskalleos:
@Alleny244 Added the vars, quickly tested it, and the debug is set:

```
Jul 24 08:32:58 quadlet.example.com pulp-api[62380]: pulp [None]: pulpcore.app.entrypoint:DEBUG: Api App '7@pulp-api.quadlet.local' heartbeat written, sleeping for '45.0' seconds
^C
Jul 24 08:33:01 quadlet.example.com pulp-worker-2[62975]: pulp [None]: pulpcore.tasking.worker:DEBUG: Worker heartbeat from '1@pulp-worker-2.quadlet.local' at time 2026-07-24 08:32:51.336690+00:00
```

Alleny244:
Tested with `foremanctl-3.0.0.develop-3.20260724083602629106.pr697.141.gd40791c.el9.noarch`.

Verified that `--pulp-log-level` now works as expected. Thanks!

ehelms:
> * foremanctl deploy --pulp-log-level

The PR description only mentions `deploy`, but `deploy-proxy` would need this Pulp logging configuration as well. Is that a follow up or you want to handle that here for consistency?

stejskalleos:
@ehelms `deploy-proxy` already has [logging included](https://github.com/theforeman/foremanctl/blob/master/src/playbooks/deploy-proxy/deploy-proxy.yaml#L15), so it should work:

```
foremanctl deploy-proxy --help
...
  --log-level {debug,info,warn,error,fatal}
                        Default log level for all services. (persisted)
  --pulp-log-level {debug,info,warning,error,critical}
                        Log level for Pulp. Defaults to the global log level. (persisted)
```

src/playbooks/_logging/metadata.obsah.yaml.orig
arvind4501, line 1: i think this is pushed by mistake 😄

stejskalleos: OMG

evgeni (approved):
ack pending lint fixes

stejskalleos:
> ack pending lint fixes

Interesting, I don't have any lint errors on the local :thinking:

evgeni:
> > ack pending lint fixes
> 
> Interesting, I don't have any lint errors on the local 🤔

ansible-lint complained about too many blank lines: https://github.com/theforeman/foremanctl/actions/runs/30262335773/job/89964930845
I think your most recent push fixes that.

arvind4501 (approved):

PR #698, opened 2026-07-23
transfer oauth creds as part of certificate bundle

arvind4501:

Proxy deployment currently requires passing `--oauth-consumer-key` and `--oauth-consumer-secret` manually with `deploy-proxy` command. Those credentials already exist on the Foreman server and are required for the proxy to communicate with Foreman.
With these changes oauth credentials will  travel with the bundle instead of being supplied as extra CLI parameters.

* Include Foreman OAuth consumer key/secret in the tarball produced by `certificate-bundle`
* During `deploy-proxy`, read OAuth credentials from the extracted bundle (`/var/lib/foremanctl/certs/`) when `flavor == foreman-proxy-content`
* Remove `--oauth-consumer-key` / `--oauth-consumer-secret` from `deploy-proxy`

Steps to reproduce:

* Deploy Foreman, then generate a proxy bundle:
  `./foremanctl auth-bundle proxy.example.com`

* Confirm the tarball includes OAuth files:
  `tar -tzvf /var/lib/foremanctl/certs/bundles/proxy.example.com.tar.gz | grep oauth`

* Deploy the proxy without OAuth CLI flags:
  `./foremanctl deploy-proxy --flavor foreman-proxy-content --auth-bundle <bundle> --foreman-fqdn <foreman-fqdn>`

* Confirm extracted files exist on the proxy:
  `/var/lib/foremanctl/certs/foreman-oauth-consumer-key`

#### Checklist


src/roles/foreman_proxy/tasks/oauth_from_bundle.yml
```diff
@@ -0,0 +1,34 @@
...
+  when: foreman_proxy_oauth_secret_file.stat.exists
+  block:
+    - name: Slurp OAuth consumer secret
+      ansible.builtin.slurp:
+        src: "{{ certificates_ca_directory }}/foreman-oauth-consumer-secret"
+      register: foreman_proxy_oauth_secret
+
+    - name: Set OAuth consumer secret from certificate bundle
+      ansible.builtin.set_fact:
+        foreman_proxy_oauth_consumer_secret: "{{ foreman_proxy_oauth_secret.content | b64decode | trim }}"
```
Gauravtalreja1: If the OAuth files are missing from the bundle in cases of old bundle or user accidently deletes them or something else, there's no CLI fallback anymore, the deploy will fail later with an unclear undefined variable error during proxy registration. so how about a better to fail early with a clear message telling the user to regenerate the bundle if these oauth secret and key are missing?

tests/flavor/foreman-proxy-content/oauth_test.py
```diff
@@ -0,0 +1,17 @@
...
+CERTS_DIR = '/var/lib/foremanctl/certs'
+
+
+@pytest.mark.parametrize("oauth_file", OAUTH_FILES)
+def test_oauth_credentials_extracted_from_bundle(server, oauth_file):
+    """Proxy deploy must extract OAuth credentials from the certificate bundle."""
+    oauth_path = f'{CERTS_DIR}/{oauth_file}'
+    oauth = server.file(oauth_path)
+    assert oauth.exists, f'{oauth_path} was not extracted from the certificate bundle'
+    assert oauth.size > 0
```
Gauravtalreja1, line 17: should we also test if it got a right permissions as specified in code?

src/roles/foreman_proxy/tasks/oauth_from_bundle.yml
```diff
@@ -0,0 +1,34 @@
...
+- name: Check for OAuth consumer secret in certificate bundle
+  ansible.builtin.stat:
+    path: "{{ certificates_ca_directory }}/foreman-oauth-consumer-secret"
+  register: foreman_proxy_oauth_secret_file
+
+- name: Read OAuth consumer secret from certificate bundle
+  when: foreman_proxy_oauth_secret_file.stat.exists
+  block:
+    - name: Slurp OAuth consumer secret
+      ansible.builtin.slurp:
```
Gauravtalreja1: Can we use a loop here instead of two separate stat+slurp blocks?

src/roles/certificate_bundle/tasks/main.yml
```diff
@@ -77,6 +77,20 @@
         remote_src: true
         mode: '0440'
 
+    - name: Copy OAuth consumer key
+      ansible.builtin.copy:
+        src: "{{ certificate_bundle_oauth_consumer_key_file }}"
+        dest: "{{ certificate_bundle_build_directory.path }}/certs/foreman-oauth-consumer-key"
```
Gauravtalreja1: Just a thought, should the OAuth files go into the `private/` subdirectory to stay consistent with how other secrets (keys) are organized in the bundle instead of `certs/`?

src/roles/foreman_proxy/tasks/oauth_from_bundle.yml
arvind4501: i think #90 is not released yet so no old users had used bundle in older releases. for user deleting those is same as user deleting certs inside bundle, so i don't  think we should handle that

src/roles/certificate_bundle/tasks/main.yml
arvind4501: I don't think so, currently it looks like

```
tar -tzvf proxy.example.com.tar.gz
-r--r--r-- root/root      1903 2026-06-19 13:55 certs/proxy.example.com.crt
-r--r--r-- root/root      1818 2026-06-19 05:29 certs/ca-bundle.crt
-r--r--r-- root/root      1903 2026-06-19 13:55 certs/proxy.example.com-client.crt
-r--r--r-- root/root      1818 2026-06-19 05:29 certs/server-ca.crt
-r--r--r-- root/root      1818 2026-06-19 05:29 certs/ca.crt
-r--r----- root/root        33 2026-07-23 06:36 foreman-oauth-consumer-key
-r--r----- root/root        33 2026-07-23 06:36 foreman-oauth-consumer-secret
-r--r----- root/root      3247 2026-06-19 13:55 private/proxy.example.com.key
-r--r----- root/root      3243 2026-06-19 13:55 private/proxy.example.com-client.key

```
which keeps oauth creds sepreate from both `certs` and `private` and ideally we should not mix those.
so Keeping them at the bundle root makes it clear they’re separate

src/roles/foreman_proxy/tasks/oauth_from_bundle.yml
Gauravtalreja1: I agree, but it would be a nice cosmetic check to validate files are present early, before we start a deploy with bundle, else it would fail later during proxy registration

arvind4501: sorry, what do you mean by early? in `checks` role? unfortunately at that time we have even not extracted the bundle so no way to check that early

src/roles/certificate_bundle/tasks/main.yml
evgeni: I would have done a separate `oauth` folder, but the root works too. `private` is for "private keys" in the sense of X.509

src/roles/foreman_proxy/tasks/oauth_from_bundle.yml
Gauravtalreja1: I don't mean checks role, instead a fail msg like here if either oauth secret or key are missing
```
  - name: Fail if OAuth credentials are missing from the certificate bundle
    ansible.builtin.fail:
      msg: >-
        OAuth credentials not found in the certificate bundle.
        Please regenerate the bundle on the Foreman server using
        'foremanctl certificate-bundle {{ ansible_facts['fqdn'] }}'
        and redeploy.
    when:
      - not foreman_proxy_oauth_key_file.stat.exists or not foreman_proxy_oauth_secret_file.stat.exists
 ```

Comment (arvind4501) [reply src/roles/foreman_proxy/tasks/oauth_from_bundle.yml]: It exists now, and way earlier than before, as we can't do that before extraction of certificates it feels a good place

File: src/roles/certificate_bundle/tasks/main.yml (hunk 4 above)
Comment (ehelms) [reply src/roles/certificate_bundle/tasks/main.yml]: I agree with @evgeni -- separate folder is clear.

**Comment by ehelms** [comment]
I think it would be nice to combine the extraction into a single extraction step of the bundle for everything in it, getting placed in the right locations.

File: src/roles/oauth_from_bundle/tasks/credential.yml [hunk 5]
```diff
@@ -0,0 +1,27 @@
+---
+- name: Load OAuth credential from certificate bundle - {{ oauth_from_bundle_credential }}
```
ehelms: The way I wanted to design the bundle was for it to be "throw away" after initial use. Can the oauth be loaded into files in the state directory and stored for reading like it is in the deploy Foreman case?

evgeni: The way it's implemented right now, the bundle.tar.gz is throw away. The reading here happens from a file on disk after extraction from a sub-folder of our state directory on the target node.

Emphasis on "target node", as that breaks the "can it be the same as Foreman" case.
In the Foreman case we store the data on the control node (lookups run on the control node), not the target node.

We could "double extract" the bundle: once the certs part on the target, and the oauth part on the controller, but I'm not sure that complexity is worth it?

tests/flavor/foreman-proxy-content/oauth_test.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+OAUTH_FILES = [
+    'foreman-oauth-consumer-key',
+    'foreman-oauth-consumer-secret',
+]
+
+OAUTH_DIR = '/var/lib/foremanctl/certs/oauth'
```
evgeni: I think I'd prefer if the oauth files would be in `/var/lib/foremanctl/oauth`, but that'd mean we'd need to re-do how the bundle is generated/extracted a bit more.

Not blocking, can be done in a follow up or filed under "shut up Evgeni, you're too picky" :)

arvind4501: Sure, i gave it a try in https://github.com/theforeman/foremanctl/pull/698/changes/6e6f7c7c0f9a7f1a7715c611790e0032077a475f

src/roles/certificates/tasks/extract.yml
```diff
@@ -11,7 +11,20 @@
...
     dest: "{{ certificates_ca_directory }}"
+    exclude:
+      - oauth
+      - oauth/*
+
+- name: Extract OAuth credentials from bundle
+  ansible.builtin.unarchive:
+    src: "{{ certificates_bundle }}"
+    dest: "{{ certificates_oauth_directory | dirname }}"
+    exclude:
```
evgeni, line 26: you can use `include` with the oauth paths, that's easier than exclude

src/roles/certificates/tasks/extract.yml
```diff
@@ -11,7 +11,18 @@
...
+      - oauth
+      - oauth/*
+
+- name: Extract OAuth credentials from bundle
+  ansible.builtin.unarchive:
+    src: "{{ certificates_bundle }}"
+    dest: "{{ certificates_oauth_directory | dirname }}"
+    include:
+      - oauth
+      - oauth/*
```
evgeni: looks like include wants real paths, not wildcards :(
sorry

ehelms:
We could switch from `certificate_bundle` to `auth_bundle` but that's an idle thought :)

src/roles/certificate_bundle/defaults/main.yml
```diff
@@ -1,3 +1,5 @@
 ---
 certificate_bundle_output_directory: "{{ certificates_ca_directory }}/bundles"
 certificate_bundle_server_ca_certificate: "{{ certificate_bundle_ca_certificate }}"
+certificate_bundle_oauth_consumer_key_file: # noqa: no-empty-defaults
+certificate_bundle_oauth_consumer_secret_file: # noqa: no-empty-defaults
```
ehelms: Should we just omit these if they are empty? What's the value of having empty defaults?

src/roles/certificate_bundle/tasks/main.yml
```diff
@@ -77,6 +78,20 @@
         remote_src: true
         mode: '0440'
 
+    - name: Copy OAuth consumer key
+      ansible.builtin.copy:
+        src: "{{ certificate_bundle_oauth_consumer_key_file }}"
+        dest: "{{ certificate_bundle_build_directory.path }}/certs/oauth/foreman-oauth-consumer-key"
+        mode: '0440'
+      when: certificate_bundle_oauth_consumer_key_file is exists
```
ehelms: If this file doesn't exist should we skip over that? Or let it error? Seems fairly serious if it doesn't exist.

src/roles/oauth_from_bundle/tasks/credential.yml
ehelms: Sounds like what you are really saying is that the certificates should exist on the control node, and normalize that everything in `/var/lib/foremanctl` should be state?

evgeni: maaaaaaaaaaybe. (it's friday afternoon)

src/roles/certificate_bundle/defaults/main.yml
arvind4501: Removed them. The empty defaults only had value when paired with the `when: ... is exists` guards on the copy tasks, the empty value helped to safely skip tasks when caller did not provide these vars, but in our case `certificate-bundle` playbook always sends these vars so good to remove

arvind4501:
> We could switch from `certificate_bundle` to `auth_bundle` but that's an idle thought :)

maybe, with that do we want to change the command from `foremanctl certificate-bundle <hostname>` to `foremanctl auth-bundle <hostname>`, or just the role name? i think it boils down to what command name we want to expose to users :)

evgeni:
> > We could switch from `certificate_bundle` to `auth_bundle` but that's an idle thought :)
> 
> maybe, with that do we want to change the command from `foremanctl certificate-bundle <hostname>` to `foremanctl auth-bundle <hostname>`, or just the role name? i think it boils down to what command name we want to expose to users :)

Given the bundle now has more than certificates, I think changing the command name makes sense.

.github/workflows/test.yml
```diff
@@ -420,15 +420,15 @@ jobs:
             --initial-admin-password=changeme \
             --tuning development \
             --add-feature foreman-proxy
-      - name: Generate certificates bundle on quadlet
+      - name: Generate auth bundle on quadlet
         run: |
-          ./foremanctl certificate-bundle proxy.example.com \
+          ./foremanctl auth-bundle proxy.example.com \
```
evgeni, line 427: This will require updates to
- https://github.com/theforeman/foreman-documentation/pull/5097/
- https://github.com/SatelliteQE/robottelo/blob/11ba2027ae74a1f905b6fc9b56e9b13ca1b0b7e5/tests/foreman/foremanctl/test_install_foremanctl.py#L568

evgeni (approved):

.github/workflows/test.yml
arvind4501: https://github.com/SatelliteQE/robottelo/pull/22277
 for doc PR, i hope to get this PR in before that :smile:

src/playbooks/certificate-bundle/certificate-bundle.yaml
```diff
@@ -1,31 +0,0 @@
----
-- name: Generate a certificate bundle for a hostname
-  hosts:
-    - quadlet
-  become: true
```
Gauravtalreja1, line 5: For such renames we should consider using `git mv <source> <destination>` instead of deleting and creating a new file; it preserves the old file history and makes it easier to track changes across commits

evgeni: gits rename detection should do that automatically if the file contents match enough (`git mv` is also just an `rm` and `add` under the hood), which they don't in this case.

`src/playbooks/{certificate-bundle => auth-bundle}/metadata.obsah.yaml` is displayed as a rename for me, but not the main file :(

arvind4501: agree with evgeni, i expected git to take care of such moves, but i think since content of those files is changed significantly, they don't come under move
looking at https://git-scm.com/docs/gitdiffcore#_rename_detection, by default if > 50% file content is changed, would be considered as delete and add

src/roles/oauth_from_bundle/tasks/credential.yml
arvind4501: is that something we need to tackle now?

evgeni: IMHO no

Gauravtalreja1 (approved):
ACK, Tested with `foremanctl-3.0.0.develop-3.20260724120313055342.pr698.143.g745e9f5.el9.noarch` 🍏

**Steps**: deployed Foreman server → generated auth bundle → verified OAuth files in tarball → deployed proxy without --oauth-consumer-key/secret → verified OAuth files extracted on proxy with correct permissions → confirmed proxy registered with Foreman and features refresh works → verified old CLI flags are rejected.

**One minor issue found**: the fail-early check for missing OAuth files doesn't trigger when using a bundle generated before this PR (old certificate-bundle), since the extraction step excludes files outside certs/ and private/, which isn't properly handled in Ansible stdout as shown below:

```
TASK [certificates : Extract OAuth credentials from bundle] ***********************************************************************************************************************************
    changed: false
    msg: |-
        Failed to find handler for "/root/.ansible/tmp/ansible-tmp-1785238624.9872594-1828915-110760222717154/source". Make sure the required command to extract the file is installed.
        Command "/usr/bin/gtar" could not handle archive: Unable to list files in the archive: tar (child): bzip2: Cannot exec: No such file or directory
        tar (child): Error is not recoverable: exiting now
        /usr/bin/gtar: Child returned status 2
        /usr/bin/gtar: Error is not recoverable: exiting now

        Command "/usr/bin/unzip" could not handle archive:   End-of-central-directory signature not found.  Either this file is not
          a zipfile, or it constitutes one disk of a multi-part archive.  In the
          latter case the central directory and zipfile comment will be found on
          the last disk(s) of this archive.
        note:  /root/.ansible/tmp/ansible-tmp-1785238624.9872594-1828915-110760222717154/source may be a plain executable, not an archive
        unzip:  cannot find zipfile directory in one of /root/.ansible/tmp/ansible-tmp-1785238624.9872594-1828915-110760222717154/source or
                /root/.ansible/tmp/ansible-tmp-1785238624.9872594-1828915-110760222717154/source.zip, and cannot find /root/.ansible/tmp/ansible-tmp-1785238624.9872594-1828915-110760222717154/source.ZIP, period.

        Command "/usr/bin/gtar" could not handle archive: Unable to list files in the archive: tar (child): zstd: Cannot exec: No such file or directory
        tar (child): Error is not recoverable: exiting now
        /usr/bin/gtar: Child returned status 2
        /usr/bin/gtar: Error is not recoverable: exiting now

        Command "/usr/bin/gtar" found no files in archive. Empty archive files are not supported.
        Command "/usr/bin/gtar" could not handle archive: Unable to list files in the archive: xz: (stdin): File format not recognized
        /usr/bin/gtar: Child returned status 1
        /usr/bin/gtar: Error is not recoverable: exiting now
```
But this is a minor edge case which I discussed with @arvind4501 ; users won't have old auth-bundles in practice so we can safely ignore this.

**LGTM** 🍏

docs/user/certificates.md
```diff
@@ -138,20 +138,20 @@ foremanctl deploy --certificate-ca-renew
 
 Both flags are **not persisted** in foremanctl’s answers file (one-shot).
 
-### Certificate Bundle for Secondary Systems
+### Auth Bundle for Secondary Systems
```
stejskalleos, line 141: In the [parameters.md](https://github.com/theforeman/foremanctl/blob/master/docs/user/parameters.md?plain=1#L142-L143), we still show `--foreman-proxy-oauth-consumer-key` & `--foreman-proxy-oauth-consumer-secret` in the Undetermined section. Should we update that as well?

src/roles/oauth_from_bundle/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+oauth_from_bundle_directory: "{{ certificates_oauth_directory }}"
```
stejskalleos, line 2: Why not just directly use the `certificates_oauth_directory`?

src/vars/base.yaml
```diff
@@ -3,10 +3,13 @@ certificates_hostnames:
   - "{{ ansible_facts['fqdn'] }}"
   - localhost
 
+oauth_directory: "{{ obsah_state_path }}/oauth"
+
 certificates_ca_password_file: "{{ obsah_state_path }}/certificates-ca-password"
 certificates_ca_password: "{{ lookup('ansible.builtin.password', certificates_ca_password_file, chars=['ascii_letters', 'digits']) }}"
 candlepin_oauth_secret_file: "{{ obsah_state_path }}/candlepin-oauth-secret"
 candlepin_oauth_secret: "{{ lookup('ansible.builtin.password', candlepin_oauth_secret_file, chars=['ascii_letters', 'digits'], length=32) }}"
+certificates_oauth_directory: /var/lib/foremanctl/oauth
```
stejskalleos, line 12: What's the difference here between `certificates_oauth_directory` & `oauth_directory`. One is relative and the other absolute?
Why do we need two variables?

docs/user/certificates.md
arvind4501: good catch, in which section should we add droped parameters which existed on foreman-installer but not in foremanctl? or just remove them from undetermined

src/roles/oauth_from_bundle/defaults/main.yml
arvind4501: the reason was to keep role's default vars to have control for the `oauth_from_bundle` role and not directly take any var.

docs/user/certificates.md
evgeni: In https://github.com/theforeman/foremanctl/pull/532 I used "unmapped" to explicitly call them out. how about that?

src/vars/base.yaml
arvind4501: `certificates_oauth_directory` is the place on target host where we store oauth creds after extraction . and oauth_directory is the place on controller where we store generated credentials from `foremanctl deploy`

docs/user/certificates.md
arvind4501: perfect, thanks

arvind4501:
> One minor issue found: the fail-early check for missing OAuth files doesn't trigger when using a bundle generated before this PR (old certificate-bundle), since the extraction step excludes files outside certs/ and private/, which isn't properly handled in Ansible stdout as shown below:

IMHO, since bundle generation changes are not released in any version, we won't need to handle upgrades. also i expect generated bundle to have `oauth/` in it and we are validating that after extraction

docs/user/parameters.md
```diff
@@ -132,15 +132,20 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--templates-listen-on` | Templates proxy to listen on https, http, or both | `--foreman-proxy-templates-listen-on` |
 | `--templates-url` | URL that hosts will use to contact the proxy for provisioning templates | `--foreman-proxy-templates-url` |
 
+### Unmapped
+
+| foreman-installer Parameter | Description | Reason |
+| --------------------------- | ----------- | ------ |
+| `--foreman-proxy-oauth-consumer-key` | OAuth consumer key for Smart Proxy | Not supported |
+| `--foreman-proxy-oauth-consumer-secret` | OAuth consumer secret for Smart Proxy | Not supported |
```
evgeni: Technically "Not required, automated", right?

ekohl:
> maybe, with that do we want to change the command from `foremanctl certificate-bundle <hostname>` to `foremanctl auth-bundle <hostname>`, or just the role name? i think it boils down to what command name we want to expose to users :)

Bike shedding a bit: `proxy-identity-bundle` perhaps? This makes it clear it's for a proxy, not for the Foreman server itself.

arvind4501:
> > maybe, with that do we want to change the command from `foremanctl certificate-bundle <hostname>` to `foremanctl auth-bundle <hostname>`, or just the role name? i think it boils down to what command name we want to expose to users :)
> 
> Bike shedding a bit: `proxy-identity-bundle` perhaps? This makes it clear it's for a proxy, not for the Foreman server itself.

when i see `proxy-identity-bundle` i wonder is there any certificate bundle we need for foreman server(i guess no?). as if no other entity(except proxy) needs auth-bundle we might not need to add `proxy-*`, also we generate bundle from server, i am happy to update. opinions?

evgeni:
I like the current `auth-bundle`.
(Also, technically, there is nothing proxy-specific in the bundle, you could use it to setup another Foreman [without Canldepin] too)

ekohl:
That's fair. Given long term I'd love to get rid of the bundle anyway I can live the `auth-bundle` name.

stejskalleos (approved):

PR #699, opened 2026-07-23
Add non-root container user validation

archanaserver:
The [SAT-47250](https://redhat.atlassian.net/browse/SAT-47250) identified container images configured to run as root, and foremanctl deploys images from multiple repositories, including images whose build pipelines are outside its control. So this adds deployment-level regression coverage to ensure that every running container is explicitly configured to use a non-root user.

* add a foremanctl test that discovers all running containers using `podman ps`
* inspect each container’s image and configured user using `podman inspect`

steps to reproduce:

* deploy a foremanctl test environment.
* run the focused test: `./forge test --pytest-args="tests/container_user_test.py -vv"`
* verify that the test:
    * passes when all running containers have a non-root `Config.User`.
    * fails and reports the container name, image, and configured user when any running container uses root.

#### Checklist

### DIFF
--- a/tests/container_user_test.py
+++ b/tests/container_user_test.py
@@ -1,0 +1,32 @@
+import shlex
+
+
+def test_all_containers_run_nonroot(server):
+    """Verify all running containers use non-root users."""
+    
+    result = server.run("podman ps --format '{{.Names}}'")
+    assert result.succeeded, result.stderr
+
+    container_names = result.stdout.split()
+    assert container_names, "No running containers found"
+
+    root_containers = []
+
+    for container_name in container_names:
+        result = server.run(
+            f"podman inspect {shlex.quote(container_name)} "
+            "--format '{{.ImageName}}|{{.Config.User}}'"
+        )
+        assert result.succeeded, f"Failed to inspect container {container_name}: {result.stderr}"
+
+        image_name, configured_user = result.stdout.strip().split("|", maxsplit=1)
+        user = configured_user.split(":", maxsplit=1)[0].lower()
+
+        if user in {"", "0", "root"}:
+            root_containers.append(
+                f"{container_name} (image={image_name}, Config.User={configured_user or '<empty>'})"
+            )
+
+    assert not root_containers, (
+        "Running containers configured to use root:\n" + "\n".join(root_containers)
+    )
--- a/tests/container_user_test.py
+++ b/tests/container_user_test.py
@@ -1,15 +1,21 @@
 import shlex
 
+EXPECTED_ROOT_IMAGES = {
+    "quay.io/foreman/pulp:foreman-nightly",
+    "quay.io/iop/puptoo:foreman-3.18",
+    "quay.io/iop/yuptoo:foreman-3.18",
+}
 
-def test_all_containers_run_nonroot(server):
-    """Verify all running containers use non-root users."""
-    
+
+def test_root_containers_match_expected(server):
+    """Verify that root containers match the temporary known-broken baseline."""
     result = server.run("podman ps --format '{{.Names}}'")
     assert result.succeeded, result.stderr
 
     container_names = result.stdout.split()
     assert container_names, "No running containers found"
 
+    expected_root_containers = []
     root_containers = []
 
     for container_name in container_names:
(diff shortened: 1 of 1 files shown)


tests/container_user_test.py
```diff
@@ -0,0 +1,36 @@
...
+    assert result.succeeded, result.stderr
+
+    container_names = result.stdout.split()
+    assert container_names, "No running containers found"
+
+    root_containers = []
+
+    for container_name in container_names:
+        if container_name in non_runnable:
+            continue
```
evgeni: Given you're using `podman ps` to get the list, I would expect that the non runnable containers don't show up there?

tests/container_user_test.py
```diff
@@ -0,0 +1,36 @@
...
+
+        result = server.run(
+            f"podman inspect {shlex.quote(container_name)} "
+            "--format '{{.ImageName}}|{{.Config.User}}'"
+        )
+        assert result.succeeded, f"Failed to inspect container {container_name}: {result.stderr}"
+
+        image_name, configured_user = result.stdout.strip().split("|", maxsplit=1)
+        user = configured_user.split(":", maxsplit=1)[0].lower()
```
evgeni, line 31: Under which circumstances does the `User` property have a colon?

At least in my deployment it does not:
```
quay.io/sclorg/postgresql-16-c10s:latest|26
quay.io/sclorg/valkey-8-c10s:latest|1001
quay.io/foreman/pulp:foreman-nightly|
quay.io/foreman/candlepin:foreman-nightly|tomcat
quay.io/foreman/pulp:foreman-nightly|
quay.io/foreman/foreman:nightly|foreman
quay.io/foreman/pulp:foreman-nightly|
quay.io/foreman/foreman:nightly|foreman
quay.io/foreman/pulp:foreman-nightly|
quay.io/foreman/pulp:foreman-nightly|
quay.io/foreman/foreman:nightly|foreman
quay.io/foreman/pulp:foreman-nightly|
quay.io/foreman/foreman:nightly|foreman
```

(Not saying this is wrong, just curious)

evgeni:
Tests are failing as expected, as pulp is running as `root` (well, unset in the definition, resulting in being root).

:+1:

tests/container_user_test.py
archanaserver: yes I see but I was thinking the `Config.User` can also contain a group, using formats such as user:group or uid:gid? so isn't splitting here ensures that values such as root:1001 or 0:1001 are still detected as root, while 1001:0 is correctly treated as a non-root user,

tests/container_user_test.py
archanaserver: yes i just check, dropping

tests/container_user_test.py
evgeni: Ah, just looked at https://docs.docker.com/reference/dockerfile/#user and yes it can be `<uid>:<gid>`, cool, makes sense to play safe here, even if unused today.

tests/container_user_test.py
```diff
@@ -0,0 +1,32 @@
+import shlex
+
+
+def test_all_containers_run_nonroot(server):
+    """Verify all running containers use non-root users."""
```
evgeni: Can we split the code in a way that we have a helper that finds the container names (via `podman ps`) and then runs the `test_container_runs_nonroot` parametrized? Then we could mark the ones that are currently failing (`quay.io/iop/yuptoo:foreman-3.18`, `quay.io/iop/puptoo:foreman-3.18`, `quay.io/foreman/pulp:foreman-nightly`) as `xfail` as we know they are broken right now?

(Emphasis on "can we", as I always find dynamic parameters in Pytest a bit strange)

archanaserver: yeah this make sense to me, also as it should also make handling the known failing images with xfail cleaner. so let me give it a try.

archanaserver: @evgeni so I took a look at the parametrized approach, but since I see the container list comes from `podman ps` on the provisioned `server` fixture here, it looks like it would require pytest's dynamic collection mechanisms. And since you mentioned dynamic parametrization, I'm wondering if that adds more complexity than it's worth here?
 Do you think it would be okay to keep the runtime loop and just split the container discovery and inspection into helper functions instead, or is there an existing pytest pattern you'd prefer?

evgeni: Yeah, if we can't easily write it in a way that allows us to skip individual containers/tests, we can keep it as is.

Maybe, instead of
```python
    assert not root_containers, (
        "Running containers configured to use root:\n" + "\n".join(root_containers)
    )
```
we can use
```python
    expected_root_containers = [pulp stuff, yuptoo, puptoo]
    assert expected_root_containers == root_containers …
```

So we have a list of things we expect to run as root (for now) and see it fail (and thus force us to update the list, at some point leading to `[]`) when one of the containers is updated to use non-root.

evgeni (approved):

ekohl:
In https://github.com/theforeman/foremanctl/pull/716 I took a stab at simplifying the test.

PR #592, opened 2026-06-23
Fixes #589 - Add IOP services to foreman.target

sjha4:
#### Why are you introducing these changes? (Problem description, related links)
IOP services should also be managed under foreman.target when deployed.
#### What are the changes introduced in this pull request?

* Adds IOP services to foreman.target

#### How to test this pull request
systemctl list-dependencies foreman.target
Try systemctl stop/start foreman.target

Steps to reproduce:

* Today systemctl stop foreman.target doesn't target IOP

#### Checklist
* [ ] Tests added/updated (if applicable)
* [ ] Documentation updated (if applicable)

PR #701, opened 2026-07-23
Test foreman.target stops and starts IOP services

vsedmik:
## Summary
- Adds a test under `tests/feature/iop/` verifying that `systemctl stop foreman.target` brings down all IOP containers and `systemctl start foreman.target` brings them back.
- Auto-skipped when IOP is not enabled (feature guarding via conftest).
- Marked as `slow` since it involves a full stop/start cycle (~2 min).

## Context
Companion test for #592 which adds `PartOf=foreman.target` and `WantedBy=foreman.target` to all IOP quadlet units.

### DIFF
--- a/tests/feature/iop/test_target_lifecycle.py
+++ b/tests/feature/iop/test_target_lifecycle.py
@@ -1,0 +1,41 @@
+import time
+
+import pytest
+
+TARGET_ACTIVE_RETRIES = 90
+TARGET_ACTIVE_DELAY = 10
+
+pytestmark = pytest.mark.slow
+
+
+def _wait_for_target_active(server, target="foreman.target"):
+    for _ in range(TARGET_ACTIVE_RETRIES):
+        if server.service(target).is_running:
+            return
+        time.sleep(TARGET_ACTIVE_DELAY)
+    raise AssertionError(f"{target} did not become active after lifecycle operation")
+
+
+def _wait_for_no_iop_containers(server, retries=30, delay=2):
+    for _ in range(retries):
+        result = server.run("podman ps --filter name=iop --format '{{.Names}}'")
+        if not result.stdout.strip():
+            return
+        time.sleep(delay)
+    raise AssertionError(f"IOP containers still running after stop: {result.stdout}")
+
+
+def test_foreman_target_stop_starts_iop_services(server):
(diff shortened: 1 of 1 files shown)


sjha4:
Looks good to me..Covers start and stop :)

ehelms:
Is there a cheaper way to test this? Can we verify that the iop services are correctly part of the `foreman.target` and then trust that systemd is going to do the right thing?

vsedmik:
> Is there a cheaper way to test this? Can we verify that the iop services are correctly part of the `foreman.target` and then trust that systemd is going to do the right thing?

Probably we could dynamically discover all `iop-core-*` and `iop-service-*` services via `systemctl list-units`, skip timer-triggered ones (cleanup, vmaas-sync), and assert that each remaining service has `PartOf=foreman.target`. This verifies the unit file wiring without touching any running services and takes ~21 seconds (instead of ~2 minutes).

tests/feature/iop/test_target_lifecycle.py
```diff
@@ -0,0 +1,20 @@
...
+
+    missing = []
+    for service in services:
+        # Skip timer-triggered services (cleanup, vmaas-sync)
+        triggered_by = server.run(f"systemctl show {service} -p TriggeredBy --value")
+        if triggered_by.stdout.strip():
+            continue
+        result = server.run(f"systemctl show {service} -p PartOf --value")
+        if "foreman.target" not in result.stdout:
+            missing.append(service)
```
evgeni, line 18: testinfra has `service.systemd_properties`, so I assume (untested) the following should work:
```
for service_name in services:
    if service.systemd_properties.get("TriggeredBy"):
        continue
    assert "foreman.target" in service.systemd_properties.get("PartOf")
```

evgeni (approved):

tests/feature/iop/test_target_lifecycle.py
evgeni: https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.service.Service.systemd_properties

aidenfine:
Can we close this in favor of? https://github.com/theforeman/foremanctl/pull/711

ekohl:
I've merged https://github.com/theforeman/foremanctl/pull/711.

PR #702, opened 2026-07-23
Increase ping endpoint timeout to avoid false failures during startup

sjha4:
The default URI module timeout of 30 seconds is sometimes not enough for getting results back from ping api. I have run into this mostly with rh_cloud and IOP services enabled.

* Increase ping API timeout to 120 secs for long running call.

Make sure the tasks `Wait for Foreman service to be accessible` and `Wait for Foreman tasks to be ready` that run during deploy pass. I have generally seen these timeout during deploy with IOP enabled which causes /ping to take longer than 30 secs to respond.

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -318,6 +318,7 @@
   ansible.builtin.uri:
     url: '{{ foreman_url }}/api/v2/ping'
     ca_path: '{{ foreman_ca_certificate }}'
+    timeout: 120
   until: foreman_status.status == 200
   retries: 60
   delay: 5
@@ -337,6 +338,7 @@
   ansible.builtin.uri:
     url: '{{ foreman_url }}/api/v2/ping'
     ca_path: '{{ foreman_ca_certificate }}'
+    timeout: 120
   until:
     - foreman_tasks_status.status == 200
     - foreman_tasks_status.json['results']['katello']['services']['foreman_tasks']['status'] == 'ok'


ehelms:
Do you have a sense as to why their existence is making this take longer? Is it simply system resources or something architectural in the ping API?

sjha4:
My best guess is the ping endpoint loads all plugins but not all are started or healthy if you run deploy with services stopped by the time the foreman role runs and checks ping causing longer waits on the stopped services. 
Mostly for roles that run after the foreman role during deploy: https://github.com/theforeman/foremanctl/blob/39d676d8a6ee2cd8959da0d7640058edac2c5be0/src/playbooks/deploy/deploy.yaml#L35 

With the ping step itself, it just needs a 200 ok return, doesn't care if services are not healthy inside the result.

sjha4:
Did some more debugging with claude, basically 2 steps to reproduce:
[quadlet] systemctl stop foreman.target
[controller] ./foremanctl deploy

Pulp services which haven't started yet in the deploy playbook are still checked for status thanks to Katello ping checking these.The pulp ping happens with http proxying it to the pulp API and that http request hangs till timeout. So it's not fail fast..Ping took 34 secs in this case.

ehelms:
> Pulp services which haven't started yet in the deploy playbook are still checked for status thanks to Katello ping checking these.The pulp ping happens with http proxying it to the pulp API and that http request hangs till timeout. So it's not fail fast..Ping took 34 secs in this case.

So we need this change, and we should create an issue to track potential improvements to our ping API?

ehelms (approved):

PR #703, opened 2026-07-24
Always use the certs from the quadlet machine in the tests

evgeni:

Our tests talk to the Smart Proxy API (see #519 and the BMC change in this PR), which requires auth. In the case of an external proxy deployment, the certs that were used were of the proxy, but that is not allowed to auth against itself.

* Fetch the certs from the quadlet machine and use those

Steps to reproduce:

* See the BMC feature test work on external proxy

#### Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -131,6 +131,17 @@ def default_certificates(certificate_source):
 
 
 @pytest.fixture(scope="module")
+def quadlet_client_certificate():
+    # this intentionally uses get_paramiko_host directly, as we want the cert of the quadlet box
+    # not the one "server" points at, as that can be the proxy
+    quadlet = get_paramiko_host('quadlet')
+    hostname = quadlet.run("hostname -f").stdout.strip()
+    cert = quadlet.file(f"/var/lib/foremanctl/certs/certs/{hostname}-client.crt").content_string
+    key = quadlet.file(f"/var/lib/foremanctl/certs/private/{hostname}-client.key").content_string
+    return (cert, key)
+
+
+@pytest.fixture(scope="module")
 def database_mode(obsah_params):
     return obsah_params.get('database_mode', 'internal')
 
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -340,13 +355,17 @@ def proxy_base_url(server_fqdn):
 
 
 @pytest.fixture(scope="module")
-def curl_request(server, certificates, server_fqdn):
+def curl_request(server, certificates, quadlet_client_certificate, server_fqdn):
+    cert, key = quadlet_client_certificate
```
evgeni, line 359: I didn't check if there are cases where we *actually* want the proxy cert here, but we should do that before merging

evgeni:
Upgrade failure unrelated, fix in #705

tests/conftest.py
ShimShtein: If you want to test server APIs that are reserved for proxy auth, for example fact upload.

tests/conftest.py
```diff
@@ -340,13 +355,17 @@ def proxy_base_url(server_fqdn):
 
 
 @pytest.fixture(scope="module")
-def curl_request(server, certificates, server_fqdn):
+def curl_request(server, certificates, quadlet_client_certificate, server_fqdn):
```
ShimShtein, line 358:
```suggestion
def curl_request(server, certificates, client_certificate=quadlet_client_certificate, server_fqdn):
```
This will enable a sane default with ability to properly override

ShimShtein (changes requested):

tests/conftest.py
evgeni: Those are Pytest fixtures, not normal Python args, I don't think you can override them like this.

tests/conftest.py
evgeni: Sure, but we don't have such a test here today.

tests/conftest.py
evgeni: Yeah, tests say that doesn't work.

ShimShtein: Sorry, for pytest you can do it like this:
``` py
@pytest.fixture
def client_certificate
  return quadlet_client_certificate
```
This will be the default, and later, if you want to change the value, you will do something like:

``` py
@pytest.mark.parameterize("client_certificate", ["/path/to/host/cert"])
def test_facts(curl_request)
# ...
```

evgeni: Tests still say no:
```
TypeError: cannot unpack non-iterable FixtureFunctionDefinition object
```

Can we use my working code and worry about possible overrides later when we actually need them?

ShimShtein: ACK, let's get a working version out and worry about the overrides later

evgeni: Dropped the last commit again.

evgeni:
Given @ShimShtein  is out today, and the override was his only concern, I am going to merge this, to unblock #706 

There was also a concern raised by @jeremylenz about the hard-coded "quadlet" name here, which I will follow up in a separate PR.

PR #705, opened 2026-07-27
explicitly chown Foreman data volume on start

evgeni:

https://github.com/theforeman/foreman-oci-images/pull/70 made the UID/GID of the foreman user static, but on upgrades the previous version might have had a different ID, resulting in permission errors.

* Explicitly chown the data volume on container start

Steps to reproduce:

* CI (upgrade test)

#### Checklist

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -116,7 +116,7 @@
     network: host
     hostname: "{{ ansible_facts['hostname'] }}.local"
     volume:
-      - 'foreman-data-run:/var/run/foreman:rw,z'
+      - 'foreman-data-run:/var/run/foreman:rw,z,U'
     secrets:
       - 'foreman-database-url,type=env,target=DATABASE_URL'
       - 'foreman-encryption-key,type=env,target=ENCRYPTION_KEY'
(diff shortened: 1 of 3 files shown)


arvind4501:
thanks @evgeni for catching this :). a curious thought for upgrades should we add these permissions changes as pre upgrade tasks?

evgeni:
Well, I hoped podman will just do it for us, and we don't need any further tasks. Let's see what CI will tell

src/roles/foreman/tasks/main.yaml
```diff
@@ -168,7 +168,7 @@
     network: host
     hostname: "{{ ansible_facts['hostname'] }}.local"
     volume:
-      - 'foreman-data-run:/var/run/foreman:rw,z'
+      - 'foreman-data-run:/var/run/foreman:rw,z,U'
```
arvind4501, line 171: with `U` option, it chown's recursively(which i think does not hurt), but it does same thing each time container volume mounts which has impact if foreman-data-run volume is big( it needs to traverse each file and chown it).

evgeni: Correct. In the case of foreman-data, I don't expect it to be big, so it should be fine. Probably not something we want to do with pulp data though :)

arvind4501: also we only need this for upragdes from 3.19 only

arvind4501 (approved):

Odilhao:
I think we will need to do something similar to pulp as well, should we prepare for that as well?

evgeni:
> I think we will need to do something similar to pulp as well, should we prepare for that as well?

I don't think recursively chowning all of pulp data on every start is a smart idea.

Odilhao (approved):
LGTM

PR #706, opened 2026-07-28
Add tests and update parameters for templates feature

shubhamsg199:
This is a followup PR for https://github.com/theforeman/foremanctl/pull/519

* Incorporating remaining changes from https://github.com/theforeman/foremanctl/pull/519
* Adding test for testing the functionality of the feature

Dependency: Added changes from https://github.com/theforeman/foremanctl/pull/703/changes for the tests to work, can be easily rebased once it is merged

### DIFF
--- a/docs/user/parameters.md
+++ b/docs/user/parameters.md
@@ -135,7 +135,6 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
 | `--add-feature templates` | Enable Templates feature on Smart Proxy | `--foreman-proxy-templates` |
-| `--templates-listen-on` | Templates proxy to listen on https, http, or both | `--foreman-proxy-templates-listen-on` |
 | `--templates-url` | URL that hosts will use to contact the proxy for provisioning templates | `--foreman-proxy-templates-url` |
 
 ### Unmapped
@@ -144,6 +143,7 @@ There are multiple use cases from the users perspective that dictate what parame
 | --------------------------- | ----------- | ------ |
 | `--foreman-proxy-oauth-consumer-key` | OAuth consumer key for Smart Proxy | Not required(Managed automatically via auth bundle) |
 | `--foreman-proxy-oauth-consumer-secret` | OAuth consumer secret for Smart Proxy | Not required(Managed automatically via auth bundle) |
+| `--foreman-proxy-templates-listen-on` | Templates proxy to listen on https, http, or both | No longer supported |
 
 ### Undetermined
 
--- a/src/playbooks/_foreman_proxy/metadata.obsah.yaml
+++ b/src/playbooks/_foreman_proxy/metadata.obsah.yaml
@@ -10,13 +10,6 @@ variables:
     parameter: --bmc-redfish-verify-ssl
     help: Verify SSL certificates for Redfish BMC connections.
     type: Boolean
-  foreman_proxy_templates_listen_on:
-    parameter: --templates-listen-on
-    help: Templates proxy to listen on https, http, or both.
-    choices:
-      - http
-      - https
-      - both
   foreman_proxy_templates_url:
     parameter: --templates-url
-    help: URL that hosts will use to contact the proxy for provisioning templates (e.g. http://<proxy-fqdn>:8000). Required when templates feature is enabled.
+    help: URL that hosts will use to contact the proxy for provisioning templates.
--- a/src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
+++ b/src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
@@ -1,4 +1,3 @@
 ---
-:enabled: {% if foreman_proxy_templates_listen_on | default('both') == 'both' %}true{% else %}{{ foreman_proxy_templates_listen_on }}{% endif %}
-
+:enabled: true
 :template_url: {{ foreman_proxy_templates_url }}
--- a/src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
+++ b/src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
@@ -1,3 +1,3 @@
 ---
-:enabled: true
+:enabled: {{ feature_enabled }}
 :template_url: {{ foreman_proxy_templates_url }}
--- a/tests/feature/foreman-proxy/base_test.py
+++ b/tests/feature/foreman-proxy/base_test.py
@@ -72,6 +72,17 @@ def test_bmc_default_provider(proxy_v2_features):
 
 
 @pytest.mark.feature('templates')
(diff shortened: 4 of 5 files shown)


docs/user/parameters.md
```diff
@@ -129,8 +129,7 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--bmc-ipmi-implementation` | IPMI implementation to use for BMC | `--foreman-proxy-bmc-default-provider` |
 | `--bmc-redfish-verify-ssl` | Verify SSL certificates for Redfish BMC connections | `--foreman-proxy-bmc-redfish-verify-ssl` |
 | `--add-feature templates` | Enable Templates feature on Smart Proxy | `--foreman-proxy-templates` |
-| `--templates-listen-on` | Templates proxy to listen on https, http, or both | `--foreman-proxy-templates-listen-on` |
```
ekohl, line 138: This should be in its own Unmapped category as something we no longer support.

src/playbooks/_foreman_proxy/metadata.obsah.yaml
```diff
@@ -10,13 +10,6 @@ variables:
...
-    parameter: --templates-listen-on
-    help: Templates proxy to listen on https, http, or both.
-    choices:
-      - http
-      - https
-      - both
   foreman_proxy_templates_url:
     parameter: --templates-url
-    help: URL that hosts will use to contact the proxy for provisioning templates (e.g. http://<proxy-fqdn>:8000). Required when templates feature is enabled.
+    help: URL that hosts will use to contact the proxy for provisioning templates. Defaults to http://<proxy-fqdn>:8000.
```
ekohl: In other places we don't describe the defaults. I'd like to avoid doing that because it's inconsistent with the rest.

ekohl (changes requested):
Thanks for picking up the parameter changes. IMHO this PR is doing 2 things: change parameters and add tests. The PR title should reflect that. On git commits I'd recommend separating it into 2 separate commits.

src/playbooks/_foreman_proxy/metadata.obsah.yaml
shubhamsg199: Removed

docs/user/parameters.md
shubhamsg199: Updated

tests/feature/foreman-proxy/base_test.py
```diff
@@ -72,6 +72,17 @@ def test_bmc_default_provider(proxy_v2_features):
 
 
 @pytest.mark.feature('templates')
-def test_templates_template_url(obsah_params, server_fqdn):
-    template_url = obsah_params.get('foreman_proxy_templates_url')
-    assert template_url == f'http://{server_fqdn}:8000'
+def test_templates_fetch_template_url(proxy_v2_features, proxy_base_url):
```
stejskalleos: I don't see `proxy_base_url` used in the `def`

tests/feature/foreman-proxy/base_test.py
```diff
@@ -72,6 +72,17 @@ def test_bmc_default_provider(proxy_v2_features):
...
+    assert 'templates' in proxy_v2_features
+    settings = proxy_v2_features['templates'].get('settings', {})
+    assert settings.get('template_url') == 'http://proxy.example.com:8000'
```
stejskalleos:
```suggestion
    assert settings.get('template_url') == obsah_params.get('foreman_proxy_templates_url')
```

stejskalleos (changes requested):

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
```diff
@@ -1,4 +1,3 @@
 ---
-:enabled: {% if foreman_proxy_templates_listen_on | default('both') == 'both' %}true{% else %}{{ foreman_proxy_templates_listen_on }}{% endif %}
-
+:enabled: true
```
stejskalleos: What if the user wants to remove the feature? Will this config file be removed and therefore the feature removed from the proxy, or will it stay?

tests/conftest.py
```diff
@@ -130,25 +130,40 @@ def default_certificates(certificate_source):
         pytest.skip("Only applies to non-custom certificate deployments")
 
 
+@pytest.fixture(scope="module")
+def quadlet_client_certificate():
+    # this intentionally uses get_paramiko_host directly, as we want the cert of the quadlet box
+    # not the one "server" points at, as that can be the proxy
```
stejskalleos, line 136: > as we want the cert of the quadlet box
    # not the one "server" points at, as that can be the proxy

Is this something relevant to our test suite?

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
evgeni: We do deploy a static `enabled: false` config in that case:
https://github.com/theforeman/foremanctl/blob/e7f157473483778ee0f61353d75cc08722b4931b/src/roles/foreman_proxy/tasks/feature.yaml#L6

But I guess it's still nicer to use `{{ feature_enabled }}` here

tests/conftest.py
evgeni: that's a question to discuss in #703, IMHO

src/roles/foreman_proxy/templates/settings.d/templates.yml.j2
shubhamsg199: Agreed. Updated to `{{ feature_enabled }}`

tests/feature/foreman-proxy/base_test.py
shubhamsg199: Removed `proxy_base_url`

tests/feature/foreman-proxy/base_test.py
shubhamsg199: Updated

tests/conftest.py
stejskalleos: Ok, let's continue there.

tests/conftest.py
```diff
@@ -130,25 +130,40 @@ def default_certificates(certificate_source):
...
+    quadlet = get_paramiko_host('quadlet')
```
jeremylenz, line 137: Hard-coding `quadlet` here makes me uneasy. This assumes that there is a box named `quadlet` that is present, something which is not guaranteed on all foremanctl setups. (for instance, if you use `forge deploy-dev --target-host localhost`.) Maybe we could add a new `--quadlet-hostname` CLI option (defaulting to "quadlet")?

jeremylenz (changes requested):

tests/conftest.py
evgeni: IMHO (but this is more for #703) the current `--server-hostname` should be `--target-hostname` and `--server-hostname` should be what you describe as `--quadlet-hostname` here

But that's probably a bigger task that should be tackled outside of #703

evgeni: Also, can you even run those tests against a dev setup?

evgeni (approved):

tests/conftest.py
evgeni: #715 is fixing this :)

ekohl (approved):

PR #710, opened 2026-07-29
Fix libvirt dependency

michalgritzbach:

Foreman and Foreman's dev setup document assumes that libvirt is installed in the system. Without libvirt, Foreman's test suite fails to run (specifically https://github.com/theforeman/foreman/blob/develop/test/integration/host_js_test.rb#L4 - the require immediately stops the suite).

* adds libvirt-devel to foreman_development_packages
* changes bundle install command in foreman_development role to include libvirt gem group

Steps to reproduce:

* libvirt-devel is present, fog-libvirt (+ ruby-libvirt) are installed on a fresh system

#### Checklist

### DIFF
--- a/development/roles/foreman_development/tasks/main.yaml
+++ b/development/roles/foreman_development/tasks/main.yaml
@@ -23,11 +23,19 @@
     name: nodejs
     state: present
 
-- name: Enable CodeReady Builder repository
+- name: Ensure the crb repository is enabled
   community.general.rhsm_repository:
-    name: "codeready-builder-for-rhel-{{ ansible_distribution_major_version }}-{{ ansible_architecture }}-rpms"
+    name: crb
     state: enabled
-  when: ansible_distribution == 'RedHat'
+  when:
+    - ansible_facts['distribution'] != "RedHat"
+
+- name: Ensure the crb repository is enabled
+  community.general.rhsm_repository:
+    name: "codeready-builder-for-rhel-{{ ansible_facts['distribution_major_version'] }}-{{ ansible_facts['architecture'] }}-rpms"
+    state: enabled
+  when:
+    - ansible_facts['distribution'] == "RedHat"
 
 - name: Install development packages
   ansible.builtin.package:
--- a/development/roles/foreman_development/tasks/main.yaml
+++ b/development/roles/foreman_development/tasks/main.yaml
@@ -180,7 +180,7 @@
 
 - name: Install Ruby dependencies
   ansible.builtin.command:
-    cmd: bundle install --path .vendor --jobs 3 --without "libvirt:journald"
+    cmd: bundle install --path .vendor --jobs 3 --without "journald"
     chdir: "{{ foreman_development_foreman_dir }}"
   become: true
   become_user: "{{ foreman_development_user }}"
--- a/development/roles/foreman_development/tasks/main.yaml
+++ b/development/roles/foreman_development/tasks/main.yaml
@@ -24,13 +24,13 @@
     state: present
 
 - name: Ensure the crb repository is enabled
-  community.general.rhsm_repository:
+  community.general.dnf_config_manager:
     name: crb
     state: enabled
   when:
     - ansible_facts['distribution'] != "RedHat"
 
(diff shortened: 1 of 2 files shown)


evgeni (approved):
One could argue those tests could skip if there is no libvirt, instead of failing, but I am not in an arguing mood ;)

evgeni:
```
No package libvirt-devel available.
```

I think on CentOS you need to enable `crb` repo for this.

ares:
Wouldn't it be nice if I could get dev environment ready to hack on any part of the Foreman? Perhaps forge can be extended with features-like concept if needed. While we can also start skipping tests, I'd still like to have a standard way to deploy dev environment for libvirt integration and more feature in future.

development/roles/foreman_development/tasks/main.yaml
```diff
@@ -23,11 +23,21 @@
...
+- name: Ensure the crb repository is enabled
+  community.general.dnf_config_manager:
+    name: crb
     state: enabled
-  when: ansible_distribution == 'RedHat'
+  when:
+    - ansible_facts['distribution'] != "RedHat"
+
+- name: Ensure the crb repository is enabled
+  ansible.builtin.command:
```
evgeni: why the change from `community.general.rhsm_repository` to `command`?

evgeni:
> Wouldn't it be nice if I could get dev environment ready to hack on any part of the Foreman? Perhaps forge can be extended with features-like concept if needed. While we can also start skipping tests, I'd still like to have a standard way to deploy dev environment for libvirt integration and more feature in future.

What exactly is your ask to this PR here?

development/roles/foreman_development/tasks/main.yaml
michalgritzbach: I originally used a snippet from another task yaml; rolled back to rhsm_respository today.

development/roles/foreman_development/tasks/main.yaml
```diff
@@ -23,11 +23,19 @@
     name: nodejs
     state: present
 
-- name: Enable CodeReady Builder repository
+- name: Ensure the crb repository is enabled
   community.general.rhsm_repository:
```
evgeni: rhsm_repository only works for systems that are subscribed (either to RH directly, or Katello/Satellite) -- that's what the code before your change was handling.

for CentOS and friends, you need to use `dnf_config_manager` like you did before the last force push.

development/roles/foreman_development/tasks/main.yaml
```diff
@@ -23,11 +23,19 @@
...
-    name: "codeready-builder-for-rhel-{{ ansible_distribution_major_version }}-{{ ansible_architecture }}-rpms"
+    name: crb
     state: enabled
-  when: ansible_distribution == 'RedHat'
+  when:
+    - ansible_facts['distribution'] != "RedHat"
+
+- name: Ensure the crb repository is enabled
```
evgeni:
```suggestion
- name: Ensure the CodeReady Builder repository is enabled
```

the RHEL name is "CodeReady Builder" :)

ares:
> What exactly is your ask to this PR here?

I want these dependencies to be installed in dev setup even if they are optional. I hinted on possible solution if we don't want this to be the default dev experience. And I'd like the same approach to be taken for future things that make our dev environment more complete.

From your comment I was not sure if you want the tests to be skipped (and just resigned) or not.

evgeni:
> From your comment I was not sure if you want the tests to be skipped (and just resigned) or not.

I want our tests to be smarter (while providing a complete dev experience).

There are only 4 tests in https://github.com/theforeman/foreman/blob/develop/test/integration/host_js_test.rb that use Libvirt, yet the way the file is written today *all* tests in there fail if fog-libvirt is not available.

I am totally with you that if we say "run this to get a full dev box", this should include *all* deps.

At the same time, if someone has a limited env (and maybe just doesn't need libvirt at all), the other host tests should remain green.

ares:
So perhaps both should happen - make the tests smart (more robust) and make this to deploy full dev env (done already). @michalgritzbach could you please make the tests skipped if the dependency is missing?

PR #711, opened 2026-07-29
Test foreman.target stops and starts IOP services

aidenfine:
Note: duplicate of #701 due to PTO.

## Summary
- Adds a test under `tests/feature/iop/` verifying that `systemctl stop foreman.target` brings down all IOP containers and `systemctl start foreman.target` brings them back.
- Auto-skipped when IOP is not enabled (feature guarding via conftest).
- Marked as `slow` since it involves a full stop/start cycle (~2 min).

## Context
Companion test for #592 which adds `PartOf=foreman.target` and `WantedBy=foreman.target` to all IOP quadlet units.

### DIFF
--- a/development/requirements.txt
+++ b/development/requirements.txt
@@ -4,6 +4,7 @@ jinja2
 paramiko
 passlib
 ruff
+pytest-subtests
 pytest-testinfra
 pytest-durations
 python-dateutil
--- a/development/requirements.txt
+++ b/development/requirements.txt
@@ -4,7 +4,6 @@ jinja2
 paramiko
 passlib
 ruff
-pytest-subtests
 pytest-testinfra
 pytest-durations
 python-dateutil
--- a/tests/feature/iop/test_target_lifecycle.py
+++ b/tests/feature/iop/test_target_lifecycle.py
@@ -1,0 +1,41 @@
+import time
+
+import pytest
+
+TARGET_ACTIVE_RETRIES = 90
+TARGET_ACTIVE_DELAY = 10
+
+pytestmark = pytest.mark.slow
+
+
+def _wait_for_target_active(server, target="foreman.target"):
+    for _ in range(TARGET_ACTIVE_RETRIES):
+        if server.service(target).is_running:
+            return
+        time.sleep(TARGET_ACTIVE_DELAY)
+    raise AssertionError(f"{target} did not become active after lifecycle operation")
+
+
+def _wait_for_no_iop_containers(server, retries=30, delay=2):
+    for _ in range(retries):
+        result = server.run("podman ps --filter name=iop --format '{{.Names}}'")
+        if not result.stdout.strip():
+            return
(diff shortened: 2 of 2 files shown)


tests/feature/iop/test_target_lifecycle.py
```diff
@@ -0,0 +1,17 @@
+def test_iop_services_part_of_foreman_target(server):
+    result = server.run(
+        "systemctl list-units 'iop-core-*' 'iop-service-*'"
+        " --type=service --no-legend --all --plain"
+    )
+    assert result.rc == 0
+    services = [line.split()[0] for line in result.stdout.strip().splitlines()]
+    assert services, "No IOP services found"
+
+    missing = []
```
ekohl: Nothing ever adds anything. Left over of a previous implementation?

tests/feature/iop/test_target_lifecycle.py
```diff
@@ -0,0 +1,17 @@
...
+    for service_name in services:
+        service = server.service(service_name)
+        if service.systemd_properties.get("TriggeredBy"):
+            continue
+        assert "foreman.target" in service.systemd_properties.get("PartOf")
```
ekohl: You can use [subtests](https://docs.pytest.org/en/stable/how-to/subtests.html) here to make sure every test is shown as its own. That way you also test all of them, even if a single test fails.

development/requirements.txt
```diff
@@ -4,6 +4,7 @@ jinja2
 paramiko
 passlib
 ruff
+pytest-subtests
```
ekohl: I didn't realize this was a separate dependency. I'm not sure it's worth adding it then.

aidenfine: I agree with this, for one test I don't think its worth it.

evgeni: https://docs.pytest.org/en/stable/how-to/subtests.html says it *was* a separate plugin, but is now (since 9.0) in core pytest.

I actually like the functionality and it would make the tests we added in #699 a tad nicer to look at.

ekohl: I couldn't help myself: https://github.com/theforeman/foremanctl/pull/716

aidenfine: Thats nice then, I added back the subtest part, but left out the not needed `pytest-subtests` dep.

tests/feature/iop/test_target_lifecycle.py
```diff
@@ -0,0 +1,15 @@
+def test_iop_services_part_of_foreman_target(server, subtests):
+    result = server.run(
+        "systemctl list-units 'iop-core-*' 'iop-service-*'"
+        " --type=service --no-legend --all --plain"
+    )
+    assert result.rc == 0
```
ekohl: Isn't this easier to read?
```suggestion
    assert result.succeeded
```

ekohl (approved):

PR #713, opened 2026-07-30
Add a test for webhooks

adamruzicka:

The original webhooks PR only had "is the feature enabled?" test. 

* Adding a test to test the actual webhook functionality.

Steps to reproduce:

* Run CI
* Wait for it to turn green

#### Checklist


tests/feature/foreman/webhooks_test.py
```diff
@@ -0,0 +1,76 @@
...
+    # Start a one-shot netcat listener that:
+    # 1. Accepts a single connection
+    # 2. Sends back a minimal HTTP 200 response (so Foreman considers the webhook delivered)
+    # 3. Saves the received request to output_file
+    # We run it in the background via nohup + shell
+    server.run(
+        f"nohup bash -c '"
+        f'printf "HTTP/1.1 200 OK\\r\\nContent-Length: 0\\r\\n\\r\\n" '
+        f"| nc -l {LISTENER_PORT} > {output_file} 2>&1 &' "
+        f">/dev/null 2>&1"
```
evgeni: this is the wildest httpd I've seen, and I've seen a lot.

adamruzicka: What can I say, except... you're welcome? 

Jokes aside, I only need to check that the webhook is received, netcat doesn't actually have to respond to it, which simplifies things a bit, but it is still a bit hacky.

tests/feature/foreman/webhooks_test.py
```diff
@@ -0,0 +1,71 @@
...
+        'http_content_type': 'application/json',
+        'webhook_template_id': webhook_template['id'],
+        'enabled': True,
+        'ssl_verification': False,
+    })
+    yield hook
+    foremanapi.delete('webhooks', hook)
+
+
+@pytest.mark.feature('webhooks')
```
evgeni: Should this file just live as `tests/feature/webhooks/base_test.py`? Then you don't need to manually mark :)

evgeni (approved):

tests/feature/foreman/webhooks_test.py
evgeni: (you then also have to touch `tests/feature/webhooks/__init__.py` to make it a "real" package)

tests/feature/webhooks/base_test.py
```diff
@@ -0,0 +1,78 @@
...
+            "enabled": True,
+            "ssl_verification": False,
+        },
+    )
+    yield hook
+    foremanapi.delete("webhooks", hook)
+
+
+def test_foreman_webhooks(foremanapi):
+    plugins = [plugin["name"] for plugin in foremanapi.list("plugins")]
```
evgeni: You'll hate me, but if we move
```
@pytest.fixture
def foreman_plugins(foremanapi):
    return [plugin['name'] for plugin in foremanapi.list('plugins')]
```
out of plugin_test.py to conftest.py, you'll be able to re-use that fixture here, without re-listing everything :)

adamruzicka: > You'll hate me, but

Nah, I take it as learning pytest, one tiny refactoring at a time

evgeni (approved):

PR #715, opened 2026-07-31
Introduce --test-target to more explicitly declare where tests run

evgeni:

Using `--server-hostname` always felt wrong, as we're not actually changing the hostname of the (Foreman) server, just the system we're running the tests against.

* Introduce `--test-target` which now declares what the `server` fixture points at
* The old `--server-hostname` remains if we ever need to change the *server* name
* Tests still use the `server` fixture, refactoring this is left for later

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -53,7 +53,8 @@ class UserParameters:
 
 
 def pytest_addoption(parser):
-    parser.addoption("--server-hostname", action="store", default="quadlet", help="Hostname of the server VM to test against")
+    parser.addoption("--test-target", action="store", default="quadlet", help="Hostname of the system to test against")
+    parser.addoption("--server-hostname", action="store", default="quadlet", help="Hostname of the Foreman server")
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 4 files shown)


evgeni:
:eyes: @jeremylenz

tests/conftest.py
```diff
@@ -151,8 +157,8 @@ def get_paramiko_host(hostname):
 
 
 @pytest.fixture(scope="module")
-def server(server_hostname):
-    yield get_paramiko_host(server_hostname)
+def server(test_target):
+    yield get_paramiko_host(test_target)
```
jeremylenz, line 161: With this in place, it'd be a small change to also support `--test-target=localhost` for testing deploy-dev setups — just branch on the hostname in `get_paramiko_host`:

```python
def get_paramiko_host(hostname):
    if hostname == 'localhost':
        return testinfra.get_host('local://', sudo=True)
    return testinfra.get_host(f'paramiko://{hostname}', sudo=True, ssh_config=SSH_CONFIG)
```

Since `deploy-dev` already supports `--target-host=localhost`, this would make the test side consistent with it.

evgeni: Sure, if that works that sounds like a good addition

ehelms:
This looks like a good change, feel free to switch it to ready

PR #716, opened 2026-07-31
Refactor container user test to use better helpers

ekohl:

In https://github.com/theforeman/foremanctl/pull/711#discussion_r3688690952 it was pointed out this could be improved, which also made spot the possible testinfra improvement.

This uses [pytest's subtests](https://docs.pytest.org/en/stable/how-to/subtests.html) to simplify the testing. It still reports each individual container that fails, but written in a cleaner way. Then uses [testinfra's podman infrastructure](https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.podman.Podman.get_containers) to simplify it even further.

Steps to reproduce:

* Run the test suite
* Verify it still runs all the tests

#### Checklist

### DIFF
--- a/tests/container_user_test.py
+++ b/tests/container_user_test.py
@@ -1,5 +1,4 @@
-import shlex
-
+# These still need to be fixed
 EXPECTED_ROOT_IMAGES = {
     "quay.io/foreman/pulp:foreman-nightly",
     "quay.io/iop/puptoo:foreman-3.18",
@@ -7,46 +6,15 @@ EXPECTED_ROOT_IMAGES = {
 }
 
 
-def test_root_containers_match_expected(server):
+def test_root_containers_match_expected(server, subtests):
     """Verify that root containers match the temporary known-broken baseline."""
-    result = server.run("podman ps --format '{{.Names}}'")
-    assert result.succeeded, result.stderr
+    containers = server.podman.get_containers(status="running")
+    assert containers, "No running containers found"
 
-    container_names = result.stdout.split()
-    assert container_names, "No running containers found"
-
-    expected_root_containers = []
-    root_containers = []
-
-    for container_name in container_names:
-        result = server.run(
-            f"podman inspect {shlex.quote(container_name)} "
-            "--format '{{.ImageName}}|{{.Config.User}}'"
-        )
-        assert result.succeeded, (
-            f"Failed to inspect container {container_name}: {result.stderr}"
-        )
-
-        image_name, configured_user = result.stdout.strip().split("|", maxsplit=1)
-        user = configured_user.split(":", maxsplit=1)[0].lower()
-
-        container_details = (
-            f"{container_name} "
-            f"(image={image_name}, "
-            f"Config.User={configured_user or '<empty>'})"
-        )
-
-        if image_name in EXPECTED_ROOT_IMAGES:
-            expected_root_containers.append(container_details)
-
-        if user in {"", "0", "root"}:
-            root_containers.append(container_details)
-
-    expected_output = "\n".join(sorted(expected_root_containers)) or "<none>"
-    actual_output = "\n".join(sorted(root_containers)) or "<none>"
-
-    assert sorted(expected_root_containers) == sorted(root_containers), (
-        "Root containers differ from the expected temporary baseline:\n"
-        f"Expected:\n{expected_output}\n"
-        f"Actual:\n{actual_output}"
-    )
+    for container in containers:
+        with subtests.test(container.name):
+            config = container.inspect()['Config']
+            if config['Image'] in EXPECTED_ROOT_IMAGES:
+                assert config['User'] in {"", "0", "root"}
+            else:
+                assert config['User'] not in {"", "0", "root"}
--- a/development/requirements.txt
+++ b/development/requirements.txt
@@ -4,6 +4,7 @@ jinja2
 paramiko
 passlib
 ruff
+pytest>=9.0
 pytest-testinfra
 pytest-durations
 python-dateutil


tests/container_user_test.py
```diff
@@ -1,52 +1,16 @@
...
-
-        if image_name in EXPECTED_ROOT_IMAGES:
-            expected_root_containers.append(container_details)
-
-        if user in {"", "0", "root"}:
-            root_containers.append(container_details)
-
-    expected_output = "\n".join(sorted(expected_root_containers)) or "<none>"
-    actual_output = "\n".join(sorted(root_containers)) or "<none>"
+    containers = server.podman.get_containers(state="running")
```
evgeni: I guess it should be `status="running"`?

tests/container_user_test.py
```diff
@@ -1,52 +1,16 @@
...
 
-    assert sorted(expected_root_containers) == sorted(root_containers), (
-        "Root containers differ from the expected temporary baseline:\n"
-        f"Expected:\n{expected_output}\n"
-        f"Actual:\n{actual_output}"
-    )
+    for container in containers:
+        with subtests.test(container.name):
+            user = container.inspect()['Config.User']
+            assert user not in {"", "0", "root"} or container.name in EXPECTED_ROOT_IMAGES
```
evgeni: can subtests xfail? so we notice when a container image listed in `EXPECTED_ROOT_IMAGES` starts using a user?

tests/container_user_test.py
ekohl: That's what you get for quickly submitting something.

tests/container_user_test.py
ekohl: Yes you can:
```python
import pytest


def test_xfail(subtests):
    for test in ('success', 'failure'):
        with subtests.test(test):
            if test == 'failure':
                pytest.xfail('all good')
            else:
                assert True
```

Then running it:
```console
$ pytest tests/bla.py  -v
============================= test session starts ==============================
platform linux -- Python 3.14.6, pytest-9.0.3, pluggy-1.6.0 -- /home/ekohl/.virtualenvs/foremanctl/bin/python
cachedir: .pytest_cache
rootdir: /home/ekohl/dev/foremanctl
configfile: pytest.toml
plugins: testinfra-10.2.2, durations-1.6.2
collected 1 item                                                               

tests/bla.py::test_xfail SUBPASSED[success]                              [100%]
tests/bla.py::test_xfail SUBXFAIL[failure] (all good)                    [100%]
tests/bla.py::test_xfail PASSED                                          [100%]
```

evgeni (approved):

tests/container_user_test.py
evgeni: Perfect, thanks

tests/container_user_test.py
```diff
@@ -7,46 +7,15 @@
...
-        "Root containers differ from the expected temporary baseline:\n"
-        f"Expected:\n{expected_output}\n"
-        f"Actual:\n{actual_output}"
-    )
+    containers = server.podman.get_containers(status="running")
+    assert containers, "No running containers found"
+
+    for container in containers:
+        with subtests.test(container.name):
+            user = container.inspect()['Config.User']
```
ekohl: Re-reading this: should we still check this even for known bad images (and find out if they've been fixed) or put this line in the `else` branch?

evgeni: oh, huh, I miss-read the original code of yours.

yeah, I think it should be something like:
```python
user = container.inspect()['Config.User']
if container.name in EXPECTED_ROOT_IMAGES:
    assert user in in {"", "0", "root"}
else:
    assert user not in {"", "0", "root"}
```

and then we don't even need xfail 🙈 

or you go fancy and do:
```python
user = container.inspect()['Config.User']
if container.name in EXPECTED_ROOT_IMAGES:
    pytest.xfail(f"{container.name} needs update to use non-root")
assert user not in {"", "0", "root"}
```
Edit: but that sadly doesn't catch the change to non-root again :/

evgeni: https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail says it should be marked `XPASS` (unexpectedly passing), but it does not.

```
% cat lol.py 
import pytest


def test_lol():
    pytest.xfail("lol")
    assert 1 == 2


def test_lol2():
    pytest.xfail("lol")
    assert 2 == 2


def test_sub(subtests):
    for x in [1,2,3]:
        with subtests.test(x):
            if x == 2:
                pytest.xfail("lol, 2")
            assert x != 2 or x == 2

$ pytest -v lol.py    
============================= test session starts ==============================
platform linux -- Python 3.14.6, pytest-9.1.1, pluggy-1.6.0 -- /home/evgeni/Devel/theforeman/foremanctl/.venv/bin/python
cachedir: .pytest_cache
rootdir: /home/evgeni/Devel/theforeman/foremanctl
configfile: pytest.toml
plugins: testinfra-10.2.2, durations-1.7.0
collected 3 items                                                              

lol.py::test_lol XFAIL (lol)                                             [ 33%]
lol.py::test_lol2 XFAIL (lol)                                            [ 66%]
lol.py::test_sub SUBPASSED[1]                                            [100%]
lol.py::test_sub SUBXFAIL[2] (lol, 2)                                    [100%]
lol.py::test_sub SUBPASSED[3]                                            [100%]
lol.py::test_sub PASSED                                                  [100%]

============================= fixture duration top =============================
total          name        num med            min           
0:00:00.000038 grand total   1 0:00:00.000038 0:00:00.000038
============================ test call duration top ============================
total          name        num med            min           
0:00:00.001230 grand total   3 0:00:00.000053 0:00:00.000041
=========================== test setup duration top ============================
total          name        num med            min           
0:00:00.000234 grand total   3 0:00:00.000054 0:00:00.000033
========================== test teardown duration top ==========================
total          name        num med            min           
0:00:00.000122 grand total   3 0:00:00.000038 0:00:00.000033
=============== 1 passed, 3 xfailed, 2 subtests passed in 0.02s ================
```

evgeni: It works with the `pytest.mark.xfail` decorator, but we can't use that in subtests. :table_flip:

evgeni:
```suggestion
            user = container.inspect()['Config']['User']
```

ekohl: Ended up with your first example, avoiding `xfail` altogether.

tests/container_user_test.py
```diff
@@ -1,52 +1,20 @@
...
-        f"Expected:\n{expected_output}\n"
-        f"Actual:\n{actual_output}"
-    )
+    containers = server.podman.get_containers(status="running")
+    assert containers, "No running containers found"
+
+    for container in containers:
+        with subtests.test(container.name):
+            user = container.inspect()['Config']['User']
+            if container.name in EXPECTED_ROOT_IMAGES:
```
evgeni:
```suggestion
            if container.inspect()['Config']['Image'] in EXPECTED_ROOT_IMAGES:
```

or we need to update the list to use names, but I think images is more representative

ekohl: Agreed, images are better. Only downside is that when we branch, we'll need to update `nightly` to the stable version but I think that is acceptable.

PR #717, opened 2026-07-31
Add pyproject.toml and require Python 3.12 for local dev

Odilhao:

`./setup-environment` fails on CentOS Stream 9/RHEL 9/UBI9 because their default Python is 3.9, and `development/requirements.txt` pins `pytest>=9.0`, which dropped Python 3.9 support (requires 3.10+). Downgrading the pytest pin would just be papering over a version mismatch: downstream packaging already targets Python 3.12 (`python-obsah` builds against `python3.12`, `foremanctl.spec` requires `python3.12-obsah`) and CI already sets up Python 3.12 for every job. EL9's default 3.9 was never actually a supported target for this project, and EL10 ships 3.12 by default anyway.

* Add `pyproject.toml` declaring `requires-python = ">=3.12"` plus project metadata and dependencies. This is metadata-only (`foremanctl`/`forge` stay as bash wrappers, no console-script entry points, no bundling of `src/`/`development/` as package data). Version is resolved via `setuptools_scm` from `git describe`, mirroring what `VERSION`/`export-subst` does for release tarballs but working correctly from a plain checkout too (a custom `tag_regex` handles this repo's non-PEP-440 tags like `3.0.0-develop`). The build backend (`setuptools`) is left unpinned so it uses whatever version already ships with the running system rather than one pip would fetch from PyPI.
* `setup-environment` now checks for `python3.12` explicitly (instead of relying on whatever `python` resolves to on PATH) and fails fast with a clear, actionable error if it's missing.
* `src/requirements.txt` / `development/requirements.txt` are left unchanged for now — this is a first step, not a full migration.
* `DEVELOPMENT.md` now lists `Python - 3.12+` under Requirements.

Steps to reproduce:

* On a machine/container without `python3.12` on PATH (e.g. plain `quay.io/centos/centos:stream9`), run `./setup-environment` and confirm it fails fast with `Error: python3.12 is required ...` instead of pip's confusing "no matching distribution" error.
* Install `python3.12`/`python3.12-devel` (CS9) or just use CS10 (ships `python3.12` as default `python3`), run `./setup-environment`, and confirm `.venv/bin/python --version` reports 3.12 and `pytest>=9.0` installs successfully.
* In that venv, run `pip install --upgrade setuptools setuptools_scm && pip install --no-build-isolation -e '.[dev]'` and confirm it builds using the system's own `setuptools` (check `pip show setuptools` matches the OS package version) and resolves a valid version via `pip show foremanctl`.

This was verified end-to-end with `podman run` on both `quay.io/centos/centos:stream9` and `:stream10`.

#### Checklist
* [x] Tests added/updated (if applicable) — verified manually via podman on CS9/CS10; no automated test suite covers `setup-environment` itself.
* [x] Documentation updated (if applicable) — `DEVELOPMENT.md`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "foremanctl"
 description = "Deployment tool for Foreman and Katello using Podman quadlets and Ansible"
-requires-python = "~=3.12.0"
+requires-python = ">=3.12"
 dynamic = ["version"]
 dependencies = [
     "obsah>=1.9.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "foremanctl"
 description = "Deployment tool for Foreman and Katello using Podman quadlets and Ansible"
-requires-python = ">=3.12"
+requires-python = "~=3.12.0"
 dynamic = ["version"]
 dependencies = [
     "obsah>=1.9.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,38 @@
+[build-system]
+requires = ["setuptools", "setuptools_scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "foremanctl"
+description = "Deployment tool for Foreman and Katello using Podman quadlets and Ansible"
+requires-python = ">=3.12"
+dynamic = ["version"]
+dependencies = [
+    "obsah>=1.9.1",
+    "requests",
+    "requests-oauthlib",
+]
+
+[project.optional-dependencies]
+dev = [
+    "ansible-lint",
+    "apypie>=0.5.0",
+    "jinja2",
+    "paramiko",
+    "passlib",
+    "pytest>=9.0",
+    "pytest-testinfra",
+    "pytest-durations",
+    "python-dateutil",
+    "pyyaml",
+    "ruff",
+]
+
+[tool.setuptools]
+packages = []
+
+# Repo tags look like "3.0.0-develop" (not a PEP 440 version on their own);
+# only capture the leading numeric part and let setuptools_scm compute the
+# dev/distance suffix from there.
+[tool.setuptools_scm]
+tag_regex = "^(?P<version>\\d+\\.\\d+(?:\\.\\d+)?)"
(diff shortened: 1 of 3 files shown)


setup-environment
```diff
@@ -2,8 +2,15 @@
 
 set -eou pipefail
 
+REQUIRED_PYTHON=python3.12
```
jeremylenz: > setup-environment now checks for python3.12 explicitly (instead of relying on whatever python resolves to on PATH) and fails fast with a clear, actionable error if it's missing.
src/requirements.txt / development/requirements.txt are left unchanged for now — this is a first step, not a full migration.
> DEVELOPMENT.md now lists Python - 3.12+ under Requirements.

So which is it? 3.12+ or 3.12 exactly?

Odilhao: Changed to one dynamic resolution that will failed if it's not 3.12+, when I tried to run on CS9 with only python3.9 it failed, and later worked with 3.12.

jeremylenz (approved):

pyproject.toml
```diff
@@ -0,0 +1,38 @@
+[build-system]
+requires = ["setuptools", "setuptools_scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "foremanctl"
+description = "Deployment tool for Foreman and Katello using Podman quadlets and Ansible"
+requires-python = ">=3.12"
```
ekohl, line 8: Runtime wise I think this is lower. It's just for testing now.

ekohl (changes requested):
> ./setup-environment fails on CentOS Stream 9/RHEL 9/UBI9 because their default Python is 3.9, and development/requirements.txt pins pytest>=9.0, which dropped Python 3.9 support (requires 3.10+).

I'm sorry for adding that. I didn't realize it when I bumped it for the subtests feature.

I don't like duplicating the dependencies in 2 places. IMHO that needs to be solved.

ekohl:
I think https://github.com/theforeman/foremanctl/pull/720 is a lighter solution.

ehelms:
> I don't like duplicating the dependencies in 2 places. IMHO that needs to be solved.

I agree with that. @Odilhao if you could handle that and give more of an explanation of the value of usage of `pyproject.toml` for the project I think that would help.

ehelms:
Closing for now. @Odilhao I know you have expressed wanting to add a pyproject.toml. Please re-open or create a new one if you pursue that.

PR #719, opened 2026-08-03
Reuse obsah_params fixture to read parameters

ekohl:

While working on https://github.com/theforeman/foremanctl/pull/511 I wanted to move the file to a new location. It broke due to deriving the parameters location from the test filename.

* Reuse `obsah_params` fixture that's already defined to avoid duplicating logic

Steps to reproduce:

* Run the test suite
* Verify it passes

#### Checklist

### DIFF
--- a/tests/pulp_test.py
+++ b/tests/pulp_test.py
@@ -1,25 +1,9 @@
 import json
-import os
 
 import pytest
-import yaml
 
 PULP_API_SOCKET = '/run/httpd.pulp-api.sock'
 PULP_CONTENT_SOCKET = '/run/httpd.pulp-content.sock'
-
-
-def load_pulp_paths_from_parameters():
-    test_dir = os.path.dirname(os.path.abspath(__file__))
-    foremanctl_dir = os.path.dirname(test_dir)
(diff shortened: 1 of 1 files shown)


arvind4501:
I think then https://github.com/theforeman/foremanctl/blob/9ffe3604f01662fcd67edc77b322556fdb1881fd/tests/migration_test.py#L7-L9 should also be using shared fixture

ekohl:
I looked at that, but the test asserts that the file exists. There's a subtle on paths difference. Compare:
https://github.com/theforeman/foremanctl/blob/9ffe3604f01662fcd67edc77b322556fdb1881fd/tests/conftest.py#L19
https://github.com/theforeman/foremanctl/blob/9ffe3604f01662fcd67edc77b322556fdb1881fd/tests/migration_test.py#L9

I didn't want to untangle that too and for now only focus on reducing the size of https://github.com/theforeman/foremanctl/pull/511.

arvind4501:
Sorry i did not catch that `.var/*` and `/var*` earlier, and that make sense, thank you :)

arvind4501 (approved):

ekohl:
I must admit I didn't either until you asked me to look at it again. Only then I spotted it.

PR #721, opened 2026-08-03
turn off host culling in IoP host inventory

pfreyburg:

There are multiple errors with host culling in IoP host-inventory so we decided to turn it off. Host culling is only cloud feature and the control of host culling should be on foreman side. 

The culling is disabled using environment variables because disabling in code base would cause big difference between IoP branches and master. The PR set environment variables to infinite values.

* set environment variables for Host Inventory MQ Service Container to enable hosts with unlimited staleness.

Steps to reproduce:

* the host won't be unregistered. There is now quick way to test that.

#### Checklist


pfreyburg:
Please note this will work only with version of host-inventory which consists this https://github.com/RedHatInsights/insights-host-inventory/pull/4674. This PR also doesn't solve the current hosts.

src/roles/iop_inventory/tasks/main.yaml
```diff
@@ -73,6 +73,12 @@
       KAFKA_BOOTSTRAP_SERVERS: "PLAINTEXT://iop-core-kafka:9092"
       USE_SUBMAN_ID: "true"
       INVENTORY_DB_SSL_MODE: "disable"
+      CONVENTIONAL_TIME_TO_STALE_SECONDS: "99999999999"
```
brantleyr: I think you're also going to need to add these to the `iop-core-host-inventory-api` below. The API can also create/update hosts via REST, which access the same models.

brantleyr (changes requested):
Two things with the `inventory.hosts` VIEW at the bottom of this file:

1. Hardcoded staleness intervals (related to this PR) - The VIEW computes staleness as `h.last_check_in + INTERVAL '29 hours'` instead of reading the actual columns. Downstream services reading from the VIEW will still see 29-hour staleness regardless of the env vars or SQL migration. This could be fixed in this PR by using the column values directly (`h.stale_timestamp, h.stale_warning_timestamp, h.deletion_timestamp AS culled_timestamp`). Note that the same change would also need to be made in [puppet-iop (manifests/core_host_inventory.pp)](https://github.com/theforeman/puppet-iop/blob/master/manifests/core_host_inventory.pp), since both define this VIEW with CREATE OR REPLACE and whichever runs last wins.

2. `system_profiles_static mismatch` (separate issue, but related, should probably be a different PR, or we do all of the VIEW updates in one PR) - The VIEW joins `hbi.system_profiles_static` for `owner_id`, `operating_system`, and `host_type`, but foreman-3.18 HBI doesn't write to that table - it writes to `system_profile_facts JSONB`. Those fields will be NULL for every host. This is a pre-existing issue that should be tracked separately - either the VIEW needs to be reverted to read from JSONB (matching what foreman-3.18 HBI actually writes), or HBI on the foreman branches needs to be updated to populate `system_profiles_static`. Worth noting that master removed this VIEW entirely as part of the Cyndi decommission - the long-term direction is REST APIs, not FDW.

So we need to decide if we want to roll-back the VIEW changes that worked, or fix the actual API/Service consuming logic for the applicable Foreman branches, then we can apply these changes.

brantleyr:
A potential simpler alternative: A new `DISABLE_HOST_CULLING=true`

Instead of six `*_TIME_TO_STALE_SECONDS` env vars, HBI's existing `should_host_stay_fresh_forever()` could be expanded with a single `DISABLE_HOST_CULLING=true` check. One env var, no int() cast dependency, clearer intent. 

Would need a small HBI change - see [comment on #4688](https://github.com/RedHatInsights/insights-host-inventory/pull/4688#issuecomment-5171740294).
Either way, this also needs to be set on the `iop-core-host-inventory-api` container (same code path).

pfreyburg:
> A potential simpler alternative: A new `DISABLE_HOST_CULLING=true`
> 
> Instead of six `*_TIME_TO_STALE_SECONDS` env vars, HBI's existing `should_host_stay_fresh_forever()` could be expanded with a single `DISABLE_HOST_CULLING=true` check. One env var, no int() cast dependency, clearer intent.
> 
> Would need a small HBI change - see [comment on #4688](https://github.com/RedHatInsights/insights-host-inventory/pull/4688#issuecomment-5171740294). Either way, this also needs to be set on the `iop-core-host-inventory-api` container (same code path).

@brantleyr https://github.com/RedHatInsights/insights-host-inventory/pull/4728

pfreyburg:
This needs to be merged first https://github.com/RedHatInsights/insights-host-inventory/pull/4728

pfreyburg:
Merge it after https://github.com/theforeman/foremanctl/pull/725. Then I will rebase this PR

brantleyr (approved):

pfreyburg:
@ehelms The https://github.com/theforeman/foremanctl/pull/725 and https://github.com/RedHatInsights/insights-host-inventory/pull/4728 was merged. Can you please merge this?

ehelms (approved):

PR #724, opened 2026-08-04
Fix Candlepin TLS cipher list for the OpenSSL FFM connector

Odilhao:

To be able to merge https://github.com/theforeman/candlepin-oci-images/pull/66 :smile: 

Candlepin 5.0's Tomcat uses the new OpenSSL FFM connector, which fails to handle the IANA-style cipher names (TLS_ECDHE_...) we were using, leaving the connector with no usable ciphers and every TLS handshake failing. Switch to the legacy OpenSSL cipher naming (ECDHE-ECDSA-AES256-GCM-SHA384, ...) to make it work.

Verified locally with foremanctl deploy against the Candlepin 5.0.2/ el10 build ([candlepin-oci-images#66](https://github.com/theforeman/candlepin-oci-images/pull/66)): Candlepin now reports (healthy) and /candlepin/status responds 200 consistently.

Steps to reproduce:

Apply the following diff and run with forge:

```bash
diff --git a/src/roles/candlepin/defaults/main.yml b/src/roles/candlepin/defaults/main.yml
index 3d4b139..c2be48a 100644
--- a/src/roles/candlepin/defaults/main.yml
+++ b/src/roles/candlepin/defaults/main.yml
@@ -5,10 +5,10 @@ candlepin_tls_versions:
   - "TLSv1.2"
   - "TLSv1.3"
 candlepin_ciphers:
-  - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
-  - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
-  - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
-  - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+ - "PROFILE=SYSTEM"
 candlepin_secret_mount_opts: "mode=0440,uid=0,gid=53,type=mount"
diff --git a/src/vars/images.yml b/src/vars/images.yml
index 0782530..8dc47ec 100644
--- a/src/vars/images.yml
+++ b/src/vars/images.yml
@@ -2,8 +2,8 @@ images_quadlet_dir: /etc/containers/systemd
 images_registry_auth_file: /etc/foreman/registry-auth.json
 
 container_tag_stream: "nightly"
-candlepin_container_image: quay.io/foreman/candlepin
-candlepin_container_tag: "foreman-{{ container_tag_stream }}"
+candlepin_container_image: quay.io/foreman/stage/candlepin
+candlepin_container_tag: "pull-request-66"
 foreman_proxy_container_image: "quay.io/foreman/foreman-proxy"

```

### DIFF
--- a/src/roles/candlepin/defaults/main.yml
+++ b/src/roles/candlepin/defaults/main.yml
@@ -5,10 +5,7 @@ candlepin_tls_versions:
   - "TLSv1.2"
   - "TLSv1.3"
 candlepin_ciphers:
-  - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
-  - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
-  - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
-  - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+  - "PROFILE=SYSTEM"
 candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "4.4.14"
 candlepin_secret_mount_opts: "mode=0440,uid=0,gid=53,type=mount"


evgeni:
> Candlepin 5.0's Tomcat uses the new OpenSSL FFM connector, which fails to handle the IANA-style cipher names (TLS_ECDHE_...) we were using, leaving the connector with no usable ciphers and every TLS handshake failing. Switch to the legacy OpenSSL cipher naming (ECDHE-ECDSA-AES256-GCM-SHA384, ...) to make it work.

We're already using FFM with Tomcat 9 and Candlepin 4.8, where the current notation works, so this explanation seems to be not complete.

Looking at both [Tomcat 9](https://tomcat.apache.org/tomcat-9.0-doc/config/http.html#SSL_Support_-_SSLHostConfig) and [Tomcat 10](https://tomcat.apache.org/tomcat-10.1-doc/config/http.html#SSL_Support_-_SSLHostConfig) documentation, I only see the OpenSSL notation of ciphers mentioned, and that's what we're already using.

Odilhao:
> > Candlepin 5.0's Tomcat uses the new OpenSSL FFM connector, which fails to handle the IANA-style cipher names (TLS_ECDHE_...) we were using, leaving the connector with no usable ciphers and every TLS handshake failing. Switch to the legacy OpenSSL cipher naming (ECDHE-ECDSA-AES256-GCM-SHA384, ...) to make it work.
> 
> We're already using FFM with Tomcat 9 and Candlepin 4.8, where the current notation works, so this explanation seems to be not complete.
> 
> Looking at both [Tomcat 9](https://tomcat.apache.org/tomcat-9.0-doc/config/http.html#SSL_Support_-_SSLHostConfig) and [Tomcat 10](https://tomcat.apache.org/tomcat-10.1-doc/config/http.html#SSL_Support_-_SSLHostConfig) documentation, I only see the OpenSSL notation of ciphers mentioned, and that's what we're already using.

I'm not sure what changed on Candlepin from 4.8 to 5.0, the cipher change made the container work locally for, but it showed another problem on Katello + Candlepin communication that @ianballou is already aware of.

ianballou:
This change was required for the Candlepin 5.0.2 container to come up healthy in my environment.

```
...
        <SSLHostConfig certificateVerification="optional"
                       protocols="TLSv1.2,TLSv1.3"
                       caCertificateFile="/etc/candlepin/certs/candlepin-ca.crt"
                       ciphers="ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256">
...
```
---
```
    Waiting for Tomcat to start...
                           ciphers="TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384:TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384:TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256:TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256">
    ---
    *   Trying ::1:23443...
    * connect to ::1 port 23443 failed: Connection refused
    *   Trying 127.0.0.1:23443...
    * Connected to localhost (127.0.0.1) port 23443 (#0)
    * ALPN, offering h2
    * ALPN, offering http/1.1
    *  CAfile: /etc/pki/tls/certs/ca-bundle.crt
    ... (2 lines hidden, /toggle to show all)
    * TLSv1.3 (OUT), TLS handshake, Client hello (1):
    } [512 bytes data]
    * TLSv1.2 (IN), TLS header, Unknown (21):
    { [5 bytes data]
    * TLSv1.3 (IN), TLS alert, handshake failure (552):
    { [2 bytes data]
    * error:0A000410:SSL routines::ssl/tls alert handshake failure
    * Closing connection 0
```

src/roles/candlepin/defaults/main.yml
```diff
@@ -5,10 +5,10 @@ candlepin_tls_versions:
   - "TLSv1.2"
   - "TLSv1.3"
 candlepin_ciphers:
```
ehelms, line 7: I wonder if we should drop these all together. From what I understand, Tomcat should be inheriting the system crypto policy through the default Java configuration.

Odilhao: Are we changing crypto policies for containers at the moment?

ehelms: We will be at some point -- but what does that have to do with this? Our goal is to inherit and rely on the system crypto policy. We are doing this across other services as well, so should we just do that here?

ehelms: @evgeni @ekohl your thoughts would be appreciated

ekohl: If on a technical level you can actually use crypto-policies, then please do. IMHO the only reason we previously didn't is that we couldn't (or didn't know how). So `PROFILE=SYSTEM` for the protocols and leave ciphers empty would have my preference.

And if we ever settle on https://github.com/theforeman/candlepin-oci-images/pull/3 then I'd remove the parameters from foremanctl altogether: users shouldn't override it that way.

ehelms: Let's try setting the ciphers value to `ciphers="PROFILE=SYSTEM`. Since FFM uses OpenSSL, this should in theory work the same as how we do it with Apache.

ianballou: I just tested with the following on a patched CP 5 image:

```
        <SSLHostConfig certificateVerification="optional"
                       protocols="TLSv1.2,TLSv1.3"
                       caCertificateFile="/etc/candlepin/certs/candlepin-ca.crt"
                       ciphers="PROFILE=SYSTEM">
            <Certificate certificateFile="/etc/candlepin/certs/tomcat.crt"
                         certificateKeyFile="/etc/candlepin/certs/tomcat.key" />
        </SSLHostConfig>
```

I was able to import a manifest, refresh it, and bind some entitlements.

ianballou: I did also try this on my PQC machine with only TLS1.3 and it threw a warning, it seems with TLS 1.3 we cannot use PROFILE=SYSTEM at the moment.

ekohl: Can you share the warning?

ianballou:
```
WARNING [main] org.apache.tomcat.util.net.openssl.panama.OpenSSLContext.init Some or all of cipher suite [PROFILE=SYSTEM] for TLS 1.3+ could not be enabled
```

ehelms: If you could also include `nmap` output to show what ciphers are being offered in this state.

ianballou: For completion, here are the list of ciphers returned by the working TLS 1.2 + 1.3 enabled machine:

```
$ nmap --script ssl-enum-ciphers -p 23443 localhost
Starting Nmap 7.92 ( https://nmap.org ) at 2026-08-12 20:30 UTC
Nmap scan report for localhost (127.0.0.1)
Host is up (0.00012s latency).
Other addresses for localhost (not scanned): ::1

PORT      STATE SERVICE
23443/tcp open  unknown
| ssl-enum-ciphers: 
|   TLSv1.2: 
|     ciphers: 
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA (secp256r1) - A
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 (secp256r1) - A
|       TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (secp256r1) - A
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA (secp256r1) - A
|       TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 (secp256r1) - A
|       TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 (secp256r1) - A
|     compressors: 
|       NULL
|     cipher preference: client
|     warnings: 
|       Key exchange (secp256r1) of lower strength than certificate key
|   TLSv1.3: 
|     ciphers: 
|       TLS_AKE_WITH_AES_128_CCM_SHA256 (secp256r1) - A
|       TLS_AKE_WITH_AES_128_GCM_SHA256 (secp256r1) - A
|       TLS_AKE_WITH_AES_256_GCM_SHA384 (secp256r1) - A
|       TLS_AKE_WITH_CHACHA20_POLY1305_SHA256 (secp256r1) - A
|     cipher preference: client
|_  least strength: A

Nmap done: 1 IP address (1 host up) scanned in 0.24 seconds
```

And here it is on the TLS 1.3 only PQC machine:

```
#  nmap --script ssl-enum-ciphers -p 23443 localhost
Starting Nmap 7.92 ( https://nmap.org ) at 2026-08-12 17:26 EDT
Nmap scan report for localhost (127.0.0.1)
Host is up (0.000091s latency).
Other addresses for localhost (not scanned): ::1

PORT      STATE SERVICE
23443/tcp open  unknown

Nmap done: 1 IP address (1 host up) scanned in 0.23 seconds
```

Seems `nmap` doesn't show anything on the TLS 1.3 only PQC machine.

Odilhao: What do I need to update now? Is this is PR good as is?

src/roles/candlepin/defaults/main.yml
```diff
@@ -5,10 +5,10 @@ candlepin_tls_versions:
...
-  - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
-  - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
-  - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
-  - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+  - ECDHE-ECDSA-AES256-GCM-SHA384
+  - ECDHE-RSA-AES256-GCM-SHA384
+  - ECDHE-ECDSA-AES128-GCM-SHA256
+  - ECDHE-RSA-AES128-GCM-SHA256
```
ianballou: This worked fine on a regular, non-PQC containerized box. I would need to do more testing to say if the issues on the PQC box were caused by other configuration hacks I made.

```suggestion
  - "PROFILE=SYSTEM"
```

src/roles/candlepin/defaults/main.yml
ianballou: Commented below with the suggestion.

src/roles/candlepin/defaults/main.yml
Odilhao: Applied.

jeremylenz (approved):

ianballou:
I dug up some more info about the TLS 1.3 situation.

Firstly, I found https://community.theforeman.org/t/smart-proxy-cipher-setting-for-tls-1-3-on-machines-with-ruby-3-2/46924

This post discusses `SSL_CTX_set_cipher_list()` in openssl for TLS 1.2 vs `SSL_CTX_set_ciphersuites()` for TLS 1.3. Okay, so there are different OpenSSL methods used for the different TLS versions.

Then, I found https://src.fedoraproject.org/rpms/openssl/blob/rawhide/f/0005-RH-Add-support-for-PROFILE-SYSTEM-system-default-cip.patch, a patch to OpenSSL that updates `ssl_create_cipher_list()` to accept `PROFILE=SYSTEM`, which `SSL_CTX_set_cipher_list()` uses:

```c
int SSL_CTX_set_cipher_list(SSL_CTX *ctx, const char *str)
{
    STACK_OF(SSL_CIPHER) *sk;

    sk = ssl_create_cipher_list(ctx, ctx->tls13_ciphersuites,
        &ctx->cipher_list, &ctx->cipher_list_by_id, str,
        ctx->cert);
```

https://github.com/openssl/openssl/blob/openssl-3.5/ssl/ssl_lib.c#L3360-L3366

So, that is to say, it looks like TLS 1.3 really is incompatible with PROFILE=SYSTEM. It may be updated in the future.

But, that isn't bad news and I think we can keep using it since the default ciphers are the fall back option as I found out in the `nmap` above.

I'm pretty certain that, on the PQC box, `nmap` failed to return anything because of the ML-DSA certs in the equation.

To sum up - I think we can continue with this PR, but we should maintain an understanding that the default tls 1.3 ciphers will be returned.

ianballou:
```
[2](https://github.com/theforeman/foremanctl/actions/runs/31822658069/job/94840098170?pr=724#step:18:263)
=================================== FAILURES ===================================
__________________________________ test_ping ___________________________________
[gw1] linux -- Python 3.12.3 /home/runner/smoker/venv/bin/python3

api = <apypie.api.Api object at 0x7f353a0d6810>

    def test_ping(api):
        ping = api.resource('ping').call('ping')['results']
    
        assert ping['foreman']['database']['active']
    
        for subsystem, data in ping.items():
            try:
>               assert data['status'] == 'ok', f"{subsystem} is ok"
E               AssertionError: foreman_rh_cloud is ok
E               assert 'FAIL' == 'ok'
E                 
E                 - ok
E                 + FAIL
```

foreman_rh_cloud related? Seems outside of this PR.

ehelms (approved):

PR #725, opened 2026-08-05
update fdw in iop_inventory

pfreyburg:

The VIEW inventory.hosts should match between:

- foremanctl - [foremanctl/src/roles/iop_inventory/tasks/main.yaml at master · theforeman/foremanctl](https://github.com/theforeman/foremanctl/blob/master/src/roles/iop_inventory/tasks/main.yaml) 
- insights-host-inventory foreman-3.18 - [insights-host-inventory/jobs/add_inventory_view.py at foreman-3.18 · RedHatInsights/insights-host-inventory](https://github.com/RedHatInsights/insights-host-inventory/blob/foreman-3.18/jobs/add_inventory_view.py) 
- insights-host-inventory foreman-3.16 - [insights-host-inventory/jobs/add_inventory_view.py at foreman-3.16 · RedHatInsights/insights-host-inventory](https://github.com/RedHatInsights/insights-host-inventory/blob/foreman-3.16/jobs/add_inventory_view.py)

* `system_profile_facts as system_profile` instead of computing it from columns which are NULL

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/roles/iop_inventory/tasks/main.yaml
+++ b/src/roles/iop_inventory/tasks/main.yaml
@@ -241,17 +241,15 @@
         h.display_name,
         h.created_on as created,
         h.modified_on as updated,
-        h.last_check_in + INTERVAL '29 hours' AS stale_timestamp,
-        h.last_check_in + INTERVAL '7 days' AS stale_warning_timestamp,
-        h.last_check_in + INTERVAL '30 days' AS culled_timestamp,
+        h.stale_timestamp,
+        h.stale_warning_timestamp,
+        h.deletion_timestamp AS culled_timestamp,
         h.tags_alt as tags,
         h.system_profile_facts as system_profile,
-        h.insights_id,
+        (h.canonical_facts ->> 'insights_id')::uuid as insights_id,
         h.reporter,
         h.per_reporter_staleness,
         h.org_id,
-        h.groups,
-        h.last_check_in
+        h.groups
       FROM hbi.hosts h
-      LEFT JOIN hbi.system_profiles_static sps ON sps.org_id = h.org_id AND sps.host_id = h.id
-      WHERE h.insights_id != '00000000-0000-0000-0000-000000000000';
+      WHERE (h.canonical_facts->'insights_id' IS NOT NULL);
--- a/src/roles/iop_inventory/tasks/main.yaml
+++ b/src/roles/iop_inventory/tasks/main.yaml
@@ -232,7 +232,6 @@
     login_host: localhost
     # TODO(RHINENG-26911): remove this view once Cyndi decommission completes
     # across all IoP services.
-    # Staleness intervals are hardcoded HBI defaults (29h, 7d, 30d).
     # Per-org custom staleness from hbi.staleness is not supported.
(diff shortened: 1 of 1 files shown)


pfreyburg:
@vkrizan @brantleyr Can you please look at this PR?

First we need to decide on https://github.com/theforeman/puppet-iop/pull/106 how we deal with these questions. Then we need to make these changes also here.

brantleyr (approved):
Approved pending updates on https://github.com/theforeman/puppet-iop/pull/106 and comments.

brantleyr:
@pfreyburg approved. We're going to need to make sure we also set this environment variable in puppet-iop:

- https://github.com/theforeman/puppet-iop/blob/master/manifests/core_host_inventory.pp#L215-L217
- https://github.com/theforeman/puppet-iop/blob/master/manifests/core_host_inventory.pp#L262-L267

vkrizan (approved):

src/roles/iop_inventory/tasks/main.yaml
```diff
@@ -241,17 +241,15 @@
...
         h.tags_alt as tags,
-        jsonb_build_object('operating_system', sps.operating_system, 'host_type', sps.host_type, 'owner_id', sps.owner_id) as system_profile,
-        h.insights_id,
+        h.system_profile_facts as system_profile,
+        (h.canonical_facts ->> 'insights_id')::uuid as insights_id,
         h.reporter,
         h.per_reporter_staleness,
         h.org_id,
-        h.groups,
-        h.last_check_in
```
vkrizan, line 254: JFYI: Removal of a column will have a fatal cause for an existing installation.

From PG docs https://www.postgresql.org/docs/current/sql-createview.html:

> CREATE OR REPLACE VIEW is similar, but if a view of the same name already exists, it is replaced. The new query must generate the same columns that were generated by the existing view query (that is, the same column names in the same order and with the same data types), but it may add additional columns to the end of the list. The calculations giving rise to the output columns may be completely different.

ehelms: > JFYI: Removal of a column will have a fatal cause for an existing installation.

As in, upgrades will fail?

vkrizan: @ehelms Yes. However, this currently only concerns existing stream installations. Technically we have a wiggle room if we require all stream installations to do a manual fix (removal of the view before upgrade).

ehelms: What are stream installations? CentOS Stream?

`iop` was included in foremanctl 2.Y release stream and master is tracking the 3.0 release. So the question is what happens for anyone upgrading from 2.Y?

vkrizan: No, I've meant Foreman Stream, i.e. the develop branch.

Petr has kept the field for continuity. There should not be an impact between foreman 3.18 and foreman develop/5.0. This could have been an issue only if one would stay on develop branch.

ehelms (approved):
The devel failures look unrelated.

PR #675, opened 2026-07-17
Add Registration smart proxy feature (SAT-44972)

jeremylenz:
#### Why are you introducing these changes? (Problem description, related links)

Adds the Registration smart proxy feature for foremanctl (SAT-44972). Registration proxies host registration requests through an external smart proxy to Foreman, enabling hosts to register via their nearest proxy.

Depends on 
* https://github.com/Katello/katello/pull/11807 (Katello-side registration proxy support).
* https://github.com/theforeman/foreman/pull/11124 (finding colocated smart proxies)

#### What are the changes introduced in this pull request?

* Adds `registration` feature to `features.yaml` with a dependency on `templates` (auto-enabled)
* HTTPS-only: `registration.yml.j2` sets `:enabled: https`
* Adds optional `--registration-url` parameter for load-balanced proxy setups
* Includes `registration` in the `foreman-proxy-content` flavor (external smart proxies only — not deployed on the main Foreman server)
* Updates `parameters.md` to document the new parameters
* Adds tests: feature presence check in `test_foreman_proxy_features` and config file verification in `test_registration_enabled`

#### Set up your external smart proxy

Make sure `foremanctl` is updated to current

On Foreman server:
```
./foremanctl auth-bundle $PROXY_FQDN
scp /var/lib/foremanctl/certs/bundles/$PROXY_FQDN.tar.gz $PROXY_FQDN:/root/
```

On your external smart proxy machine:

```
./foremanctl deploy-proxy --foreman-fqdn "$FOREMAN_FQDN" --flavor foreman-proxy-content --auth-bundle ~/$PROXY_FQDN.tar.gz
```

* Verify `registration` appears in the smart proxy features list
* Verify `/etc/foreman-proxy/settings.d/registration.yml` contains `:enabled: https`
* Optionally pass `--registration-url https://loadbalancer.example.com` and verify the setting

#### Checklist
* [x] Tests added/updated (if applicable)
* [x] Documentation updated (if applicable)

PR #726, opened 2026-08-05
Satellite flavor

evgeni:

* Adds Satellite flavor, based on #628

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -502,8 +502,6 @@ jobs:
       - name: Deploy features
         run: |
           ./foremanctl deploy \
-            --add-feature azure-rm \
-            --add-feature google \
             --add-feature bmc
       - name: Run tests
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -502,8 +502,6 @@ jobs:
       - name: Deploy features
         run: |
           ./foremanctl deploy \
-            --add-feature hammer \
-            --add-feature foreman-proxy \
             --add-feature azure-rm \
             --add-feature google \
             --add-feature bmc
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -465,6 +465,76 @@ jobs:
           limit-access-to-actor: true
           wait-timeout-minutes: 5
 
+  satellite-tests:
+    runs-on: ubuntu-24.04
+    name: "Tests Satellite Deployment"
+    steps:
+      - uses: actions/checkout@v7
+      - name: Set up Python
+        uses: actions/setup-python@v7
+        with:
+          python-version: '3.12'
+      - name: Setup libvirt for Vagrant
+        uses: voxpupuli/setup-vagrant@v0
+        with:
+          configure_dns: true
+      - name: Install Ansible
+        run: pip install --upgrade ansible-core
+      - name: Setup environment
+        run: ./setup-environment
+      - name: Start VMs
+        run: |
+          ./forge vms start --vms "quadlet client"
+      - name: Configure repositories
+        run: |
+          ./forge setup-repositories
+      - name: Run image pull
+        run: |
+          ./foremanctl pull-images
+      - name: Run deployment
+        run: |
+          ./foremanctl deploy \
+            --flavor satellite \
+            --initial-admin-password=changeme \
+            --initial-organization "Foreman CI" \
+            --initial-location "Internet" \
+            --tuning development
+      - name: Deploy features
+        run: |
+          ./foremanctl deploy \
+            --add-feature hammer \
+            --add-feature foreman-proxy \
+            --add-feature azure-rm \
+            --add-feature google \
+            --add-feature bmc
+      - name: Run tests
+        run: |
+          ./forge test
+      - name: Run smoker
+        run: |
+          ./forge smoker
+      - name: Archive smoker report
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: smoker-satellite
+          path: "/home/runner/smoker/report/"
+      - name: Generate sos reports
+        if: ${{ always() }}
+        run: ./forge sos
+      - name: Archive sos reports
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: sosreport-satellite
+          path: sos/
+      - name: Setup upterm session
+        if: ${{ failure() }}
+        uses: owenthereal/action-upterm@v1
+        with:
+          limit-access-to-actor: true
+          wait-timeout-minutes: 5
+
   # A dummy job that you can mark as a required check instead of each individual test
   test-suite:
     if: always()
@@ -474,6 +544,7 @@ jobs:
       - upgrade
       - migration
       - foreman-proxy-content-tests
+      - satellite-tests
       - ansible-lint
       - python-lint
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -482,6 +482,8 @@ jobs:
         run: pip install --upgrade ansible-core
       - name: Setup environment
         run: ./setup-environment
+      - name: Apply vendor overrides
+        run: cp -a vendor_overrides/satellite src/playbooks/_vendor_overrides
       - name: Start VMs
         run: |
           ./forge vms start --vms "quadlet client"
--- a/src/filter_plugins/foremanctl.py
+++ b/src/filter_plugins/foremanctl.py
@@ -73,18 +73,25 @@ def available_foreman_plugins(_value):
 def list_all_features(enabled_features, only_enabled=False):
     enabled_list = []
     available_list = []
+    list_internal = os.environ.get('FOREMANCTL_FEATURES_LIST_INTERNAL', '') == 'true'
     for name, meta in FEATURE_MAP.items():
-        if meta.get('internal', False) and not os.environ.get('FOREMANCTL_FEATURES_LIST_INTERNAL', '') == 'true':
+        internal = meta.get('internal', False)
+        if internal and not list_internal:
             continue
         description = meta.get('description', '')
         if name in enabled_features:
-            enabled_list.append((name, 'enabled', description))
+            enabled_list.append((name, 'enabled', internal, description))
         elif not only_enabled:
-            available_list.append((name, 'available', description))
+            available_list.append((name, 'available', internal, description))
 
-    output = [f"{'FEATURE':<25} {'STATE':<12} DESCRIPTION"]
-    for name, state, description in enabled_list + available_list:
-        output.append(f"{name:<25} {state:<12} {description}")
+    if not list_internal:
+        output = [f"{'FEATURE':<25} {'STATE':<12} DESCRIPTION"]
+        for name, state, _internal, description in enabled_list + available_list:
+            output.append(f"{name:<25} {state:<12} {description}")
+    else:
+        output = [f"{'FEATURE':<25} {'STATE':<12} {'INTERNAL':<8} DESCRIPTION"]
+        for name, state, internal, description in enabled_list + available_list:
+            output.append(f"{name:<25} {state:<12} {internal:<8} {description}")
 
     return "\n".join(output)
 
--- a/src/filter_plugins/foremanctl.py
+++ b/src/filter_plugins/foremanctl.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 
 __metaclass__ = type
 
+import os
 import pathlib
 
 import yaml
@@ -15,7 +16,7 @@ features_yaml = _SRC_ROOT / 'features.yaml'
 with features_yaml.open() as features_file:
     FEATURE_MAP = yaml.safe_load(features_file)
 
-# load additional feature files under features.d  
+# load additional feature files under features.d
 _features_d = _SRC_ROOT / 'features.d'
 if _features_d.is_dir():
     for _overlay in sorted(_features_d.glob('*.yaml')):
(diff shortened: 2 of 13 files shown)


tests/conftest.py
```diff
@@ -286,7 +290,8 @@ def pytest_configure(config):
 
 def pytest_collection_modifyitems(config, items):
     active_flavor = config.user_parameters.flavor
-    active_flavor_dir = FLAVOR_TESTS_DIR / active_flavor
+    # if there is an override, use that, otherwise use the flavor verbatim
+    active_flavor_dir = FLAVOR_TESTS_DIR / FLAVOR_TESTS_DIR_OVERRIDES.get(active_flavor, active_flavor)
```
evgeni, line 307: this means we can't have a `tests/flavor/satellite` that adds satellite-specific tests. I don't think that's bad (we can place them somewhere else, if we ever have any), but wanted to call it out here.

.github/workflows/test.yml
```diff
@@ -465,6 +465,74 @@ jobs:
...
+            --flavor satellite \
+            --initial-admin-password=changeme \
+            --initial-organization "Foreman CI" \
+            --initial-location "Internet" \
+            --tuning development
+      - name: Deploy features
+        run: |
+          ./foremanctl deploy \
+            --add-feature azure-rm \
+            --add-feature google \
```
shubhamsg199: Shouldn't azure-rm and google both be enabled by default as per the satellite flavor?

evgeni: good catch, yes they should

tests/backup_test.py
```diff
@@ -8,23 +8,24 @@
...
+        databases.append('foreman')
 
-    # Foreman-proxy-content flavor only has pulp
-    elif flavor == 'foreman-proxy-content':
-        databases = ['pulp']
+    if 'candlepin' in enabled_features:
+        databases.append('candlepin')
+
+    if any([feature.startswith('content/') for feature in enabled_features]):
+        databases.append('pulp')
```
ekohl: There's the internal feature `pulp`. In https://github.com/theforeman/foremanctl/pull/722 I did some work to expose internal features to the testing and may be better than the `content` features. If this gets merged first, I can update my PR to do that.

tests/conftest.py
```diff
@@ -35,7 +35,8 @@ def features(self):
         # FEATURE                   STATE               DESCRIPTION
         # $feature                  enabled/available   $description
         output = subprocess.check_output(['./foremanctl', 'features'], cwd=self._config.rootdir,
-                                         universal_newlines=True)
+                                         universal_newlines=True,
+                                         env=dict(**os.environ, **{'FOREMANCTL_FEATURES_LIST_INTERNAL': 'true'}))
```
ekohl: In my change I went with `os.environ | {'FOREMANCTL_FEATURES_LIST_INTERNAL': 'true'}` which works since Python 3.9. IMHO it's more readable.

tests/backup_test.py
evgeni: Updated to use `pulp`

tests/conftest.py
ekohl: In https://github.com/theforeman/foremanctl/pull/722 I wanted to go all the way to feature-based testing anyway and reduce the flavor-based testing. I see here you already took the approach of exposing internal features (just like I did) so perhaps we can just get rid of the flavor based tests anyway?

`tests/flavor/katello/postgresql_test.py` already has logic that is more automated in `tests/backup_test.py` while `tests/flavor/katello/images_test.py` could also be easily written to look at enabled features.

The `foreman-proxy-content` are a bit more special, but that's not a concern for this particular PR.

tests/conftest.py
evgeni: I always forget about this modern shit

src/vars/flavors/satellite.yml
```diff
@@ -0,0 +1,34 @@
...
+  - content/container
+  - content/rpm
+  - ansible
+  - remote-execution
+  - rh-cloud
+  - theme-satellite
+  - webhooks
+  - hammer
+  - foreman-proxy
+
```
stejskalleos, line 19:
```suggestion
  - bootdisk
  - google
  - leapp
  - discovery
  - openscap
  - templates #foreman_templates 
  - virt_who

 ```

I know some of them are not ready yet, but do we have an issue to track the missing features?

Comment (stejskalleos) [reply src/vars/flavors/satellite.yml:19]: From my Satellite:

Image

**Comment by shubhamsg199** [comment]
https://github.com/SatelliteQE/robottelo/pull/22356#issuecomment-5195758819 So --remove-feature does not work when we use --flavor satellite ?

**Comment by evgeni** [comment]
> [SatelliteQE/robottelo#22356 (comment)](https://github.com/SatelliteQE/robottelo/pull/22356#issuecomment-5195758819) So --remove-feature does not work when we use --flavor satellite ?

You can't remove a feature that is part of the flavor, only those you added yourself.

**Comment by stejskalleos** [comment]
Running `foremanctl deploy --flavor satellite --initial-admin-password=changeme --tuning development` on nightly with the [packit](https://dashboard.packit.dev/jobs/copr/3833499) does not enable satellite theme:

image

**Comment by evgeni** [comment]
Where exactly did you expect to get the theme from? You need to use Satellite-based containers for that.

**Comment by stejskalleos** [comment]
> Where exactly did you expect to get the theme from? You need to use Satellite-based containers for that.

So the `--flavor satellite` should be available only on Satellite containers as well?

**Comment by evgeni** [comment]
> So the `--flavor satellite` should be available only on Satellite containers as well?

Ideally, yes. I am still working on this aspect (see also #643)

File: src/vars/flavors/satellite.yml (hunk 5 above)
Comment (evgeni) [reply src/vars/flavors/satellite.yml:19]: That would mean also actually defining those features, and people would see them, try to enable them, and complain that it has no effect.

IMHO we should not.

File: .github/workflows/test.yml [hunk 6]
```diff
@@ -465,6 +465,74 @@ jobs:
...
+      - name: Setup libvirt for Vagrant
+        uses: voxpupuli/setup-vagrant@v0
+        with:
+          configure_dns: true
+      - name: Install Ansible
+        run: pip install --upgrade ansible-core
+      - name: Setup environment
+        run: ./setup-environment
+      - name: Apply vendor overrides
+        run: cp -a vendor_overrides/satellite src/playbooks/_vendor_overrides
```
arvind4501, line 486: Is this copied to only test the vendor supplied flavors and with that we don't want to expose satellite flavor upstream? or did i misread something

evgeni: Correct. A "regular" (upstream) foremanctl deployment should never expose the Satellite flavor. Similarly, a Satellite deployment should not expose the Katello flavor (that's why this file fully overrides the `choices`).

I am still a bit torn whether I want this file to live here in the repo at the top level, or to put it as a fixture to the tests and let downstream ship an own version.

We could build `foremanctl-overrides-<vendor>` sub-packages from the files here, but not ship them in our repos, but then let downstreams ship them as needed. That's probably a question to @zjhuntin and @Odilhao how they want the downstream shipping to happen.

arvind4501: I was thinking about `features.d` implementation, where downstream package just need to add the dropin, simiar to that maybe like `src/flavors.d/satellite/metadata.obsah.yml`  drop in could work, but i am not sure how complex is that.
Also we could teach obsah to have a `choices_directory` from where it should pull flavor choices and we could differ that in upstream and downstream, 

these are some thoughts i had, so added them here, but i don't want to block this PR on those as PR already looks neat to me.

ekohl: Going to the extreme: should upstream also be a vendor?

evgeni: Yeah, there are several ways we could make this easier/cleaner if we teach obsah a more explicit way to recognize overrides (right now I am merely abusing the fact that includes work and support partial overrides of previously defined stuff). If you want to experiment with this, please :)

evgeni: > Going to the extreme: should upstream also be a vendor?

Not on a Friday, no. :stuck_out_tongue_closed_eyes:

evgeni:
@coderabbitai review

src/filter_plugins/foremanctl.py
```diff
@@ -72,18 +73,25 @@ def available_foreman_plugins(_value):
...
+            available_list.append((name, 'available', internal, description))
+
+    if not list_internal:
+        output = [f"{'FEATURE':<25} {'STATE':<12} DESCRIPTION"]
+        for name, state, _internal, description in enabled_list + available_list:
+            output.append(f"{name:<25} {state:<12} {description}")
+    else:
+        output = [f"{'FEATURE':<25} {'STATE':<12} {'INTERNAL':<8} DESCRIPTION"]
+        for name, state, internal, description in enabled_list + available_list:
+            output.append(f"{name:<25} {state:<12} {internal:<8} {description}")
```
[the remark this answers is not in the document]
evgeni: The `:<8` already does the int conversion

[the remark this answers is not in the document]
evgeni: Explain:
```
>>> print(f"{False:<8}")
0       
```

.github/workflows/test.yml
arvind4501: https://github.com/theforeman/obsah/pull/135 this is what i had in mind

ehelms: > I am still a bit torn whether I want this file to live here in the repo at the top level, or to put it as a fixture to the tests and let downstream ship an own version.
> 
> We could build `foremanctl-overrides-<vendor>` sub-packages from the files here, but not ship them in our repos, but then let downstreams ship them as needed.

Fixture vs. direct inclusion I am not sure it really matters, so maybe just start with the easiest solution.

As for the sub-package shipping, I don't think we should get into that game yet without more compelling use cases to do so.

Odilhao: > Yeah, there are several ways we could make this easier/cleaner if we teach obsah a more explicit way to recognize overrides (right now I am merely abusing the fact that includes work and support partial overrides of previously defined stuff). If you want to experiment with this, please :)

Everything is a vendor, even the vendor is the vendor. We are in Inception now. https://tenor.com/0xNi.gif

> We could build foremanctl-overrides-<vendor> sub-packages from the files here, but not ship them in our repos, but then let downstreams ship them as needed. That's probably a question to @zjhuntin and @Odilhao how they want the downstream shipping to happen.

I don't have any problem if we need to add one override file at one .d/ folder downstream to ensure that the override is present, since we will always have one `opinionated` package downstream.

arvind4501 (approved):
LGTM

ekohl (approved):
Approving the general design and the features it enables.

src/roles/post_install/tasks/message.yaml
Gauravtalreja1, line 9: Should we also change it as `s/Foreman/Satellite`?

src/vars/flavors/satellite.yml
```diff
@@ -0,0 +1,36 @@
...
+  - candlepin
+  - content/ansible
+  - content/container
+  - content/rpm
+  - ansible
+  - azure-rm
+  - google
+  - remote-execution
+  - rh-cloud
+  - theme-satellite
```
Gauravtalreja1, line 15: To make theme-satellite work, do we still need to have changes in`/usr/share/foremanctl/src/features.d` or am I missing something?

arvind4501: I think it needs `foreman-theme-satellite` package installed in foreman container, and which is not available in upstream images but should be in downstream

evgeni: It is in `features.yaml`: https://github.com/theforeman/foremanctl/pull/726/changes#diff-e37c700aeb814cffa6c1048fa70e5a1b32c4657de49774c80da26983c5b7fde9R94

But you need a container with the plugin actually present for it to work correctly.

src/roles/post_install/tasks/message.yaml
evgeni: We do that in the `satellite-installer`, and we could, but I prefer to defer this for a future time. (The less we brand, the less can break :) )

Gauravtalreja1: Sure, but can we create a GH issue to track this change so we don't miss?

ekohl: I'd say we can ignore it. It really is the Foreman service running there, not Pulp, not Candlepin. A foremanctl (or satellitectl user) should learn to differentiate the parts since they'll also run `foremanctl health` to see individual services.

evgeni: https://github.com/theforeman/foremanctl/issues/748

src/vars/flavors/satellite.yml
```diff
@@ -0,0 +1,36 @@
+---
+flavor_features:
+  - foreman
+  - katello
```
Gauravtalreja1, line 4: I see we're missing `registration` and `templates` features which are enabled by default on SAT today, or is this intentional?

src/roles/post_install/tasks/message.yaml
Gauravtalreja1: thanks!

src/vars/flavors/satellite.yml
evgeni: Not totally.
I missed that the registration feature was merged last week (#675).
Templates was already implemented in #519.

I'll add those.

evgeni: Oh wait, no!

Those are only needed on a capsule, not on the Satellite. And that'll be handled in #735

Gauravtalreja1: I think those should work when user has foreman-proxy enabled on SAT, right?

evgeni: They are not required on a Satellite, as there the client can talk directly to Sat.

You only need them on a Capsule to forward requests for a client.

Gauravtalreja1:
ACK, tested with `foremanctl-3.0.0.develop-4.20260806101125697491.pr726.188.ge09da29.el9.noarch` 🍏 
Scenarios tested:
  - Satellite flavor deployment with vendor overrides -> works
  - Health check passes
  - Feature listing correctly hides/shows enabled features
  - `--add-feature bmc` works; --remove-feature on flavor-provided features is a silent no-op
  - Backup works with feature-based database detection
  - No regressions on katello flavor
 
Issues found: https://github.com/theforeman/foremanctl/issues/748

Gauravtalreja1 (approved):

PR #728, opened 2026-08-05
Move co-located containers onto shared foreman-core-network

Alleny244:
## Problem Statement

Embedded PostgreSQL defaults to `listen_addresses = '*'`, exposing port `5432` on public interfaces when containers use host networking.

https://redhat.atlassian.net/browse/SAT-48503

## Changes

- Add a shared Podman bridge network `foreman-core-network` (`10.130.0.0/24`) via the `foreman_core_network` role, created early in deploy.
- Move co-located containers off host networking onto that network (PostgreSQL, Valkey, Candlepin, Pulp, Foreman/Dynflow, IOP services).
- Reach services by container name on the shared network (e.g. `postgresql`, `valkey`, `candlepin`, `iop-core-gateway`).
- Publish host access on loopback only where needed (e.g. `127.0.0.1:5432`, `127.0.0.1:6379`, Candlepin/IOP gateway ports) so ports are not exposed on `0.0.0.0` / public interfaces.
- Remove the IOP-only `iop_network` / `iop-core-network`; IOP uses `foreman-core-network`.
- Update TLS SANs and Smart Proxy URL so Foreman can reach Candlepin and the IOP gateway by container name.
- Leave `foreman_proxy` on host networking (edge/DHCP/DNS/TFTP/REX).
- Add/update tests to ensure PostgreSQL (and Valkey) are not published on `0.0.0.0:5432` / `[::]:5432` (and equivalent for Valkey).

## Tests

    ./forge test --pytest-args="tests/postgresql_test.py -vv"

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (`docs/iop.md`)


src/roles/iop_core/tasks/main.yaml
```diff
@@ -7,6 +7,34 @@
   ansible.builtin.include_role:
     name: iop_kafka
 
+# Kafka is the first container on iop-core-network, so the gateway IP
```
ehelms: This, as you are discovering, is tricky and the same issue I ran into when I tried to do this before. This creates a rather fragile experience.

ehelms:
Here's two other ideas: 

1. Put all the containers on the same network. 

The `iop` network was created back in the foreman-installer for isolation, to reduce the need for TLS everywhere and because we were learning. Then we copied the idea here to start with parity. However, back when we implemented `iop` we always said this should get easier to manage once everything is in containers because we can normalize.

2. Put postgresql on both networks and have it listen to both.

Alleny244:
> Here's two other ideas:
> 
> 1. Put all the containers on the same network.
> 
> The `iop` network was created back in the foreman-installer for isolation, to reduce the need for TLS everywhere and because we were learning. Then we copied the idea here to start with parity. However, back when we implemented `iop` we always said this should get easier to manage once everything is in containers because we can normalize.
> 
> 2. Put postgresql on both networks and have it listen to both.

I’ll go with PostgreSQL on both networks here. Putting all containers on the same network would require broader changes, so a separate PR would be right?

ehelms:
> I’ll go with PostgreSQL on both networks here. Putting all containers on the same network would require broader changes, so a separate PR would be right?

Yes and no, you are ultimately trying to solve this broader problem of running only what's needed on the public interface. I'd rather us solve the architecture question than patch it and change it again right after.

I was reading on how `pods` accomplish the networking issue, and they create a lightweight "infra" container that basically does nothing but is the first container to always start and is present on the network to ensure it's available. I do not think we should use pods here as that will complicate things, but perhaps we can learn from this idea. The trickiest part would be getting the ordering right. But I do think if they are all on the same network, things get easier and I do not see a reason anymore to split them across networks.

Alleny244:
> > I’ll go with PostgreSQL on both networks here. Putting all containers on the same network would require broader changes, so a separate PR would be right?
> 
> Yes and no, you are ultimately trying to solve this broader problem of running only what's needed on the public interface. I'd rather us solve the architecture question than patch it and change it again right after.
> 
> I was reading on how `pods` accomplish the networking issue, and they create a lightweight "infra" container that basically does nothing but is the first container to always start and is present on the network to ensure it's available. I do not think we should use pods here as that will complicate things, but perhaps we can learn from this idea. The trickiest part would be getting the ordering right. But I do think if they are all on the same network, things get easier and I do not see a reason anymore to split them across networks.

Yeah, makes sense. I’ll look into the single network approach and see how we can handle the startup ordering.

Alleny244:
> > I’ll go with PostgreSQL on both networks here. Putting all containers on the same network would require broader changes, so a separate PR would be right?
> 
> Yes and no, you are ultimately trying to solve this broader problem of running only what's needed on the public interface. I'd rather us solve the architecture question than patch it and change it again right after.
> 
> I was reading on how `pods` accomplish the networking issue, and they create a lightweight "infra" container that basically does nothing but is the first container to always start and is present on the network to ensure it's available. I do not think we should use pods here as that will complicate things, but perhaps we can learn from this idea. The trickiest part would be getting the ordering right. But I do think if they are all on the same network, things get easier and I do not see a reason anymore to split them across networks.

Should I reuse the existing iop-core-network or create a new general purpose network ( if naming conventions are being looked into) for the containers

ehelms:
> Should I reuse the existing iop-core-network or create a new general purpose network ( if naming conventions are being looked into) for the containers

New general purpose network name that is more descriptive.

development/roles/foreman_development/defaults/main.yaml
```diff
@@ -29,11 +29,15 @@ foreman_development_rails_command: "puma -w 2 -p {{ foreman_development_rails_po
 
 foreman_development_url: "http://{{ ansible_facts['fqdn'] }}:{{ foreman_development_rails_port }}"
 
-foreman_development_database_host: "localhost"
+# Prefer 127.0.0.1 over localhost: Postgres is published on IPv4 only, and
+# "localhost" often tries ::1 first (connection refused).
+foreman_development_database_host: "127.0.0.1"
```
stejskalleos, line 32: The commit title says:

> Restrict PostgreSQL listen_addresses to localhost and IOP gateway

But I don't see any change in the `listen_addresses`. Please update the commit message & PR title/description to reflect the current state of the PR.

src/roles/iop_fdw/defaults/main.yaml
```diff
@@ -6,9 +6,12 @@ iop_fdw_remote_database_name: "{{ undef(hint='You must specify the remote FDW da
 iop_fdw_remote_user: "{{ undef(hint='You must specify the remote FDW database user') }}"
 iop_fdw_remote_password: "{{ undef(hint='You must specify the remote FDW database password') }}"
 
-# Optional parameters - can use defaults
-iop_fdw_database_host: "localhost"
+# Optional parameters - can use defaults.
+# Use 127.0.0.1: Postgres is published on IPv4 only (not ::1), and TCP
+# requires a password now that it is no longer on host networking.
+iop_fdw_database_host: "127.0.0.1"
```
stejskalleos: The [iop.md](https://github.com/Alleny244/foremanctl/blob/fix/postgresql-listen-localhost/docs/iop.md) architecture diagram (mermaid) still draws Postgres under "Host System", outside the iop-core-network subgraph, while the prose now says Postgres is reached via postgresql:5432 on that network.

Should we update that as well?

src/roles/iop_fdw/tasks/main.yaml
```diff
@@ -9,12 +9,14 @@
     name: postgres_fdw
     login_db: "{{ iop_fdw_database_name }}"
     login_user: postgres
+    login_password: "{{ iop_fdw_login_password }}"
```
stejskalleos, line 12: This will log the password value in the logs, especially when using `-vvv`. Should we use `no_log: true` or something similar (secrets?) to mask the value of the password?

src/vars/database.yml
```diff
@@ -35,7 +36,8 @@ foreman_database_port: "{{ database_port }}"
 foreman_database_ssl_mode: "{{ database_ssl_mode }}"
 foreman_database_ssl_ca: "{{ database_ssl_ca }}"
 
-iop_database_host: host.containers.internal
+# Co-located Postgres joins iop-core-network; IOP reaches it by container name.
+iop_database_host: postgresql
```
stejskalleos, line 47: Suggestion from AI:

> Pre-existing (not introduced by this PR) architectural gap: `iop_database_host` in `src/vars/database.yml` is unconditionally set to `postgresql` (a container name only resolvable on iop-core-network), with no branching on database_mode. Combining --database-mode external with the iop feature was already questionable before this PR (nothing in features.yaml or obsah metadata blocks that combination), and this PR doesn't change that — just flagging it's still an open gap, not something to necessarily fix in this PR.

src/vars/database.yml
```diff
@@ -101,42 +103,45 @@ databases:
...
   - name: iop_vulnerability
     database: "{{ iop_vulnerability_database_name }}"
-    host: "{{ iop_vulnerability_database_host }}"
+    host: "{{ database_host }}"
     port: "{{ iop_vulnerability_database_port }}"
     user: "{{ iop_vulnerability_database_user }}"
     password: "{{ iop_vulnerability_database_password }}"
     feature: iop
 
+
```
stejskalleos:
```suggestion
```

src/roles/postgresql/tasks/main.yml
```diff
@@ -44,7 +45,6 @@
         PartOf=foreman.target
   notify:
     - Restart postgresql
```
stejskalleos, line 49: Why remove the line?

```suggestion
    - Restart postgresql

```

stejskalleos (changes requested):

development/roles/foreman_development/defaults/main.yaml
Alleny244: Yeah, sure. I’ll update it. Also, the current implementation was based on putting PostgreSQL on both networks. Based on the discussion above, we decided to move forward with creating a new network and placing all the containers on it instead. I’ve been working on that implementation and will push the changes shortly.

src/vars/database.yml
Alleny244: Removed

pablomh:
Heads up: this overlaps significantly with #403 ("Switch container networking from host to named bridge networks"), which just came out of draft and is ready for review. Both PRs solve the same underlying problem (services listening/publishing on the public interface under `network: host`) by moving containers onto Podman bridge networks, but with materially different architectures, and touch almost the exact same set of files (`candlepin`, `foreman`, `postgresql`, `pulp`, `valkey`, most `iop_*` roles, `certificates/tasks/issue.yml`, the `check_*` roles, `vars/base.yaml`, `vars/database.yml`, and the corresponding tests). Landing both as-is isn't really possible — whichever merges second will need a substantial rebase/rework, if not an outright redesign.

The key architectural difference, for whoever needs to reconcile these:

- **This PR**: one shared `foreman-core-network` (`10.130.0.0/24`) for everything co-located (Foreman, Postgres, Valkey, Candlepin, Pulp, IOP), with `foreman_proxy` left on host networking, and host access to Postgres/Valkey/Candlepin/the IOP gateway restored via `127.0.0.1`-only publish. `iop_network`/`iop-core-network` is removed and folded into the shared network.
- **#403**: multiple purpose-scoped bridge networks — `foreman-db` (isolated, Postgres-only), `foreman-cache` (isolated, Valkey-only), `foreman-app` (Foreman/Candlepin/Pulp/httpd), `foreman-proxy` (moves the smart proxy off host networking too, when the feature is enabled), and a dedicated `iop-core-network` (dual-homed gateway onto `foreman-app` + the IOP network). Postgres/Valkey are not published to the host at all in the internal-database case; internal-network Postgres access goes through the Unix socket instead of TCP.

I saw the discussion above converged on "put everything on one network, don't see a reason to split them" — worth being aware that #403 deliberately went the other way (isolating the DB/cache tier from the app tier, and not publishing DB/cache ports even on loopback) before that gets baked in further here. Given the amount of surface area both PRs touch, it'd be good for a maintainer to pick one direction (or explicitly merge the best of both) before either goes further, rather than have both mature in parallel and conflict at merge time.

src/roles/postgresql/tasks/main.yml
Alleny244: oh my bad, fixed. I didnt notice that

src/roles/iop_fdw/defaults/main.yaml
Alleny244: Yeah, it's updated now

src/roles/iop_fdw/tasks/main.yaml
Alleny244: Yes, added no_log property here

stejskalleos:
> Given the amount of surface area both PRs touch, it'd be good for a maintainer to pick one direction (or explicitly merge the best of both) before either goes further, rather than have both mature in parallel and conflict at merge time.

@ehelms @ekohl we need a decision on which way (PR) we want to go.

src/roles/certificates/tasks/issue.yml
```diff
@@ -1,7 +1,27 @@
 ---
 - name: Issue server certificate
   when:
-    - (certificates_source != 'custom_server') or (certificates_hostname == 'localhost')
+    # localhost/candlepin are always issued (service-name TLS on the shared network).
```
stejskalleos: I don't understand why these changes are necessary for this PR. Can you please explain?

Alleny244: Since the services are now on the same network, they communicate using the container names instead of localhost. So we need to add those names to the certificate SANs as well, else TLS hostname verification will fail, right?

stejskalleos: make sense, thx.

ekohl:
I'd like to avoid introducing additional networks at this point. It's a major architectural change and currently I'm much more interested in actually having feature parity with our existing installation. That will allow us to deprecate the package-based installations, which in turn unlocks dropping Ruby 3.1 support, which in turn allows upgrading Ruby on Rails to at least Rails 7.2. That is needed to get CVE fixes.

At this point in the release cycle even this change is IMHO very risky. If we're going to introduce networks, I think we should do it properly and well thought out.

IMHO neither this nor https://github.com/theforeman/foremanctl/pull/403 do what I would expect: have 1 network per service.

When I say service, I mean:
* Candlepin
* Foreman
* Foreman Proxy
* Pulp
* IoP

But then there are more complexities, like do you run 1 instance of Valkey for both Foreman and Pulp or isolate them from each other? PostgreSQL is in a similar boat.

pablomh:
On "1 network per service" — I think the tricky part is exactly the Valkey/Postgres question you raised.

If Foreman, Candlepin and Pulp all need the same Postgres, there are really only two ways to handle it: hook Postgres up to all three networks, or give each service its own Postgres.

The multi-homing option doesn't end up buying much isolation though — Postgres is reachable from all three networks either way, so "who can reach the DB" looks basically the same as just having one `foreman-db` network with those three on it. Same picture, just more bridges to draw it.

Separate Postgres/Valkey per service would give real isolation, but that feels like a much bigger change than what we're actually trying to fix (containers listening on `0.0.0.0`).

That's the thinking behind going with networks per *workflow* rather than per named service: `foreman-db`/`foreman-cache` are the DB/cache-access boundary — one shared instance, one network, and membership is controlled per consumer. `foreman-app` covers Foreman/Candlepin/Pulp/httpd, since those already need to talk to each other directly regardless of how many networks exist. And `foreman-proxy` is split off on its own, since it's the one piece here talking to arbitrary remote hosts and probably shouldn't have a path to the database.

That proxy split feels like the boundary that matters most. The rest ends up as one network because those services are already tightly coupled, and splitting them further wouldn't really stop anything from reaching anything — just add more bridges for the same result.

ekohl:
> And `foreman-proxy` is split off on its own, since it's the one piece here talking to arbitrary remote hosts and probably shouldn't have a path to the database.

https://github.com/theforeman/foremanctl/pull/611 would disagree with you.

ehelms:
I'll give both PRs a re-read, but the thing I am missing is *why* or what value individual networks per service gives us. 

Networks aren't for free, they will require our users to ensure that address space is available and the more introduce the more we potentially clash with any existing configuration they might have. We should be thoughtful about this trade-off when considering what additional networks bring us in value given most of our services run locally.

The biggest reason I pushed for a single, custom podman network is to allow easy container-to-container DNS name resolution as this is not available on the default podman network.

ekohl:
> I'll give both PRs a re-read, but the thing I am missing is _why_ or what value individual networks per service gives us.

My initial thought was that you would have a good isolation. However, given all the cross-service talk it's questionable how much benefit we would actually get.

> Networks aren't for free, they will require our users to ensure that address space is available and the more introduce the more we potentially clash with any existing configuration they might have. We should be thoughtful about this trade-off when considering what additional networks bring us in value given most of our services run locally.

If you make the network IPv6-only then the address space is less of a concern. Also note that with systemd socket activation for Foreman and Pulp they don't have any ingress, only egress.

But sure, it's also additional complexity that can break.

development/roles/foreman_development/defaults/main.yaml
```diff
@@ -29,11 +29,13 @@ foreman_development_rails_command: "puma -w 2 -p {{ foreman_development_rails_po
...
 
-foreman_development_database_host: "localhost"
+foreman_development_database_host: "127.0.0.1"
 foreman_development_database_port: 5432
 foreman_development_database_name: "foreman_development"
 foreman_development_database_user: "foreman"
-foreman_development_database_password: "foreman" # noqa: no-static-secrets
+# Must match the DB role password set by the postgresql role (random from state file).
+# With bridge+publish, only TCP auth is available — a hardcoded password no longer works.
+foreman_development_database_password: "{{ foreman_database_password }}"
```
ehelms: This is using a role variable from another role.

src/roles/candlepin/defaults/main.yml
```diff
@@ -13,7 +13,14 @@ candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "4.4.14"
 candlepin_secret_mount_opts: "mode=0440,uid=0,gid=53,type=mount"
 
-candlepin_database_host: localhost
+
+candlepin_networks:
+  - "{{ foreman_core_network_name | default('foreman-core-network') }}"
+candlepin_publish:
+  - "127.0.0.1:{{ candlepin_ssl_port }}:{{ candlepin_ssl_port }}"
```
ehelms: Given they are all on the same network, do we need to publish Candlepin? Shouldn't Foreman just be able to access it directly now using the container name?

src/roles/foreman/defaults/main.yaml
```diff
@@ -5,13 +5,21 @@ foreman_container_name: foreman
 
 foreman_database_name: foreman
 foreman_database_user: foreman
-foreman_database_host: localhost
+# On the shared network, reach Postgres by container name (not host loopback).
+foreman_database_host: "{{ 'postgresql' if (database_mode | default('internal') == 'internal') else database_host }}"
```
ehelms: `database_host` doesn't exist in this role, the default should juts be `postgresql`

src/roles/foreman/defaults/main.yaml
```diff
@@ -55,24 +63,7 @@ foreman_env:
   FOREMAN_PUMA_WORKERS: "{{ foreman_puma_workers }}"
 
 foreman_dynflow_extra_env:
-  DYNFLOW_REDIS_URL: "redis://localhost:6379/6"
+  DYNFLOW_REDIS_URL: "{{ foreman_dynflow_redis_url }}"
```
ehelms: Why do we need another variable to represent this? Could it just update to `redis://valkey:6379/6` ? Similar question for the other places the code is now doing this.

src/roles/foreman/defaults/main.yaml
```diff
@@ -5,13 +5,21 @@ foreman_container_name: foreman
...
+foreman_database_host: "{{ 'postgresql' if (database_mode | default('internal') == 'internal') else database_host }}"
 foreman_database_port: 5432
 foreman_database_pool: 9
 foreman_database_ssl_mode: disable
 foreman_database_ssl_ca: # noqa: no-empty-defaults
 foreman_database_ssl_ca_path: /etc/foreman/db-ca.crt
 
+
+foreman_networks:
+  - "{{ foreman_core_network_name | default('foreman-core-network') }}"
```
ehelms: This is a variable outside the role. We can consider making it a role-wide variable. I do not think the `| default(` is needed.

src/roles/foreman/tasks/main.yaml
```diff
@@ -113,11 +113,21 @@
     image: foreman.image
     state: quadlet
     sdnotify: true
-    network: host
+    network: "{{ foreman_networks if foreman_networks else omit }}"
```
ehelms:
```suggestion
    network: "{{ foreman_networks }}"
```

I do not think we should omit this. Same for other places.

src/roles/foreman/templates/katello.yaml.j2
```diff
@@ -3,7 +3,7 @@
   :rest_client_timeout: 3600
 
   :candlepin:
-    :url: https://localhost:23443/candlepin
+    :url: {{ foreman_candlepin_url }}
```
ehelms: Why does this need to be a variable now?

src/roles/iop_advisor/defaults/main.yaml
```diff
@@ -5,5 +5,5 @@ iop_advisor_container_tag: "foreman-3.18"
 iop_advisor_database_name: advisor_db
 iop_advisor_database_user: advisor_user
 iop_advisor_database_password: "{{ undef(hint='Set a secure database password') }}"
-iop_advisor_database_host: host.containers.internal
+iop_advisor_database_host: "{{ iop_database_host | default('postgresql') }}"
```
ehelms:
```suggestion
iop_advisor_database_host:  postgresql
```

I think you get the pattern by now. While I get the temptation to use a variable, the whole point of defaults is to default to something that is meaningful. And not to be reliant on some variable to exist somewhere external.

src/roles/iop_fdw/defaults/main.yaml
ehelms: Why do some places use 127.0.0.1 and others use `postgresql` ?

src/roles/postgresql/defaults/main.yml
```diff
@@ -13,4 +12,9 @@ postgresql_max_connections: 500
 postgresql_shared_buffers: 512MB
 postgresql_effective_cache_size: 1GB
 
+postgresql_networks:
+  - "{{ foreman_core_network_name | default('foreman-core-network') }}"
+postgresql_publish:
+  - "127.0.0.1:5432:5432"
```
ehelms, line 18: Does this need to be published if everything is on the network that needs it?

ehelms:
This raises the question of:

- Do we need TLS for Candlepin?
- Do we need TLS for the iop-gateway?

These are both services that are only internally communicated with by other services on the network aka the foreman container. We currently do not have TLS between any of the other services (e.g. iop-*, postgresql, redis).

development/roles/foreman_development/defaults/main.yaml
Alleny244: Yeah, deploy-dev playbook already include  these variables and is passed as a variable in foreman_development role so we can drop it from here

src/roles/candlepin/defaults/main.yml
Alleny244: For production deployment this scenario works, but for dev (foreman_development) foreman runs on the host , so it still need port publishing right?

src/roles/foreman/defaults/main.yaml
Alleny244: yeah, got it. I ll wire in external mode in deploy playbook  by passing foreman_database_host: "{{ database_host }}" when database_mode == 'external'

src/roles/foreman/defaults/main.yaml
Alleny244: Yeah , sure

src/roles/foreman/defaults/main.yaml
Alleny244: Yeah sure, makes sense

src/roles/foreman/tasks/main.yaml
Alleny244: yeah , sure

src/roles/foreman/templates/katello.yaml.j2
Alleny244: yeah, will drop the variable here

src/roles/iop_advisor/defaults/main.yaml
Alleny244: yeah agreed, have updated all to use default postgresql

src/roles/postgresql/defaults/main.yml
Alleny244: For host side consumers, they still require port publication right?

Alleny244:
> This raises the question of:
> 
> * Do we need TLS for Candlepin?
> * Do we need TLS for the iop-gateway?
> 
> These are both services that are only internally communicated with by other services on the network aka the foreman container. We currently do not have TLS between any of the other services (e.g. iop-*, postgresql, redis).

TLS here is mainly because Candlepin is currently configured to use HTTPS, and the IOP gateway is used as a Smart Proxy, which Foreman expects to access over HTTPS right?. It’s not required by the shared network itself

src/roles/postgresql/defaults/main.yml
ehelms: What host side consumers are you expecting to exist?

ehelms:
> TLS here is mainly because Candlepin is currently configured to use HTTPS, and the IOP gateway is used as a Smart Proxy, which Foreman expects to access over HTTPS right?. It’s not required by the shared network itself

Fair point on the IOP gateway.
My point with Candlepin is -- should it continue to be? I suppose we do not have to answer that now, but it's a valid question to consider. We do not, for example, lock down postgresql over TLS. 

The dedicated network does offer some security enhancements.

src/roles/postgresql/defaults/main.yml
Alleny244: Mainly for deploy-dev workflow which runs forman on the host itself

Alleny244:
> > TLS here is mainly because Candlepin is currently configured to use HTTPS, and the IOP gateway is used as a Smart Proxy, which Foreman expects to access over HTTPS right?. It’s not required by the shared network itself
> 
> Fair point on the IOP gateway. My point with Candlepin is -- should it continue to be? I suppose we do not have to answer that now, but it's a valid question to consider. We do not, for example, lock down postgresql over TLS.
> 
> The dedicated network does offer some security enhancements.

yeah maybe we could leave TLS posure as a follow up?

src/roles/candlepin/defaults/main.yml
ehelms: yes, but we would want to only do that in the development setup

src/roles/certificates/tasks/issue.yml
ehelms: Why does this role need to know this though? Just like localhost, can `candlepin` not be specified in the list of hosts in the playbook to generate certificates for?

ehelms: Here -- https://github.com/theforeman/foremanctl/blob/master/src/vars/base.yaml#L2

src/roles/candlepin/defaults/main.yml
Alleny244: oh yeah, agreed. I ll change the scope here , I think some tests are relying on localhost for testing connectivity, I ll update those too

src/roles/postgresql/defaults/main.yml
Alleny244: Just a doubt here, please correct me if I’m wrong, since the Ansible task `community.postgresql.postgresql_user` uses `login_host: 127.0.0.1`, doesn't PostgreSQL need to have its port published on the host  for this task to work?

src/roles/certificates/tasks/issue.yml
Alleny244: Oh yes, it's updated now

src/roles/postgresql/defaults/main.yml
ehelms: Fair point, I suppose that is the best we can do right now.

ehelms:
Can you introduce a new document that lays out the architecture of the network? Take a look at https://github.com/theforeman/foremanctl/pull/771 as I'd like it to go into that folder structure.

Alleny244:
> Can you introduce a new document that lays out the architecture of the network? Take a look at #771 as I'd like it to go into that folder structure.

Sure I ll take a look into that and formulate one

Alleny244:
One thing that I noticed was , TLS issue affects both Candlepin and IOP: their certs are issued for candlepin / iop-core-gateway, but deploy-dev Foreman connects via localhost.  Currently we are using  those hostnames instead of localhost so hostname verification succeeds for production. For the host Foreman, I guess we’d also need to add both names to /etc/hosts pointing to 127.0.0.1 for TLS

ehelms:
> One thing that I noticed was , TLS issue affects both Candlepin and IOP: their certs are issued for candlepin / iop-core-gateway, but deploy-dev Foreman connects via localhost. Currently we are using those hostnames instead of localhost so hostname verification succeeds for production. For the host Foreman, I guess we’d also need to add both names to /etc/hosts pointing to 127.0.0.1 for TLS

For devel setup, could the certificates be issued for localhost as they are today?

development/playbooks/deploy-dev/deploy-dev.yaml
```diff
@@ -5,6 +5,14 @@
...
     flavor: katello
     pulp_register_foreman_proxy: false
+    postgresql_publish:
+      - "127.0.0.1:5432:5432"
+    valkey_publish:
+      - "127.0.0.1:6379:6379"
+    candlepin_publish:
+      - "127.0.0.1:23443:23443"
+
+
```
ehelms:
```suggestion
```

src/roles/candlepin/defaults/main.yml
```diff
@@ -10,7 +10,12 @@ candlepin_container_image: quay.io/foreman/candlepin
 candlepin_container_tag: "4.4.14"
 candlepin_secret_mount_opts: "mode=0440,uid=0,gid=53,type=mount"
 
-candlepin_database_host: localhost
+
+candlepin_networks:
+  - foreman-core-network
+
+# On the shared network, reach Postgres by container name (not host loopback).
```
ehelms: I think this can be dropped.

development/playbooks/deploy-dev/deploy-dev.yaml
Alleny244: Reverted

Alleny244:
> > One thing that I noticed was , TLS issue affects both Candlepin and IOP: their certs are issued for candlepin / iop-core-gateway, but deploy-dev Foreman connects via localhost. Currently we are using those hostnames instead of localhost so hostname verification succeeds for production. For the host Foreman, I guess we’d also need to add both names to /etc/hosts pointing to 127.0.0.1 for TLS
> 
> For devel setup, could the certificates be issued for localhost as they are today?

Yeah it is covered now. The Candlepin and IOP certs include both the container names (candlepin / iop-core-gateway) and localhost as SANs. So production can connect using the container names, while devel setup can continue using localhost (no /etc/hosts changes are needed)

src/roles/iop_fdw/defaults/main.yaml
stejskalleos: @Alleny244 this should be addressed, as you use both in IOP.md:

```
flowchart LR
    Ansible["Ansible / tests on host"] -->|"127.0.0.1:5432<br/>published"| PG
    Apps["Foreman, Pulp, IOP apps"] -->|"postgresql:5432<br/>bridge DNS"| PG
    FDW["postgres_fdw inside PG"] -->|"127.0.0.1:5432<br/>container loopback"| PG
    PG[(postgresql container)]
```

src/vars/database.yml
stejskalleos: @Alleny244, are we planning to fix it? If not, can we at least create a GH issue to track it?

tests/feature/katello/candlepin_test.py
```diff
@@ -27,20 +27,33 @@ def test_candlepin_runs_as_tomcat(server):
...
     assert status.stdout == '200'
 
 
-def test_candlepin_logs_in_journal(server, certificates):
+def test_candlepin_logs_in_journal(server):
     server.run(
-        f"curl --cacert {certificates['ca_certificate']} --silent --output /dev/null "
-        f"https://localhost:23443/candlepin/status"
+        "podman exec foreman curl --cacert /etc/foreman/katello-default-ca.crt"
+        "--silent --output /dev/null https://candlepin:23443/candlepin/status"
```
stejskalleos:
```suggestion
        "podman exec foreman curl --cacert /etc/foreman/katello-default-ca.crt "
        "--silent --output /dev/null https://candlepin:23443/candlepin/status"
```

missing space, this can't work IMO

tests/feature/katello/candlepin_test.py
```diff
@@ -50,8 +63,8 @@ def test_candlepin_logs_in_journal(server, certificates):
 
 def test_candlepin_tomcat_logs_in_journal(server, certificates):
     server.run(
-        f"curl --cacert {certificates['ca_certificate']} --silent --output /dev/null "
-        f"https://localhost:23443/candlepin/status"
+        f"podman exec foreman curl --cacert {certificates['ca_certificate']} "
```
stejskalleos: From AI:

certificates['ca_certificate'] resolves to a host-side path (rendered from src/vars/certificates.yml, e.g.,/var/lib/foremanctl/certs/certs/ca.crt), but this is run with podman exec foreman, i.e., inside the foreman container's filesystem, where that path doesn't exist. 

The sibling test test_candlepin_status correctly uses the in-container path /etc/foreman/katello-default-ca.crt — this one should too.

stejskalleos (changes requested):
Some things has to be addressed before we can merge.

Also, the failing CI:
```
TASK [pre_install : Install other dependencies] ********************************
[ERROR]: Task failed: Module failed: Depsolve Error occurred: 
 Problem: package python3.12-pyOpenSSL-26.2.0-2.el10.noarch from pulpcore requires (python3.12dist(cryptography) < 49~~ with python3.12dist(cryptography) >= 46), but none of the providers can be installed
  - package python3.12-requests-2.33.1-3.el10.noarch from pulpcore requires python3.12-pyOpenSSL >= 0.14, but none of the providers can be installed
  - cannot install both python3-cryptography-49.0.0-1.el10.x86_64 from baseos and python3-cryptography-48.0.0-3.el10.x86_64 from baseos
  - package python3-cryptography-49.0.0-1.el10.x86_64 from baseos obsoletes python3.12-cryptography < 49.0.0-1.el10 provided by python3.12-cryptography-46.0.7-3.el10.x86_64 from pulpcore
  - 
  ```
do we track it somewhere? Or is there a fix / person working on it?

File: src/playbooks/deploy/deploy.yaml [hunk 22]
```diff
@@ -15,6 +15,13 @@
     - "../../vars/foreman.yml"
     - "../../vars/logging.yml"
     - "../../vars/base.yaml"
+  pre_tasks:
+    - name: Set external database hosts
+      ansible.builtin.set_fact:
+        foreman_database_host: "{{ database_host }}"
+        candlepin_database_host: "{{ database_host }}"
+        pulp_database_host: "{{ database_host }}"
+      when: database_mode == 'external'
```
ehelms: Why is this needed here now? If the user provides the parameter this should cascade to modifying these in the `src/vars/database.yml` file.

src/playbooks/deploy-proxy/deploy-proxy.yaml
```diff
@@ -14,6 +14,11 @@
     - "../../vars/database.yml"
     - "../../vars/logging.yml"
     - "../../vars/base.yaml"
+  pre_tasks:
+    - name: Set external database hosts
+      ansible.builtin.set_fact:
+        pulp_database_host: "{{ database_host }}"
+      when: database_mode == 'external'
```
ehelms: We do not support external database with the `deploy-proxy` action.

src/vars/base.yaml
```diff
@@ -13,8 +20,8 @@ certificates_oauth_directory: /var/lib/foremanctl/oauth
 
 candlepin_ca_key: "{{ ca_key }}"
 candlepin_ca_certificate: "{{ ca_certificate }}"
-candlepin_tomcat_key: "{{ localhost_key }}"
-candlepin_tomcat_certificate: "{{ localhost_certificate }}"
+candlepin_tomcat_key: "{{ candlepin_key }}"
```
ehelms: Trying to remember why we set any of these certificate things here and not in `src/vars/certificates.yaml`.

src/vars/certificates.yml
```diff
@@ -14,9 +14,11 @@ localhost_key: "{{ certificates_ca_directory }}/private/localhost.key"
 localhost_certificate: "{{ certificates_ca_directory }}/certs/localhost.crt"
 localhost_client_key: "{{ certificates_ca_directory }}/private/localhost-client.key"
 localhost_client_certificate: "{{ certificates_ca_directory }}/certs/localhost-client.crt"
+candlepin_key: "{{ certificates_ca_directory }}/private/candlepin.key"
+candlepin_certificate: "{{ certificates_ca_directory }}/certs/candlepin.crt"
```
ehelms: I think these are a bit confusing to name this way and not use `tomcat` in the name. Since these are for tomcat whereas Candlepin itself does get key/cert that are CAs for it's other use cases. I would align on `tomcat` naming. And see https://github.com/theforeman/foremanctl/pull/728/changes#r3872587738 I think we need some alignment of certs vars into the same file.

src/vars/database.yml
```diff
@@ -28,22 +28,19 @@ foreman_proxy_container_gateway_db_password: >-
 postgresql_admin_password_file: "{{ obsah_state_path }}/postgresql-admin-password"
 postgresql_admin_password: "{{ lookup('ansible.builtin.password', postgresql_admin_password_file, chars=['ascii_letters', 'digits']) }}"
 
-candlepin_database_host: "{{ database_host }}"
```
ehelms, line 31: Why are we dropping these here?

src/vars/database.yml
```diff
@@ -1,5 +1,5 @@
 ---
-database_host: localhost
+database_host: 127.0.0.1
```
ehelms: I would have thought this would be `postgresql` now?

Alleny244: Shouldn't this stay as `127.0.0.1`? `database_host` is used for host side tasks(login_host, pg_isready), while the containers use their own `*_database_host` defaults with `postgresql` on `foreman-core-network`.

src/vars/database.yml
Alleny244: Those variables used to point to localhost but with new shared network I thought of keep them under their respective roles default variables, but based on this feedback(https://github.com/theforeman/foremanctl/pull/728#discussion_r3872556214) I think it is better to have it in the initial place itself

src/playbooks/deploy/deploy.yaml
Alleny244: yeah agreed, have moved it back to the database.yml file

src/playbooks/deploy-proxy/deploy-proxy.yaml
Alleny244: Have removed it

src/vars/certificates.yml
Alleny244: yeahh, makes sense have made the change

tests/feature/katello/candlepin_test.py
Alleny244: yeah, thanks, have added the space

tests/feature/katello/candlepin_test.py
Alleny244: Yeahh, changed!!

src/roles/iop_fdw/defaults/main.yaml
Alleny244: Thanks, addressed this in the docs. The difference is the network context: host-side Ansible/tests use the published `127.0.0.1:5432`, while the app containers use `postgresql:5432` through the bridge. The FDW runs inside the PostgreSQL container, so its `127.0.0.1` refers to the PostgreSQL container itself. I’ve added a note and linked the “PostgreSQL from three vantage points” section.

src/vars/database.yml
Alleny244: IOP is internal-only, so `iop_database_host: postgresql` is intentional. `deploy.yaml` already skips `iop_core` for external DB mode; the remaining issue is that the CLI accepts the combination and skips it silently. I think opening a separate issue to make it fail explicitly would be better?

stejskalleos: > I think opening a separate issue to make it fail explicitly would be better?

Fine by me. Feel free to raise an issue in GH, and link it here when resolving the comment :)

stejskalleos (approved):
My comments have been addressed; running foremanctl deploy works fine, and CI is green.

🍏 **LGTM** from my side, waiting for @ehelms for final ack

src/vars/database.yml
Alleny244: Yeah sure, I have created a new issue for the same
https://github.com/theforeman/foremanctl/issues/801

src/vars/database.yml
ehelms: That is a fair question as `database_host` now has two meanings with this change. I believe it should mean "this is the `host` that services find that database at". In an external database scenario, this should represent the host of the external database. I would turn it around then, and consider if a new variable is needed to represent the host where database tooling connects to.

ehelms: Maybe like `database_management_host` ?

Alleny244: Yup, this makes sense. `database_host` should  represent the host that the services connect to, and we can use `database_management_host` for host side database tooling.  I ll update the pr

Alleny244: Done!!

stejskalleos (approved):
🍏 **LGTM**

All comments have been addressed; the CI is green. The only thing missing is an ack from @ehelms .

src/vars/database.yml
```diff
@@ -28,22 +29,22 @@ foreman_proxy_container_gateway_db_password: >-
 postgresql_admin_password_file: "{{ obsah_state_path }}/postgresql-admin-password"
 postgresql_admin_password: "{{ lookup('ansible.builtin.password', postgresql_admin_password_file, chars=['ascii_letters', 'digits']) }}"
 
-candlepin_database_host: "{{ database_host }}"
+candlepin_database_host: "{{ database_host if database_mode == 'external' else 'postgresql' }}"
```
ehelms: I think all of these lines can now just be `{{ database_host }}` with your introduction of the `database_management_host` ?

ehelms:
Just this one little nit to ensure the code reads clean -- https://github.com/theforeman/foremanctl/pull/728#discussion_r3969398992 

The rest of the changes look good!

src/vars/database.yml
Alleny244: No, it would fail for migration, database_host can be localhost after migration, so if we are omitting it, this would break

ehelms: > No, it would fail for migration, database_host can be localhost after migration, so if we are omitting it, this would break

I do not understand. After what migration?

Alleny244: I mean `foremanctl migrate --apply` , it can leave `database_host: localhost` in `parameters.yaml`. So using `{{ database_host }}` directly can break migrated installs, right?

ehelms: The migration should not attempt to migrate a database host that is set to localhost. That indicates the internal database is being used and we should prefer our default values so that everything get's setup and configured properly with postgresql running in a container and on the foreman network. 

Right now, I think that migrate logic tries to migrate the database host parameter regardless. However, I think that should change.

Alleny244: Yeah got it, have updated that part now

ehelms (approved):

src/roles/foreman_core_network/tasks/main.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Create foreman-core-network
```
vkrizan, line 2: @ehelms @Alleny244 Do we have an upgrade path from iop network to core?

When I've run it on previous changes this one fails:
```
TASK [foreman_core_network : Create foreman-core-network] *****************************************************************************************************************************************************************************************************
    changed: false
    msg: Can't create network foreman-core-network
    stderr: |-
        Error: subnet 10.130.0.0/24 is already used on the host or by another config
    stderr_lines: <omitted>
    stdout: ''
    stdout_lines: <omitted>
```

ekohl: Also reported in https://community.theforeman.org/t/foremanctl-deploy-encounters-an-error-subnet-10-130-0-0-24-is-already-used-on-the-host-or-by-another-config/47651

ehelms: Upgrade is a bit different than the thread. Let's let the thread play out. As for upgrades, @Alleny244 I think we may need to delete the iop network before we create the new one.

Alleny244: Yeah makes sense , I’ll add cleanup of the old network before creating foreman-core-network

ehelms: I opened https://github.com/theforeman/foremanctl/actions/runs/35116535804/job/104863270280?pr=839 to add iop to the upgrades as a feature, this highlights it and would have shown it before.

ekohl: > Yeah makes sense , I’ll add cleanup of the old network before creating foreman-core-network

Should we have used https://docs.podman.io/en/latest/markdown/podman-network.unit.5.html#networkdeleteonstop-true-defaults-to-false for it? In the overall description it also mentions:

> Please note that stopping the corresponding service will not remove the Podman network by default - this can be overridden by the NetworkDeleteOnStop=true option.

Alleny244: > > Yeah makes sense , I’ll add cleanup of the old network before creating foreman-core-network
> 
> Should we have used https://docs.podman.io/en/latest/markdown/podman-network.unit.5.html#networkdeleteonstop-true-defaults-to-false for it? In the overall description it also mentions:
> 
> > Please note that stopping the corresponding service will not remove the Podman network by default - this can be overridden by the NetworkDeleteOnStop=true option.

But this setting applies to podman network managed by quadlets , but in our case it's not quadlet

Alleny244: PR for upgrade issue https://github.com/theforeman/foremanctl/pull/840

PR #734, opened 2026-08-10
Fix Pulp volume ownership for non-root containers

archanaserver:
Set ownership of Pulp storage directories to UID 700 to support the non-root Pulp container image. This ensures the container running as UID 700 can write to bind-mounted volumes.

Fixes: SAT-47249

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -7,13 +7,19 @@
     path: "{{ item | split(':') | first }}"
     state: directory
     mode: "0755"
+    owner: "700"
+    group: "700"
+    recurse: true
   loop: "{{ pulp_volumes }}"
 
 - name: Create Pulp storage subdirs
   ansible.builtin.file:
-    path: "/var/lib/pulp/{{ item }}"
+    path: "{{ pulp_storage_path }}/{{ item }}"
     state: directory
     mode: "0755"
+    owner: "700"
+    group: "700"
+    recurse: true
   loop:
     - tmp
     - assets
@@ -24,6 +30,9 @@
     path: "{{ item }}"
     state: directory
     mode: "0755"
+    owner: "700"
+    group: "700"
+    recurse: true
   loop: "{{ pulp_default_import_paths + pulp_import_paths }}"
   when: not pulp_mirror
 
@@ -32,6 +41,9 @@
     path: "{{ item }}"
     state: directory
     mode: "0755"
+    owner: "700"
+    group: "700"
+    recurse: true
   loop: "{{ pulp_default_export_paths + pulp_export_paths }}"
   when: not pulp_mirror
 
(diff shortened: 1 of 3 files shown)


src/roles/pulp/tasks/main.yaml
```diff
@@ -35,6 +35,13 @@
   loop: "{{ pulp_default_export_paths + pulp_export_paths }}"
   when: not pulp_mirror
 
+- name: Set ownership of Pulp directories for non-root container
+  ansible.builtin.file:
+    path: "{{ pulp_storage_path }}"
```
evgeni: I don't think that doing a recursive run on all of `pulp_storage_path` on every deploy run is a good idea, as with increasing amount of data in there, this will take longer and longer.

src/roles/restore/tasks/restore_pulp_content.yaml
```diff
@@ -24,6 +24,13 @@
         dest: "{{ pulp_storage_path }}"
         remote_src: true
 
+    - name: Fix ownership after restore for non-root containers
```
evgeni, line 32: Do we need this? Shouldn't the permissions be correct in the backup already?

evgeni (changes requested):

stejskalleos:
Failing CI is caused by the changes in this PR.

@arvind4501 I remember you had a presentation on how to tackle the issue, proposing several strategies how to resolve the issue. If I remember correctly, changing the chmod to `0700` is not enough, right?

I had the same issue in my TFTP feature, where there was a collision between permissions for the quadlet and the services running inside it. See https://github.com/theforeman/foremanctl/pull/533#discussion_r3441949280

arvind4501:
> @arvind4501 I remember you had a presentation on how to tackle the issue, proposing several strategies how to resolve the issue. If I remember correctly, changing the chmod to 0700 is not enough, right?


I think since pulp runs as `pulp` user with` UID =700` (https://github.com/theforeman/pulp-oci-images/pull/76/changes#diff-212aba00ec12bfbcbd949fcee12b42af8e5648c5a624c28094a12b4dc3f69e99R14), chowing the mounted directory would rightly give pulp user inside container, access to the host directory.

but having said that we also need think about the migration path for 3.19  containers(which runs as root), as `/var/lib/pulp` directory on host needs to chowned to 700 before starting the pulp container to work with mounted volume directory

stejskalleos:
Hardcoded user ID number has been pointed out as a potential issue in [this comment](https://github.com/theforeman/foremanctl/pull/533#discussion_r3461914893)

archanaserver:
> 
> but having said that we also need think about the migration path for 3.19 containers(which runs as root), as `/var/lib/pulp` directory on host needs to chowned to 700 before starting the pulp container to work with mounted volume directory

Yes that's a criticial I see, the host `/var/lib/pulp` directory must be owned by 700:700 before starting the non-root container, otherwise bind mounts fail. but for the existing deployments, ownership is fixed on the next `foremanctl deploy` run (dir are recreated with correct ownership during the directory creation tasks). WDYT?

archanaserver:
> Hardcoded user ID number has been pointed out as a potential issue in [this comment](https://github.com/theforeman/foremanctl/pull/533#discussion_r3461914893)

So the the TFTP collision you referenced was about choosing an already-system-allocated UID. UID 700 is in the system range and is already reserved by upstream Pulp, so there's no collision risk I believe, correct me if I'm pointing wrong here.

Umm so is there a specific concern with 700, or would you recommend a different approach we can look for?

src/roles/restore/tasks/restore_pulp_content.yaml
archanaserver: Guess that's the problem. When restoring a backup files are correctly extracted as root:root. But the new container runs as UID 700 and can't access them.

So now I've made it conditional and it only runs when files are owned by `UID 0` (from old root-based backups). If they're already UID 700, it skips entirely.

archanaserver:
@coderabbitai review

arvind4501:
> but for the existing deployments, ownership is fixed on the next foremanctl deploy run (dir are recreated with correct ownership during the directory creation tasks).


By recreating you mean backup and restore? if not then we may end up lossing all content data and only have a empty directory. IMO a migration task from root to non-root could simply be something like, check if the host dir(/var/lib/pulp) is owned by root, if yes then chown the whole dir recursively(as part of foremanctl deploy), this would be one time only. else run the deployment normally so that upcoming version migrations (from 5.1 container to 5.2 containers, which both have non root UID) does not trigger chowning the dirs as they are alredy owned by non root.

archanaserver:
> > but for the existing deployments, ownership is fixed on the next foremanctl deploy run (dir are recreated with correct ownership during the directory creation tasks).
> 
> By recreating you mean backup and restore? if not then we may end up lossing all content data and only have a empty directory. IMO a migration task from root to non-root could simply be something like, check if the host dir(/var/lib/pulp) is owned by root, if yes then chown the whole dir recursively(as part of foremanctl deploy), this would be one time only. else run the deployment normally so that upcoming version migrations (from 5.1 container to 5.2 containers, which both have non root UID) does not trigger chowning the dirs as they are alredy owned by non root.

So by "recreated with correct ownership," I mean the existing data and dir structure are preserved, and we only fix the ownership recursively using `chown 700:700` on the existing content during the deploy task, and i mean with not deleting and recreating from scratch.

archanaserver:
Here theimplementation uses numeric `UID/GID 700` to prepare the host bind mounts for the non-root Pulp container. `skopeo inspect` confirms that `quay.io/foreman/pulp:foreman-nightly` currently has no `USER` configured and therefore defaults to root, so Pulp remains temporarily listed in `EXPECTED_ROOT_IMAGES`.

Once https://github.com/theforeman/pulp-oci-images/pull/76 is merged and the updated image, declaring `USER pulp:pulp`, which resolves to `UID/GID 700` and is published to `quay.io/foreman/pulp:foreman-nightly`, this test exception can be removed.

archanaserver:
The current CI failure appears unrelated to this PR. `./foremanctl pull-images` fails while installing host dependencies, before the images are pulled or the Pulp ownership changes are exercised. So that can be looked separately but for this is unrelated, but before that https://github.com/theforeman/pulp-oci-images/pull/76 needs to be merged.

src/roles/pulp/tasks/main.yaml
```diff
@@ -7,13 +7,19 @@
     path: "{{ item | split(':') | first }}"
     state: directory
     mode: "0755"
+    owner: "700"
+    group: "700"
```
lzap, line 11: Instead of hardcoding 700, these could be defaults of the pulp role.:

```
pulp_container_uid: 700
pulp_container_gid: 700
```

src/roles/restore/tasks/restore_pulp_content.yaml
lzap: I think what @evgeni thinks it that since there is no major release based on foremanctl yet, this patch fixes the permissions, therefore all production backups will be all fine.

We can probably do a followup and drop this.

ehelms (approved):
Test failures are unrelated.

PR #735, opened 2026-08-10
Capsule flavor

arvind4501:

*

Steps to reproduce:

*

#### Checklist


src/vars/flavors/capsule.yml
```diff
@@ -0,0 +1,24 @@
+---
+flavor_features:
+  - foreman-proxy
+  - pulp
+  - content/ansible
+  - content/container
+  - content/rpm
+  - remote-execution
```
jeremylenz, line 8: Now that `templates` and `registration` features are available, Capsule flavor should have them 🙂

jeremylenz (commented):
Thanks @arvind4501 

Looking forward to this!

src/vars/flavors/capsule.yml
arvind4501: Sure we can add those, but in [SAT-47662](https://redhat.atlassian.net/browse/SAT-47662) these are not part of the enabled features by default. having said that i think we can surely enable those as they were in [foreman-installer](https://github.com/theforeman/foreman-installer/blob/develop/config/foreman-proxy-content-answers.yaml#L29-L30).
@ekohl what do you think about enabling these features by default in capsule deployment?

ekohl: Similar to #726 I think we should first aim at parity with the existing install where possible. I'll modify the downstream issue accordingly.

arvind4501: > Similar to https://github.com/theforeman/foremanctl/pull/726 I think we should first aim at parity with the existing install where possible

does that mean, we should add those(templates, registration) feature in capsule defaults?

ekohl: Yes

vendor_overrides/satellite/deploy-proxy/metadata.obsah.yaml
evgeni, line 1: If you put this under `verndor_overrides/satellite`, you only need one `cp` command in the setup. I've originally meant the `satellite` here as the product, not the software.

Gauravtalreja1: Or if we decide to keep it separate we might need packaging change as similar to satellite flavor?

arvind4501: we need packaging change both ways, i opted what evgeni suggested to put overrides under `verndor_overrides/satellite`.

evgeni: It needs a packaging change downstream either way (which Arvind already has prepared).

The files here are only for CI.

.github/workflows/test.yml
```diff
@@ -532,6 +532,73 @@ jobs:
...
+            --flavor satellite \
+            --initial-admin-password=changeme \
+            --tuning development
+      - name: Generate auth bundle on quadlet
+        run: |
+          ./foremanctl auth-bundle proxy.example.com
+      - name: Fetch auth bundle from quadlet
+        run: |
+          ./forge fetch-bundle proxy.example.com
+      - name: Deploy capsule
```
evgeni, line 578: For proxy-content, we purge params in https://github.com/theforeman/foremanctl/blob/6bbcfa991d57564badb0064f1b00dfdce632a7e6/.github/workflows/test.yml#L436-L438
Should we do the same here? Just to be safe?

evgeni (approved):
one tests nitpick, but ack either way

Gauravtalreja1 (approved):
ACK, tested with `foremanctl-3.0.0.develop-4.20260814050952506859.pr735.207.g06e3c13.el10.noarch` 🍏

Scenarios tested:
- Satellite + Capsule deployment with vendor overrides -> works
- Auth bundle workflow (generate, fetch/copy, deploy-proxy) -> works
- Health check passes
- Pulp API, services, mirror mode, RHSM URL -> correct
- Feature listing correctly hides/shows enabled features
- Capsule uses auth bundle certs, no CA key present -> works
- Capsule registered on Satellite correctly, and content sync works
- No regressions on foreman-proxy-content flavor

PR #736, opened 2026-08-10
Foreman environment fixes

evgeni:

recurring timers didn't start as they tried to enable ansible, it escalated from there

* move environment setup to variables, so all containers get the same env

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/development/roles/foreman_development/defaults/main.yaml
+++ b/development/roles/foreman_development/defaults/main.yaml
@@ -156,6 +156,7 @@ foreman_development_packages:
   - ruby-devel
   - npm
   - postgresql-devel
+  - libcap-devel
   - libxml2-devel
   - libxslt-devel
   - libcurl-devel
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -128,14 +128,7 @@
       - 'foreman-client-cert,type=mount,target=/etc/foreman/client_cert.pem'
       - 'foreman-client-key,type=mount,target=/etc/foreman/client_key.pem'
       - 'foreman-db-ca,type=mount,target={{ foreman_database_ssl_ca_path }}'
-    env:
-      SEED_ORGANIZATION: "{{ foreman_initial_organization }}"
-      SEED_LOCATION: "{{ foreman_initial_location }}"
-      FOREMAN_PUMA_THREADS_MIN: "{{ foreman_puma_threads_min }}"
-      FOREMAN_PUMA_THREADS_MAX: "{{ foreman_puma_threads_max }}"
-      FOREMAN_PUMA_WORKERS: "{{ foreman_puma_workers }}"
-      FOREMAN_ENABLED_PLUGINS: "{{ foreman_plugins | join(' ') }}"
-      FOREMAN_LOGGING_LEVEL: "{{ foreman_log_level }}"
+    env: "{{ foreman_env }}"
     quadlet_options:
       - |
         [Unit]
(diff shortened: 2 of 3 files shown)


src/roles/foreman/tasks/main.yaml
```diff
@@ -238,10 +227,7 @@
       - 'foreman-ca-cert,type=mount,target=/etc/foreman/katello-default-ca.crt'
       - 'foreman-client-cert,type=mount,target=/etc/foreman/client_cert.pem'
       - 'foreman-client-key,type=mount,target=/etc/foreman/client_key.pem'
-    env:
-      SEED_ORGANIZATION: "{{ foreman_initial_organization }}"
-      SEED_LOCATION: "{{ foreman_initial_location }}"
-      FOREMAN_LOGGING_LEVEL: "{{ foreman_log_level }}"
+    env: "{{ foreman_base_env }}"
```
evgeni: Should this also use `foreman_env`? Setting the extra Puma vars shouldn't hurt here.

src/roles/foreman/tasks/main.yaml
```diff
@@ -272,10 +258,7 @@
     sdnotify: false
     network: host
     command: bash -c "bin/rails db:migrate && bin/rails db:seed"
-    env:
-      FOREMAN_ENABLED_PLUGINS: "{{ foreman_plugins | join(' ') }}"
-      SEED_ORGANIZATION: "{{ foreman_initial_organization }}"
-      SEED_LOCATION: "{{ foreman_initial_location }}"
+    env: "{{ foreman_base_env }}"
```
evgeni: Should this also use `foreman_env`? Setting the extra Puma vars shouldn't hurt here.

adamruzicka (approved):
Haven't tested, but looks reasonable

development/roles/foreman_development/defaults/main.yaml
```diff
@@ -156,6 +156,7 @@ foreman_development_packages:
   - ruby-devel
   - npm
   - postgresql-devel
+  - libcap-devel
```
adamruzicka, line 159: ~Why though? Why do we need a library for packet capture?~

What difference a single letter can make. Never mind me

evgeni: I'll still answer: https://github.com/eitoball/net-ping/pull/44

src/roles/foreman/tasks/main.yaml
```diff
@@ -179,11 +172,7 @@
       - 'foreman-client-key,type=mount,target=/etc/foreman/client_key.pem'
       - 'foreman-dynflow-worker-hosts-queue-yaml,type=mount,target=/etc/foreman/dynflow/worker-hosts-queue.yml'
       - 'foreman-db-ca,type=mount,target={{ foreman_database_ssl_ca_path }}'
-    env:
-      DYNFLOW_REDIS_URL: "redis://localhost:6379/6"
-      REDIS_PROVIDER: "DYNFLOW_REDIS_URL"
-      FOREMAN_ENABLED_PLUGINS: "{{ foreman_plugins | join(' ') }}"
-      FOREMAN_LOGGING_LEVEL: "{{ foreman_log_level }}"
+    env: "{{ foreman_dynflow_env }}"
```
arvind4501, line 175: this will also mount other foreman_env env's, is that a issue to pass extra envs than needed?

evgeni: No it should be no problem. The only "extra" vars I see are the Puma ones, and those are ignored when Puma is not running (like in the Dynflow case).

arvind4501: ok, thanks

arvind4501 (approved):

PR #737, opened 2026-08-10
Add execution tests for recurring timers

archanaserver:

Previously, tests only verified that recurring timers were enabled, running, and that their services existed.
This didn't catch failures where timers were configured but failed to execute or completed with errors.

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/tests/feature/foreman/base_test.py
+++ b/tests/feature/foreman/base_test.py
@@ -1,4 +1,5 @@
 import json
+import time
 
 import pytest
 
@@ -75,6 +76,44 @@ def test_foreman_recurring_timers_enabled_and_running(server, instance):
 def test_foreman_recurring_services_exist(server, instance):
     service = server.service(f"foreman-recurring@{instance}.service")
     assert service.exists
+
+
+@pytest.mark.parametrize("instance", RECURRING_INSTANCES)
+def test_foreman_recurring_timer_last_trigger(server, instance):
+    """Verify that timers have a valid last trigger time (if they've run)."""
+    timer_name = f"foreman-recurring@{instance}.timer"
+    timer = server.service(timer_name) 
+    assert "LastTriggerUSec" in timer.systemd_properties
+
+
+@pytest.mark.parametrize("instance", RECURRING_INSTANCES)
+def test_foreman_recurring_timer_next_trigger(server, instance):
+    """Verify that timers have a scheduled next trigger time."""
+    timer_name = f"foreman-recurring@{instance}.timer"
+    timer = server.service(timer_name)
+    assert timer.systemd_properties["NextElapseUSecRealtime"] != "0"
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize("instance", RECURRING_INSTANCES)
+def test_foreman_recurring_timer_execution(server, instance):
+    """Trigger a timer manually and verify it executes successfully."""
+    service_name = f"foreman-recurring@{instance}.service"
+
+    server.check_output(f"systemctl start {service_name}")
+
+    # Wait for the service to complete (these are oneshot services)
+    # Poll the service status until it's no longer active
+    max_wait = 60  # Maximum wait time in seconds
+    poll_interval = 2
+    waited = 0
+
+    service = server.service(service_name)
+    while service.is_running and waited < max_wait:
+        time.sleep(poll_interval)
+        waited += poll_interval
+
+    assert service.systemd_properties["Result"] == "success"
 
 
 def test_foreman_delivery_method_setting(foremanapi):


tests/feature/foreman/base_test.py
```diff
@@ -77,6 +78,52 @@ def test_foreman_recurring_services_exist(server, instance):
     assert service.exists
 
 
+@pytest.mark.parametrize("instance", RECURRING_INSTANCES)
+def test_foreman_recurring_timer_last_trigger(server, instance):
+    """Verify that timers have a valid last trigger time (if they've run)."""
+    timer_name = f"foreman-recurring@{instance}.timer"
+    result = server.run(f"systemctl show {timer_name} --property=LastTriggerUSec")
```
evgeni: I think you can use `server.service(timer_name)` and then look at the `systemd_properties`: https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.service.Service.systemd_properties

(applies to the other tests too)

evgeni:
This should fail until #736 is in, which is always a good thing for tests :)

tests/feature/foreman/base_test.py
```diff
@@ -77,6 +78,52 @@ def test_foreman_recurring_services_exist(server, instance):
...
+    assert next_trigger != "0"
+
+
+@pytest.mark.parametrize("instance", RECURRING_INSTANCES)
+def test_foreman_recurring_timer_execution(server, instance):
+    """Trigger a timer manually and verify it executes successfully."""
+    service_name = f"foreman-recurring@{instance}.service"
+
+    result = server.run(f"systemctl start {service_name}")
+    assert result.succeeded or result.rc == 0
```
ekohl: `succeeded` and `rc == 0` is the same (https://github.com/pytest-dev/pytest-testinfra/blob/04cccc3969ab786e88e2d67462c409c8778005e8/testinfra/backend/base.py#L68), but you can also use `check_output` which ensures exit code is 0 or raises an exception:
```suggestion
    server.check_output(f"systemctl start {service_name}")
```

Or `run_expect`:

```suggestion
    server.run_expect([0], f"systemctl start {service_name}")
```

I think the former is a bit easier and the overhead of calling `rstrip` on the output is negligible in the whole scale of our testing.

tests/feature/foreman/base_test.py
```diff
@@ -77,6 +78,52 @@ def test_foreman_recurring_services_exist(server, instance):
...
+
+    max_wait = 60
+    poll_interval = 2
+    waited = 0
+
+    while waited < max_wait:
+        is_active = server.run(f"systemctl is-active {service_name}")
+        if is_active.stdout.strip() == "inactive":
+            break
```
ekohl: You can use the `is_running` method on the service to determine this:
```suggestion
        if not server.service(service_name).is_running:
            break
```

evgeni:
closing and re-opening to see that the tests now pass

evgeni:
extdb is failing, as we were not passing the DB CA to the tasks container, fix in https://github.com/theforeman/foremanctl/pull/741

evgeni:
All green now, nice

archanaserver:
> All green now, nice

Thanks @evgeni :)

tests/feature/foreman/base_test.py
```diff
@@ -77,6 +78,53 @@ def test_foreman_recurring_services_exist(server, instance):
...
+    waited = 0
+
+    while waited < max_wait:
+        service = server.service(service_name)
+        if not service.is_running:
+            break
+        time.sleep(poll_interval)
+        waited += poll_interval
+
+    service = server.service(service_name)
```
ekohl: I don't think there is caching on the `is_running` so you can also write it:
```suggestion
    service = server.service(service_name)
    while service.is_running and waited < max_wait:
        time.sleep(poll_interval)
        waited += poll_interval
```

tests/feature/foreman/base_test.py
```diff
@@ -77,6 +78,53 @@ def test_foreman_recurring_services_exist(server, instance):
...
+
+    assert "ExecMainStatus" in service.systemd_properties
+    assert "Result" in service.systemd_properties
```
ekohl: I don't think these add a lot: down below you check the value and they'll raise a `KeyError` if they're not in there.
```suggestion
```

tests/feature/foreman/base_test.py
```diff
@@ -77,6 +78,53 @@ def test_foreman_recurring_services_exist(server, instance):
...
+
+    assert service.systemd_properties["ExecMainStatus"] == "0"
+    assert service.systemd_properties["Result"] == "success"
```
ekohl: I'm reading https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.systemd1.html and it states:
> Result encodes the execution result of the last run of the service. It is useful to determine the reason a service failed if it is in the "failed" state (see ActiveState above). The following values are currently known: "success" is set if the unit did not fail. "resources" indicates that not enough resources were available to fork off and execute the service processes. "timeout" indicates that a timeout occurred while executing a service operation. "exit-code" indicates that a service process exited with an unclean exit code. "signal" indicates that a service process exited with an uncaught signal. "core-dump" indicates that a service process exited uncleanly and dumped core. "watchdog" indicates that a service did not send out watchdog ping messages often enough. "start-limit-hit" indicates that a service has been started too frequently in a specific time frame (as configured in StartLimitInterval, StartLimitBurst). "oom-kill" indicates that a service was terminated due to an out-of-memory (OOM) situation.

So perhaps asserting `success` is sufficient here?

tests/feature/foreman/base_test.py
```diff
@@ -77,6 +78,53 @@ def test_foreman_recurring_services_exist(server, instance):
...
+
+
+@pytest.mark.parametrize("instance", RECURRING_INSTANCES)
+def test_foreman_recurring_timer_next_trigger(server, instance):
+    """Verify that timers have a scheduled next trigger time."""
+    timer_name = f"foreman-recurring@{instance}.timer"
+    timer = server.service(timer_name)
+    assert "NextElapseUSecRealtime" in timer.systemd_properties
+    next_trigger = timer.systemd_properties.get("NextElapseUSecRealtime")
+    assert next_trigger != "0"
```
ekohl: Perhaps a very mind nit, but this is a plain dict so you can also rely on `KeyError` to be raised if it's missing.
```suggestion
    assert timer.systemd_properties["NextElapseUSecRealtime"] != "0"
```

tests/feature/foreman/base_test.py
```diff
@@ -77,6 +78,53 @@ def test_foreman_recurring_services_exist(server, instance):
...
+
+
+@pytest.mark.parametrize("instance", RECURRING_INSTANCES)
+def test_foreman_recurring_timer_execution(server, instance):
```
ekohl, line 99: Perhaps mark this as a slow test?
```suggestion
@pytest.mark.slow
def test_foreman_recurring_timer_execution(server, instance):
```

ekohl (approved):
Since @evgeni is out and I started a review I've taken another look. Overall I think it looks good. If this passes, I have no objection to merging this.

The comments are mostly stylistic and a way for me to write down my thought process. I tried to focus on the tests being short and focused. Minimal code and only a single assert per test.

tests/feature/foreman/base_test.py
archanaserver: yeah make sense

tests/feature/foreman/base_test.py
archanaserver: yes that seems sufficient and correct here.

ekohl (approved):
Looks good, but it feels like I'm partially approving my own code. Perhaps someone else could give it a look as well?

ehelms (approved):

PR #740, opened 2026-08-11
forge iso - Install Satellite/foremanctl from ISO

stejskalleos:

Allow developers to install Satellite / foremanctl from an ISO (disconnected scenario)

* `forge iso`

Steps to reproduce:

* `forge iso --help`

#### TODOs

* Tests (?)
* Documentation

#### Checklist

### DIFF
--- a/development/playbooks/iso/iso.yaml
+++ b/development/playbooks/iso/iso.yaml
@@ -1,0 +1,61 @@
+---
+- name: Prepare a disconnected installation from an ISO
+  hosts:
+    - quadlet
+  become: true
+  vars:
+    iso_download_path: /root/satellite.iso
+    iso_mount_path: /mnt/satellite
+  tasks:
+    - name: Install packages
+      ansible.builtin.package:
+        name:
+          - podman
+          - skopeo
+        state: present
+
+    - name: Check if ISO already exists on target
+      ansible.builtin.stat:
+        path: "{{ iso_download_path }}"
+      register: iso_stat
+
+    - name: Upload the ISO from local path
+      ansible.builtin.copy:
+        src: "{{ iso_path }}"
+        dest: "{{ iso_download_path }}"
+        mode: "0644"
+      when: not iso_stat.stat.exists
+
+    - name: Ensure ISO mount point exists
+      ansible.builtin.file:
+        path: "{{ iso_mount_path }}"
+        state: directory
+        mode: "0755"
+
+    - name: Mount the ISO, run prepare_system, then always unmount
+      block:
+        - name: Mount the ISO
+          ansible.posix.mount:
+            path: "{{ iso_mount_path }}"
+            src: "{{ iso_download_path }}"
+            fstype: iso9660
+            opts: loop,ro
+            state: ephemeral
+
+        - name: Run the prepare_system script from the mounted ISO
+          ansible.builtin.command:
+            argv: "{{ ['python3', 'prepare_system'] + (['--nogpgcheck'] if iso_nogpgcheck | default(false) else []) }}"
+            chdir: "{{ iso_mount_path }}"
+          register: iso_prepare_system_result
+          changed_when: true
+
+        - name: Show prepare_system output
+          ansible.builtin.debug:
+            msg:
+              stdout: "{{ iso_prepare_system_result.stdout }}"
+              stderr: "{{ iso_prepare_system_result.stderr }}"
+      always:
+        - name: Unmount the ISO
+          ansible.posix.mount:
+            path: "{{ iso_mount_path }}"
+            state: unmounted
--- a/development/playbooks/iso/metadata.obsah.yaml
+++ b/development/playbooks/iso/metadata.obsah.yaml
@@ -1,0 +1,21 @@
+---
+help: |
+  Prepare a disconnected installation from an ISO.
+
+  Uploads the ISO (unless already present), mounts it, and runs the
+  prepare_system script shipped on the ISO to install the foremanctl
+  package and import container images into local storage. The ISO is
+  unmounted afterwards. Once complete, run foremanctl deploy (as
+  suggested by the script) to finish the installation.
+
+  Use --nogpgcheck when installing from unsigned/nightly composes.
+
+variables:
+  iso_path:
+    parameter: iso_path
+    help: Path to a local ISO file to upload to the target machine.
+    type: AbsolutePath
+  iso_nogpgcheck:
+    parameter: --nogpgcheck
+    help: Skip GPG signature verification when installing packages (for unsigned/nightly composes).
+    action: store_true


ekohl (commented):
The ISO is large so it may help to have it cached locally in case you want to tear down the whole target machine. At first I thought about pointing it to an ISO on the local filesystem well, but that complicates the code.

How about a README somewhere that you can use `python -m http.server` to easily host local HTTP server and download from there?

stejskalleos:
> At first I thought about pointing it to an ISO on the local filesystem well, but that complicates the code.

I thought about another approach: add `--iso-url` and `--iso-path`, so devs could choose between those two.
The Python server could be a solution as well; an update to the README is an easy way to do it.

arvind4501:
While looking at PR, the first thought that comes to mind is where to get iso from? is it only available downstream or we also have upstream iso?

ekohl:
We don't have upstream ISOs. While investigating this @ShimShtein and I have looked at it. IMHO the best path forward is to build a process where an upstream Katello's Inter-Satellite Sync (ISS) can generate all the media. If we teach foremanctl how to import that then we would no longer need a downstream ISO.

But in the interest of time we'll continue with the ISO path for the next release before building the long term path.

stejskalleos:
> While looking at PR, the first thought that comes to mind is where to get iso from?

Yeah, it's from an internal URL that can't be shared publicly. Customers can download the Satellite iso from the Customer Portal.

> But in the interest of time we'll continue with the ISO path for the next release before building the long term path.

Ok, will add it in the next commit.

development/playbooks/iso/metadata.obsah.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  Downloads the ISO (unless already present), mounts it, and runs the
+  prepare_system script shipped on the ISO to install the foremanctl
+  package and import container images into local storage. The ISO is
+  unmounted afterwards. Once complete, run foremanctl deploy (as
+  suggested by the script) to finish the installation.
+
+variables:
+  iso_url:
+    parameter: --iso-url
```
evgeni:
```suggestion
    parameter: iso_url
```

This will change the invocation to `./forge iso <iso_url>` and make the parameter *required*, so you can drop the `ansible.builtin.assert` in your playbook.

development/playbooks/iso/metadata.obsah.yaml
```diff
@@ -0,0 +1,36 @@
+---
+help: |
+  Prepare a disconnected Satellite installation from a Satellite ISO.
```
evgeni: Do we need to call out Satellite here? (And in other parts of this PR?)

I'd prefer not to have any product branding in foremanctl by default.

development/playbooks/iso/iso.yaml
```diff
@@ -0,0 +1,95 @@
...
+                 + (['--enhanced_reporting'] if enhanced_reporting | default(false) | bool else [])
+                 + (['--nogpgsigs'] if nogpgsigs | default(false) | bool else []) }}
+            chdir: "{{ iso_mount_path }}"
+          register: iso_install_packages_result
+          changed_when: true
+
+        - name: Show install_packages output
+          ansible.builtin.debug:
+            msg: "{{ iso_install_packages_result.stdout_lines }}"
+        - name: Run the setup_containers script from the mounted ISO
```
evgeni:
```suggestion

        - name: Run the setup_containers script from the mounted ISO
```

development/playbooks/iso/metadata.obsah.yaml
```diff
@@ -0,0 +1,36 @@
...
+    action: store_true
+    persist: false
+  enhanced_reporting:
+    help: Passed through to prepare_system. Install the enhanced reporting feature.
+    action: store_true
+  nogpgsigs:
+    help: Passed through to prepare_system. Ignore package GPG signatures when installing.
+    action: store_true
+  registry:
+    help: "Passed through to prepare_system. Container registry to use (script defaults to registry.redhat.io)."
```
evgeni: Why can we set a registry? I thought the point of this is that everything is contained in the ISO.

development/playbooks/iso/metadata.obsah.yaml
```diff
@@ -0,0 +1,36 @@
...
+  insecure:
+    help: Do not verify TLS certificates when downloading the ISO.
+    action: store_true
+  force:
+    help: Re-download the ISO even if a previously downloaded copy already exists at /root/satellite.iso.
+    action: store_true
+    persist: false
+  enhanced_reporting:
+    help: Passed through to prepare_system. Install the enhanced reporting feature.
+    action: store_true
```
evgeni: What is enhanced reporting? Looking at the code of the script you run I could not understand it either.

development/playbooks/iso/metadata.obsah.yaml
```diff
@@ -0,0 +1,36 @@
...
+  nogpgsigs:
+    help: Passed through to prepare_system. Ignore package GPG signatures when installing.
+    action: store_true
+  registry:
+    help: "Passed through to prepare_system. Container registry to use (script defaults to registry.redhat.io)."
+  containers_path:
+    help: Passed through to prepare_system as --containers-path. Absolute path to the containers directory.
+    type: AbsolutePath
```
evgeni: Why would one want to override this?

development/playbooks/iso/metadata.obsah.yaml
```diff
@@ -0,0 +1,36 @@
...
+  force:
+    help: Re-download the ISO even if a previously downloaded copy already exists at /root/satellite.iso.
+    action: store_true
+    persist: false
+  enhanced_reporting:
+    help: Passed through to prepare_system. Install the enhanced reporting feature.
+    action: store_true
+  nogpgsigs:
+    help: Passed through to prepare_system. Ignore package GPG signatures when installing.
+    action: store_true
```
evgeni: Shouldn't the ISO come with sigs and this shouldn't be needed?

evgeni (commented):
Conceptually, this reads fine.

I'd probably not include any params besides iso_url, insecure and force, focusing on the one happy path that's supposed to work for now, as I see no good reason to expose the other parameters and also do not expect other consumers of the ISO to use them. But wouldn't block on that.

stejskalleos:
Added the `--iso-path` arg

development/playbooks/iso/metadata.obsah.yaml
stejskalleos: From my testing, it doesn't import the sigs by default.
Unless you do it yourself, without `nogpgsigs`, it doesn't work.

development/playbooks/iso/metadata.obsah.yaml
stejskalleos: IDK, but the script does support that.

development/playbooks/iso/metadata.obsah.yaml
stejskalleos: Ok, will remove the Satellite mention.

development/playbooks/iso/metadata.obsah.yaml
stejskalleos: So we don't want to maintain the original functionality of install_packages and setup_containers?
Not that I'm against it, just want to make sure I'm on the same page.

evgeni: I would prefer to expose as few parameters/knobs as possible for now.

The task was "As a developer I want to be able to test "install from ISO" scenario with forge command" -- if this parameter is not needed for this, we should not make our lifes more complicated by supporting it.

development/playbooks/iso/metadata.obsah.yaml
evgeni: A parameter you *always* have to pass is a bad default, as it leaves no choice to the user *and* confuses the user why it's not working without passing it.

I'd argue that `prepare_system` script should be setting up GPG keys correctly and a `prepare_system` user should not have to pass `--nogpgsigs` to it in *most* cases. If it doesn't, that's a bug in `prepare_system`, which we *can* workaround here by *always* passing `--nogpgsigs` for now while we're fixing the real bug.

stejskalleos: yop agree, I'm looking into it right now.

stejskalleos:
```
rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
[root@quadlet satellite]# ./prepare_system 
This script will install the satellite packages on the current machine.
   - Ensuring we are in an expected directory.
   - Copying installation files.
   - Creating Satellite Repository File
   - Creating Maintenance Repository File
   - Checking to see if foremanctl is already installed.
   - Importing the gpg key.
Package foremanctl-3.0.0.develop-4.20260811160122303547.master.193.g893d4a0.el9sat.noarch.rpm is not signed
Package python3.12-obsah-1.10.0-2.el9sat.noarch.rpm is not signed
Package satellitectl-6.20.0-2.el9sat.noarch.rpm is not signed
Error: GPG check FAILED
Error while executing command: 'dnf install -y satellitectl'
```

This will make the script not working for test environment, but if you want to, i will remove the option.

stejskalleos:
Applied comments from the 2nd round, reflecting the latest changes in https://gitlab.cee.redhat.com/spmm/composes/-/merge_requests/553

development/playbooks/iso/metadata.obsah.yaml
evgeni: Ooooh, could it be it's because that's a nightly ISO, the packages are unsigned?

stejskalleos: Probably, yes.

arvind4501: I also encountered the same issue at gpg key import step
```
./prepare_system 2>&1
This script will install the satellite packages on the current machine.
   - Ensuring we are in an expected directory.
   - Copying installation files.
   - Creating Satellite Repository File
   - Creating Maintenance Repository File
   - Checking to see if satellitectl is already installed.
   - Importing the gpg key.
Package ansible-core-2.16.19-1.el9sat.noarch.rpm is not signed
Package foremanctl-3.0.0.develop-4.20260811160122303547.master.193.g893d4a0.el9sat.noarch.rpm is not signed
Package python3.12-cffi-2.1.1-1.el9pc.x86_64.rpm is not signed
Package python3.12-cryptography-46.0.7-2.el9pc.x86_64.rpm is not signed
Package python3.12-jinja2-3.1.5-2.el9sat.noarch.rpm is not signed
Package python3.12-markupsafe-3.0.2-2.el9sat.x86_64.rpm is not signed
Package python3.12-obsah-1.10.0-2.el9sat.noarch.rpm is not signed
Package python3.12-packaging-26.2-2.el9pc.noarch.rpm is not signed
Package python3.12-pycparser-3.0-2.el9pc.noarch.rpm is not signed
Package python3.12-pyparsing-3.3.2-2.el9pc.noarch.rpm is not signed
Package python3.12-pyyaml-6.0.3-2.el9pc.x86_64.rpm is not signed
Package python3.12-resolvelib-1.0.1-3.el9sat.noarch.rpm is not signed
Package satellitectl-6.20.0-2.el9sat.noarch.rpm is not signed
Error: GPG check FAILED
Error while executing command: 'dnf install -y satellitectl': Command '['dnf', 'install', '-y', 'satellitectl']' returned non-zero exit status 1
```

I think we shoule keep the `--nogpgsigs` flag as it was before as a option to skip gpg signature, until we are sure that all packages are signed by default

stejskalleos:
`--nogpgcheck` is back!

arvind4501:
The `./forge iso iso-path`  where the iso is stream iso(which still does not contain updated prepare_system script) does not directly work, but for now we can manually mount the stream iso, something like
```
mount the ISO and copy its contents to a writable location:

  mount -o loop,ro /root/satellite.iso /mnt/satellite
  cp -a /mnt/satellite /root/satellite
  fetch prepare_system script
  run `./prepare_system` with `--nogpgcheck`
```
 and copy the prepare_system script onto mounted directory and then we can run the script and see verify all the images exists on the machine.
and then from there user can follow the usual flow of deploying using foremantcl/satellitectl will work as we have all the bits and pieces.

development/playbooks/iso/iso.yaml
```diff
@@ -0,0 +1,59 @@
...
+        - name: Run the prepare_system script from the mounted ISO
+          ansible.builtin.command:
+            argv: "{{ ['python3', 'prepare_system'] + (['--nogpgcheck'] if iso_nogpgcheck | default(false) else []) }}"
+            chdir: "{{ iso_mount_path }}"
+          register: iso_prepare_system_result
+          changed_when: true
+
+        - name: Show prepare_system output
+          ansible.builtin.debug:
+            msg: "{{ iso_prepare_system_result.stdout_lines }}"
```
ekohl: Won't you miss stderr this way?

Odilhao:
I know this is probably not the PR for it, but would make for us to have `foremanctl` itself to be able to generate the ISO from the list of images that we define?

arvind4501:
There is some [dependencies](https://github.com/theforeman/foremanctl/blob/master/src/roles/pre_install/tasks/main.yaml#L23) that needs to be installed on machine in order to run deployment, should these packages be part of ISO?

evgeni:
> There is some [dependencies](https://github.com/theforeman/foremanctl/blob/master/src/roles/pre_install/tasks/main.yaml#L23) that needs to be installed on machine in order to run deployment, should these packages be part of ISO?

Those come from RHEL, and the expectation is that you also have a RHEL iso mounted (or somehow otherwise available)

stejskalleos:
Rebased & squashed, added `stdout` & `stderr` to print the `prepare_system` output

evgeni:
My concerns were addressed, I'll leave the final review to @arvind4501 as I will be out the next two weeks.

ekohl:
> I know this is probably not the PR for it, but would make for us to have `foremanctl` itself to be able to generate the ISO from the list of images that we define?

We have talked about that and I'd love to do that somehow in the long run. I have some ideas and I wrote them down in https://github.com/theforeman/foremanctl/issues/754

development/playbooks/iso/iso.yaml
```diff
@@ -0,0 +1,61 @@
+---
+- name: Prepare a disconnected installation from an ISO
+  hosts:
+    - quadlet
+  become: true
+  vars:
+    iso_download_path: /root/satellite.iso
+    iso_mount_path: /mnt/satellite
```
ehelms, line 8: If you change these two not to say `satellite` then this becomes generic of an action so long as an ISO is provided that meets the criteria of having a `prepare_system` python command that loads containers into storage and installs foremanctl. Meaning, in effect, this would then expect an "interface" that any ISO could be generated to match.

arvind4501 (approved):
This PR looks good, i manually tried this on quadlet vm with manually pulling `prepare_system` script to the vm. so acking for that

There are two things that block a proper testing of this:

- we need to get prepare_script in
-  we need the downstream image overides for all images so deployment on disconnected does not try to pull from quay.

development/playbooks/iso/iso.yaml
ekohl: In our current upstream documentation we use `/mnt/sat6`. This is already the more generic `/mnt/satellite` that we should aim for IMHO. This mount is in addition to `/mnt/rhel` for the actual OS.

For now I think this is good enough. Long term I'd love to move away from the ISO anyway (see #754). We can always iterate on it later.

But if you have a suggestion for a better name, I'd welcome it.

PR #741, opened 2026-08-11
Set foreman secrets for all containers

### LINKED ISSUE #648: Consolidate Foreman env vars and secrets into role variables
While reviewing [#640](https://github.com/theforeman/foremanctl/pull/640) (log levels for Foreman & Proxy), [@evgeni noted](https://github.com/theforeman/foremanctl/pull/640#discussion_r3558953043) that adding a new environment variable to Foreman containers requires editing multiple podman_container tasks in the same file. The same duplication exists for Podman secrets.

### Proposed solution
Follow the pattern already used in the pulp role (`src/roles/pulp/defaults/main.yaml`).

Define reusable variable maps in src/roles/foreman/defaults/main.yaml (or a dedicated vars file), e.g.:
* `foreman_container_base_env` — env vars common to most/all Foreman containers
* `foreman_container_base_secrets` — secrets common to most/all Foreman containers

evgeni:

recurring tasks were not setting the DB CA secret correctly

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -117,17 +117,7 @@
     hostname: "{{ ansible_facts['hostname'] }}.local"
     volume:
       - 'foreman-data-run:/var/run/foreman:rw,z,U'
-    secrets:
-      - 'foreman-database-url,type=env,target=DATABASE_URL'
-      - 'foreman-encryption-key,type=env,target=ENCRYPTION_KEY'
-      - 'foreman-seed-admin-user,type=env,target=SEED_ADMIN_USER'
-      - 'foreman-seed-admin-password,type=env,target=SEED_ADMIN_PASSWORD'
-      - 'foreman-settings-yaml,type=mount,target=/etc/foreman/settings.yaml'
-      - 'foreman-katello-yaml,type=mount,target=/etc/foreman/plugins/katello.yaml'
-      - 'foreman-ca-cert,type=mount,target=/etc/foreman/katello-default-ca.crt'
-      - 'foreman-client-cert,type=mount,target=/etc/foreman/client_cert.pem'
-      - 'foreman-client-key,type=mount,target=/etc/foreman/client_key.pem'
-      - 'foreman-db-ca,type=mount,target={{ foreman_database_ssl_ca_path }}'
+    secrets: "{{ foreman_secrets }}"
     env: "{{ foreman_env }}"
     quadlet_options:
       - |
@@ -162,16 +152,7 @@
     hostname: "{{ ansible_facts['hostname'] }}.local"
     volume:
       - 'foreman-data-run:/var/run/foreman:rw,z,U'
-    secrets:
-      - 'foreman-database-url,type=env,target=DATABASE_URL'
-      - 'foreman-encryption-key,type=env,target=ENCRYPTION_KEY'
-      - 'foreman-settings-yaml,type=mount,target=/etc/foreman/settings.yaml'
-      - 'foreman-katello-yaml,type=mount,target=/etc/foreman/plugins/katello.yaml'
-      - 'foreman-ca-cert,type=mount,target=/etc/foreman/katello-default-ca.crt'
-      - 'foreman-client-cert,type=mount,target=/etc/foreman/client_cert.pem'
-      - 'foreman-client-key,type=mount,target=/etc/foreman/client_key.pem'
-      - 'foreman-dynflow-worker-hosts-queue-yaml,type=mount,target=/etc/foreman/dynflow/worker-hosts-queue.yml'
-      - 'foreman-db-ca,type=mount,target={{ foreman_database_ssl_ca_path }}'
+    secrets: "{{ foreman_dynflow_secrets }}"
     env: "{{ foreman_dynflow_env }}"
     command: "/usr/libexec/foreman/sidekiq-selinux -e production -r ./extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/%i.yml"
     quadlet_options:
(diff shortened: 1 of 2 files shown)


src/roles/foreman/tasks/main.yaml
```diff
@@ -162,16 +152,7 @@
...
       - 'foreman-data-run:/var/run/foreman:rw,z,U'
-    secrets:
-      - 'foreman-database-url,type=env,target=DATABASE_URL'
-      - 'foreman-encryption-key,type=env,target=ENCRYPTION_KEY'
-      - 'foreman-settings-yaml,type=mount,target=/etc/foreman/settings.yaml'
-      - 'foreman-katello-yaml,type=mount,target=/etc/foreman/plugins/katello.yaml'
-      - 'foreman-ca-cert,type=mount,target=/etc/foreman/katello-default-ca.crt'
-      - 'foreman-client-cert,type=mount,target=/etc/foreman/client_cert.pem'
-      - 'foreman-client-key,type=mount,target=/etc/foreman/client_key.pem'
-      - 'foreman-dynflow-worker-hosts-queue-yaml,type=mount,target=/etc/foreman/dynflow/worker-hosts-queue.yml'
```
ekohl, line 173: Should this have been using `%i` as well? The command executes `-C /etc/foreman/dynflow/%i.yml`.

evgeni: no? we don't need to override the other configs.

ekohl: It was my impression that for every dynflow worker you need a dedicated config file. That's what the current installer does: https://github.com/theforeman/puppet-foreman/blob/4325bc72a20ff7b8e0d747460bc3a8a573aa46ab/manifests/dynflow/worker.pp#L37-L53

So I guess both `dynflow-sidekiq@orchestrator` and `dynflow-sidekiq@worker` don't need a config file and falls back on hardcoded defaults. Then the mounted config file for `dynflow-sidekiq@worker-hosts-queue` will be unused in those containers.

When we do support other queues we will need to resolve this somehow.

evgeni: Oh sure, they need a config, but the one inside the RPM, based on what's in https://github.com/theforeman/foreman/tree/develop/config/dynflow is sufficient.

And yes, for those containers the hosts queue one will be cruft. But also doesn't hurt :woman_shrugging:

evgeni: Technically, we could even pass those to the normal foreman containers too at no real cost. As that's what is happening today with RPM deployments anyway.

adamruzicka: > Oh sure, they need a config, but the one inside the RPM, based on what's in https://github.com/theforeman/foreman/tree/develop/config/dynflow is sufficient.

What if we had katello package carry the definition for the hosts queue worker to not have to deal with it here in a custom way?

src/roles/foreman/tasks/main.yaml
```diff
@@ -162,16 +152,7 @@
...
-      - 'foreman-db-ca,type=mount,target={{ foreman_database_ssl_ca_path }}'
+    secrets: "{{ foreman_dynflow_secrets }}"
```
stejskalleos, line 155: This will now include secrets that are not needed by Dynflow:

```
  - 'foreman-seed-admin-user,type=env,target=SEED_ADMIN_USER'
  - 'foreman-seed-admin-password,type=env,target=SEED_ADMIN_PASSWORD'
  - ... and other?
```

IMO `foreman_dynflow_secrets` should be defined from scratch and not extend the `foreman_secrets`

stejskalleos (changes requested):

src/roles/foreman/tasks/main.yaml
evgeni: And that hurts?

Not defining it from scratch is exactly the point of this PR :)

src/roles/foreman/tasks/main.yaml
ekohl: In the old installer we had support for multiple host worker queues and were named with numbers: `worker-hosts-queue-1`, worker-hosts-queue-2`, etc. Those all needed their own configs.

Downstream there's also an RFE for a dedicated queue for webhook delivery. We never had a good way to easily support those in our installer. That's what's I'm now considering: can we do better in this design?

evgeni: Sure, we can talk about making this better.

But this PR doesn't change anything about the deployment layout of our dynflow workers, only where we define the configs they get.

evgeni: > > Oh sure, they need a config, but the one inside the RPM, based on what's in https://github.com/theforeman/foreman/tree/develop/config/dynflow is sufficient.
> 
> What if we had katello package carry the definition for the hosts queue worker to not have to deal with it here in a custom way?

If we don't need to change anything in there, that'd work

ekohl: > Sure, we can talk about making this better.
> 
> But this PR doesn't change anything about the deployment layout of our dynflow workers, only where we define the configs they get.

I've added context in https://github.com/theforeman/foremanctl/issues/72.

evgeni: Thanks!

src/roles/foreman/tasks/main.yaml
stejskalleos: > And that hurts?

Nah, I don't think it's an issue or a security flaw; it's just about exposing stuff the container needs and nothing more.

ekohl: Foreman on startup will check if needs seeding (https://github.com/theforeman/foreman/blob/develop/config/initializers/seeds.rb). While it normally shouldn't happen, there is a code path that could trigger it. Better safe than sorry.

evgeni: but then we need to remove that from *ALL* containers that are not our "db:seed" container. right?

evgeni: also, not setting those doesn't mean seeding won't happen, it will just happen with different values:

https://github.com/theforeman/foreman/blob/f7e9a6cd0869f76b438a9a9d591eb30eee41489a/db/seeds.d/035-admin.rb#L44-L71

ekohl: That was the point I implied: I think Dynflow can actually perform seeding and not setting the values could trigger incorrect seeding. I'd be in favor of being safe and always set the secrets (as the current code does).

evgeni: Oh, that way round :)

ehelms (approved):

PR #742, opened 2026-08-11
Agentic skill to setup foreman development

arvind4501:

* Adds `.agents/skills/dev-setup/` skill  for interactive dev environment setup with foremanctl

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/.agents/skills/dev-setup/SKILL.md
+++ b/.agents/skills/dev-setup/SKILL.md
@@ -1,0 +1,193 @@
+---
+name: dev-setup
+description: >-
+  Interactive foremanctl development environment setup. Set up a Foreman/Katello development environment with abilty to enable plugins.
+---
+
+# Foremanctl Dev Setup
+
+Set up a Foreman/Katello development environment via `forge deploy-dev`. Foreman is installed from git and runs directly on the VM as a Rails server. Backend services (PostgreSQL, Valkey, Candlepin, Pulp, HTTPD) run in containers.
+
+**Always use this skill** for foremanctl dev environment tasks — including partial workflows. Do not reimplement steps manually via ad-hoc commands when this skill covers the task.
+
+## Workflow routing
+
+Steps are composable. Ask what the user already has, then run only the matching steps:
+
+| User goal | What they need | Steps to run |
+|-----------|----------------|--------------|
+| Full environment from scratch | Nothing yet | Steps 1 → 2 → 3 → 4 → (optional 5) |
+| VMs already running, deploy Foreman | Vagrant VMs up, inventory exists | Step 3 → 4 → (optional 5) |
+| Re-deploy with different plugins | Vagrant VMs up | Step 3 (re-run with new plugin selection) → 4 |
+| Add hammer to existing deployment | Deployed environment | Step 5 |
+| Verify deployment health | Deployed environment | Step 4 |
+| Provision VMs only | Vagrant + libvirt installed | Steps 1 → 2 |
+
+When the user asks for a single step (e.g. "just run tests"), use the matching row above — do not skip the skill.
+
+## Prerequisites
+
+Before starting, verify the following are installed on the local machine:
+
+- **Python 3** — virtualenv and dependencies
+- **Ansible 2.14+** — automation runtime
+- **Vagrant 2.2+** — VM lifecycle management
+- **vagrant-libvirt** — Vagrant plugin for libvirt/KVM backend
+- **libvirt** — hypervisor for local VMs
+- **Virtualization** — enabled in BIOS/UEFI
+
+Follow [instructions](https://github.com/theforeman/forklift/blob/master/docs/vagrant.md) to install Vagrant and libvirt.
+
+Check with:
+```bash
+command -v python3 && command -v vagrant && vagrant plugin list | grep -q vagrant-libvirt && echo "All prerequisites met"
+```
+
+## Step 1 — Set up environment
+
+Set up the Python virtualenv and build Ansible collections.
+
+Run from the project root:
+
+```bash
+./setup-environment
+source .venv/bin/activate
+```
+
+Validate that `build/collections/foremanctl` and `build/collections/forge` directories exist. Report Python and Ansible versions.
+
+`./setup-environment` is idempotent — safe to re-run if `.venv` already exists.
+
+## Step 2 — Provision VMs
+
+Provision Vagrant VMs (quadlet + client) via libvirt.
+
+Check that `vagrant` is in PATH and `vagrant-libvirt` plugin is installed (`vagrant plugin list | grep vagrant-libvirt`). Show `vagrant status`.
+
+```bash
+source .venv/bin/activate
+./forge vms start
+```
+
+Validate that `inventories/local_vagrant` was generated — show its contents.
+
+**Troubleshooting:**
+- **vagrant-libvirt not found** — install with `vagrant plugin install vagrant-libvirt`
+- **libvirt permission denied** — ensure the user is in the `libvirt` group
+- **No inventory generated** — check `./forge vms start` output for provisioning errors
+
+## Step 3 — Deploy dev environment
+
+Deploy Foreman and all supporting services to the target VM.
+
+**Plugins** — each selected plugin is git-cloned into its own directory under the deployment dir (`/home/<dev-user>/`, e.g. `/home/vagrant/katello/`, `/home/vagrant/foreman_remote_execution/`). Plugins are local path gems — source edits are picked up immediately by the Rails server. Ask the user to select using `AskUserQuestion` with `multiSelect: true`:
+
+- `foreman_remote_execution` — Remote Execution
+- `foreman_ansible` — Ansible integration
+- `foreman_rh_cloud` — Red Hat Cloud
+- `foreman_discovery` — Bare-metal host discovery
+- `foreman_openscap` — OpenSCAP security audits
+- `foreman_bootdisk` — Boot disk provisioning
+- `foreman_theme_satellite` — Satellite theme
+- `foreman_tasks` — Task management
+- `foreman_webhooks` — Webhook notifications
+- `foreman_templates` — Template sync
+- `foreman_leapp` — RHEL in-place upgrades (Leapp)
+- `foreman_puppet` — Puppet integration
+
+Default plugins (always included): `katello`, `foreman_remote_execution`. Make this clear in the question.
+
+**Features** — enable additional infrastructure services. Ask if the user wants any:
+
+- `hammer` — Foreman CLI (git checkout of hammer-cli + plugins)
+- `foreman-proxy` — Smart Proxy (git checkout, registered into Foreman)
+
+Then ask (skip if user provides no input):
+- **GitHub username** — for additional git remotes on checkouts
+- **Target host** — defaults to `quadlet` VM; set to hostname/IP for any other host
+- **Development user** — defaults to `vagrant`; set when the target host uses a different user
+- **Manage repos** — defaults to `true`; set to `false` to skip git cloning (useful when repos are already checked out)
+
+Build the `./forge deploy-dev` command:
+- Each plugin: `--foreman-development-enabled-plugin <name>`
+- Each feature: `--add-feature <name>`
+- GitHub user: `--foreman-development-github-username <user>`
+- Target host: `--target-host <host>`
+- Dev user: `--foreman-development-user <user>`
+- Manage repos: `--manage-repos false`
+
+**SSH authentication** — when `--target-host` is set (non-Vagrant), test SSH key auth before deploying:
+
+```bash
+ssh -o BatchMode=yes -o ConnectTimeout=5 <user>@<host> "echo OK" 2>/dev/null
+```
+
+- If this succeeds, key-based auth works — no password needed.
+- If this fails, the host requires password authentication. Show the full command prefixed with `ANSIBLE_ASK_PASS=true` and tell the user to run it themselves using the `!` prefix so the interactive password prompt works within this session. Do NOT attempt to run `ANSIBLE_ASK_PASS=true` commands directly — the password prompt requires an interactive terminal.
+
+Example:
+```
+! source .venv/bin/activate && ANSIBLE_ASK_PASS=true ./forge deploy-dev [args...]
+```
+
+## Step 4 — Verify deployment
+
+Verify SSH access, container health, systemd services, and Foreman API on the deployed host.
+
+Determine the host and SSH details: if Vagrant was used, extract from `inventories/local_vagrant`; if `--target-host` was used, use that value directly.
+
+```bash
+python3 -c "
+import yaml
+with open('inventories/local_vagrant') as f:
+    inv = yaml.safe_load(f)
+host = inv['all']['hosts']['quadlet']
+print(host.get('ansible_host', ''))
+print(host.get('ansible_ssh_private_key_file', ''))
+"
+```
+
+Run these checks and report results:
+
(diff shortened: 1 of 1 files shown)


.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+- **Vagrant 2.2+** — VM lifecycle management
+- **vagrant-libvirt** — Vagrant plugin for libvirt/KVM backend
+- **libvirt** — hypervisor for local VMs
+- **Virtualization** — enabled in BIOS/UEFI
+
+Follow [instructions](https://github.com/theforeman/forklift/blob/master/docs/vagrant.md) to install Vagrant and libvirt.
+
+Check with:
+```bash
+command -v python3 && command -v vagrant && vagrant plugin list | grep -q vagrant-libvirt && echo "All prerequisites met"
```
stejskalleos, line 43: from `man`:

> The command builtin runs command with args suppressing the normal  shell  func‐
tion  lookup  for command.  Only builtin commands or commands found in the PATH
named command are executed.  If the -p option is supplied, the search for  com‐
mand is performed using a default value for PATH that is guaranteed to find all
of the standard utilities.

I struggle to understand of the benefits here, why not just call `python` or `vagrant`?

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+## Prerequisites
+
+Before starting, verify the following are installed on the local machine:
+
+- **Python 3** — virtualenv and dependencies
+- **Ansible 2.14+** — automation runtime
+- **Vagrant 2.2+** — VM lifecycle management
+- **vagrant-libvirt** — Vagrant plugin for libvirt/KVM backend
+- **libvirt** — hypervisor for local VMs
+- **Virtualization** — enabled in BIOS/UEFI
```
stejskalleos, line 37: Who will maintain the information in this file? Can we point to another document that does this (if we have it in the repo), or at least remove the versioning?

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+**Troubleshooting:**
+- **vagrant-libvirt not found** — install with `vagrant plugin install vagrant-libvirt`
+- **libvirt permission denied** — ensure the user is in the `libvirt` group
+- **No inventory generated** — check `./forge vms start` output for provisioning errors
+
+## Step 3 — Deploy dev environment
+
+Deploy Foreman and all supporting services to the target VM.
+
+**Plugins** — git-clone Foreman plugins into the Rails app as local dev gems. Ask the user to select using `AskUserQuestion` with `multiSelect: true`:
```
stejskalleos: > git-clone Foreman plugins into the Rails app as local dev gems.

This sounds like it will really clone them into the foreman directory. IMHO we should mention the dir structure:

```
/home/vagrant/foreman
/home/vagrant/foreman_plugin
```

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+
+- `foreman_remote_execution` — Remote Execution
```
stejskalleos, line 85: There are many other plugins that can be enabled. IMO, we should explain the logic for getting the plugin, where to clone it, and how to install it, and that's it.

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+## Step 5 — Add features to existing deployment (optional)
+
+Re-run deploy-dev with `--add-feature` to add `hammer` or `foreman-proxy` after initial deployment:
+
+```bash
+source .venv/bin/activate
+./forge deploy-dev --add-feature hammer --add-feature foreman-proxy
+```
+
+## Step 6 — Tear down
```
stejskalleos: Why to tear down VMs after the setup is done? I want to use them.

.agents/skills/dev-setup/SKILL.md
stejskalleos: ... and I don't see dependencies for the rails app itself: ruby, rvm/rbenv(?), postgresql, ...

Is setting up the Ruby on Rails environment part of this skill as well?

stejskalleos (changes requested):

.agents/skills/dev-setup/SKILL.md
arvind4501: good point, looking at `DEVELOPMENT.md` do you think its a right place to point to, that would also need maintanince for version.
 

> ... and I don't see dependencies for the rails app itself: ruby, rvm/rbenv(?), postgresql, ...

https://github.com/theforeman/foremanctl/blob/44d2d2d322c4fe2843af547797ff8c4be5ae73a6/development/roles/foreman_development/tasks/main.yaml#L40 take care of it alredy

.agents/skills/dev-setup/SKILL.md
arvind4501: although step 6 is not part of the complete deployment but still it confuses so i will remove the step and keep it a seprate way if user explictely ask to destory development setup

.agents/skills/dev-setup/SKILL.md
arvind4501: the logic is in https://github.com/theforeman/foremanctl/blob/44d2d2d322c4fe2843af547797ff8c4be5ae73a6/development/roles/foreman_development/tasks/main.yaml#L173 so as a user i would just select the available plugins which i want and rest is taken care by https://github.com/theforeman/foremanctl/blob/44d2d2d322c4fe2843af547797ff8c4be5ae73a6/development/roles/foreman_development/tasks/plugin.yaml

.agents/skills/dev-setup/SKILL.md
arvind4501: I have looked at internal `sat-foreman-dev-setup` and took reference from there for this as that also uses 
```
command -v broker && command -v sshpass && command -v jq && echo "All prerequisites met
```

so thought of reusing similar concept

.agents/skills/dev-setup/SKILL.md
arvind4501: simplified the description, also cloning and setting them on bundler.d is taken care by the role itself, so agent just need to pass the asked plugins

lzap:
I am not huge fan of copying instructions from README to AGENTS or skills. Instead, I prefer brief instructions for agents, like basic guardrails what to do (always run tests) or not to do (never delete cache directory). But then, I write all the technical details as links to README or documentation.

arvind4501:
> I am not huge fan of copying instructions from README to AGENTS or skills. Instead, I prefer brief instructions for agents, like basic guardrails what to do (always run tests) or not to do (never delete cache directory). But then, I write all the technical details as links to README or documentation.

I am not exactly sure what links you are pointing to. can you please point out things that you think can be added just as link. i did for some https://github.com/theforeman/foremanctl/pull/742/changes#diff-ed2297c45e6d70cc402fa884421b6d9cad698cd6fa64d9f745d79bb0ca0ee517R39. thanks

lzap:
Pretty much any command from the skill that is useful information for humans I would move into READMEs. There are some good examples agent-only commands or instructions, do not get me wrong (e.g. how to connect to the box).

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+- Manage repos: `--manage-repos false`
+
+**SSH authentication** — when `--target-host` is set (non-Vagrant), test SSH key auth before deploying:
+
+```bash
+ssh -o BatchMode=yes -o ConnectTimeout=5 <user>@<host> "echo OK" 2>/dev/null
+```
+
+- If this succeeds, key-based auth works — no password needed.
+- If this fails, the host requires password authentication. Show the full command prefixed with `ANSIBLE_ASK_PASS=true` and tell the user to run it themselves using the `!` prefix so the interactive password prompt works within this session. Do NOT attempt to run `ANSIBLE_ASK_PASS=true` commands directly — the password prompt requires an interactive terminal.
```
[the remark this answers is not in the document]
arvind4501: i tend to agree, but we should ideally not describe all possible failure scenarios for ssh?

arvind4501:
> Pretty much any command from the skill that is useful information for humans I would move into READMEs. There are some good examples agent-only commands or instructions, do not get me wrong (e.g. how to connect to the box).

thanks, i have tried to consolidate and use more links.

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+## Step 1 — Set up environment
+
+Set up the Python virtualenv and build Ansible collections.
+
+Run from the project root:
+
+```bash
+./setup-environment
+source .venv/bin/activate
+```
```
lzap: We have this already twice, this introduced a 3rd copy:

```
[lzapleta@yuki foremanctl]$ ag setup-environment
docs/developer/development-environment.md
19:- Run `./setup-environment` and activate the virtual environment

DEVELOPMENT.md
27:./setup-environment
60:./setup-environment
```

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+## Step 2 — Provision VMs
+
+Provision Vagrant VMs (quadlet + client) via libvirt.
+
+Check that `vagrant` is in PATH and `vagrant-libvirt` plugin is installed (`vagrant plugin list | grep vagrant-libvirt`). Show `vagrant status`.
+
+```bash
+source .venv/bin/activate
+./forge vms start
+```
```
lzap, line 47: Again, third copy:

```
[lzapleta@yuki foremanctl]$ ag bin/activate
setup-environment
7:  source .venv/bin/activate
13:printf "Activate the virtualenv 'source .venv/bin/activate'.\n"

DEVELOPMENT.md
28:source .venv/bin/activate
61:source .venv/bin/activate
```

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+- **Development user** — defaults to `vagrant`; set when the target host uses a different user
+- **Manage repos** — defaults to `true`; set to `false` to skip git cloning (useful when repos are already checked out)
+
+Build the `./forge deploy-dev` command:
+- Each plugin: `--foreman-development-enabled-plugin <name>`
+- Each feature: `--add-feature <name>`
+- GitHub user: `--foreman-development-github-username <user>`
+- Target host: `--target-host <host>`
+- Dev user: `--foreman-development-user <user>`
+- Manage repos: `--manage-repos false`
```
lzap, line 76: Idea: This is a copy of `./forge deploy-dev --help` which is a very nice and extensive. You could instruct agents to run the help command to figure out the mapping. But this is not bad.

I wonder if this is not too much information to ask, I can imagine user of LLM probably wants to just get a dev environment and question for "target host" might be confusing. I can imagine that only github user would be the only question + the instruction to use help screen if user asks to further parametrize the deployment.

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+Only perform container, systemd, or log checks if deployment failed or the API does not respond. See the [deployment verification instructions](../../../docs/developer/development-environment.md#verifying-the-deployment) for troubleshooting.
+
+## Step 5 — Add features to existing deployment (optional)
+
+Re-run deploy-dev with `--add-feature` to add `hammer` or `foreman-proxy` after initial deployment:
+
+```bash
+source .venv/bin/activate
+./forge deploy-dev --add-feature hammer --add-feature foreman-proxy
+```
```
lzap, line 104: I think any LLM will figure out this command, this could be dropped.

.agents/skills/dev-setup/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Tear down
+
+Stop and destroy the Vagrant VMs. Only run when the user explicitly asks to tear down:
+
+```bash
+source .venv/bin/activate
+./forge vms stop
+```
```
lzap, line 112:
```
$ ag "vms stop"
DEVELOPMENT.md
79:./forge vms stop
```

lzap (commented):
Thanks for the links, I think you could go further.

It looks like we already have some instructions twice, perhaps a chance to consolidate this in the PR.

.agents/skills/dev-setup/SKILL.md
arvind4501: IMHO ```./setup-environment` is idempotent — safe to re-run if `.venv` already exists.``` line adds value to skill and that is the place we repeat it.

.agents/skills/dev-setup/SKILL.md
arvind4501: thats only used once

.agents/skills/dev-setup/SKILL.md
arvind4501: > Idea: This is a copy of ./forge deploy-dev --help which is a very nice and extensive. You could instruct agents to run the help command to figure out the mapping. But this is not bad.

yes, that can be used sure.


> I wonder if this is not too much information to ask, I can imagine user of LLM probably wants to just get a dev environment and question for "target host" might be confusing. I can imagine that only github user would be the only question + the instruction to use help screen if user asks to further parametrize the deployment.

to me a user here is developer, and as a developer i think its okay to keep these queations as they give a hint of how much customization can be done to their setup. the default experience remains.

.agents/skills/dev-setup/SKILL.md
arvind4501: I think it depends on what model we are using, LLM's are smart to find that out, but i think writing this down helps to reduce halucination.

.agents/skills/dev-setup/SKILL.md
ehelms: I do not see having it here as a bad thing since this skill is intending to define a specific workflow in a specific order. While I am not a huge fan of repeating information either, there is a balance to strike between the main documentation, referencing that and just having a concise repeat of the exact action that the agent needs to take.

.agents/skills/dev-setup/SKILL.md
ehelms: I tend to think it helps rather than hurts to be specific when specific is possible.

ehelms (approved):

.agents/skills/dev-setup/SKILL.md
lzap: Yes, there is no doubt about that, but then we could make a reference to the user-facing documentation for the command.

arvind4501: yes, i added the reference:

> Follow the [development environment prerequisites and setup instructions](../../../DEVELOPMENT.md#development-environment-setup). Run `./setup-environment` from the project root and activate `.venv`; the setup script is idempotent and safe to re-run.

.agents/skills/dev-setup/SKILL.md
lzap: When a link is longer than the command it refers to, sure and generally I agree. I tend to move the balance point a bit against AI because it can always follow links correctly and technically there is no need to keep agent instructions readable for humans. However, less readable instruction file means harder review.

For the record, I submitted these as comments, not as changes requested.

lzap:
For the record, my comments are just idea feel free to consider them and carry on.

ehelms (approved):

ehelms:
I think this looks like a solid start, and we can iterate on agentic things as folks experience and use them.

PR #746, opened 2026-08-11
Add feature removal validation and removability metadata

ehelms:

`--remove-feature` silently does nothing when targeting a flavor feature (e.g. `--remove-feature foreman` on the katello flavor), gives no feedback when targeting a feature that has no teardown code, and `foremanctl features` provides no indication of which features can actually be removed. This makes feature removal confusing and error-prone.

* Change `--remove-feature` to store removal requests in a separate `remove_features` variable (`persist: false`) instead of silently removing from the `features` list via `dest: features`. This preserves the requests for validation before they take effect.
* Compute `enabled_features` as `(flavor_features + features) | difference(remove_features)` so removals are applied after validation.
* Add `validate_feature_removals` filter that blocks three cases: removing a flavor feature, removing an unknown feature, and removing a feature not marked `removable: true` in `features.yaml`.
* Add `unsatisfied_dependencies` filter that validates all feature dependencies are satisfied in the final `enabled_features` set.
* Add a persistence task in `check_features` that updates `parameters.yaml` to remove features from the persisted `features` list after validation passes.
* Add a REMOVABLE column to `foremanctl features` output showing "no (flavor)", "yes", or "no" per feature.
* Add `removable` metadata support to `features.yaml` (no features are marked removable yet -- that happens per-feature as teardown code is written in follow-up work).

Steps to reproduce:

* Run `foremanctl features` and verify the output includes a REMOVABLE column with "no (flavor)" for flavor features and "no" for all others.
* Run `foremanctl deploy --remove-feature foreman` and verify it fails with an error about flavor features.
* Run `foremanctl deploy --remove-feature nonexistent` and verify it fails with an unknown feature error.
* Run `foremanctl deploy --remove-feature ansible` (without first adding it) and verify it fails with a "does not support removal" error.
* Run filter plugin tests: `python -c "import sys; sys.path.insert(0, 'src/filter_plugins'); from foremanctl import validate_feature_removals; print(validate_feature_removals(['foreman'], ['foreman', 'katello']))"`

#### Checklist

### DIFF
--- a/src/filter_plugins/foremanctl.py
+++ b/src/filter_plugins/foremanctl.py
@@ -71,30 +71,45 @@ def available_foreman_plugins(_value):
     return compact_list(plugins)
 
 
-def list_all_features(enabled_features, only_enabled=False):
+def list_all_features(enabled_features, only_enabled=False, flavor_features=None):
     enabled_list = []
     available_list = []
     list_internal = os.environ.get('FOREMANCTL_FEATURES_LIST_INTERNAL', '') == 'true'
+    flavor_features_set = set(flavor_features or [])
+
     for name, meta in FEATURE_MAP.items():
         internal = meta.get('internal', False)
         if internal and not list_internal:
             continue
         description = meta.get('description', '')
+
+        if name in flavor_features_set:
+            removable = 'flavor'
+        elif meta.get('removable', False):
+            removable = 'yes'
+        else:
+            removable = 'no'
+
         if has_feature(enabled_features, name):
-            enabled_list.append((name, 'enabled', internal, description))
+            enabled_list.append((name, 'enabled', internal, removable, description))
         elif not only_enabled:
-            available_list.append((name, 'available', internal, description))
+            available_list.append((name, 'available', internal, removable, description))
 
     if not list_internal:
-        output = [f"{'FEATURE':<25} {'STATE':<12} DESCRIPTION"]
-        for name, state, _internal, description in enabled_list + available_list:
-            output.append(f"{name:<25} {state:<12} {description}")
+        output = [f"{'FEATURE':<25} {'STATE':<12} {'REMOVABLE':<13} DESCRIPTION"]
+        for name, state, _internal, removable, description in enabled_list + available_list:
+            output.append(f"{name:<25} {state:<12} {removable:<13} {description}")
     else:
-        output = [f"{'FEATURE':<25} {'STATE':<12} {'INTERNAL':<8} DESCRIPTION"]
-        for name, state, internal, description in enabled_list + available_list:
-            output.append(f"{name:<25} {state:<12} {internal:<8} {description}")
+        output = [f"{'FEATURE':<25} {'STATE':<12} {'INTERNAL':<8} {'REMOVABLE':<13} DESCRIPTION"]
+        for name, state, internal, removable, description in enabled_list + available_list:
+            output.append(f"{name:<25} {state:<12} {internal:<8} {removable:<13} {description}")
 
     return "\n".join(output)
+
+
+def is_feature_removable(feature_name):
+    """Check if a feature supports removal."""
+    return FEATURE_MAP.get(feature_name, {}).get('removable', False)
 
 
 def invalid_features(features):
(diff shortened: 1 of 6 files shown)


ehelms:
I am slo tempted to remove the `--remove-feature` flag all together until we actually have something.

ekohl:
> I am slo tempted to remove the `--remove-feature` flag all together until we actually have something.

I think that's a good idea. There are some plugins we can't remove yet anyway because in Foreman we can't roll back the database.

ehelms:
> I think that's a good idea. There are some plugins we can't remove yet anyway because in Foreman we can't roll back the database.

@evgeni reminded me that there are some foreman-proxy features that are already removable

src/filter_plugins/foremanctl.py
```diff
@@ -103,6 +117,59 @@ def conflicting_features(features):
...
+    errors = []
+    enabled_set = set(enabled_features)
+
+    for feature in enabled_features:
+        dependencies = get_dependencies_for_feature(feature)
+        for dep in dependencies:
+            if dep not in enabled_set:
+                errors.append(
+                    f"Feature '{feature}' requires '{dep}' which is not enabled"
+                )
```
ekohl: Would it make sense to report all the missing dependencies for a feature?

You could determine the missing ones easily:
```python
missing = set(get_dependencies_for_feature(feature)) - enabled_set
```

src/filter_plugins/foremanctl.py
```diff
@@ -103,6 +117,59 @@ def conflicting_features(features):
...
+        # Check 1: Cannot remove flavor features
+        if feature in flavor_features:
+            errors.append(
+                f"Cannot remove '{feature}' — it is a core feature of the current flavor. "
+                f"Flavor features cannot be removed."
+            )
+            continue
+
+        # Check 2: Feature must exist
+        if feature not in FEATURE_MAP:
```
ekohl: I'm never a fan of `continue`
```suggestion
        # Check 2: Feature must exist
        elif feature not in FEATURE_MAP:
```

src/filter_plugins/foremanctl.py
```diff
@@ -103,6 +117,59 @@ def conflicting_features(features):
...
+            errors.append(
+                f"Cannot remove unknown feature '{feature}'. "
+                f"Run 'foremanctl features' to see available features."
+            )
+            continue
+
+        # Check 3: Feature must support removal
+        if not is_feature_removable(feature):
```
ekohl:
```suggestion
        # Check 3: Feature must support removal
        elif not is_feature_removable(feature):
```

src/vars/defaults.yml
```diff
@@ -3,4 +3,5 @@ certificates_source: default
 database_mode: internal
 tuning: default
 features: []
-enabled_features: "{{ (flavor_features + features) }}"
+remove_features: []
```
ekohl, line 6: Is the idea behind this that you can run some active code to remove the features? I can see that.

I've considered a generic framework in foreman, like `foreman-rake plugin:remove[$plugin]`. This could be used for it.

ekohl (commented):
Overall I like the structure/concept.

Probably not a blocker, but today we can easily remove Foreman Proxy features. So we can easily remove `smart_proxy_ansible`, but not `foreman_ansible`. I know you didn't mark it as removable now so we can deal with that in the future.

jeremylenz:
> > I am slo tempted to remove the `--remove-feature` flag all together until we actually have something.
> 
> I think that's a good idea.

Please also note we've already started to add `--remove-feature` to documentation, so I wouldn't want to see it go away.

jeremylenz (commented):
Can we also [add this to deploy-dev](https://github.com/theforeman/foremanctl/pull/783/changes/c101d50ec9af4d17ae182f25635d4be6e07066eb)?

ehelms:
> Can we also [add this to deploy-dev](https://github.com/theforeman/foremanctl/pull/783/changes/c101d50ec9af4d17ae182f25635d4be6e07066eb)?

I guess... although I don't know what will happen in dev with this. As the `--add-feature` is not seamless (see https://github.com/theforeman/foremanctl/pull/443). Think about the case where you deploy a `foreman-proxy` vs. a smart-proxy via git. What does removing features mean in those two use cases in dev? 

Honestly, I'd prefer to keep this to production for now and solve features cleanly in `deploy-dev`.

jeremylenz (approved):
Okay, prod-only is fine for now as long as you promise to think about devs ;)

Tested and works for me 👍

src/playbooks/_flavor_features/metadata.obsah.yaml
```diff
@@ -6,6 +6,6 @@ variables:
     action: append_unique
   remove_features:
     parameter: --remove-feature
-    help: Additional features to disable in this deployment.
-    action: remove
-    dest: features
+    help: Features to remove from this deployment.
+    action: append_unique
+    persist: false
```
evgeni, line 11: If you don't persist it, won't it come back on next deployment?

src/roles/check_features/tasks/main.yaml
```diff
@@ -1,26 +1,69 @@
...
+      {% for conflict in check_features_conflicts %}
         - {{ conflict }}
       {% endfor %}
 
       These features cannot be enabled together.
   vars:
-    found_conflicts: "{{ enabled_features | conflicting_features }}"
+    check_features_conflicts: "{{ enabled_features | conflicting_features }}"
+
+- name: Persist feature removals
```
evgeni, line 54: Ooooh.
Ewwww. (Sorry)

ehelms: Is that a "ooh here it is" or a "ewww, why does it need to happen here, maybe it should be in Obsah" ?

evgeni: It's both. My other comment was "If you don't persist it, won't it come back on next deployment?" which is answered by the "ooooh, it's persisted here", quickly followed by the "ewwww, this looks like a terrible hack" (it reminded me of https://github.com/theforeman/obsah/pull/125)

PR #749, opened 2026-08-12
Don't mention Foreman in the finish message

evgeni:

*

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/roles/post_install/tasks/message.yaml
+++ b/src/roles/post_install/tasks/message.yaml
@@ -6,5 +6,5 @@
       - "{{ _post_install_url_msg }}"
       - "{{ _post_install_cred_msg if (post_install_done_flag is not exists) else '' }}"
   vars:
-    _post_install_url_msg: "Foreman is running at {{ post_install_foreman_server_url }}"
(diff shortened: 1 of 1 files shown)


ekohl (approved):

src/roles/post_install/tasks/message.yaml
```diff
@@ -6,5 +6,5 @@
       - "{{ _post_install_url_msg }}"
       - "{{ _post_install_cred_msg if (post_install_done_flag is not exists) else '' }}"
   vars:
-    _post_install_url_msg: "Foreman is running at {{ post_install_foreman_server_url }}"
+    _post_install_url_msg: "You can log in at {{ post_install_foreman_server_url }}"
```
Gauravtalreja1, line 9: Or we could also mention "Server is running at ..."?

ekohl: We should avoid using `server`. Foreman is a service, not a server. In the future, a service may be served by multiple servers behind a load balancer. Even if we're not there, let's get our phrasing right.

PR #750, opened 2026-08-12
Reduce deploy time by ~60s with parallel restarts and faster healthchecks

ehelms:

The deploy playbook spends ~60 unnecessary seconds on serial service restarts, redundant API polling loops, and slow healthcheck intervals. The three dynflow-sidekiq instances restart one at a time (~30s total), the Foreman API readiness check runs as two separate polling loops with 5s delays, and PostgreSQL/Candlepin containers use the default 30s Podman healthcheck interval with `sdnotify: healthy`, meaning systemd can wait up to 30s after the service is actually ready before being notified.

* Restart dynflow-sidekiq@ instances in parallel using async/poll instead of a serial loop, matching the pattern already used for service starts
* Combine the two sequential Foreman API wait tasks (ping + foreman_tasks readiness) into a single polling loop with a 2s delay instead of two loops with 5s delays
* Set `healthcheck_interval: 5s` on PostgreSQL and Candlepin containers so the `sdnotify: healthy` signal fires within 5s of the service becoming ready instead of up to 30s

Steps to reproduce:

* Deploy with `./foremanctl deploy --foreman-initial-admin-password=changeme --tuning development`
* Enable `ansible.posix.profile_tasks` callback to compare task timings before and after
* Verify PostgreSQL restart drops from ~31s to ~6s
* Verify dynflow-sidekiq restart no longer appears in the top-20 slowest tasks
* Verify Candlepin restart drops from ~31s to ~21s
* Verify Foreman services start and the API becomes accessible
* Verify Pulp smart proxy registration completes successfully

#### Checklist

### DIFF
--- a/src/roles/candlepin/tasks/main.yml
+++ b/src/roles/candlepin/tasks/main.yml
@@ -98,6 +98,7 @@
         [Service]
         TimeoutStartSec=300
     healthcheck: curl --fail --insecure --noproxy localhost https://localhost:23443/candlepin/status
+    healthcheck_interval: 5s
     sdnotify: healthy
 
 - name: Run daemon reload to make Quadlet create the service files
--- a/src/roles/foreman/handlers/main.yml
+++ b/src/roles/foreman/handlers/main.yml
@@ -9,11 +9,27 @@
     name: foreman
     state: restarted
 
-- name: Restart dynflow-sidekiq@
+- name: Restart dynflow-sidekiq@ instances
+  listen: Restart dynflow-sidekiq@
   ansible.builtin.systemd:
     name: "dynflow-sidekiq@{{ item }}"
     state: restarted
+  async: 120
+  poll: 0
   loop:
     - orchestrator
     - worker
     - worker-hosts-queue
+  register: foreman_dynflow_restart
+
+- name: Wait for dynflow-sidekiq@ restarts
+  listen: Restart dynflow-sidekiq@
+  ansible.builtin.async_status:
+    jid: "{{ item.ansible_job_id }}"
+  register: foreman_dynflow_restart_result
+  until: foreman_dynflow_restart_result is finished
+  retries: 120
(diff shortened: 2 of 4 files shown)


Alleny244:
# Test Results

Tested with `foremanctl-3.0.0.develop-4.20260812194851262255.pr750.203.gad54455.el9` on RHEL 9.8 (Satellite stream 207.0, containerized).

The deploy completed successfully:

```text
ok=233 changed=19 failed=0
```

## Issue Identified: `Wait for dynflow-sidekiq@ restarts` Handler Never Runs

I verified this on a live Satellite by deleting a dynflow secret to trigger the `Restart dynflow-sidekiq@` handler.

* The `Restart dynflow-sidekiq@` handler ran for all 3 instances.
* The `Wait for dynflow-sidekiq@ restarts` handler using `async_status` never ran.
* The two handlers have different names, so the second handler is never notified.
* Ansible only runs handlers that are directly notified or are listening to the notified topic.
* Since the async restart uses `poll: 0`, the playbook continues without waiting for the restarts to finish.

The instances happened to restart within ~1 second during testing, so nothing failed afterwards. However, there is a race condition that could cause failures under load.

**Fix:** Add `listen: Restart dynflow-sidekiq@` to both handlers.


## Additional Verification

* PostgreSQL `HealthInterval=5s`: restart took ~16s, down from ~31s.
* Candlepin `HealthInterval=5s`: restart took ~14s, down from ~31s.
* `dynflow-sidekiq` instances restart in parallel using `async` / `poll: 0`.
* Combined API readiness check is working with the old task removed and a 2s delay.
* `foreman_tasks` is healthy (`hammer ping`).
* 2 smart proxies are registered and working.
* All 16 containers are running after deployment.

ehelms:
Updated, thanks @Alleny244

Alleny244 (approved):

PR #704, opened 2026-07-24
Update iop documentation

ehelms:
#### Why are you introducing these changes? (Problem description, related links)

The IOP documentation covered the basics (enabling, service list, databases, config) but was missing significant architectural detail that exists in the source code. Operators and developers had to read Ansible roles to understand data flow, deployment ordering, FDW setup, the CVE map downloader, systemd integration, and backup behavior.

#### What are the changes introduced in this pull request?

* Add data flow section with end-to-end pipeline narrative and Kafka topic table
* Add deployment order section documenting the 16-step orchestration from `iop_core`
* Add smart proxy registration details (OAuth, relay config)
* Add systemd integration section (foreman.target, oneshot/timer patterns, timer table)
* Add foreign data wrapper (FDW) section with `iop_fdw` role explanation and ASCII diagram
* Add CVE map downloader section with online/offline modes and reposync trigger
* Add VMAAS-Katello integration section explaining repo list sync
* Add backup and restore section listing the 5 database dumps
* Add container inventory summary table (~23 running + 3 oneshot/timer)
* Expand services table with inventory sub-containers and vulnerability container breakdown
* Add inventory frontend (was missing from the diagram and frontend assets section)
* Update configuration table with OAuth consumer key/secret variables
* Expand certificates table with all IOP cert paths including CVE map downloader and VMAAS
* Note Kafka's separate image, image pre-pull, and engine `console.redhat.com` hosts mapping

#### How to test this pull request

Steps to reproduce:

* Read `docs/iop.md` and verify the new sections are accurate against the source roles in `src/roles/iop_*`
* Verify the deployment order matches `src/roles/iop_core/tasks/main.yaml`
* Verify the Kafka topics match `src/roles/iop_kafka/files/kafka/init`
* Verify the FDW description matches `src/roles/iop_fdw/tasks/main.yaml`
* Verify the CVE map downloader description matches `src/roles/iop_cvemap_downloader/`

#### Checklist
* [ ] Tests added/updated (if applicable)
* [x] Documentation updated (if applicable)

PR #755, opened 2026-08-14
Ensure /var/lib/foreman exists for IOP manual override files

jeremylenz:

Investigating SAT-48087, the cvemap/vex "manual override" files used for offline/disconnected installs were hardcoded to `/var/lib/foreman/cvemap.xml` and `/var/lib/foreman/vex-latest.tar.zst`, but nothing in foremanctl ever created that directory or guaranteed it existed, so disconnected installs had no reliable way to seed vulnerability data.

(This PR originally relocated the files to `/var/lib/foremanctl/iop/`; after review discussion it was decided to keep `/var/lib/foreman` to avoid breaking any existing disconnected-mode users on upgrade, and instead just ensure the directory exists with proper permissions.)

* Add a task to each role (`iop_cvemap_downloader`, `iop_vex_downloader`) that explicitly creates `/var/lib/foreman` (root:root, mode `0755`) on deploy
* Introduce `iop_cvemap_downloader_manual_dir` / `iop_vex_downloader_manual_dir` role defaults holding just the directory, with well-known filenames (`cvemap.xml`, `vex-latest.tar.zst`) appended where needed, so additional manual-override files can be added later without duplicating the path
* Convert the previously static systemd path units (and the vex downloader script) to Jinja templates so they reference these defaults instead of a hardcoded path
* Update the `iop_cvemap_downloader` / `iop_vex_downloader` feature tests to assert the directory is created
* Document the offline/disconnected workflow in `docs/iop.md` (previously undocumented)

Refs SAT-48087

Steps to reproduce:

* Deploy a dev VM with the IOP/Satellite flavor enabled
* Confirm `/var/lib/foreman` exists (root:root, mode `0755`) after deploy
* Manually place a `cvemap.xml` (or `vex-latest.tar.zst`) file at `/var/lib/foreman/`
* Confirm the corresponding systemd path unit (`iop-cvemap-download.path` / `iop-vex-download.path`) fires and the download/copy service picks up the manual file without a manual restart
* Confirm `/var/www/html/pub/...` is updated with the copied file

#### Checklist
* [x] Tests added/updated (feature tests updated for directory creation; `ansible-lint` passes for changed files)
* [x] Documentation updated (`docs/iop.md`)
* [ ] `./forge test` run against a deployed dev VM to validate the systemd path-watcher behavior end-to-end (not yet run in this environment)

### DIFF
--- a/src/roles/iop_cvemap_downloader/defaults/main.yaml
+++ b/src/roles/iop_cvemap_downloader/defaults/main.yaml
@@ -3,7 +3,7 @@ iop_cvemap_downloader_base_url: "https://security.access.redhat.com/data/meta/v1
 iop_cvemap_downloader_timer_interval: "24h"
 iop_cvemap_downloader_output_dir: "/var/www/html/pub"
 iop_cvemap_downloader_relative_path: "iop/data/meta/v1/cvemap.xml"
-iop_cvemap_downloader_manual_file: "/var/lib/foremanctl/iop/cvemap.xml"
+iop_cvemap_downloader_manual_dir: "/var/lib/foremanctl/iop"
 
 iop_cvemap_downloader_client_cert: "/var/lib/foremanctl/certs/certs/{{ ansible_facts['fqdn'] }}-client.crt"
 iop_cvemap_downloader_client_key: "/var/lib/foremanctl/certs/private/{{ ansible_facts['fqdn'] }}-client.key"
--- a/src/roles/iop_cvemap_downloader/defaults/main.yaml
+++ b/src/roles/iop_cvemap_downloader/defaults/main.yaml
@@ -3,6 +3,7 @@ iop_cvemap_downloader_base_url: "https://security.access.redhat.com/data/meta/v1
 iop_cvemap_downloader_timer_interval: "24h"
 iop_cvemap_downloader_output_dir: "/var/www/html/pub"
 iop_cvemap_downloader_relative_path: "iop/data/meta/v1/cvemap.xml"
+iop_cvemap_downloader_manual_file: "/var/lib/foremanctl/iop/cvemap.xml"
 
 iop_cvemap_downloader_client_cert: "/var/lib/foremanctl/certs/certs/{{ ansible_facts['fqdn'] }}-client.crt"
 iop_cvemap_downloader_client_key: "/var/lib/foremanctl/certs/private/{{ ansible_facts['fqdn'] }}-client.key"
--- a/src/roles/iop_cvemap_downloader/defaults/main.yaml
+++ b/src/roles/iop_cvemap_downloader/defaults/main.yaml
@@ -3,7 +3,7 @@ iop_cvemap_downloader_base_url: "https://security.access.redhat.com/data/meta/v1
 iop_cvemap_downloader_timer_interval: "24h"
 iop_cvemap_downloader_output_dir: "/var/www/html/pub"
 iop_cvemap_downloader_relative_path: "iop/data/meta/v1/cvemap.xml"
-iop_cvemap_downloader_manual_dir: "/var/lib/foremanctl/iop"
+iop_cvemap_downloader_manual_dir: "/var/lib/foreman"
 
 iop_cvemap_downloader_client_cert: "/var/lib/foremanctl/certs/certs/{{ ansible_facts['fqdn'] }}-client.crt"
 iop_cvemap_downloader_client_key: "/var/lib/foremanctl/certs/private/{{ ansible_facts['fqdn'] }}-client.key"
--- a/src/roles/iop_cvemap_downloader/tasks/main.yaml
+++ b/src/roles/iop_cvemap_downloader/tasks/main.yaml
@@ -1,4 +1,10 @@
 ---
+- name: Create IOP manual override directory
+  ansible.builtin.file:
+    path: "{{ iop_cvemap_downloader_manual_file | dirname }}"
+    state: directory
+    mode: '0755'
+
 - name: Install cvemap download script
   ansible.builtin.template:
     src: iop-cvemap-download.sh.j2
@@ -33,8 +39,8 @@
   notify: Restart cvemap download timer
 
 - name: Create cvemap download path unit
-  ansible.builtin.copy:
-    src: iop-cvemap-download.path
+  ansible.builtin.template:
+    src: iop-cvemap-download.path.j2
     dest: /etc/systemd/system/iop-cvemap-download.path
     mode: '0644'
   notify: Restart cvemap download path
--- a/src/roles/iop_cvemap_downloader/tasks/main.yaml
+++ b/src/roles/iop_cvemap_downloader/tasks/main.yaml
@@ -3,6 +3,8 @@
   ansible.builtin.file:
     path: "{{ iop_cvemap_downloader_manual_dir }}"
     state: directory
+    owner: root
+    group: root
     mode: '0755'
 
 - name: Install cvemap download script
--- a/src/roles/iop_cvemap_downloader/tasks/main.yaml
+++ b/src/roles/iop_cvemap_downloader/tasks/main.yaml
@@ -1,7 +1,7 @@
 ---
 - name: Create IOP manual override directory
   ansible.builtin.file:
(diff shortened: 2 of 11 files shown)


docs/iop.md
```diff
@@ -141,3 +141,18 @@ iop_engine_extra_packages:
   - "prodsec.rules"
   - "telemetry.rules.plugins"
 ```
+
+### Offline Vulnerability Data (Disconnected Installs)
```
vkrizan: Potential conflict with #704

src/roles/iop_cvemap_downloader/defaults/main.yaml
```diff
@@ -3,6 +3,7 @@ iop_cvemap_downloader_base_url: "https://security.access.redhat.com/data/meta/v1
 iop_cvemap_downloader_timer_interval: "24h"
 iop_cvemap_downloader_output_dir: "/var/www/html/pub"
 iop_cvemap_downloader_relative_path: "iop/data/meta/v1/cvemap.xml"
+iop_cvemap_downloader_manual_file: "/var/lib/foremanctl/iop/cvemap.xml"
```
vkrizan: Since we're gonna add a few more files, could the variables carry just the folder? Filenames under would then be well known.

ehelms:
> /var/lib/foreman is not a real path in foremanctl's containerized architecture

For awareness, the RPM creates it -- https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/foremanctl/foremanctl.spec#L65 so I think this statement needs qualification of which use case you mean.

If `/var/lib/foreman` doesn't exist, then how would a sub-directory exist? Putting all of these files under a sub-directory is not a bad idea. I want to ensure we capture the real reasons behind the change.

jeremylenz:
> For awareness, the RPM creates it -- https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/foremanctl/foremanctl.spec#L65 so I think this statement needs qualification of which use case you mean.

The RPM creates `/var/lib/foremanctl/` but does not create `/var/lib/foreman/`, right? I think it's the right move to use `foremanctl/` rather than `foreman/`

jeremylenz:
@vkrizan Addressed your comments as well.

ehelms:
> The RPM creates `/var/lib/foremanctl/` but does not create `/var/lib/foreman/`, right? I think it's the right move to use `foremanctl/` rather than `foreman/`

Oh right, keeping it as `/var/lib/foreman` would help disconnected users that upgrade. I also wonder about using a podman volume to store these files and make the timer a proper container that runs like the others. And then having the gateway mount and serve these files. CC @vkrizan

vkrizan:
> I also wonder about using a podman volume to store these files and make the timer a proper container that runs like the others. And then having the gateway mount and serve these files. 

@ehelms That is what we are exploring for the content that Insights produces. For frequently updated content, use of containers as volumes isn't ideal, as it requires restarts of the container that serves them, plus there is some maint. costs. We could move the serving of the content from the Apache httpd to IoP Gateway. However, for the cvemap.xml and other frequently updated content I would stick to regular (shared) volumes.

ehelms:
> @ehelms That is what we are exploring for the content that Insights produces. For frequently updated content, use of containers as volumes isn't ideal, as it requires restarts of the container that serves them, plus there is some maint. costs. We could move the serving of the content from the Apache httpd to IoP Gateway. However, for the cvemap.xml and other frequently updated content I would stick to regular (shared) volumes.

A podman volume is exactly a shared volume. What am I missing?

vkrizan:
@ehelms Oh, right. I've misread, I thought you've meant container as volume... Yeah, a podman volume could do. Generally, these ideally need to be accessed and managed from a container. Therefore, all operations for the disconnected instances use should have a container run to copy&distribute the content. I think this needs a bit more discussion, as I'm assuming that the processes to update disconnected instances are on the system level done by an sysadmin.

jeremylenz:
Rebased in hopes that the new CI run will be greener.

jeremylenz:
:green_apple: :green_apple:

ehelms:
I keep coming back to not being a huge fan of tieing the runtime data to the deployment directory. I think I'd still rather just update the code to ensure `/var/lib/foreman` exists right now. That also keeps upgrades cleaner. That would simplify your change here as something just needs to ensure that directory exists.


> @ehelms Oh, right. I've misread, I thought you've meant container as volume... Yeah, a podman volume could do. Generally, these ideally need to be accessed and managed from a container. Therefore, all operations for the disconnected instances use should have a container run to copy&distribute the content. I think this needs a bit more discussion, as I'm assuming that the processes to update disconnected instances are on the system level done by an sysadmin.

I am going to post to an issue some of my design ideas as I think they are rather large for this PR and distracting it.

jeremylenz:
> I keep coming back to not being a huge fan of tieing the runtime data to the deployment directory. I think I'd still rather just update the code to ensure `/var/lib/foreman` exists right now. That also keeps upgrades cleaner. That would simplify your change here as something just needs to ensure that directory exists.

I don't think `/var/lib/foreman` is a logical place to put these files in the new architecture. It kinda implies that it's a functional directory and somehow related to the `foreman` container when it's not actually. `/var/lib/foremanctl` fits much better IMO.

(also, we already tie runtime data to the deployment directory - for example, auth-bundle tarballs are provided by the user and then placed by `src/roles/certificates/tasks/main.yml` in `/var/lib/foremanctl/certs`.)

However, your concern about upgrades is valid. How about something like this:

* Keep `/var/lib/foremanctl/iop` as the watched location
* Add a one-time migration task that copies the file from the old location (if it exists) to the new one
* Don't copy again if the new location already has a file
* Persistent deprecation warning if the old file still exists, warning that the old path is no longer watched
* Migration task and warning could be dropped some time in the future

thoughts?

ehelms:
> (also, we already tie runtime data to the deployment directory - for example, auth-bundle tarballs are provided by the user and then placed by `src/roles/certificates/tasks/main.yml` in `/var/lib/foremanctl/certs`.)

This isn't runtime data -- it's deployment data tied to foremanctl. So I am still hesitant to tie application runtime data to the deployment data directory. 


> I don't think `/var/lib/foreman` is a logical place to put these files in the new architecture. It kinda implies that it's a functional directory and somehow related to the `foreman` container when it's not actually.

By that logic, the directory should be `/var/lib/iop` if we want to isolate data based on what it directly relates to. We have previously used `/var/lib/foreman` to store persistent data related to anything related to foreman. I do think we should also look at serving these from the gateway instead of Apache. I can't recall why we didn't do that the first time.

jeremylenz:
> By that logic, the directory should be `/var/lib/iop` if we want to isolate data based on what it directly relates to. We have previously used `/var/lib/foreman` to store persistent data related to anything related to foreman. I do think we should also look at serving these from the gateway instead of Apache. I can't recall why we didn't do that the first time.

Since it looks like you feel strongly against the `foremanctl` location, I feel we can do one of the following _for now_:

1. keep `/var/lib/foreman` and change the PR here as you suggested, to simply ensure it exists
2. use `/var/lib/iop`

And later, we can decide to serve that from a container, which I agree is the best long-term solution.

(1), while I still dislike it, would possibly require fewer documentation updates.
(2) seems like a logical place for it and I prefer it slightly over (1), but again, based mostly on feelings.

@ehelms thoughts?

jeremylenz:
Okay fiiiiine, _for now_ I've updated this to use `/var/lib/foreman/`. Currently it creates it with root:root but I am not sure that will be correct for all use cases (I'm thinking rh_cloud inventory reports.)

ehelms (approved):

ehelms:
@jeremylenz can you rebase this?

src/roles/iop_cvemap_downloader/tasks/main.yaml
```diff
@@ -1,4 +1,10 @@
 ---
+- name: Create IOP manual override directory
+  ansible.builtin.file:
+    path: "{{ iop_cvemap_downloader_manual_dir }}"
+    state: directory
+    mode: '0755'
```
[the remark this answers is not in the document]
ehelms: I agree with this change.

PR #756, opened 2026-08-15
stop pulp services in case of databse migration failure

bochi:
When upgrading from 3.18 to 3.19 with a major pulpcore bump (3.85 -> 3.105) the 0143_require_app_lock_zdu failed while migrating the database schema because of a guard in the migration and the old pulpcore services still running

* Stop pulp services in case of database migration failure and retry 

Steps to reproduce:

* Upgrade from 3.18 to 3.19 using foremanctl by changing the image versions, see the pulp db migration hangs

#### Checklist

### DIFF
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -270,9 +270,21 @@
     daemon_reload: true
 
 - name: Migrate the Pulp database
-  ansible.builtin.systemd:
-    name: pulpcore-manager-migrate.service
-    state: restarted
+  block:
+    - name: Run Pulp database migration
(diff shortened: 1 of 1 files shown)


ehelms:
Did you stop all services using the `foreman.target` before you ran deploy when doing the upgrade? See https://github.com/theforeman/foremanctl/blob/master/docs/developer/deployment.md#container-changes-upgrades

bochi:
You are right I did not follow the docs :) The good news is that an online (almost zero downtime) upgrade appears to be working though with these changes. But I agree it is better to tear down foreman.target prior to an upgrade. So, closing.

PR #757, opened 2026-08-17
Add forge host-setup and macOS/Apple Silicon support

### LINKED ISSUE #685: Add forge subcommand to bootstrap host prerequisites (Vagrant, libvirt / vagrant-qemu)
## Problem

Setting up a `forge`-based development environment currently requires manually installing Vagrant and a hypervisor provider on the host before running `./forge vms start`. `DEVELOPMENT.md` just links out to forklift's [`docs/vagrant.md`](https://github.com/theforeman/forklift/blob/master/docs/vagrant.md) for this, and there's no automated path — unlike forklift, which has a dedicated [`roles/vagrant`](https://github.com/theforeman/forklift/blob/master/roles/vagrant/tasks/main.yml) role that installs Vagrant/libvirt/vagrant-libvirt, enables `libvirtd`, and adds the user to the `libvirt` group.

This came up while working on [theforeman/jenkins-jobs#588](https://github.com/theforeman/jenkins-jobs/pull/588), which provisions a `forge`-driven install test on Duffy hosts — those hosts need the same host-level bootstrap before `forge`/`foremanctl` can run.

## Proposal

Add a new, **opt-in** `forge` subcommand (e.g. `forge host setup` — naming TBD) that provisions the *host itself* for local development. This is a distinct, one-time step in the environment setup flow — run manually by the user *before* `./setup-environment` / `./forge vms start`, not automatically triggered by them. Users who already have their host provisioned (or want to manage it themselves) simply skip it.

- **Linux (Fedora / CentOS Stream):** Install Vagrant + the libvirt provider stack (mirrors forklift's `roles/vagrant`: `libvirt-daemon-kvm`, `vagrant-libvirt`, enabling `libvirtd`/`virtnetworkd`, COPR repos where needed). Add the invoking user to the `libvirt` group.
- **macOS on Apple Silicon (ARM):** Install Vagrant + the [`vagrant-qemu`](https://github.com/ppggff/vagrant-qemu) provider, since `vagrant-libvirt` isn't an option there 😂 (yes, I have an M-series MacBook now, don't @ me). Needs its own box/provider selection in the `Vagrantfile`/vars, since boxes aren't 1:1 portable between libvirt and qemu providers.
- Runs with elevated privileges since it's modifying host packages/services — clearly separate from the existing `forge vms` playbooks, which only manage guest VMs and should never implicitly touch the host.

Open questions to resolve during implementation:
- Reimplement the logic in `development/roles/`, or depend on/reuse the `theforeman.forklift` collection's `vagrant` role directly for the Linux/libvirt path (foremanctl already pulls in the forklift collection transitively for VM lifecycle)? The qemu/macOS path would need new logic regardless, since forklift doesn't support it today.
- How does box selection in the `Vagrantfile` need to branch per-provider/per-arch (libvirt boxes vs qemu-compatible boxes)?
- Any macOS-specific package manager assumptions (Homebrew) to bake in for installing Vagrant/vagrant-qemu?
- Where does this fit relative to `./setup-environment` — a flag on it, or a fully separate command documented as an earlier optional step in `DEVELOPMENT.md`?

## Why this matters

- Reduces friction for new contributors following `DEVELOPMENT.md`.
- Needed for automated CI hosts (e.g. Duffy-provisioned boxes in jenkins-jobs#588) that need a repeatable, non-interactive way to get `forge` runnable from a bare host.
- Opens up local development on Apple Silicon MacBooks, which currently isn't documented/supported at all.

Odilhao:

Closes #685. There was no automated way to bootstrap a host for `forge`, and no way to run it on Apple Silicon at all: foremanctl's containers are x86_64-only, and vagrant-qemu needs a different box format and networking setup than vagrant-libvirt.

* `forge bootstrap-vagrant`, a new command that installs Vagrant and a hypervisor provider (libvirt on Linux, vagrant-qemu on macOS) through forklift's vagrant role, see theforeman/forklift#1981
* On Apple Silicon, the Vagrantfile boots a native arm64 CentOS Stream 10 guest through vagrant-qemu instead of emulating the whole VM
* socket_vmnet gives the VM a real IP the same way libvirt does on Linux, and vagrant-hostmanager keeps `/etc/hosts` in sync so the UI is reachable by hostname
* A new `container_binfmt` role registers qemu-user-static on the guest so it can still run the existing x86_64 container images
* Fixed `setup-environment`, which called `python` instead of `python3` and never worked on macOS at all

This depends on theforeman/forklift#1981, which isn't merged yet, so CI/local testing against a fresh checkout won't pick up the new vagrant role until that lands.

Steps to reproduce:

* On an Apple Silicon Mac, run `./forge bootstrap-vagrant` once
* `./forge vms start` to bring up the quadlet VM
* `./foremanctl deploy --initial-admin-password=changeme --tuning development`
* Confirm `https://quadlet.example.com` loads in a browser

Timed a fresh run on an M-series Mac, box image already cached: `forge vms start` took about 1m20s, `foremanctl deploy` about 16m, UI reachable right after.

Only the quadlet VM works this way for now, client/database/proxy still need libvirt. Also worth flagging: `./forge test` against the qemu VM currently fails several API tests, because the test suite assumes the Foreman API is reachable at the same address as SSH, which is true for libvirt but not for vagrant-qemu's separate NAT and vmnet interfaces. That's a known gap, not something this PR fixes.

#### Checklist


development/playbooks/bootstrap-vagrant/metadata.obsah.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
ehelms, line 1: I think `host-setup` could be confused with other meanings. How about `setup-vagrant` or `bootstrap-vagrant` ?

development/playbooks/container_binfmt.yaml
```diff
@@ -0,0 +1,45 @@
...
+            content: |
+              [Service]
+              TimeoutStartSec=600
+            mode: "0644"
+          loop:
+            - foreman.service
+            - "dynflow-sidekiq@.service"
+          notify: Reload systemd after adding binfmt timeout drop-ins
+
+  handlers:
```
ehelms: Given this is a playbook, why is the handler needed? This could just run the reload as the next task.

development/playbooks/container_binfmt.yaml
```diff
@@ -0,0 +1,45 @@
...
+- name: Register binfmt support for foreign-architecture containers
+  become: true
+  hosts:
+    - all
+  gather_facts: true
+  tasks:
+    - name: Register binfmt support
+      ansible.builtin.include_role:
+        name: container_binfmt
+      when: ansible_facts['architecture'] != 'x86_64'
```
ehelms, line 9: Given this when condition, you could move this down into the block? Why not put what's in this playbook into the role since they share the same name?

docs/developer/development-environment.md
```diff
@@ -13,9 +13,11 @@ The development environment provides:
 
 ## Prerequisites
 
+- A host with Vagrant and a hypervisor provider installed. Run `./forge host-setup` once to bootstrap this (libvirt on Linux, vagrant-qemu on Apple Silicon macOS).
```
ehelms: This is technically only true in some cases, that is why the statement after this says "A running EL9 virtual machine..". I'd prefer to see us update this to indicate Vagrant as a target, then add a Vagrant section with these Vagrant specific details.

Vagrantfile
```diff
@@ -1,8 +1,53 @@
...
+    FileUtils.mkdir_p(FOREMANCTL_CACHE_DIR)
+    system('ssh-keygen', '-t', 'ed25519', '-N', '', '-C', 'foremanctl-quadlet-qemu', '-f', private_key_path, '-q') || raise('Failed to generate ssh keypair for the qemu quadlet VM')
+  end
+  [private_key_path, "#{private_key_path}.pub"]
+end
+
 Vagrant.configure("2") do |config|
   config.vm.synced_folder ".", "/vagrant"
 
+  if ENV['VAGRANT_DEFAULT_PROVIDER'] == 'qemu' && Vagrant.has_plugin?('vagrant-hostmanager')
```
ehelms, line 63: Can you expand on why this is being added? In a previous PR, this plugin was discussed and not implemented. Here are the details to help: https://github.com/theforeman/foremanctl/pull/500#pullrequestreview-4317360459

Vagrantfile
```diff
@@ -17,17 +62,63 @@ Vagrant.configure("2") do |config|
...
-      override.vm.box_url = "https://cloud.centos.org/centos/10-stream/x86_64/images/CentOS-Stream-Vagrant-10-latest.x86_64.vagrant-libvirt.box"
+    unless ENV['VAGRANT_DEFAULT_PROVIDER'] == 'qemu'
+      override.vm.box = ENV.fetch("FOREMANCTL_BASE_BOX", "centos/stream9")
+      if override.vm.box == "centos/stream10"
+        override.vm.box_url = "https://cloud.centos.org/centos/10-stream/x86_64/images/CentOS-Stream-Vagrant-10-latest.x86_64.vagrant-libvirt.box"
+      end
     end
     override.vm.hostname = "quadlet.#{DOMAIN}"
 
+    if ENV['VAGRANT_DEFAULT_PROVIDER'] == 'qemu'
```
ehelms, line 99: Can line 71 be moved above the `unless` block and then have the `unless` block merged with this one to form an if-else block?

Vagrantfile
Odilhao: This part is quite complicated on Mac, we don't have the libvirtd option, I tried exploring with libkrun but I did not found anything for DNS. I had one idea during the implementation of having a small container in the mac host running a dnsServer, but it felt too hacky.

development/playbooks/bootstrap-vagrant/metadata.obsah.yaml
Odilhao: Updated to `bootstrap-vagrant`

Vagrantfile
ehelms: CC @ekohl given you had strong thoughts about the plugin.

Odilhao: This was the only working method that I found to keep vagrant as part of the provisioning on M-series, we have https://github.com/libkrun/libkrun, that will use [HVF](https://developer.apple.com/documentation/hypervisor) instead of KVM, I could do some exploration to check how the network part would work on HVF, vagrant-qemu also run with HVF but with no libvirt DNS backend :cry:

Vagrantfile
```diff
@@ -1,11 +1,88 @@
...
+FOREMANCTL_CACHE_DIR = File.join(Dir.home, '.cache', 'foremanctl').freeze
+
+def fetch_qemu_guest_image
+  image_path = File.join(FOREMANCTL_CACHE_DIR, 'CentOS-Stream-GenericCloud-10-latest.aarch64.qcow2')
+  unless File.exist?(image_path)
+    FileUtils.mkdir_p(FOREMANCTL_CACHE_DIR)
+    system('curl', '-fL', '-o', image_path, CENTOS_STREAM10_AARCH64_IMAGE_URL) || raise("Failed to download #{CENTOS_STREAM10_AARCH64_IMAGE_URL}")
+  end
+  image_path
+end
```
ekohl, line 20: Why can't you just pass in the native file path and let Vagrant do its thing?

Vagrantfile
```diff
@@ -1,11 +1,88 @@
...
+    # plugin with the same custom ip_resolver pattern for non-default NICs.
+    config.hostmanager.enabled = true
+    config.hostmanager.manage_host = true
+    config.hostmanager.manage_guest = true
+    config.hostmanager.include_offline = true
+    config.hostmanager.ip_resolver = proc do |vm, _resolving_vm|
+      next unless vm.ssh_info && vm.ssh_info[:host]
+
+      result = ''
+      vm.communicate.execute('ip addr show eth0') { |type, data| result << data if type == :stdout }
```
ekohl, line 76: `result` is a string and `<<` relies on mutable strings. Why not:
```suggestion
      vm.communicate.execute('ip addr show eth0') { |type, data| result = data if type == :stdout }
```

Vagrantfile
```diff
@@ -1,11 +1,88 @@
...
+  end
+end
+
 Vagrant.configure("2") do |config|
   config.vm.synced_folder ".", "/vagrant"
 
-  config.vm.provision("etc_hosts", type: 'ansible') do |ansible|
-    ansible.playbook = "development/playbooks/etc_host.yml"
-    ansible.compatibility_mode = "2.0"
+  if ENV['VAGRANT_DEFAULT_PROVIDER'] == 'qemu' && Vagrant.has_plugin?('vagrant-hostmanager')
```
ekohl, line 63: Normally I'd think you'd use:
```suggestion
  config.vm.provider "qemu" do |qemu|
     if Vagrant.has_plugin?('vagrant-hostmanager')
```

Vagrantfile
```diff
@@ -1,11 +1,88 @@
...
+    config.hostmanager.ip_resolver = proc do |vm, _resolving_vm|
+      next unless vm.ssh_info && vm.ssh_info[:host]
+
+      result = ''
+      vm.communicate.execute('ip addr show eth0') { |type, data| result << data if type == :stdout }
+      (ip = /inet (\d+\.\d+\.\d+\.\d+)/.match(result)) && ip[1]
+    end
+  end
+
+  if ENV['VAGRANT_DEFAULT_PROVIDER'] != 'qemu'
```
ekohl, line 81: I'd love to get rid of this and mandate https://community.theforeman.org/t/setting-up-libvirt-with-dns-resolution-for-vagrant/46567 but otherwise:

```suggestion
  config.vm.provider "libvirt" do |libvirt, provider|
```

ekohl (changes requested):
I had kind of hoped that with EL 10 we would also start building ARM packages and avoid the need for all of that emulation.

Right now this feels very complex.

At least all of the code that uses `VAGRANT_DEFAULT_PROVIDER` needs to be using the proper integration per https://developer.hashicorp.com/vagrant/docs/providers/configuration.

PR #761, opened 2026-08-19
Closes #39652 - unconfigure SMTP as an email delivery method

michalgritzbach:
Part of an effort to remove sendmail delivery method altogether. SAT-43109.

* Rolls back https://github.com/theforeman/foremanctl/pull/386/changes

Since SMTP is now the default, it will be set even without explicitly setting it.

#### Checklist


ekohl (changes requested):
> Since SMTP is now the default, it will be set even without explicitly setting it.

https://github.com/theforeman/foreman/pull/10870 is still open, so it is an option. By having the value set in `settings.yaml` it becomes read-only in the UI/API. Until the option is gone, we should keep this.

PR #762, opened 2026-08-19
Fix: Support systemd property compatibility for CentOS Stream 10

Odilhao:

While testing https://github.com/Katello/katello/pull/11835 at https://github.com/theforeman/foreman-oci-images/pull/90 I found one testing missing that is related to https://github.com/theforeman/foremanctl/commit/1d1617b20fb1a61597c097ebccf6ff31703566a9, looks like the issue started switched to CS10 base containers.

The LastTriggerUSec property was renamed in newer systemd versions. CentOS Stream 10 uses systemd 256+ which reports LastTriggerUSecMonotonic and LastTriggerUSecRealtime instead. Support all three property names to maintain compatibility across systemd versions.

Test Integration needs to run with CS10 based containers.

### DIFF
--- a/tests/feature/foreman/base_test.py
+++ b/tests/feature/foreman/base_test.py
@@ -82,8 +82,9 @@ def test_foreman_recurring_services_exist(server, instance):
 def test_foreman_recurring_timer_last_trigger(server, instance):
     """Verify that timers have a valid last trigger time (if they've run)."""
     timer_name = f"foreman-recurring@{instance}.timer"
-    timer = server.service(timer_name) 
-    assert "LastTriggerUSec" in timer.systemd_properties
+    timer = server.service(timer_name)
+    props = timer.systemd_properties
+    assert 'LastTriggerUSec' in props or 'LastTriggerUSecRealtime' in props
 
 
 @pytest.mark.parametrize("instance", RECURRING_INSTANCES)


ehelms (approved):

tests/feature/foreman/base_test.py
```diff
@@ -82,8 +82,11 @@ def test_foreman_recurring_services_exist(server, instance):
 def test_foreman_recurring_timer_last_trigger(server, instance):
     """Verify that timers have a valid last trigger time (if they've run)."""
     timer_name = f"foreman-recurring@{instance}.timer"
-    timer = server.service(timer_name) 
-    assert "LastTriggerUSec" in timer.systemd_properties
+    timer = server.service(timer_name)
+    props = timer.systemd_properties
+    # LastTriggerUSec was renamed in newer systemd versions (CS10 uses 256+)
+    # Support all three property names for compatibility across systemd versions
+    assert 'LastTriggerUSec' in props or 'LastTriggerUSecRealtime' in props or 'LastTriggerUSecMonotonic' in props
```
ekohl: When I read https://wiki.archlinux.org/title/Systemd/Timers#Examples we don't configure a monotonic timer, only a realtime timer so I think it shouldn't check for `LastTriggerUSecMonotonic`.
```suggestion
    assert 'LastTriggerUSec' in props or 'LastTriggerUSecRealtime' in props
```

tests/feature/foreman/base_test.py
```diff
@@ -82,8 +82,11 @@ def test_foreman_recurring_services_exist(server, instance):
 def test_foreman_recurring_timer_last_trigger(server, instance):
     """Verify that timers have a valid last trigger time (if they've run)."""
     timer_name = f"foreman-recurring@{instance}.timer"
-    timer = server.service(timer_name) 
-    assert "LastTriggerUSec" in timer.systemd_properties
+    timer = server.service(timer_name)
+    props = timer.systemd_properties
+    # LastTriggerUSec was renamed in newer systemd versions (CS10 uses 256+)
+    # Support all three property names for compatibility across systemd versions
```
ekohl: IMHO this is implicit
```suggestion
```

ekohl (commented):
Looks like it was changed in https://github.com/systemd/systemd/commit/dedabea4b3d61a87cedb5c8d7ccce5b86ea84afe so systemd 212.

ekohl:
> The LastTriggerUSec property was renamed in newer systemd versions. CentOS Stream 10 uses systemd 256+ which reports LastTriggerUSecMonotonic and LastTriggerUSecRealtime instead. Support all three property names to maintain compatibility across systemd versions.

This is the wrong way around. systemd prior to 212 used `LastTriggerUSecMonotonic` and `LastTriggerUSecRealtime`, 212+ uses `LastTriggerUSec`.

Odilhao:
> > The LastTriggerUSec property was renamed in newer systemd versions. CentOS Stream 10 uses systemd 256+ which reports LastTriggerUSecMonotonic and LastTriggerUSecRealtime instead. Support all three property names to maintain compatibility across systemd versions.
> 
> This is the wrong way around. systemd prior to 212 used `LastTriggerUSecMonotonic` and `LastTriggerUSecRealtime`, 212+ uses `LastTriggerUSec`.

Fixed the wording on the commit and dropped the comment in the latest commit.

ekohl (approved):
I think a comment that systemd 212 changed it can be useful, but it's also in the commit message which git blame should point out. Thanks!

PR #773, opened 2026-08-21
Add --server-alias support to auth-bundle for load-balanced proxies

pablomh:
## Summary
- Proxies behind a TLS-passthrough load balancer need the load balancer's FQDN as a SAN on their own certificate, since the LB never terminates TLS.
- `--server-alias` previously only existed on `deploy` (main Foreman server). Extracted it into a shared `_server_aliases` obsah fragment and included it from `auth-bundle` too, reusing the existing `certificates_server_aliases` plumbing that already builds SANs generically for whichever hostname is being issued a certificate.
- No changes needed to `deploy-proxy`: it never generates/re-issues certificates, it only consumes the bundle produced by `auth-bundle`.

## Test plan
- [x] `ansible-lint` passes on changed/new obsah metadata
- [x] `ruff check tests/auth_bundle_test.py` passes
- [x] `./foremanctl auth-bundle --help` shows `--server-alias` / `--reset-server-alias`
- [x] `./foremanctl deploy --help` still shows `--server-alias` (no regression from the `_foreman` refactor)
- [ ] CI: `tests/auth_bundle_test.py::test_server_certificate_includes_server_alias_san` (new test asserting the alias SAN on the generated proxy cert)


pablomh:
The CI errors are not related to this change.

tests/auth_bundle_test.py
```diff
@@ -111,3 +113,21 @@ def test_server_ca_and_default_ca_differ_for_custom_deployment(server, generate_
     default_ca = server.run(f'tar xzf {TARBALL} -O certs/ca.crt')
     assert default_ca.succeeded
     assert server_ca.stdout != default_ca.stdout
+
+
+@pytest.fixture(scope="module")
+def generate_bundle_with_alias(server, default_certificates):
+    """--server-alias only applies to certificates issued by the internal CA."""
```
ehelms, line 120: Should `--server-alias` be a parameter that isn't allowed to be used if the user has identified they are using custom certificates?

pablomh: Yeah, agreed, it's a real footgun. Traced it: `issue.yml` skips the whole cert-issuing block when `certificate_source == custom_server`, so `--server-alias` just silently does nothing in that case - no error, no warning, cert ends up without the SAN and you only find out when the LB fails hostname verification.

Good news is there's already a `forbidden_if` constraint in this same file for the mirror case (`certificates_source == default` forbids the custom cert params), so we can just add the symmetric one:

```yaml
forbidden_if:
  - ['certificates_source', 'default', ['certificates_custom_server_certificate', 'certificates_custom_server_key']]
  - ['certificates_source', 'custom_server', ['server_aliases']]
```

Fails fast at CLI parse time instead of silently no-op-ing.

pablomh:
CI is finally green :)

ehelms (approved):

PR #774, opened 2026-08-21
Fix IOP/Postgres dependency race on foreman.target restart

pablomh:
## Summary
- Every IOP container that connects directly to Postgres (`iop_advisor`, `iop_inventory`, `iop_vmaas`, `iop_vulnerability`, `iop_remediation`) was joined to `foreman.target`'s restart lifecycle without ever declaring `After=`/`Wants=postgresql.service`. During a coordinated `foreman.target` stop/start cycle, this let the affected containers race against Postgres's own restart and crash (`django.db.utils.OperationalError: FATAL: the database system is shutting down`), then burn through systemd's default crash-loop rate limit and get left permanently failed.
- Fixed by adding the same `After=`/`Wants=postgresql.service` pattern already used by the `foreman`, `pulp`, and `candlepin` roles in this repo. `iop-core-host-inventory-api` additionally needed `iop-core-host-inventory-migrate.service` (matching its sibling MQ container's existing pattern), and `iop-service-advisor-backend-api` additionally needed `iop-core-host-inventory-api.service` (it calls that REST endpoint directly via `INVENTORY_SERVER_URL`).

## Test plan
- [x] `ansible-lint` passes on all changed files
- [x] Applied live on a real deployment (satcpt): confirmed `hammer ping` and all affected services healthy after multiple `foreman.target` stop/start cycles
- [x] Confirmed via journal timestamps that the dependency fix correctly orders IOP container shutdown before Postgres's own stop (systemd's `After=` ordering is bidirectional - reversed on stop)

### DIFF
--- a/src/roles/iop_inventory/tasks/main.yaml
+++ b/src/roles/iop_inventory/tasks/main.yaml
@@ -54,6 +54,8 @@
       - |
         [Unit]
         Description=Database Readiness and Migration Init Container
+        After=postgresql.service
+        Wants=postgresql.service
         PartOf=foreman.target
         [Service]
         Type=oneshot
@@ -117,6 +119,8 @@
       - |
         [Unit]
         Description=IOP Core Host-Based Inventory Web Container
+        After=postgresql.service iop-core-host-inventory-migrate.service
+        Wants=postgresql.service iop-core-host-inventory-migrate.service
         PartOf=foreman.target
         [Service]
         Restart=on-failure
--- a/src/roles/iop_advisor/tasks/main.yaml
+++ b/src/roles/iop_advisor/tasks/main.yaml
@@ -68,8 +68,8 @@
       - |
         [Unit]
         Description=Advisor Backend API
-        After=iop-core-kafka.service
-        Wants=iop-core-kafka.service
+        After=postgresql.service iop-core-kafka.service iop-core-host-inventory-api.service
+        Wants=postgresql.service iop-core-kafka.service iop-core-host-inventory-api.service
         PartOf=foreman.target
         [Service]
         Restart=on-failure
@@ -98,8 +98,8 @@
       - |
         [Unit]
         Description=Advisor Backend Service
-        After=iop-core-kafka.service
-        Wants=iop-core-kafka.service
+        After=postgresql.service iop-core-kafka.service
+        Wants=postgresql.service iop-core-kafka.service
         PartOf=foreman.target
         [Service]
         Restart=on-failure
--- a/src/roles/iop_remediation/tasks/main.yaml
+++ b/src/roles/iop_remediation/tasks/main.yaml
@@ -64,8 +64,8 @@
       - |
         [Unit]
         Description=Remediations API
-        Wants=iop-core-host-inventory-api.service iop-service-advisor-backend-api.service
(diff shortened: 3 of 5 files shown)


src/roles/iop_inventory/tasks/main.yaml
```diff
@@ -117,6 +119,8 @@
       - |
         [Unit]
         Description=IOP Core Host-Based Inventory Web Container
+        After=postgresql.service iop-core-host-inventory-migrate.service
+        Wants=postgresql.service iop-core-host-inventory-migrate.service
```
vkrizan, line 123: By adding the hbi-migrate.service I assume that it would be started once by systemd (in case of a normal operations when both hbi services are started). Correct?

vkrizan (approved):
Thanks!

src/roles/iop_inventory/tasks/main.yaml
pablomh: That's correct. `iop-core-host-inventory-migrate` is `Type=oneshot` + `RemainAfterExit=true`, so once it runs it just sits there `active (exited)` and won't get re-triggered by the `Wants=` on its own. It does re-run once whenever `foreman.target` itself gets stopped and started again though (it's got `PartOf=foreman.target` already, pre-existing), since that stops it too and the next start re-triggers it.

Just double-checked this live on one of our test envs by running 10 `foreman.target` stop/start cycles back to back - `make upgrade_db` re-ran exactly once per cycle (10 for 10), ~4-5s each, no surprises. Also worth noting this isn't a new pattern from this PR - `iop-core-host-inventory` (the MQ container) already `Requires=`'d this same migrate service before this change, we're just extending the same thing to the API container too.

ehelms (approved):

pablomh:
Converted to DRAFT until I investigate the CI errors that didn't happen in my testing.

pablomh:
According to Claude:

> Root cause found via the CI run's sosreport journal — it's a pure CI flake, completely unrelated to this PR's changes:
> 
> 11:38:59  iop-advisor-image[154989]: Error: unable to copy from source docker://quay.io/iop/advisor-backend:foreman-3.18:
>           copying system image from manifest list: writing blob: storing blob to file "...": happened during read: unexpected EOF
>           (while reconnecting: Get "https://cdn01.quay.io/...": EOF)
> 11:38:59  iop-advisor-image.service: Main process exited, code=exited, status=125/n/a
> 11:38:59  systemd[1]: Dependency failed for Advisor Backend API.
> 11:38:59  iop-service-advisor-backend-api.service: Job .../start failed with result 'dependency'.
> 
> The dependency that failed is iop-advisor-image.service — the podman-quadlet auto-generated Requires=<name>-image.service unit that pulls the
> container image (this exists for every quadlet, unrelated to anything in the PR). It failed because cdn01.quay.io (Quay's S3-backed CDN edge) dropped
> the connection mid-transfer with an unexpected EOF — a transient network blip, not an auth/registry-config issue.
> 
> Confirms it's transient: exactly 1 second later, iop-service-advisor-backend-service's handler triggered the same image pull again and it completed
> cleanly (Finished iop-advisor-image.service / Started Advisor Backend Service at 11:39:13) — same image, same registry, no retry logic changed, it
> just worked the second time.

PR #776, opened 2026-08-21
Reduce PostgreSQL superuser usage

ehelms:

Per CWE-250 and PostgreSQL hardening guidance, the superuser should only be used for operations that genuinely require it (extension installation, foreign server creation, user mapping management). Using it for schema creation, index checks, backups, and restores breaks horizontal isolation between databases and increases attack surface.

* **iop_fdw**: Retain `postgres` only for DDL that requires superuser (enabling extensions, creating foreign servers, managing user mappings, granting foreign server usage). Switch schema creation, foreign table imports, view creation, and grants to the per-service database user. Add `password_required 'false'` to the FDW user mapping so non-superusers can connect through the foreign data wrapper on localhost. Remove the separate postgres user mapping that is no longer needed.
* **iop_inventory**: Switch inventory schema and view creation from `postgres` to the inventory service user.
* **iop_advisor / iop_vulnerability**: Pass `iop_fdw_database_password` when including the `iop_fdw` role so the service user can authenticate.
* **check_database_index**: Accept `check_database_index_user` and `check_database_index_password` instead of hardcoding `postgres`. Callers (checks, backup preflight) now pass the per-database user and password.
* **restore**: Use the per-database password from the database list item instead of the global `postgresql_admin_password` for `pg_restore`.
* **iop_fdw defaults**: Add `iop_fdw_database_password` as a required parameter.

Steps to reproduce:

* Deploy with IoP enabled (`foremanctl deploy` with advisor/vulnerability features) and verify the FDW setup completes without errors
* Run `foremanctl checks` and confirm database index checks pass
* Run `foremanctl backup` and `foremanctl restore` against an internal database deployment

#### Checklist

### DIFF
--- a/src/roles/restore/tasks/restore_databases.yaml
+++ b/src/roles/restore/tasks/restore_databases.yaml
@@ -27,6 +27,7 @@
       dump_file: "{{ restore_backup_metadata.database_mapping[item.name] }}.dump"
       database: "{{ item.database }}"
       user: "{{ item.user }}"
+      password: "{{ item.password }}"
   loop: "{{ databases }}"
   loop_control:
     label: "{{ item.name }}"
@@ -89,7 +90,7 @@
       --dbname={{ item.database }}
       {{ backup_dir }}/{{ item.dump_file }}
   environment:
-    PGPASSWORD: "{{ postgresql_admin_password }}"
+    PGPASSWORD: "{{ item.password }}"
   loop: "{{ restore_databases_to_restore }}"
   loop_control:
     label: "{{ item.dump_file }} → {{ item.database }}"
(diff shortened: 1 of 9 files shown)


src/roles/restore/tasks/restore_databases.yaml
ekohl, line 86: Why isn't this using `postgresql_db` with `state: restore`? https://docs.ansible.com/projects/ansible/latest/collections/community/postgresql/postgresql_db_module.html#parameter-state describes that and would be a lot more consistent with the rest.

ekohl (commented):
Would it make sense to split this into backup/restore and IOP?

PR #777, opened 2026-08-21
Simplify recurring timer test to validate container plumbing

ehelms:

The `test_foreman_recurring_timer_execution` test waited up to 60 seconds for the rake task to complete and asserted `Result == "success"`. This made the test slow, flaky (race with `Persistent=true` timers that fire on first enable), and tested more than it needed to — the rake task outcome, not the container plumbing.

The real value of this test is verifying that the quadlet definition, image, secrets, env vars, and volumes are wired up correctly so the container can start.

* Start the service, wait 5 seconds, and assert `Result != "failed"` — enough to catch container startup failures without waiting for the rake task to finish
* Remove the polling loop, `@pytest.mark.slow`, and the pre-start drain logic

Steps to reproduce:

* Deploy a VM with recurring timers enabled
* Run the recurring timer test: `./forge test -- tests/feature/foreman/base_test.py -vv -k test_foreman_recurring_timer_execution`
* Verify all four instances (hourly, daily, weekly, monthly) pass

#### Checklist

### DIFF
--- a/tests/feature/foreman/base_test.py
+++ b/tests/feature/foreman/base_test.py
@@ -98,23 +98,43 @@ def test_foreman_recurring_timer_next_trigger(server, instance):
 @pytest.mark.slow
 @pytest.mark.parametrize("instance", RECURRING_INSTANCES)
 def test_foreman_recurring_timer_execution(server, instance):
-    """Trigger a timer manually and verify it executes successfully."""
+    """Verify the recurring container is wired up correctly and can run.
+
+    Triggers a controlled run to exercise the quadlet/image/secrets/env/
+    volume plumbing. It deliberately does not wait for the rake task to
+    finish: cron:daily/weekly/monthly can run for many minutes and their
+    runtime and outcome are not what this test verifies.
+    """
     service_name = f"foreman-recurring@{instance}.service"
 
-    server.check_output(f"systemctl start {service_name}")
+    previous_invocation = server.service(service_name).systemd_properties.get("InvocationID", "")
 
-    # Wait for the service to complete (these are oneshot services)
-    # Poll the service status until it's no longer active
-    max_wait = 60  # Maximum wait time in seconds
-    poll_interval = 2
+    server.check_output(f"systemctl start --no-block {service_name}")
+
+    startup_timeout = 60
+    poll_interval = 5
     waited = 0
+    while waited < startup_timeout:
+        props = server.service(service_name).systemd_properties
+        active_state = props["ActiveState"]
+        result = props["Result"]
+        invocation = props.get("InvocationID", "")
 
-    service = server.service(service_name)
-    while service.is_running and waited < max_wait:
+        assert active_state != "failed", (
+            f"{service_name} failed to start: "
+            f"ActiveState={active_state}, Result={result}"
+        )
+
+        if invocation and invocation != previous_invocation and (
+            active_state in ("active", "activating", "deactivating")
+            or (active_state == "inactive" and result == "success")
+        ):
+            break
+
         time.sleep(poll_interval)
         waited += poll_interval
-
-    assert service.systemd_properties["Result"] == "success"
+    else:
+        pytest.fail(f"{service_name} did not start within {startup_timeout}s")
 
 
 def test_foreman_delivery_method_setting(foremanapi):


ehelms:
@archanaserver Can you take a look at this modification? I am trying to handle this scenario -- https://github.com/theforeman/foreman-oci-images/actions/runs/32482066584/job/96770396512?pr=91  where effectively a race condition exists and these tests are unpredictable in their runtime.

tests/feature/foreman/base_test.py
```diff
@@ -95,26 +95,18 @@ def test_foreman_recurring_timer_next_trigger(server, instance):
...
     service_name = f"foreman-recurring@{instance}.service"
 
     server.check_output(f"systemctl start {service_name}")
 
-    # Wait for the service to complete (these are oneshot services)
-    # Poll the service status until it's no longer active
-    max_wait = 60  # Maximum wait time in seconds
-    poll_interval = 2
-    waited = 0
+    time.sleep(5)
```
ekohl: Sleep is not a synchronization step. What if it did nothing and the previous state was already not equal to `failed`?

tests/feature/foreman/base_test.py
```diff
@@ -95,26 +95,18 @@ def test_foreman_recurring_timer_next_trigger(server, instance):
...
 
     service = server.service(service_name)
-    while service.is_running and waited < max_wait:
-        time.sleep(poll_interval)
-        waited += poll_interval
+    assert service.systemd_properties["Result"] != "failed"
```
ekohl: Should we maintain the current implementation but widen the assertion after timeout?
```suggestion
    if waited >= max_wait:
        assert service.systemd_properties["Result"] != "failed"
    else:
        assert service.systemd_properties["Result"] == "success"
```

ekohl (commented):
I think the general idea is interesting, but it has some challenges. What if the service starts starting, but fails to fully initialize.

Thinking out loud: how do you identify a new run has started? Can you identify on the service?

ehelms:
I modified the test to instead check that the initial starts of these that happened at deploy time were successful.

tests/feature/foreman/base_test.py
```diff
@@ -98,23 +98,51 @@ def test_foreman_recurring_timer_next_trigger(server, instance):
...
+    service_name = f"foreman-recurring@{instance}.service"
 
-    # Wait for the service to complete (these are oneshot services)
-    # Poll the service status until it's no longer active
-    max_wait = 60  # Maximum wait time in seconds
-    poll_interval = 2
+    def prop(name):
+        return server.check_output(
+            f"systemctl show {service_name} --property={name} --value"
+        ).strip()
```
ekohl: The problem is that `systemd_properties` is cached. Rather than reimplementing this manually I'd prefer to instantiate a new `server.service(service_name)` in every loop.

ekohl (dismissed):

ehelms:
@ekohl Cleaned it up

ekohl:
@archanaserver mind sharing your opinions in a review?

tests/feature/foreman/base_test.py
```diff
@@ -98,23 +98,40 @@ def test_foreman_recurring_timer_next_trigger(server, instance):
...
-    poll_interval = 2
+    startup_timeout = 60
+    poll_interval = 5
     waited = 0
+    while waited < startup_timeout:
+        props = server.service(service_name).systemd_properties
+        active_state = props["ActiveState"]
+        result = props["Result"]
+
+        assert result != "failed", (
```
archanaserver: I think this assertion may be checking the wrong value. while looking here https://www.freedesktop.org/software/systemd/man/252/org.freedesktop.systemd1.html?utm_source=chatgpt.com#:~:text=Result%20encodes%20the%20execution%20result%20of%20the%20last%20run%20of%20the%20service "failed" is not a valid `Result` value and it's seems an `ActiveState` value, So isn't this assertion will almost never catch failures; it's checking for a value that doesn't exist

tests/feature/foreman/base_test.py
```diff
@@ -98,23 +98,40 @@ def test_foreman_recurring_timer_next_trigger(server, instance):
...
+            f"{service_name} failed to start: "
+            f"ActiveState={active_state}, Result={result}"
+        )
+
+        if props["InvocationID"] and (
```
archanaserver: A question about the `InvocationID`: so could this still be from a previous run? Since `systemctl start --no-block` returns immediately, so my thinking is would it be safer to capture the `ID` before starting and wait for it to change, ensuring we observe the new invocation?

tests/feature/foreman/base_test.py
archanaserver: also I believe we can check `active_state == "failed"` here instead? since activating is accepted below, could the test pass before an immediate container startup failure is reported?

archanaserver (changes requested):
The overall approach makes sense. I left a few of ques around the `InvocationID` and `Result/ActiveState` handling approach, mainly to ensure the test doesn’t pass using a previous invocation.

ehelms:
@archanaserver I have made updates based on your suggestions.

archanaserver (approved):
LGTM, @ekohl if you don't have any concerns then we can merge it.

ekohl (approved):
Thanks! Checking for a change in `InvocationID` is an implementation of what I was thinking about.

PR #778, opened 2026-08-21
Add documentation blob about backup/restore in dev setups

sjha4:
Adding a section about backup/restore in developer environments where quadlet and proxy are separate VMs and not localhost like with package installed foremanctl.

There are subtle differences in backup/restore behavior in dev setups vs user setups. For user setups everything maps to localhost making the flow simpler. On dev setups, since quadlet/proxies are their own VMs, we need to add some documentation about the difference in behavior especially when managing quadlet and proxies from same controller with different branches for example..

* 

Steps to reproduce:

*

#### Checklist


qcjames53 (approved):
Thanks Samir. This looks good to me.

ehelms (approved):

PR #781, opened 2026-08-25
Resolve transitive dependencies when listing enabled features

jeremylenz:

`foremanctl features` reported dependency-only features (e.g. `tasks` and `dynflow`, pulled in by `remote-execution`) as `available` instead of `enabled`. The list checked for literal membership in `enabled_features` instead of using the existing `has_feature` dependency resolution already used elsewhere (plugin installation, health checks), so the displayed state didn't match what was actually being installed/configured.

* `list_all_features()` now checks each feature with `has_feature()` instead of `name in enabled_features`, so features enabled only as a dependency of another feature are correctly marked `enabled`.
* Added a unit test covering a dependency-only feature showing as `enabled`.

Steps to reproduce:

* Deploy with a feature that has dependencies, e.g. `--add-feature remote-execution`.
* Run `FOREMANCTL_FEATURES_LIST_INTERNAL=true ./foremanctl features`.
* Before this change: `tasks` and `dynflow` show as `available`. After this change: they show as `enabled`.
* `python -m pytest tests/unit/filter_test.py -vv`

#### Checklist

### DIFF
--- a/src/filter_plugins/foremanctl.py
+++ b/src/filter_plugins/foremanctl.py
@@ -79,7 +79,7 @@ def list_all_features(enabled_features, only_enabled=False):
         if internal and not list_internal:
             continue
         description = meta.get('description', '')
-        if name in enabled_features:
+        if has_feature(enabled_features, name):
             enabled_list.append((name, 'enabled', internal, description))
         elif not only_enabled:
             available_list.append((name, 'available', internal, description))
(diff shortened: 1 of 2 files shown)


ehelms (approved):

jeremylenz:
I think test failures are unrelated, so feel free to merge :)

PR #782, opened 2026-08-25
Fix no-change deploy restarts Foreman issue

### LINKED ISSUE #769: A no-change deploy restarts Foreman for ~57s, because the migration one-shot is restarted unconditionally and Foreman Requires= it
### Summary

A deploy that changes nothing takes Foreman down for about a minute.

`foreman-db-migrate.service` is driven with `systemd: state: restarted`, unconditionally, on every deploy. Both `foreman.container` and the `dynflow-sidekiq@` template render `Requires=foreman-db-migrate.service`, and systemd propagates a restart to units that require the restarted one. So restarting the migration one-shot stops and starts Foreman and all three dynflow workers.

### Measured

Journal from a re-deploy on an unchanged host, same image, no configuration change:

```
13:56:53  Stopping foreman.service...
13:56:54  Stopped foreman.service.
13:57:16  Starting foreman.service...
13:57:50  Started foreman.service.
```

Fifty-seven seconds of downtime. Service `ActiveEnterTimestamp` confirms it: `foreman` and the three `dynflow-sidekiq@*` units all show times inside the deploy window, while `pulp-api`, `pulp-content` and `httpd` still show their first-install times. Pulp is unaffected because its units do not declare `Requires=` on its own migration one-shot.

### Why it matters

Two consequences beyond the downtime itself.

The dynflow workers cycle with Foreman, so **in-flight tasks are lost on every deploy**: remote-execution jobs, repository syncs and content-view publishes have to be re-run. That is understood as a cost of failover; it is surprising as a cost of a no-op deploy.

And it means a deploy cannot be run at an arbitrary moment on a fleet of any size. Any managed host that checks in during the window gets a failed `subscription-manager` or package operation. It also rules out running deploy periodically to converge state, which is otherwise a natural way to use it.

### Why the obvious fix is wrong

`state: started` would be a no-op here, because the one-shot is `Type=oneshot` with `RemainAfterExit=yes` and stays active after a successful run. That is presumably exactly why `restarted` is used: on an image upgrade the unit is already active, and `started` would silently skip the migrations. So relaxing it would trade this for broken upgrades.

Removing the `Requires=` is also not right, since it is what prevents Foreman starting against an unmigrated database.

### Suggested direction

Make the restart conditional on the image having changed, rather than unconditional. The `images` role already reports `changed` when it rewrites an image file, so that status could gate the migrate restart, or the restart could become a handler notified by it. Migrations would then run exactly when a new image needs them, and an unchanged deploy would touch nothing.

### Environment

foremanctl 2.3.0.

shubhamsg199:

foremanctl deploy that changes nothing takes Foreman down for about a minute. foreman-db-migrate.service is driven with state: restarted unconditionally on every deploy. Both foreman.container and the dynflow-sidekiq@ template declare Requires=foreman-db-migrate.service, so systemd propagates the restart stopping and starting Foreman and all three dynflow-sidekiq workers even when nothing has changed.

Fixes: https://github.com/theforeman/foremanctl/issues/769

- Register the result of the image quadlet generation task in the shared deploy_image.yaml and accumulate per-image change status into an images_changed dict

- Make the Foreman migration task conditional: use state: restarted only when the foreman image actually changed, state: started otherwise. Since the migration service is Type=oneshot with RemainAfterExit=yes, started is a genuine no-op when the service is already active, no restart cascade

Steps to reproduce:

* Run `foremanctl deploy` without any changes to images

#### Checklist

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -246,7 +246,7 @@
 - name: Migrate and seed the Foreman database
   ansible.builtin.systemd:
     name: foreman-db-migrate.service
-    state: restarted
+    state: "{{ 'restarted' if (images_changed.foreman | default(false) | bool) else 'started' }}"
 
 - name: Flush handlers to restart services
   ansible.builtin.meta: flush_handlers
--- a/src/roles/images/tasks/deploy_image.yaml
+++ b/src/roles/images/tasks/deploy_image.yaml
@@ -11,6 +11,7 @@
       - |
         [Service]
         Environment=REGISTRY_AUTH_FILE={{ images_registry_auth_file }}
+  register: image_deploy_result
 
 - name: Create drop-in directory for {{ images_definition.name }}
   ansible.builtin.file:
@@ -21,3 +22,9 @@
 - name: Register deployed image name
   ansible.builtin.set_fact:
     images_deployed_names: "{{ images_deployed_names + [images_definition.name] }}"
+
+- name: Track whether image changed for {{ images_definition.name }}
+  ansible.builtin.set_fact:
+    images_changed: >-
+      {{ images_changed | default({})
+         | combine({images_definition.name: (image_deploy_result is changed)}) }}
--- a/src/roles/images/tasks/deploy_image.yaml
+++ b/src/roles/images/tasks/deploy_image.yaml
@@ -11,7 +11,7 @@
       - |
         [Service]
         Environment=REGISTRY_AUTH_FILE={{ images_registry_auth_file }}
-  register: image_deploy_result
+  register: images_deploy_result
 
 - name: Create drop-in directory for {{ images_definition.name }}
   ansible.builtin.file:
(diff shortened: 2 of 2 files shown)


src/roles/foreman/tasks/main.yaml
```diff
@@ -246,7 +246,7 @@
 - name: Migrate and seed the Foreman database
   ansible.builtin.systemd:
     name: foreman-db-migrate.service
-    state: restarted
+    state: "{{ 'restarted' if (images_changed.foreman | default(false) | bool) else 'started' }}"
```
ehelms, line 249: This is a non-role variable that should not be used inside the `foreman` role.

ehelms:
See https://github.com/theforeman/foremanctl/issues/769#issuecomment-5413636788 for some more insight into this issue.

shubhamsg199:
> See [#769 (comment)](https://github.com/theforeman/foremanctl/issues/769#issuecomment-5413636788) for some more insight into this issue.

Added my observations [here](https://github.com/theforeman/foremanctl/issues/769#issuecomment-5423734131)

shubhamsg199:
@ehelms since you took over the fix in https://github.com/theforeman/foremanctl/pull/793, I’m closing this.

PR #785, opened 2026-08-26
Support air-gapped deployments

archanaserver:

Air-gapped deployments use preloaded container images, but foremanctl could still attempt registry access. This adds an explicit air-gapped mode persisted in `/var/lib/foremanctl/parameters.yaml`.

The complete offline-media workflow depends on [SAT-49617](https://redhat.atlassian.net/browse/SAT-49617) persisting air_gapped: true during prepare_system.

Refs: [SAT-48204](https://redhat.atlassian.net/browse/SAT-48204)

* Add the `--air-gapped` parameter.
* Use `Policy=never` and remove stale pull overrides.
* Reject  `foremanctl pull-images` in air-gapped mode.
* Disable the Katello subscription connection.
* Add tests, CI coverage, and documentation.

* Preload the required images:

  ```bash
  ./foremanctl pull-images
  ```

* Deploy in air-gapped mode:

  ```bash
  ./foremanctl deploy --air-gapped
  ```

* Verify that the generated image units contain `Policy=never`:

  ```bash
  grep -R '^Policy=' /etc/containers/systemd/*.image
  ```

* Verify that no stale force-pull overrides remain:

  ```bash
  find /etc/containers/systemd -path '*.image.d/00-pull-always.conf' -print
  ```

  This command should not return any files.

* Verify that `air_gapped: true` is persisted in the foremanctl parameters file.

* Run `./foremanctl pull-images` again and verify that it is rejected without contacting a registry.

* Verify that the image units still contain `Policy=never` after the rejected command.

* Verify that `subscription_connection_enabled` is `false` and read-only.

#### Checklist


archanaserver:
Question: So in disconnected mode, should `foremanctl pull-images` fail as implemented here, or should it validate that all required images exist locally using `Policy=never`? Should this validation also run as an explicit early deploy pre-check?

lzap:
IMHO, as a user in disconnected mode I would expect that a command named `pull-images` would validate that *all* the images are present. From what I understand, `pull-images` depends on options so running it with default values might not download *all available* images and then, after the system is disconnected, something would be missing. Ideal state:

* In connected mode, `pull-images` downloads all images that are available.
* In disconnected mode, `pull-images` validates presence of all images that are available.

The `conf` extension is a bit weird if this is supposed to be just marker file, not sure if this isn't some pattern you guys use. If not, I suggest dropping the extension completely and considering using `/var` directory instead.

Sidenote: It seems that the "image validation and count steps" hunk is created, then dropped in the last commit. I am confused. :-)

src/roles/pre_install/tasks/main.yaml
```diff
@@ -1,4 +1,14 @@
 ---
+- name: Check if disconnected config file exists
+  ansible.builtin.stat:
+    path: /etc/foreman/disconnected.conf
```
stejskalleos: As @lzap mentioned in [the previous comment](https://github.com/theforeman/foremanctl/pull/785#issuecomment-5425097148), this indeed feels like a config file that should contain configuration values. My proposal is to change it to `/var/lib/foreman/disconnected`

lzap: I am still studying how `foremanctl` work, I see there are some configuration files already stored. So that looks like a good place I think, tho, why you use a directory? I would expect something like `/var/lib/foreman/disconnected_enabled` (a regular empty file).

I only have experience with dev setup so far and this is what I see:

```
(.venv) [lzapleta@yuki foremanctl]$ find .var
.var
.var/lib
.var/lib/foremanctl
.var/lib/foremanctl/parameters.yaml
.var/lib/foremanctl/foreman-db-password
.var/lib/foremanctl/candlepin-db-password
.var/lib/foremanctl/pulp-db-password
.var/lib/foremanctl/postgresql-admin-password
.var/lib/foremanctl/candlepin-oauth-secret
.var/lib/foremanctl/oauth
.var/lib/foremanctl/oauth/foreman-oauth-consumer-key
.var/lib/foremanctl/oauth/foreman-oauth-consumer-secret
.var/lib/foremanctl/foremanctl.2026-08-26-14-37-56.log
.var/lib/foremanctl/foremanctl.2026-08-26-14-38-28.log
.var/lib/foremanctl/foremanctl.log
```

Should this be `/var/lib/foremanctl/...` instead? Also why not to use `parameters.yaml`? I am not sure what that does, random thoughts.

stejskalleos: the `disconnected` is meant as an empty file. But yeah, it should be in the `/var/lib/foremanctl` directory

src/roles/foreman/templates/settings.yaml.j2
```diff
@@ -44,3 +44,8 @@
 
 # Configure SMTP as delivery method
 :delivery_method: smtp
+
+{% if disconnected_mode | bool %}
+# Disable subscription connection in disconnected mode
+:subscription_connection_enabled: false
```
ehelms: This lives in katello not Foreman, so I think it would need to go in the katello.yaml plugin config?

src/roles/images/defaults/main.yaml
```diff
@@ -2,3 +2,4 @@
 images_quadlet_dir: /etc/containers/systemd
 images_registry_auth_file: /etc/foreman/registry-auth.json
 images_deployed_names: []
+images_pull_policy: "{{ 'never' if disconnected_mode | bool else 'missing' }}"
```
ehelms: `disconnected_mode` would need to be role scoped

src/roles/images/tasks/pull.yaml
```diff
@@ -1,4 +1,11 @@
 ---
+- name: Block image pulling in disconnected mode
+  ansible.builtin.fail:
+    msg: |
+      Cannot run 'foremanctl pull-images' in disconnected mode.
+      Images must be pre-loaded using prepare_system.
```
ehelms: What is `prepare_system` ? Is that a future `foremanctl` command?

ehelms:
I think we also need to think about what `disconnected` means here, as it has a lot of meaning. For example:

- The installation can be completely air-gapped, no external internet access.
- The installation can be disconnected from the internet, but pulling images from another Foreman.
- The installation can be disconnected from services like Red Hat Lightspeed and Portal but internet connected and pulling images.

Right now this solution is combining a few of these into one disconnected mode. I think you are actually targetting "airgapped" with this initial approach. And I would lean into that nomenclature.

shubhamsg199:
> Question: So in disconnected mode, should `foremanctl pull-images` fail as implemented here, or should it validate that all required images exist locally using `Policy=never`? Should this validation also run as an explicit early deploy pre-check?

`prepare_system` should already handle pull-images validation right? So maybe fail or a message stopping the execution seems a wise choice here.

src/roles/images/tasks/pull.yaml
shubhamsg199: Its the replacement of `install_packages` script in containerized [disconnected](https://docs.redhat.com/en/documentation/red_hat_satellite/6.19/html/installing_satellite_server_in_a_disconnected_network_environment/installing-satellite-server#installing-from-the-offline-repositories_satellite)

src/roles/pre_install/tasks/main.yaml
ehelms: Why do we want an empty file?

src/roles/images/tasks/pull.yaml
ehelms: That's for Satellite. What about non-Satellite? This references some unknown term that will confuse users or is only available to Satellite users. Further, it's not clear that this is some script available in an ISO produced by Satellite. This reads like it's a command in foremanctl.

stejskalleos: @ehelms yes, it's for satellite only, since we don't have an upstream ISO (yet).

ehelms: @stejskalleos I edited your comment to drop the private link as this is a public project. 

I think my point is being lost here. Our code and help text should be clear, and understandable by users. Referencing some unknown thing that is only applicable to Satellite will lead to confusion by our users.

src/roles/pre_install/tasks/main.yaml
arvind4501: we can also persist the `disconnected: true` in `/var/lib/foremanctl/parameters.yaml` through the prepare script, so that `foremanctl deploy` automatically runs like `foremanctl deploy --disconnected` without the need to pass  `--disconnected`. originally the idea was shared by @evgeni ,when i encountered some disconnected related issues. just wanted to share, does not block for me

stejskalleos:
> prepare_system should already handle pull-images validation right?

@shubhamsg199 correct. In this scenario, "pull images" verifies the presence of images, and then `skopeo copy` runs.

src/roles/pre_install/tasks/main.yaml
stejskalleos: @arvind4501 That's a good point. Why not save it as a persistent parameter :+1:

archanaserver:
> I think we also need to think about what `disconnected` means here, as it has a lot of meaning. For example:
> 
>     * The installation can be completely air-gapped, no external internet access.
> 
>     * The installation can be disconnected from the internet, but pulling images from another Foreman.
> 
>     * The installation can be disconnected from services like Red Hat Lightspeed and Portal but internet connected and pulling images.
> 
> 
> Right now this solution is combining a few of these into one disconnected mode. I think you are actually targetting "airgapped" with this initial approach. And I would lean into that nomenclature.

Thinking on this I would say this yes this PR targets fully air-gapped installations prepared from local media, I try to make changes on that, and other disconnected scenarios might still need registry access, so yeah "airgapped" seems more precise in this case. So you want me to update the whole `disconnected` term to `airgapped` here. just to be clear? @ehelms thoughts

src/roles/pre_install/tasks/main.yaml
archanaserver: Agreed, using `/var/lib/foremanctl/parameters.yaml` is cleaner and i see it fits the existing persistent-parameter model better than an empty marker file. Thanks for the suggestions @arvind4501

src/roles/images/tasks/pull.yaml
archanaserver: Yeah I referenced it wrong I see, I'll update the error message to be more generic.

tests/disconnected_test.py
```diff
@@ -0,0 +1,52 @@
+import pytest
+
+
+def test_disconnected_image_policy_never(server, enabled_features):
+    """Verify image units have Policy=never in disconnected mode."""
+    disconnected_config = server.file("/etc/foreman/disconnected.conf")
```
nacoool: WRT above comments i think ` /etc/foreman/disconnected.conf` would be invalid here. You might need to change tests accordingly as well.

tests/disconnected_test.py
```diff
@@ -0,0 +1,52 @@
...
+        if image_file.exists:
+            assert "Policy=never" in image_file.content_string, \
+                f"{image}.image should have Policy=never in disconnected mode"
+            assert "Policy=missing" not in image_file.content_string, \
+                f"{image}.image should not have Policy=missing in disconnected mode"
+
+
+def test_subscription_connection_disabled(server, foremanapi, enabled_features):
+    """Verify subscription_connection_enabled is disabled and readonly."""
+    disconnected_config = server.file("/etc/foreman/disconnected.conf")
```
nacoool: Same changes need to be applied here as well.

tests/disconnected_test.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+    assert len(settings) == 1
+    assert settings[0]["value"] is False
+    assert settings[0]["readonly"] is True
+
+
+def test_pull_images_blocked_in_disconnected_mode(server):
+    """Verify foremanctl pull-images fails in disconnected mode."""
+    disconnected_config = server.file("/etc/foreman/disconnected.conf")
```
nacoool: Same changes need to be applied here as well.
```
 params_file = server.file("/var/lib/foremanctl/parameters.yaml")

    if not params_file.exists:
        return False

    try:
        params = yaml.safe_load(params_file.content_string)
        return params.get('disconnected', False) is True
    except (yaml.YAMLError, AttributeError):
        return False
```

nacoool: You can also create a fixture that can use and check `server.file("/var/lib/foremanctl/parameters.yaml")` at once and dont need to check it repeatedly.

```

@pytest.fixture
def disconnected_mode(server):
    """Fixture to verify server is in disconnected/airgapped mode."""
    params_file = server.file("/var/lib/foremanctl/parameters.yaml")

    if not params_file.exists:
        pytest.skip("parameters.yaml not found")

    try:
        params = yaml.safe_load(params_file.content_string)
        is_disconnected = params.get('disconnected', False)

        if not is_disconnected:
            pytest.skip("Not in disconnected mode")

        yield params  

    except yaml.YAMLError as e:
        pytest.fail(f"Failed to parse parameters.yaml: {e}")
```
and then add in current test 
```
def test_pull_images_blocked_in_disconnected_mode(server, disconnected_mode):
    """Verify foremanctl pull-images fails in disconnected mode."""

    result = server.run("foremanctl pull-images")
    pass #remaining test steps
```

nacoool (changes requested):

ehelms:
> > I think we also need to think about what `disconnected` means here, as it has a lot of meaning. For example:
> > ```
> > * The installation can be completely air-gapped, no external internet access.
> > 
> > * The installation can be disconnected from the internet, but pulling images from another Foreman.
> > 
> > * The installation can be disconnected from services like Red Hat Lightspeed and Portal but internet connected and pulling images.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Right now this solution is combining a few of these into one disconnected mode. I think you are actually targetting "airgapped" with this initial approach. And I would lean into that nomenclature.
> 
> Thinking on this I would say this yes this PR targets fully air-gapped installations prepared from local media, I try to make changes on that, and other disconnected scenarios might still need registry access, so yeah "airgapped" seems more precise in this case. So you want me to update the whole `disconnected` term to `airgapped` here. just to be clear? @ehelms thoughts

I lean towards yes.

.github/workflows/test.yml
```diff
@@ -491,15 +497,20 @@ jobs:
           ./forge setup-repositories
       - name: Run image pull
         run: |
-          ./foremanctl pull-images
+          ./foremanctl pull-images ${{ matrix.mode == 'air-gapped' && '--add-feature foreman-proxy' || '' }}
```
arvind4501, line 502: why are we adding `foreman-proxy` as a feature with pull-images?

tests/air_gapped_test.py
```diff
@@ -0,0 +1,75 @@
+import subprocess
+from pathlib import Path
+
+import pytest
+import yaml
+
+REPOSITORY_ROOT = Path(__file__).resolve().parents[1]
+PARAMETERS_FILE = (
```
arvind4501: we can use a existing fixture for getting the parameters file.
https://github.com/theforeman/foremanctl/blob/fe182fdb7a3c0e70200e9446245281fda54de6fa/tests/conftest.py#L129

src/roles/images/tasks/pull.yaml
```diff
@@ -1,4 +1,16 @@
 ---
+- name: Reload systemd after removing stale pull overrides
+  ansible.builtin.systemd:
+    daemon_reload: true
+  when: images_air_gapped | bool
```
arvind4501: I see a lot of `| bool` is that needed? or its defensive approch

archanaserver: It was added defensively, but `air_gapped` is already declared as a boolean. so I’ve kept the conversion at the playbook-to-role boundary now and removed the redundant `| bool` filters inside the images role which seems not needed.

tests/air_gapped_test.py
archanaserver: good point. i'll update the tests to use the existing `obsah_params` fixture and kept only the air-gapped mode check in the local fixture

.github/workflows/test.yml
archanaserver: so the `pull-images` command defaults doesn't include `foreman-proxy`. Since `air-gapped` mode uses `Policy=never` and can't pull images from the registry, we need to pre-load foreman-proxy explicitly. and without it, deployment fails with a missing image error. but do you have diff workflow in mind?

ehelms: `pull-images` should respect the current flavor, seems like you have found a bug in that assumption that we should fix.

.github/workflows/test.yml
```diff
@@ -481,8 +465,14 @@ jobs:
           wait-timeout-minutes: 5
 
   satellite-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        mode:
```
ehelms, line 471: This doesn't apply to our regular foreman tests?

src/roles/foreman/templates/katello.yaml.j2
```diff
@@ -10,3 +10,8 @@
 
   :pulp:
     :ca_cert_file: /etc/foreman/katello-default-ca.crt
+
+{% if air_gapped %}
```
ehelms, line 14:
```suggestion
{% if foreman_air_gapped %}
```

or opt instead for `foreman_subscription_connection_enabled` for the role as a variable and set it based on the `air_gapped` parameter in the vars files.

ehelms:
Looking good, please add documentation to `docs/`.

.github/workflows/test.yml
arvind4501: I think the issue is with https://github.com/theforeman/foremanctl/blob/master/src/playbooks/pull-images/pull-images.yaml#L6 passing katello flavor always to pull-images. as katello does not enable foreman-proxy feature by default

archanaserver: Ah yes, that make sense. So `--add-feature foreman-proxy` is just working around pull-images using the katello flavor. I guess this can be fixed separately and not scoping for this PR, as this need little adjustment?

src/playbooks/pull-images/pull-images.yaml
```diff
@@ -73,3 +73,5 @@
       ansible.builtin.include_role:
         name: images
         tasks_from: pull.yaml
+      vars:
```
arvind4501, line 76: While testing this , i found that foremanctl deploy correctly sets `policy=never` in .image units but once we run `foremanctl pull-images` it reverts back them to `policy=missing`.

The issue is because we are passing `images_air_gapped` to pull.yaml role, but the unit files are deployed before that and using the role default `images_air_gapped: false`.

correct fix IMO should be to hoist the var  to play scope.

archanaserver: Right, thanks! I’ve moved `images_air_gapped` to the play scope and so it is now available while all the image units are being generated.

.github/workflows/test.yml
ehelms: Yes it is -- and it can be fixed separately, we just need to not forget this if we merge this first. Can you file a Github issue describing this and linking to this discussion to not forget it?

This is going to also cause extra stuff to be pulled onto proxy I think.

.github/workflows/test.yml
```diff
@@ -515,15 +529,22 @@ jobs:
...
       - name: Run deployment
         run: |
           ./foremanctl deploy \
+            ${{ matrix.mode == 'air-gapped' && '--air-gapped' || '' }} \
             --flavor satellite \
             --initial-admin-password=changeme \
             --initial-organization "Foreman CI" \
             --initial-location "Internet" \
             --tuning development
+      - name: Verify air-gapped parameter persistence
```
ehelms: Is this needed? This is testing Obsah parameter handling which could be true for any parameter.

.github/workflows/test.yml
```diff
@@ -157,6 +163,7 @@ jobs:
         run: |
           ./foremanctl deploy \
             --certificate-source=${{ matrix.certificate_source }} \
+            ${{ matrix.air_gapped == true && '--air-gapped' || '' }} \
```
evgeni, line 166: Our docs call it "Installing Satellite Server in a disconnected network environment", so I think we should keep that wording and make the param (and everything around it) `--disconnected`?

Or is the idea to explicitly differentiate between "disconnected" (aka: no *direct* connection to the internet, but maybe an intermediate Satellite) and "air-gapped" (aka: absolutely no way to reach the broader internet)?

.github/workflows/test.yml
evgeni: I'd skip it

src/playbooks/deploy/deploy.yaml
```diff
@@ -42,6 +43,8 @@
     - role: foreman
       when:
         - "enabled_features | has_feature('foreman')"
+      vars:
+        foreman_air_gapped: "{{ air_gapped | bool }}"
```
evgeni: I think you can set this in `defaults.yml` or `foreman.yml`

src/playbooks/deploy/deploy.yaml
```diff
@@ -5,6 +5,7 @@
   become: true
   vars:
     flavor: katello
+    images_air_gapped: "{{ air_gapped | bool }}"
```
evgeni: I think you can set this in `defaults.yml` or `images.yml`

evgeni:
Besides the naming question, this LGTM!

evgeni:
Mhh, what's up with the tests?

```
Sep 16 09:57:33 quadlet.example.com dynflow-sidekiq-worker[54521]: 2026-09-16T09:57:33 [E|bac|8f98c905] Could not use any proxy for the ["SSH", "Script"] job. Consider configuring remote_execution_global_proxy, remote_execution_fallback_proxy in settings (RuntimeError)
```

Weird.

.github/workflows/test.yml
```diff
@@ -166,6 +173,7 @@ jobs:
             --content-import-path /custom/import \
             --content-export-path /custom/export
       - name: Deploy features
+        if: matrix.air_gapped != true
```
evgeni: This breaks the tests, as they expect the `foreman-proxy` feature (and others) to be present

evgeni: I took a stab at making such "accidental feature removal" changes more obvious: https://github.com/theforeman/foremanctl/pull/843

archanaserver: I see the issue now. Would it make sense to keep this step enabled and then we can like preload `foreman-proxy` before enabling air-gapped mode? As far as I can see, that is the only additional container image required by this feature set, and IoP is disabled for this matrix case. Separately [#843](https://github.com/theforeman/foremanctl/pull/843) should make any future mismatch in the expected features more obvious. WDYT?

.github/workflows/test.yml
archanaserver: Yeah, https://github.com/theforeman/foremanctl/issues/846
and I’ll keep the current `--add-feature foreman-proxy` workaround in this PR

.github/workflows/test.yml
archanaserver: Yeah, the idea was to differentiate between the two. This implementation targets the fully offline case: all container images must be preloaded, the image units use `Policy=never`, and `pull-images` is blocked.

@ehelms also pointed out https://github.com/theforeman/foremanctl/pull/785#issuecomment-5426265688 earlier that “disconnected” could also cover systems without direct internet access that can still reach an intermediate Satellite or registry, so I think air-gapped seemed more precise here.

But I see your point about keeping it consistent with the Satellite documentation. So would you prefer keeping `--air-gapped` and clarifying that it corresponds to a fully disconnected installation, or using `--disconnected` throughout for consistency?

.github/workflows/test.yml
evgeni: Yeah, I think it makes sense to preload the container image in that case.

.github/workflows/test.yml
evgeni: Right, being explicit is good. Maybe we can sneak in the word "disconnected" into the help text somehow so people who use search on the help output find it easier?
Something like "Enable air-gapped mode, where the system is completely disconnected from the Internet and can't fetch container images on its own"

.github/workflows/test.yml
```diff
@@ -508,8 +496,14 @@ jobs:
           wait-timeout-minutes: 5
 
   satellite-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        mode:
```
evgeni: Any reason to use `mode=(connected|air-gapped)` here, while using `air-gapped=(true|false)` for Foreman above? Makes it harder to keep the tests in sync.

.github/workflows/test.yml
```diff
@@ -534,10 +529,13 @@ jobs:
           ./forge setup-repositories
       - name: Run image pull
         run: |
-          ./foremanctl pull-images
+          # The Satellite flavor includes foreman-proxy, so preload it before
+          # enabling air-gapped mode with Policy=never.
+          ./foremanctl pull-images ${{ matrix.mode == true && '--add-feature foreman-proxy' || '' }}
```
evgeni: you're still using `matrix.mode` here and below

.github/workflows/test.yml
ehelms: I think using both words is a good compromise. Satellite documentation can also change.

docs/developer/deployment.md
```diff
@@ -197,21 +197,34 @@ No user action required beyond installing the RPM and logging into the product r
...
 ```
 
-##### Disconnected install from local media
+##### Air-gapped install from local media
 
-In air-gapped environments, container images must be brought in without network access. Images are loaded directly into Podman's container storage using `skopeo copy` before running `foremanctl deploy`.
+In fully air-gapped environments, all container images required by the selected flavor and enabled features must be loaded into local Podman storage before running `foremanctl deploy`.
 
-For each image, run:
+Product-provided offline-media tooling can preload these images automatically. For manual development or testing, an image can be imported with:
```
ehelms: This is an oddly worded, vague statement. But I get that you are trying to walk a careful line. I think we can be more specific here to help the reader out. How about:

```
Vendor-provided offline media, such as Satellite ISOs that bundle images in oci-archive format, can automatically preload container image archives into your environment. For manual development or testing, you can import an image directly using:
```

docs/developer/deployment.md
```diff
@@ -197,21 +197,34 @@ No user action required beyond installing the RPM and logging into the product r
...
 skopeo copy \
   oci-archive:/path/to/media/{image}/{tag}/oci-archive.tar \
   containers-storage:{pullspec}
 ```
 
-Where `{pullspec}` matches the effective image reference for each unit — the value from `src/vars/images.yml` (e.g. `quay.io/foreman/foreman:nightly`), or the overriding `Image=` from any drop-in present (e.g. a product RPM's `10-product.conf`). The pullspec used for `containers-storage:` must match what the merged quadlet unit will resolve to, or the image will not be found at deploy time. Once all images are loaded, run `foremanctl deploy` — the `.image` units start with `Policy=missing`, find the images already in storage, and do not attempt registry pulls.
+The destination `{pullspec}` must exactly match the effective image reference used by the generated Quadlet unit, including any `Image=` override provided by a drop-in.
 
-Do not run `foremanctl pull-images` in a disconnected environment. That command adds a `Policy=always` drop-in and restarts image services to force a registry pull, which will fail without network access.
+After all required images are available locally, enable air-gapped mode:
```
ehelms, line 214: I didn't think about it till now, `air-gapped` almost reads like a `feature`. I suppose we do not have a clear heuristic on when something is a feature and when it's a parameter that enables some "mode".

ehelms:
I mostly had a little wording nitpick, but I think this looks good!

docs/developer/deployment.md
archanaserver: Yeah, the boundary is not very clearly defined. And my thinking on keping this as a deployment parameter because it changes the behavior of the whole deployment rather than enabling an optional component.
It affects all image units, blocks pull-images, and changes the Katello configuration regardless of the selected features. So in that case does keeping it as a mode-like parameter make sense from that perspective?

docs/developer/deployment.md
archanaserver: Yeah, that is clearer and more specific. I’ve updated the wording, thanks!

.github/workflows/test.yml
archanaserver: Ah yes, I accidentally nested `air_gapped` under `mode` and missed these remaining `matrix.mode` references. Updated!

evgeni (approved):

ehelms (approved):

shubhamsg199:
@ehelms Manual testing was in progress, @archanaserver requested e2e testing for this and wasn't supposed to be merged until the test ack.

Tested with foremanctl-3.3.develop-3.20260922103944956117.pr785.34.gfb055af.el10.noarch

Scenarios tested and observations:

1. foremanctl pull-images fails in air-gapped mode as expected `Image pulling is disabled in air-gapped mode Ensure that all required images are available in local container storage`
2. Verified that the image units still contain `Policy=never` after the rejected command.
3. `foremanctl deploy --air-gapped` -> succeeded with out any issues
4. Verified the `air_gapped` is persisted in the foremanctl parameters file.
5. Verified the image units still contain Policy=never  after the deploy
6. Verified `:subscription_connection_enabled: false` through `podman exec foreman cat /etc/foreman/plugins/katello.yaml`
7. Deployed again with `foremanctl deploy --air-gapped --add-feature foreman-proxy` and verified all above points.

archanaserver:
Thanks @shubhamsg199 testing this and sharing the results! Everything looks as expected

ehelms:
> @ehelms Manual testing was in progress, @archanaserver requested e2e testing for this and wasn't supposed to be merged until the test ack.

If testing is requested, it needs to be stated on the PR and ideally, the person doing the testing add as a requested reviewer to add their approval. Despite my best efforts, I cannot read minds :grin:

PR #788, opened 2026-08-26
Pulp volume ownership followup (SAT-47249)

lzap:
Few changes as a follow-up to https://github.com/theforeman/foremanctl/pull/734

* Use variables instead harcoding 700
* Recursing pulp content is a trap! Implemented suggested solution by @arvind4501 which only recurse when permissions were previously root:root.
* Secrets should be owned by root, at least that is the pattern in the file

### DIFF
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -7,8 +7,8 @@
     path: "{{ item | split(':') | first }}"
     state: directory
     mode: "0755"
-    owner: "700"
-    group: "700"
+    owner: "{{ pulp_container_uid }}"
+    group: "{{ pulp_container_gid }}"
     recurse: true
   loop: "{{ pulp_volumes }}"
 
@@ -17,8 +17,8 @@
     path: "{{ pulp_storage_path }}/{{ item }}"
     state: directory
     mode: "0755"
-    owner: "700"
-    group: "700"
+    owner: "{{ pulp_container_uid }}"
+    group: "{{ pulp_container_gid }}"
     recurse: true
   loop:
     - tmp
@@ -30,8 +30,8 @@
     path: "{{ item }}"
     state: directory
     mode: "0755"
-    owner: "700"
-    group: "700"
+    owner: "{{ pulp_container_uid }}"
+    group: "{{ pulp_container_gid }}"
     recurse: true
   loop: "{{ pulp_default_import_paths + pulp_import_paths }}"
   when: not pulp_mirror
@@ -41,8 +41,8 @@
     path: "{{ item }}"
     state: directory
     mode: "0755"
-    owner: "700"
-    group: "700"
+    owner: "{{ pulp_container_uid }}"
+    group: "{{ pulp_container_gid }}"
     recurse: true
   loop: "{{ pulp_default_export_paths + pulp_export_paths }}"
   when: not pulp_mirror
@@ -75,8 +75,8 @@
 - name: Set secret key file permissions
   ansible.builtin.file:
     path: "{{ pulp_storage_path }}/django_secret_key"
-    owner: "700"
-    group: "700"
+    owner: "{{ pulp_container_uid }}"
+    group: "{{ pulp_container_gid }}"
     mode: '0600'
 
 - name: Create django pulp secret key secret
--- a/src/roles/pulp/tasks/main.yaml
+++ b/src/roles/pulp/tasks/main.yaml
@@ -9,7 +9,6 @@
     mode: "0755"
     owner: "{{ pulp_container_uid }}"
     group: "{{ pulp_container_gid }}"
-    recurse: true
   loop: "{{ pulp_volumes }}"
 
 - name: Create Pulp storage subdirs
@@ -19,7 +18,6 @@
     mode: "0755"
     owner: "{{ pulp_container_uid }}"
     group: "{{ pulp_container_gid }}"
-    recurse: true
   loop:
     - tmp
     - assets
@@ -32,7 +30,6 @@
     mode: "0755"
     owner: "{{ pulp_container_uid }}"
     group: "{{ pulp_container_gid }}"
-    recurse: true
   loop: "{{ pulp_default_import_paths + pulp_import_paths }}"
   when: not pulp_mirror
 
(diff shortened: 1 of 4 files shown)


src/roles/restore/tasks/restore_pulp_content.yaml
```diff
@@ -24,20 +24,30 @@
...
       when:
-        - restore_pulp_media_ownership.stat.exists
-        - restore_pulp_media_ownership.stat.uid == 0
+        - restore_pulp_storage_ownership.stat.exists
+        - >-
+          restore_pulp_storage_ownership.stat.uid == 0 or
+          (
+            restore_pulp_media_ownership.stat.exists and
+            restore_pulp_media_ownership.stat.uid == 0
+          )
```
lzap, line 65: I am not sure if checking `media` is necessary, just the root (storage) ownership might be enough.

src/roles/pulp/tasks/main.yaml
```diff
@@ -75,8 +88,8 @@
 - name: Set secret key file permissions
   ansible.builtin.file:
     path: "{{ pulp_storage_path }}/django_secret_key"
-    owner: "700"
-    group: "700"
+    owner: root
+    group: root
```
ehelms, line 93: Why do these get owned by root instead of the pulp user?

src/roles/restore/tasks/restore_pulp_content.yaml
```diff
@@ -24,20 +24,30 @@
...
+        path: "{{ pulp_storage_path }}"
+      register: restore_pulp_storage_ownership
+
+    - name: Check ownership of restored Pulp media directory
       ansible.builtin.stat:
         path: "{{ pulp_storage_path }}/media"
       register: restore_pulp_media_ownership
 
-    - name: Fix ownership after restore for non-root containers
+    - name: Migrate restored Pulp storage ownership from root to container user
```
ehelms, line 48: @sjha4 Can you take a look at the restore changes included in this PR?

src/roles/pulp/tasks/main.yaml
lzap: Follows what we are doing for the other files in that file (e.g. `database_fields.symmetric.key`).

src/roles/restore/tasks/restore_pulp_content.yaml
archanaserver: I think the `media` check make sense here. As the storage root could already be owned by the container user, so while restoring an older backup could still restore the media directory as root-owned. 

And If we only check the storage root, wouldn’t we miss that case?

lzap:
I am running into https://github.com/theforeman/smoker/pull/53 that could help.

src/roles/pulp/tasks/main.yaml
```diff
@@ -41,12 +38,29 @@
...
   loop: "{{ pulp_default_export_paths + pulp_export_paths }}"
   when: not pulp_mirror
 
+- name: Check ownership of Pulp volume paths
+  ansible.builtin.stat:
+    path: "{{ item | split(':') | first }}"
+  register: pulp_volume_path_stats
+  loop: "{{ pulp_volumes }}"
+
+- name: Migrate root-owned Pulp volume paths to container user
```
ehelms, line 52: Any idea how fast or slow this operation might be for large content sets?

lzap: This is merged code, I am only making if optional if the root directory is correctly owned.

But yes, good point, these operations can be extremely slow. I am wondering if we can actually drop all of this and simply enforce the permissions when creating the backup if we are sure all of the files should be owned by 700. Pseudo-code:

```diff
 diff --git a/src/roles/backup/tasks/pulp_content.yaml b/src/roles/backup/tasks/pulp_content.yaml
  index ... .. ...
  --- a/src/roles/backup/tasks/pulp_content.yaml
  +++ b/src/roles/backup/tasks/pulp_content.yaml
  @@ -5,6 +5,8 @@
       - name: Backup pulp content directory with encryption keys  # noqa: command-instead-of-module
         ansible.builtin.command:
           cmd: >
             tar -czf "{{ backup_dir_full }}/pulp-content.tar.gz"
  +          --owner=700
  +          --group=700
  +          --numeric-owner
             --listed-incremental="{{ backup_dir_full }}/pulp.snar"
             --directory="{{ backup_pulp_storage_path }}"
             --exclude=media/exports
```

src/roles/restore/tasks/restore_pulp_content.yaml
sjha4: Looks good..I was able to test the restore workflow and some smoke testing of pulp following that..The ownership is set correctly.

src/roles/pulp/tasks/main.yaml
ehelms: I assumed this needs to be done not just for backups but also for upgrades from when the containers did not have a Pulp user inside them?

lzap: > upgrades from when the containers did not have a Pulp user inside them?

Ah, right. I assumed 5.0 was the first container release but I forgot that we released 3.19. This rules the "backup it correctly" idea.

lzap:
I dropped the ownership commit from the patchset, now it only contains smallish things.

src/roles/pulp/tasks/main.yaml
```diff
@@ -7,8 +7,8 @@
     path: "{{ item | split(':') | first }}"
     state: directory
     mode: "0755"
-    owner: "700"
-    group: "700"
+    owner: "{{ pulp_container_uid }}"
+    group: "{{ pulp_container_gid }}"
     recurse: true
```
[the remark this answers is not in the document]
lzap: Well yeah, that is what I am trying to fix. Puhed a fix.

lzap:
All tests fail with "Box 'centos/stream10' could not be found" not sure why.

lzap:
Needs: https://github.com/theforeman/foremanctl/pull/851

ehelms (approved):

lzap:
The final solution is a bit simpler, not great but I think this is worth it to prevent possibly slow recursing. I think we should backport this to prevent problems for upgrades.

ehelms:
> The final solution is a bit simpler, not great but I think this is worth it to prevent possibly slow recursing. I think we should backport this to prevent problems for upgrades.

Agreed.

PR #789, opened 2026-08-27
Use legacy OpenSSL cipher naming for Candlepin's TLS connector

stejskalleos:
Candlepin 5.0's Tomcat uses the new OpenSSL FFM connector, which fails to handle the IANA-style cipher names (TLS_ECDHE_...) we were using, leaving the connector with no usable ciphers and every TLS handshake failing. Switch to the legacy OpenSSL cipher naming (ECDHE-ECDSA-AES256-GCM-SHA384, ...) to make it work.

(cherry picked from commit dfe678c2452f9e89aa9656cbc9252060bdf2c5d2)

### DIFF
--- a/src/roles/candlepin/defaults/main.yml
+++ b/src/roles/candlepin/defaults/main.yml
@@ -5,10 +5,7 @@ candlepin_tls_versions:
   - "TLSv1.2"
   - "TLSv1.3"
 candlepin_ciphers:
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I'd prefer to take current develop and release 3.1.0 into Foreman 5.0.

stejskalleos:
not needed anymore, 3.1.0 has been released.

PR #794, opened 2026-08-29
Optimize Kafka topic initialization on re-deploy

ehelms:

On every re-deploy, the Kafka topic initialization task (`iop_kafka : Initialize Kafka topics`) runs the expensive `init.sh --create` operation, which spawns 17 parallel JVMs to create topics. This takes ~11.4 seconds even when all topics already exist and no actual work is needed.

This pattern mirrors the puppet-iop implementation (`puppet-iop/manifests/core_kafka.pp:80-91`), which guards topic creation with an `unless` check.

* Add `Check whether Kafka topics already exist` task that runs `/opt/kafka/init.sh --check` before topic creation
* Make the `Initialize Kafka topics` task conditional on `iop_kafka_check.rc != 0`
* The `--check` operation runs a single JVM vs 17 parallel JVMs in `--create`

Steps to reproduce:

* Deploy with IoP enabled: `./foremanctl deploy --flavor katello --features iop`
* Verify topics are created: `podman exec iop-core-kafka /opt/kafka/init.sh --check` (should succeed)
* Re-deploy without changes: `./foremanctl deploy --flavor katello --features iop`
* Observe timing: `iop_kafka : Check whether Kafka topics already exist` should run in ~1-2s and `Initialize Kafka topics` should be skipped
* Compare to before: previously `Initialize Kafka topics` ran every deploy and took ~11.4s

#### Checklist

### DIFF
--- a/src/roles/iop_kafka/files/kafka/init
+++ b/src/roles/iop_kafka/files/kafka/init
@@ -38,11 +38,37 @@ create_topics() {
 
   echo -e "====================="
   echo -e "Creating Kafka topics:"
+
+  existing_topics_list=$("$kafka_cmd" --bootstrap-server "$kafka_bootstrap_server" --list 2>&1)
+  list_exit_code=$?
+  if [ $list_exit_code -ne 0 ]; then
+    echo "Error: Failed to list Kafka topics" >&2
+    echo "Output/Error:" >&2
+    echo "$existing_topics_list" >&2
+    return 1
+  fi
+
+  local pids=()
   for topic in "${topics[@]}"; do
-    echo -e "Creating topic ""$topic"
-    $kafka_cmd --create --if-not-exists --topic "$topic" --bootstrap-server $kafka_bootstrap_server --partitions 1 replication-factor 1 &
+    if ! echo "$existing_topics_list" | grep -q -x -F -- "$topic"; then
+        echo -e "Creating topic ""$topic"
+        $kafka_cmd --create --if-not-exists --topic "$topic" --bootstrap-server $kafka_bootstrap_server --partitions 1 replication-factor 1 &
+        pids+=($!)
+    else
+        echo -e "$topic exists"
+    fi
   done
-  wait
+
+  local failed=0
+  for pid in "${pids[@]}"; do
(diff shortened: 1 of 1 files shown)


src/roles/iop_kafka/tasks/main.yaml
```diff
@@ -67,10 +67,19 @@
...
+    name: iop-core-kafka
+    command: /opt/kafka/init.sh --check
+  register: iop_kafka_check
+  changed_when: false
+  failed_when: false
+
 - name: Initialize Kafka topics
   containers.podman.podman_container_exec:
     name: iop-core-kafka
     command: /opt/kafka/init.sh --create
```
evgeni, line 73: Looking at https://github.com/theforeman/foremanctl/blob/1ceae62e32b245158abd9fdc708e6f2d40bf1c45/src/roles/iop_kafka/files/kafka/init#L37-L50 I wonder if we could make create smarter instead?

The complaint is that we run `kafka_cmd` 17 times -- which is slow, but we could avoid doing that?

ehelms: I have not seen in my research any way to bulk create topics.

evgeni: There is none. But we can list them (as `--check` does) and only create the missing ones.
```
existing_topics_list=$("$kafka_cmd" --bootstrap-server "$kafka_bootstrap_server" --list 2>&1)
for topic in "${topics[@]}"; do 
if ! echo "$existing_topics_list" | grep -q -x -w "$topic"; then
    echo -e "Creating topic ""$topic"
    $kafka_cmd --create --if-not-exists --topic "$topic" --bootstrap-server $kafka_bootstrap_server --partitions 1 replication-factor 1 &
else
    echo -e "$topic exists"
fi
done
```

ehelms: OIC -- you want to make the script smarter rather than have a separate task.

evgeni: Yeah

evgeni (approved):

PR #797, opened 2026-08-30
Fix plugin list ordering for quadlet idempotency

### LINKED ISSUE #767: features_to_foreman_plugins is non-deterministic and emits duplicates, so FOREMAN_ENABLED_PLUGINS churns between deploys
### Summary

`filter_plugins/foremanctl.py` builds the plugin list by way of `get_dependencies()`, which returns a `set`, and then iterates that set. Python randomises string hashing per process, so the rendered `FOREMAN_ENABLED_PLUGINS` value changes between otherwise identical runs. The same code path also emits a duplicate whenever a feature appears both in the requested list and in another feature's dependency closure.

The practical consequence is that a repeat `foremanctl deploy` frequently rewrites the Foreman, dynflow and db-migrate quadlets for no reason and restarts Foreman, so the deploy is not idempotent.

### Cause

In `foreman_plugins()`:

```python
dependencies = list(get_dependencies(filter_features(value)))
plugins = [FEATURE_MAP.get(f, {}).get('foreman', {}).get('plugin_name')
           for f in filter_features(value + dependencies)]
```

`get_dependencies()` accumulates into a `set` and returns it, so `list(...)` order is arbitrary. And because `value + dependencies` concatenates the requested features with their closure, any feature that is both requested and depended upon is mapped twice.

### Measurement

Running the filter in six separate interpreters, with one extra feature enabled whose dependencies include an already-requested feature, produced **three distinct orderings**, with one plugin duplicated in all six:

```
katello foreman_kernel_care foreman_remote_execution katello foreman-tasks
katello foreman_kernel_care katello foreman_remote_execution foreman-tasks
katello foreman_kernel_care foreman-tasks foreman_remote_execution katello
```

Pinning `PYTHONHASHSEED=0` collapses this to a single stable ordering, which confirms set iteration order as the cause. A deployed quadlet on a live host carried the duplicate.

### Expected

A deterministic, deduplicated list. Sorting and de-duplicating before the join would fix both, and would make the rendered unit files stable across deploys.

Version: foremanctl 2.3.0.

ehelms:

Fix non-deterministic plugin list ordering that causes quadlet files to regenerate on every deploy, triggering unnecessary service restarts. Currently, the `FOREMAN_ENABLED_PLUGINS` environment variable in foreman.service has unstable ordering with duplicates, causing the quadlet file to differ between consecutive deploys even when nothing changed. This contributes to Item 5 (quadlet idempotency) from redeploy-timing-improvements-plan.md.

Example diff showing the problem:
```
< Environment='FOREMAN_ENABLED_PLUGINS=katello foreman_google foreman_remote_execution foreman_webhooks foreman_ansible foreman-tasks foreman_remote_execution katello foreman_rh_cloud'
---
> Environment='FOREMAN_ENABLED_PLUGINS=katello foreman_google foreman_remote_execution foreman_webhooks foreman_ansible katello foreman_remote_execution foreman_rh_cloud foreman-tasks'
```

Note the duplicates (katello, foreman_remote_execution) and different ordering.

* Deduplicate features before extracting plugin names using `dict.fromkeys()`
* Sort plugin lists alphabetically for stable ordering
* Apply fix to `foreman_plugins()`, `hammer_plugins()`, and `foreman_proxy_plugins()` filters
* Add unit tests for deduplication and sorting behavior

Steps to reproduce:

* Deploy foremanctl: `./foremanctl deploy --foreman-initial-admin-password=changeme --tuning development`
* Save quadlet file: `cp /etc/containers/systemd/foreman.service /tmp/foreman.service.1`
* Run second deploy: `./foremanctl deploy --foreman-initial-admin-password=changeme --tuning development`
* Save quadlet again: `cp /etc/containers/systemd/foreman.service /tmp/foreman.service.2`
* Compare: `diff /tmp/foreman.service.1 /tmp/foreman.service.2`
* Verify FOREMAN_ENABLED_PLUGINS is identical (no diff)
* Run unit tests: `cd tests/unit && python -m pytest filter_test.py::test_foreman_plugins_sorted -v`

Expected behavior:
- Plugin lists are alphabetically sorted
- No duplicates in plugin lists
- Quadlet files are byte-for-byte identical on consecutive deploys

#### Checklist

### DIFF
--- a/src/filter_plugins/foremanctl.py
+++ b/src/filter_plugins/foremanctl.py
@@ -61,8 +61,9 @@ def get_dependencies(features):
 
 def foreman_plugins(value):
     dependencies = list(get_dependencies(filter_features(value)))
-    plugins = [FEATURE_MAP.get(feature, {}).get('foreman', {}).get('plugin_name') for feature in filter_features(value + dependencies)]
-    return compact_list(plugins)
+    features = set(filter_features(value + dependencies))
+    plugins = [FEATURE_MAP.get(feature, {}).get('foreman', {}).get('plugin_name') for feature in features]
+    return sorted(compact_list(plugins))
 
 
 def available_foreman_plugins(_value):
@@ -113,14 +114,16 @@ def conflicting_features(features):
 
 def hammer_plugins(value):
     dependencies = list(get_dependencies(filter_features(value)))
-    plugins = [FEATURE_MAP.get(feature, {}).get('hammer') for feature in filter_features(value + dependencies)]
-    return compact_list(plugins)
+    features = set(filter_features(value + dependencies))
+    plugins = [FEATURE_MAP.get(feature, {}).get('hammer') for feature in features]
+    return sorted(compact_list(plugins))
 
 
 def foreman_proxy_plugins(value):
     dependencies = list(get_dependencies(filter_features(value)))
-    plugins = [FEATURE_MAP.get(feature, {}).get('foreman_proxy', {}).get('plugin_name') for feature in filter_features(value + dependencies)]
-    return compact_list(plugins)
+    features = set(filter_features(value + dependencies))
+    plugins = [FEATURE_MAP.get(feature, {}).get('foreman_proxy', {}).get('plugin_name') for feature in features]
+    return sorted(compact_list(plugins))
 
 
 def available_foreman_proxy_plugins(_value):
--- a/tests/unit/filter_test.py
+++ b/tests/unit/filter_test.py
@@ -1,5 +1,8 @@
 from foremanctl import FEATURE_MAP
 from foremanctl import conflicting_features
+from foremanctl import foreman_plugins
+from foremanctl import foreman_proxy_plugins
+from foremanctl import hammer_plugins
 from foremanctl import list_all_features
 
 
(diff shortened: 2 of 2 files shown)


ekohl:
This fixes #767, right? If not, please verify if it could.

evgeni (dismissed):

src/filter_plugins/foremanctl.py
```diff
@@ -61,8 +61,9 @@ def get_dependencies(features):
 
 def foreman_plugins(value):
     dependencies = list(get_dependencies(filter_features(value)))
-    plugins = [FEATURE_MAP.get(feature, {}).get('foreman', {}).get('plugin_name') for feature in filter_features(value + dependencies)]
-    return compact_list(plugins)
+    features = list(dict.fromkeys(filter_features(value + dependencies)))
```
evgeni: Why is this using `dict.fromkeys` and not `set` here?

src/filter_plugins/foremanctl.py
ekohl, line 1: I think this file would benefit lot from type annotations

src/filter_plugins/foremanctl.py
ehelms: My understanding is `dict.fromkeys` both ensures uniqueness and ordering consistently. I switched these to use sorted and set to ensure the list is alphabetical.

evgeni: It does, maybe, but that's not its purpose.
And I see no good reason to sort the features here anyway.
The return value is sorted, and that's what matters?

ehelms: I totally missed seeing that. Fixed.

evgeni (approved):

PR #798, opened 2026-08-30
Run database migrations only when pending

ehelms:

Foreman database migration containers currently run on deployments where their databases are already up to date. These containers add unnecessary startup time. Database migrations should be checked explicitly, and the migration containers should only run when pending migrations are reported.

* Add one-shot Foreman migration-check containers.
* Run the migration container only when the migration check reports pending migrations.
* Remove database seeding from the Foreman migration container; Foreman performs seeding during application startup.

Steps to reproduce:

* Run `ANSIBLE_COLLECTIONS_PATH="$PWD/build/collections/foremanctl" ANSIBLE_COLLECTIONS_SCAN_SYS_PATH=false bash -c 'source .venv/bin/activate && cd src && ansible-lint'`; expect no findings.
* Run `source .venv/bin/activate && ruff check tests/ src/ development/scripts/ inventories/`; expect all checks to pass.
* Run `git diff --check`; expect no whitespace errors.
* Deploy on a provisioned test VM and verify that the migration service runs when migrations are pending and is skipped when the database is current.

#### Checklist

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -129,8 +129,6 @@
         PartOf=foreman.target
         Wants=valkey.service postgresql.service candlepin.service
         After=valkey.service postgresql.service candlepin.service
-        Wants=foreman-db-migrate.service
-        After=foreman-db-migrate.service
   notify: Restart foreman
 
 - name: Install foreman-rake wrapper
@@ -163,8 +161,6 @@
         PartOf=foreman.target
         Wants=valkey.service postgresql.service
         After=valkey.service postgresql.service
-        Wants=foreman-db-migrate.service
-        After=foreman-db-migrate.service
       - |
         [Service]
         Restart=on-failure
@@ -222,6 +218,24 @@
   loop_control:
     label: "{{ item.instance }}"
 
+- name: Deploy foreman database migration check container
+  containers.podman.podman_container:
+    name: foreman-db-migration-check
+    state: quadlet
+    image: foreman.image
+    sdnotify: false
+    network: host
+    command: bash -c "bin/rails db:abort_if_pending_migrations"
+    env: "{{ foreman_env }}"
+    secrets: "{{ foreman_secrets }}"
+    quadlet_options:
+      - |
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        SuccessExitStatus=1
+        TimeoutStartSec=30m
+
 - name: Deploy foreman database migration container
   containers.podman.podman_container:
     name: foreman-db-migrate
@@ -229,7 +243,7 @@
     image: foreman.image
     sdnotify: false
     network: host
-    command: bash -c "bin/rails db:migrate && bin/rails db:seed"
+    command: bash -c "bin/rails db:migrate"
     env: "{{ foreman_env }}"
     secrets: "{{ foreman_secrets }}"
     quadlet_options:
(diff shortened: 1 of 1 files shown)


shubhamsg199 (approved):
Ack, similar approach I suggested in https://github.com/theforeman/foremanctl/issues/769#issuecomment-5423734131

src/roles/foreman/tasks/main.yaml
```diff
@@ -243,10 +243,32 @@
...
+        foreman_target_image_info.images | length == 0 or
+        foreman_running_container_info.containers | length == 0 or
+        foreman_running_container_info.containers[0].Image != foreman_target_image_info.images[0].Id
+      }}
+
 - name: Migrate and seed the Foreman database
   ansible.builtin.systemd:
     name: foreman-db-migrate.service
     state: restarted
+  when: foreman_migration_needed
```
evgeni, line 284: In `src/roles/foreman_proxy/handlers/main.yml` we use `state: "{{ (_foreman_proxy_service is changed) | ternary('started', 'restarted') }}"` and I wonder if we should do the same here.


```suggestion
    state: "{{ foreman_migration_needed | ternary('started', 'restarted') }}"
```

That way if the service is not started yet (which it should be usually, it's `Type=oneshot` with `RemainAfterExit=yes`), it will be triggered even if we messed up `foreman_migration_needed` for some reason.

OTOH, it's not enabled to be run on boot, right? So after a reboot foremanctl would still run migrations once even if unneeded?

src/roles/foreman/tasks/main.yaml
```diff
@@ -243,10 +243,32 @@
...
+  register: foreman_running_container_info
+  failed_when: false
+
+- name: Determine if Foreman migration is needed
+  ansible.builtin.set_fact:
+    foreman_migration_needed: >-
+      {{
+        foreman_target_image_info.images | length == 0 or
+        foreman_running_container_info.containers | length == 0 or
+        foreman_running_container_info.containers[0].Image != foreman_target_image_info.images[0].Id
```
evgeni: When the user pulled the image and restarted the service themselves, this won't detect a needed migration and the user needs to start foreman-db-migrate themselves too, right?
In the past, foreman-installer would "fix" this situation, regardless what mess the user produced.

evgeni:
I really wish we could run `db:abort_if_pending_migrations` and ask Rails whether migrations are needed, instead of guessing.

src/roles/foreman/tasks/main.yaml
ehelms: Correct - that is one downside of this approach. I also considered storing image digests in `parameters.yaml` (or an equivalent like file) on every foremanctl run as the comparison mechanism so it's judged based on `foremanctl` view of the world and not outside influences.

ehelms:
> I really wish we could run `db:abort_if_pending_migrations` and ask Rails whether migrations are needed, instead of guessing.

We can. I tried this approach as well and it works. It's just not as efficient because booting our environment is slow. When I researched how to make our boot time faster I basically encountered needing to go explore bootsnap which we only enable in development. Based on the time savings of other big apps (e.g. Shopify) we could possibly get our boot time down to sub 5 seconds.

evgeni:
Stupid question: Why can't we migrate on app boot (so in `config.after_initialize` or so)?

ehelms:
> Stupid question: Why can't we migrate on app boot (so in `config.after_initialize` or so)?

I think we avoided this primarily due to potential race conditions with Puma having multiple workers, the need to ensure orchestration of the foreman instance before dynflow, and to keep the door open to multi-container instances in the future.

src/roles/foreman/tasks/main.yaml
evgeni: I'd avoid adding more "state" to foremanctl

evgeni (approved):

ehelms:
So I realized there is a signal missing in this design that would cause problems and that is if a feature is enabled that enables a plugin this would not signal the migration. That leaves us with two options:

1. Keep this design but add a check if the quadlet changed and use that as a signal that migration is needed
2. Swap to check if there are pending migrations, accept the cost of the Rails environment boot up for now and try to tackle that problem instead.

ehelms:
I have reworked this to the check if there are pending migrations model as that is the most reliable. I tried out checking for image change or quadlet change but those both still felt like there could be edge cases.

src/roles/pulp/defaults/main.yaml
```diff
@@ -29,6 +29,8 @@ pulp_pulp_url: "http://{{ ansible_facts['fqdn'] }}:24817"
 pulp_enable_analytics: false
 pulp_mirror: false
 pulp_register_foreman_proxy: true
+# Pulp post-migration hooks must run even when there are no pending migrations.
```
evgeni: If it must always run, why bother checking if it needs to run? :)

ehelms: Great question, this came from https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/database.pp#L39. Re-reading the PR (https://github.com/theforeman/puppet-pulpcore/pull/351) I see we did this as a quirk of puppet idempotency.

ehelms: I removed the Pulp handling from the code, this just does Foreman now.

PR #799, opened 2026-08-30
Run only rails server in foreman container command

ehelms:

The foreman container inherits the image's default CMD, which chains three
full-boot `rails` invocations:

```
rails db:migrate && rails db:seed && rails server ...
```

Each is a separate process that loads the full Rails + plugin environment.
foremanctl already runs `db:migrate` and `db:seed` via the dedicated
`foreman-db-migrate.service` one-shot, which the foreman container `Requires=`
and runs `After=`. So the `db:migrate` and `db:seed` in the container CMD are
pure duplication — two extra full Rails boots (~14s) on every foreman start,
doing work the one-shot already did.

Journal analysis of a restart confirmed the identical env-load sequence
(`Rails cache backend: Redis` -> plugin permission registration -> app metadata)
running three times, with `=> Booting Puma` appearing only on the third; the
second pass ends with `Acquiring Advisory-Transaction-Lock 'database_seed'`.

Follow-up to the `Requires=` -> `Wants=` restart fix (#769).

* Override the foreman container command to run only `rails server`, so the
  container no longer re-runs `db:migrate` and `db:seed` on start. Migration and
  seeding remain owned by `foreman-db-migrate.service` at deploy time.

Steps to reproduce:

* Deploy foreman, then time a restart: `time systemctl restart foreman`.
* Before: ~26s, with three env-load passes in
  `journalctl -u foreman -o short-monotonic`.
* After: ~11s, a single env-load pass (one `=> Booting Puma`), workers fork in
  ~0.4s, socket activation / sdnotify / preload all intact.
* Confirm migrate/seed still run on a full `./foremanctl deploy` (the
  `Migrate and seed the Foreman database` task restarts
  `foreman-db-migrate.service`).
* Confirm the app serves `/api/v2/ping` after restart. Note a bare
  `systemctl restart foreman` intentionally no longer migrates — the one-shot
  owns that at deploy time.

#### Checklist

### DIFF
--- a/src/roles/foreman/tasks/main.yaml
+++ b/src/roles/foreman/tasks/main.yaml
@@ -115,6 +115,7 @@
     sdnotify: true
     network: host
     hostname: "{{ ansible_facts['hostname'] }}.local"
+    command: "/usr/share/foreman/bin/rails server --environment production --pid /tmp/rails.pid"
     volume:
       - 'foreman-data-run:/var/run/foreman:rw,z,U'
     secrets: "{{ foreman_secrets }}"


ehelms:
We can decide to also fix this directly in the container. I started here to be able to clearly see the effect, and because this is where the orchestration that makes it redundant happens.

src/roles/foreman/tasks/main.yaml
```diff
@@ -115,6 +115,7 @@
     sdnotify: true
     network: host
     hostname: "{{ ansible_facts['hostname'] }}.local"
+    command: "/usr/share/foreman/bin/rails server --environment production --pid /tmp/rails.pid"
```
ekohl, line 118: The `/tmp/rails.pid` is the only reason we have a shared `/run` today. Will this allow us to drop that too?

evgeni: It's not. We need a shared space so that a Rails-uploaded file can be processed by a Dynflow worker.

evgeni:
Is there a reason to override it here, instead of fixing it in https://github.com/theforeman/foreman-oci-images/blob/00f56061770e7281f227782dd3f8cb4885b89047/images/foreman/Containerfile#L22?

(I am fine with doing the change to "only rails, no migrations", just unsure where it fits best)

ehelms:
> Is there a reason to override it here, instead of fixing it in https://github.com/theforeman/foreman-oci-images/blob/00f56061770e7281f227782dd3f8cb4885b89047/images/foreman/Containerfile#L22?
> 
> (I am fine with doing the change to "only rails, no migrations", just unsure where it fits best)

I think if our goal is for the `oci-*` images to support this installation path, then we should do it in the container itself.

ekohl:
Foreman will seed on startup when needed so that's redundant. Whether the migration is also needed is a good question. I can see arguments either way.

If we change foreman-oci-images, could we document the expectations on users in README?

evgeni:
Yeah, update to the container+readme seems reasonable.

Our Pulp containers also don't manage their db stuff automatically.

ekohl:
> Our Pulp containers also don't manage their db stuff automatically.

On the other hand, Candlepin does. For better, or worse.

My mind is desperately looking for consistency. Foreman and Pulp are multi-container (web app(s) + worker) while Candlepin is single-container. Perhaps the rule can be that single-container apps may manage their own DB while we can expect multi-container apps to require some orchestration?

Foreman could support single-container if we upgrade to Sidekiq 7 and implement [embedding](https://sidekiq.org/wiki/Embedding), but we're not there.

ehelms:
Added here -- https://github.com/theforeman/foreman-oci-images/pull/92

ehelms:
This was solved at the image level.

PR #802, opened 2026-08-31
Backup/Restore with tar volume size 

sjha4:
Allow backup/restore to split pulp content archive into volumes of specific size based on user input.
Implement compress-then-split for tar-volume-size

- Use tar -czf - | split instead of tar --tape-length -M
- Provides full gzip compression PLUS volume splitting
- Better for disconnected environments (less total data to transport)
- Files named: pulp-content.tar.gz.part0000, part0001, etc.
- Restore uses: cat part* | tar -xzf -
- Works with incremental backups (.snar files)
- Updated documentation with compression benefits and air-gapped workflows

Steps to reproduce:

* foremanctl backup /var/backup --tar-volume-size 125M 
* Foremanctl restore /var/backup/backup_DIR --tar-volume-size 125M 

A more detailed testing guide: https://gist.github.com/sjha4/b79a2b007394d152a3b524b7ab7999da

#### Checklist


src/roles/backup/tasks/pulp_content.yaml
```diff
@@ -2,7 +2,34 @@
...
+      when:
+        - tar_volume_size is defined
+        - tar_volume_size | length > 0
+        - tar_volume_size is not match('^\d+[bBcGKkMPTw]?$')
+
+    - name: Backup pulp content directory with encryption keys (compressed multi-volume)  # noqa: risky-shell-pipe
+      ansible.builtin.shell:
+        cmd: >
+          tar -czf -
+          --listed-incremental="{{ backup_dir_full }}/.pulp.snar"
```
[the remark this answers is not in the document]
ianballou: I discovered this as well, and it broke testing.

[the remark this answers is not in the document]
vsedmik: PRT bumped into the same here https://github.com/SatelliteQE/robottelo/pull/22670#issuecomment-5665148548

src/roles/restore/tasks/restore_pulp_content.yaml
```diff
@@ -19,21 +25,47 @@
...
 
-    - name: Extract pulp content archive (incremental with snar)  # noqa: command-instead-of-module
+    - name: Extract pulp content archive (compressed multi-volume incremental)  # noqa: risky-shell-pipe
+      ansible.builtin.shell:
+        cmd: >
+          cat {{ restore_backup_dir }}/pulp-content.tar.gz.part* |
+          tar -xzf - --listed-incremental=/dev/null -C {{ pulp_storage_path }}
+      when:
+        - restore_pulp_content_volumes.matched > 0
+        - restore_pulp_snar_check.stat.exists
```
ianballou, line 42: Where is `restore_pulp_snar_check` defined?

src/roles/backup/tasks/pulp_content.yaml
```diff
@@ -2,7 +2,42 @@
...
+      when:
+        - tar_volume_size is defined
+        - tar_volume_size | length > 0
+      changed_when: true
+
+    - name: Backup pulp content directory with encryption keys (compressed multi-volume)  # noqa: risky-shell-pipe command-instead-of-module
+      ansible.builtin.shell:
+        cmd: >
+          tar -czf -
+          --listed-incremental="{{ backup_dir_full }}/.pulp.snar"
```
ianballou:
```suggestion
          --listed-incremental="{{ backup_dir_full }}/pulp.snar"
```

src/roles/backup/tasks/pulp_content.yaml
```diff
@@ -2,7 +2,42 @@
...
+        - tar_volume_size is defined
+        - tar_volume_size | length > 0
+        - tar_volume_size is not match('^[1-9]\d*[bBcGKkMPTw]?$')
+
+    - name: Remove existing pulp content volume files
+      ansible.builtin.shell:
+        cmd: rm -f "{{ backup_dir_full }}/pulp-content.tar.gz.part"*
+      when:
+        - tar_volume_size is defined
+        - tar_volume_size | length > 0
```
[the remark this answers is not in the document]
sjha4: We'd expect admins to handle cleanup..backup command cleaning out folders would be unexpected behavior IMO..

src/roles/restore/tasks/restore_pulp_content.yaml
sjha4: Updated in last push.. 👍🏼

docs/user/backup-restore.md
```diff
@@ -103,6 +104,80 @@ foremanctl backup /var/backup --base-backup /var/backup/foreman-backup-20260629T
...
+```
+
+**How it works:**
+1. Pulp content is compressed with gzip (typically 80-90% compression, varies by content type)
+2. The compressed stream is split into fixed-size chunks
+3. Each chunk is a valid piece of the compressed archive
+4. On restore, chunks are concatenated and decompressed
+
+**Benefits:**
+- ✅ **Full compression** - Get 80-90% space savings (varies by content type)
```
ianballou, line 131: This makes it sound like the splitting saves space over not splitting, but compression is involved in both cases right?

Also we got some emojis leaking in here :)

src/roles/restore/tasks/validate.yaml
```diff
@@ -230,12 +230,23 @@
...
-- name: Fail if insufficient Pulp disk space
+- name: Find Pulp content volume files (multi-volume)
+  ansible.builtin.find:
+    paths: "{{ restore_backup_dir }}"
+    patterns: "pulp-content.tar.gz.part*"
+  register: restore_pulp_volumes
+
+- name: Calculate total volume size (multi-volume)
+  ansible.builtin.set_fact:
+    restore_pulp_volumes_size: "{{ restore_pulp_volumes.files | map(attribute='size') | sum }}"
```
ianballou, line 246: Is this saving the compressed file size? Do we want the uncompressed size here to fail on?

src/roles/restore/tasks/restore_pulp_content.yaml
```diff
@@ -19,21 +30,51 @@
...
+    - name: Extract pulp content archive (compressed multi-volume full)  # noqa: risky-shell-pipe
+      ansible.builtin.shell:
+        cmd: >
+          set -o pipefail;
+          cat "{{ restore_backup_dir }}"/pulp-content.tar.gz.part* |
+          tar -xzf - -C "{{ pulp_storage_path }}"
+        executable: /bin/bash
+      when:
+        - restore_pulp_content_volumes.matched > 0
+        - not restore_pulp_snar_check.stat.exists
```
ianballou, line 54: Do we always pass --listed-incremental? I think if so, won't there always be a `snar` file?

ianballou (commented):
I tested the happy path and it's working! Left some comments.

Should we add a negative test for what happens if the split volume isn't complete? I wonder if the user sees something helpful.

I don't suppose some validation that that the tar volume is complete would be helpful. If users are splitting the tar they may be copying these files separately - I wonder if they'd end up with a broken restore if they forgot some of the files.

src/roles/restore/tasks/validate.yaml
sjha4: Ya..This was not right..Updated to save pulp size in metadata.yaml during backup and validating disk space based on that.

sjha4:
Had claude cook up validation around the multi-volume archive in the last push:
 During Backup:
  - Save backup_pulp_volume_count (number of volumes created)
  - Save backup_pulp_volume_size (the size parameter used)
  - Store both in metadata.yml
During Restore:
  - Volume count check: Verifies the number of .part* files matches the expected count from metadata
  - Sequential numbering check: Ensures volume files are numbered 0, 1, 2, 3... with no gaps

src/roles/restore/tasks/restore_pulp_content.yaml
sjha4: This was overly defensive..Updated..

docs/user/backup-restore.md
```diff
@@ -332,6 +407,45 @@ The restore command validates the backup chain:
 
 If any validation fails, a clear error message explains which backup is missing or incorrect.
 
+### Restore Multi-Volume Backup
+
+Restore from a compressed multi-volume backup created with `--tar-volume-size`:
+
+```bash
+foremanctl restore /var/backup/foreman-backup-20260629T120000 --tar-volume-size 5G
```
ehelms, line 415: The user has to know what tar volume size was used during the backup? That seems slightly fragile. Could this be part of the backup metadata?

docs/user/backup-restore.md
```diff
@@ -332,6 +407,45 @@ The restore command validates the backup chain:
...
+
+**How it works:**
+1. All `.part0000`, `.part0001`, etc. files are concatenated in order
+2. The combined compressed stream is piped to tar for extraction
+3. Content is decompressed and extracted to `/var/lib/pulp`
+
+**Important:**
+- All volume files (`.part0000`, `.part0001`, etc.) must be present in the backup directory
+- Files must be in the same directory and properly numbered
+- Missing volume files will cause the restore to fail
```
ehelms, line 426: I have not read ahead, is this up to the user or does the `restore` command check to ensure all expected volume files are present? If the tool does, let's mention that fact here.

src/roles/backup/tasks/pulp_content.yaml
```diff
@@ -2,7 +2,41 @@
...
+    - name: Validate tar_volume_size format
+      ansible.builtin.fail:
+        msg: "Invalid tar_volume_size format '{{ tar_volume_size }}'. Use format like '5G', '100M' (see 'man split' SIZE argument)"
+      when:
+        - tar_volume_size is defined
+        - tar_volume_size | length > 0
+        - tar_volume_size is not match('^[1-9]\d*([bB]|[KMGTPEZYkmgtpezy][bB]?)?$')
+
+    - name: Remove existing pulp content volume files
+      ansible.builtin.shell:
```
ehelms, line 14: Can the `file` module be used to ensure `absent` on this ?

src/roles/backup/tasks/pulp_content.yaml
```diff
@@ -2,7 +2,41 @@
...
+          tar -czf -
+          --listed-incremental="{{ backup_dir_full }}/pulp.snar"
+          --directory="{{ backup_pulp_storage_path }}"
+          --exclude=media/exports
+          --exclude=media/imports
+          --exclude=media/sync_imports
+          media
+          database_fields.symmetric.key
+          django_secret_key
+          | split -b {{ tar_volume_size }} -d --suffix-length=6 - "{{ backup_dir_full }}/pulp-content.tar.gz.part"
```
ehelms, line 31: Can this use full option names for split and tar?

docs/user/backup-restore.md
sjha4: The restore commands checks for 2 things during validation:

**Volume count check**: Verifies the number of .part* files matches the expected count from metadata
**Sequential numbering check**: Ensures volume files are numbered 0, 1, 2, 3... with no gaps

docs/user/backup-restore.md
sjha4: This is entirely optional and is not validated..Let me remove that to avoid confusion..

PR #804, opened 2026-09-01
Configure foreman to serve proxy deployment images

arvind4501:

* introduce a command to configure foreman to serve images needed for proxy deployment

Steps to reproduce:

*

#### Checklist


src/vars/proxy-registry-images.yml
ekohl, line 1: Could we take https://github.com/theforeman/foremanctl/blob/master/src/vars/images.yml as a base to reduce duplication?

arvind4501: in theory we can merge both in one, but i would not like to for the fact that the proxy images has significantly more metadata(which is defined in a pattern that we don't need for other non-proxy images) then we have in the images.yml.

ekohl: I'm thinking in terms of branding downstream, but we already define the images in another way (https://github.com/theforeman/foremanctl/blob/master/docs/developer/deployment.md#rpm-provided-images) so it's not like we have that information anyway and we'll have duplication regardless.

src/playbooks/setup-proxy-images/setup-proxy-images.yaml
```diff
@@ -0,0 +1,135 @@
...
+             | map(attribute='repository_name')
+             | map('regex_replace', '^', '  ' ~ (org_label | lower) ~ '/' ~ (product_label | lower) ~ '/')
+             | list }}
+
+    - name: Query the container registry catalog
+      ansible.builtin.uri:
+        url: "{{ foreman_url }}/v2/_catalog"
+        method: GET
+        user: "{{ foreman_initial_admin_username }}"
+        password: "{{ foreman_initial_admin_password }}"
```
ekohl: We can't rely on these credentials because a user can change them. Not sure which mechanism we do have.

docs/user/proxy-image-serving.md
```diff
@@ -0,0 +1,152 @@
+# Proxy Image Serving Setup
+
+This guide covers configuring Katello to serve container images for
+proxy deployments.
+
+## Overview
+
+When deploying proxies in enviornments where we don't want the proxy to consume container images(for proxy deployment) directly from internet, we can configure our foreman server to serve these images as registry.
```
ekohl:
```suggestion
When deploying proxies in environments where we don't want the proxy to consume container images(for proxy deployment) directly from internet, we can configure our foreman server to serve these images as registry.
```

ekohl (commented):
Overall I like the idea that a user is able to manage data entities from the CLI, though we should avoid trying to replicate Hammer. This very specialized implementation that helps a particular workflow is IMHO a good middle ground.

src/vars/proxy-registry-images.yml
arvind4501: by branding downstrem you mean overriding the `proxy-registry-images` ? 
for that i have thought of something like

> *Downstream / vendor images:** vendors (for example, Red Hat Satellite) ship a `src/vars/vendor/proxy-registry-images.yml` file that replaces this image set with their own — typically images synced from `registry.redhat.io`. When that file is present it is loaded automatically, so the product, repository names, upstream images, tags, and registry may differ from the default.

src/playbooks/setup-proxy-images/setup-proxy-images.yaml
arvind4501: you're right. in other apis using FAM modules i used oauth but i don't think we can use same for `/v2/_catalog` thats related to container registry

ekohl: Exactly that.

Sorry to call you out @ianballou but do you have any idea how we could obtain the same information with credentials that we can control?

docs/user/proxy-image-serving.md
```diff
@@ -0,0 +1,152 @@
...
+## Manual Verification
+
+The command already asserts that every repository is present in the container
+registry catalog before it finishes. To inspect the catalog yourself:
+
+```bash
+SERVER=foreman.example.com
+
+curl -sk -u USERNAME:PASSWORD \
+  https://$SERVER/v2/_catalog | python3 -m json.tool
```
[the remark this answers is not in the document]
arvind4501: the user running this is already a root so,

src/playbooks/setup-proxy-images/setup-proxy-images.yaml
```diff
@@ -0,0 +1,135 @@
...
+      ansible.builtin.set_fact:
+        product_label: "{{ product_info.resources[0].label }}"
+
+    - name: Build published registry paths
+      ansible.builtin.set_fact:
+        registry_repo_paths: >-
+          {{ proxy_registry_images.values()
+             | map(attribute='repository_name')
+             | map('regex_replace', '^', '  ' ~ (org_label | lower) ~ '/' ~ (product_label | lower) ~ '/')
+             | list }}
```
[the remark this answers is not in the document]
arvind4501: what do you recommend querying a catalag alredy comes with a concern raised above

> We can't rely on these credentials because a user can change them. Not sure which mechanism we do have.

consider that as well

docs/user/proxy-image-serving.md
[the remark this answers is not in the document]
arvind4501: what i meant above was, if a user is root, he is alredy have access to almost everything, so whats the point hiding the password from shell history

docs/user/proxy-image-serving.md
```diff
@@ -0,0 +1,152 @@
+# Proxy Image Serving Setup
+
+This guide covers configuring Katello to serve container images for
+proxy deployments.
+
+## Overview
+
+When deploying proxies in environments where we don't want the proxy to consume container images(for proxy deployment) directly from internet, we can configure our foreman server to serve these images as registry.
```
ehelms:
```suggestion
When deploying proxies in environments where we don't want the proxy to consume container images(for proxy deployment) directly from the internet, we can configure our foreman server (that has the content/container feature enabled) to serve these images as registry.
```

docs/user/proxy-image-serving.md
```diff
@@ -0,0 +1,152 @@
...
+- Creates a Product for the proxy container images
+- Creates a container repository for each required image
+- Syncs the repositories from the configured upstream registry
+- Verifies each repository is published in the container registry catalog
+  (`/v2/_catalog`)
+- Prints the registry paths
+
+## Prerequisites
+
+- A deployed and running Foreman/Katello server (`foremanctl deploy` completed).
```
ehelms:
```suggestion
- A Foreman server deployed with the content/container feature (`foremanctl deploy` completed).
```

docs/user/proxy-image-serving.md
```diff
@@ -0,0 +1,152 @@
...
+- Network access from the server to the upstream registry it syncs from
+  (`https://quay.io` by default).
+- Registry credentials **only if** your upstream registry requires authentication. The public `quay.io/foreman` images used by default are public and need no credentials.
+
+## What Gets Synced
+
+By default the command syncs the public upstream Foreman images from `https://quay.io` into a product named **`Proxy Container Images`**:
```
ehelms: Consider something like `Foreman Infra Container Images`

src/playbooks/setup-proxy-images/setup-proxy-images.yaml
```diff
@@ -0,0 +1,136 @@
...
+- name: Setup Proxy Container Images
+  hosts: quadlet
+  become: true
+  vars_files:
+    - "../../vars/images.yml"
+    - "../../vars/foreman.yml"
+    - "../../vars/base.yaml"
+    - "../../vars/proxy-registry-images.yml"
+
+  tasks:
```
ehelms: Given this has many tasks and default variables, consider moving it into a role.

ehelms:
Consider making this more "generic" by making the naming something like "Foreman Infra Container Images" as this could easily be used by users that have a hub-and-spoke model (a Foreman that other Foreman's pull from) and want to use the container content from Foreman's registry to manage multiple Foreman's.

docs/user/infra-image-serving.md
```diff
@@ -0,0 +1,180 @@
+# Serving Foreman Infra Container Images
+
+This guide covers configuring Foreman/Katello to serve the container images
```
ehelms:
```suggestion
This guide covers configuring Foreman with the content/container feature to serve the container images
```

docs/user/infra-image-serving.md
```diff
@@ -0,0 +1,180 @@
...
+for Foreman infrastructure deployment, so other machines can pull
+them from your server instead of the internet.
+
+## Overview
+
+When deploying foreman and proxies in environments where we don't want the them to consume container images(for deployment) directly from the internet, we can configure our foreman server (that has the content/container feature enabled) to serve these images as registry.
```
ehelms:
```suggestion
When deploying Foreman and proxies in environments where we don't want the them to consume container images(for deployment) directly from the internet, we can configure our Foreman server (that has the content/container feature enabled) to serve these images as registry.
```

docs/user/infra-image-serving.md
```diff
@@ -0,0 +1,180 @@
...
+
+- **Proxy deployments** — a proxy pulls its images from its parent Foreman
```
ehelms:
```suggestion
- **Proxy deployments** — a proxy can pull its images from its parent Foreman
```

docs/user/infra-image-serving.md
```diff
@@ -0,0 +1,180 @@
+# Serving Foreman Infra Container Images
```
ehelms, line 1: In general, the line breaks in this file don't match, can you go through the file and at least make it a full sentenace per line?

docs/user/infra-image-serving.md
```diff
@@ -0,0 +1,180 @@
...
+
+## What Gets Synced
+
+Images are published into a product named **`Foreman Infra Container Images`**.
+Which of them are published is chosen with `--image-set`:
+
+| Repository      | Upstream image                | `proxy` | `server` | `all` |
+| --------------- | ----------------------------- | :-----: | :------: | :---: |
+| `foreman`       | `foreman/foreman`             |         |    ✓     |   ✓   |
+| `foreman-proxy` | `foreman/foreman-proxy`       |    ✓    |          |   ✓   |
```
ehelms: This should have a check for server.

docs/user/proxy-image-serving.md
```diff
@@ -0,0 +1,152 @@
+# Proxy Image Serving Setup
```
ehelms: I assume this doc should be removed now?

src/vars/infra-registry-images.yml
```diff
@@ -0,0 +1,65 @@
...
+
+# Registry each image is synced from.
+infra_registry_url: https://quay.io
+
+# Which set of images to publish. Overridden with --image-set.
+infra_image_set: proxy
+
+# Image definitions available for the registry sync.
+# base_tag is the tag synced for each repository (they do not share a version).
+infra_registry_images:
```
ehelms, line 18: This now repeats some image information again. Have you thought about how to have a single source of truth?

src/vars/infra-registry-images.yml
```diff
@@ -0,0 +1,65 @@
...
+    upstream_name: sclorg/postgresql-16-c10s
+    base_tag: "{{ postgresql_container_tag }}"
+  valkey:
+    repository_name: valkey-8
+    upstream_name: sclorg/valkey-8-c10s
+    base_tag: "{{ valkey_container_tag }}"
+
+# Named selections of the images above. The 'all' set is not listed here: it is
+# every key of infra_registry_images, so it picks up vendor additions for free.
+infra_image_sets:
```
ehelms: I am wondering if we should care about this proxy / server split. Does it add value for the complexity it adds? Should we default to simple to start with and just include all images?

docs/user/infra-image-serving.md
arvind4501: by default katello flavor does not include `foreman-proxy` feature, and i kept only flavor features related images to be part of a `image-set`

src/vars/infra-registry-images.yml
arvind4501: yes, i agree with defaulting to all to make things simpler. earlier since it only started for proxy i was thinking to have a option for that. 
but including all images makes total sense, the only cost of syncing all images on server, if the only goal is to serve images for proxy deployment, is two extra image sync consuming storage. but i think that does not hurt  much.
i have updated to use all images by default now, thanks

src/roles/infra_images/tasks/main.yaml
```diff
@@ -0,0 +1,135 @@
...
+  ansible.builtin.set_fact:
+    infra_images_selected: "{{ infra_registry_images | dict2items }}"
+
+- name: Confirm at least one image is defined
+  ansible.builtin.assert:
+    that:
+      - infra_images_selected | length > 0
+    fail_msg: >-
+      No images are defined in infra_registry_images. A vendor override at
+      src/vars/vendor/infra-registry-images.yml replaces the whole dict, so it
```
ehelms, line 12: This path will be dependent on whether one is using the RPM or from git checkout.

ehelms:
Looking good!

- https://github.com/theforeman/foremanctl/pull/804/changes#r3962183512 is my biggest design question left
- Have you thought about testing for this?

src/vars/infra-registry-images.yml
arvind4501: thats a valid concern, while at first i thought of we want to define infra images metadata in a format that is easy to override by vendors. looking at  existing images.yaml , i am afraid we need to use the similar metadata in vendor override definations too.
what do you think about  using images.yaml and vendor/images.yaml as vendor overrides. i need to verify how the metadata defination changes for vendors.

sjha4:
Unclear on how versioning will work for this configured repo? Will we have different repos(unique names in katello) for different tags or should users be able to add tags to the same repo's included_tag as they go along with upgrades etc?

Second is this command can only be run once..Running it more than once will fail due to the duplicate names of product/repository unless we use the second run to update existing repo with new fields like tags.

src/playbooks/setup-proxy-images/setup-proxy-images.yaml
ianballou: I will check if there is a way to access the catalog with easier credentials, but there is also the option of using the Foreman API to get this data instead of the container registry.

ianballou: Certificates are one possible option, but it would need to be the debug cert that is used and I'm not sure it's been tested with our container registry. I would say it's an unsupported option right now.

It doesn't look like oauth will work with the container registry, so I think the choice is either username/pass or explore using the debug certificate.

So, if we're only querying for container repos, then the API might be best. But if pulling the images is involved, then we'll need to stick with username/pass or certs.

ianballou: @arvind4501 @ekohl  can we learn anything from https://github.com/Katello/katello/pull/11776 to help out here?

arvind4501:
> Unclear on how versioning will work for this configured repo? Will we have different repos(unique names in katello) for different tags or should users be able to add tags to the same repo's included_tag as they go along with upgrades etc?

for versioning we are using the existing base_tag from the image metadata in `infra-registry-images.yml` . for vendors like satellite they can create their own image override and provide a tag which they want to use. for upgrades the new image metadata with updated tag will be used to update the existing repos. it can leave the old synced repos on server  but since we updated the tag of repo, it will sync that version of repos. we can ideally clear the older one from synced storage.


> Second is this command can only be run once..Running it more than once will fail due to the duplicate names of product/repository unless we use the second run to update existing repo with new fields like tags.

no, this is idempotent, so it can be run any no. of time we  want. running it more than once won't fail and in case of upgrade the existing repo will be updated with updated tag, as we are using `present: true` not `present_with_defaults` which does not update https://docs.ansible.com/projects/ansible/latest/collections/theforeman/foreman/repository_module.html#parameter-state

src/playbooks/setup-proxy-images/setup-proxy-images.yaml
arvind4501: thanks @ianballou , i have used certs and that works

docs/user/infra-image-serving.md
```diff
@@ -0,0 +1,148 @@
+# Serving Foreman Infra Container Images
+
+This guide covers configuring Foreman with the content/container feature to serve the container images for Foreman infrastructure deployment, so other machines can pull them from your server instead of the internet.
+
+## Overview
+
+When deploying Foreman and proxies in environments where we don't want the them to consume container images(for deployment) directly from the internet, we can configure our Foreman server (that has the content/container feature enabled) to serve these images as registry.
```
ehelms, line 7:
```suggestion
When deploying Foreman and proxies in environments where we do not want the them to consume container images (for deployment) directly from the internet, we can configure our Foreman server (that has the content/container feature enabled) to serve these images as registry.
```

ehelms:
I was wondering if there was a strategy whereby the image services (generated by the `.image` quadlet`) could be the source of truth of the set of infra images and thus prevent us from having to define them twice.

PR #805, opened 2026-09-01
Fixes #39721 - Support non-RSA certificate generation

michalgritzbach:

The `certificates` role hardcoded RSA key parameters: every private key was generated with a type and a size, and every CSR requested the RSA-only `keyEncipherment` key usage. Deploying with EC keys was therefore impossible, and it blocks the eventual move to ML-DSA. `foreman-certificate-check` made the same assumption and rejected correctly generated non-RSA certificates that only carry Digital Signature.

Fixes #39721

* `certificates_algorithm_type` now accepts `ECC` next to `RSA`. It stays a single choice that applies to the whole certificate set: the CA, plus the server and client pair of every hostname in `certificates_hostnames`. The default remains RSA 4096.
* New `src/roles/certificates/vars/main.yml` maps each algorithm to the parameters `community.crypto.openssl_privatekey` expects for it — a size for RSA, a curve for ECC, neither for ML-DSA. A future algorithm, or custom certificates with PQC, can extend that table instead of the tasks.
* `ca.yml` and `issue.yml` build the CA, server and client keys from that table, passing `size` and `curve` only where the algorithm takes them.
* CSRs request `keyEncipherment` only for RSA. EC and ML-DSA keys cannot encipher keys, so their certificates carry `digitalSignature` alone.
* `foreman-certificate-check` requires the key usage matching the certificate's public key algorithm — Key Encipherment for RSA, Digital Signature otherwise — instead of always demanding Key Encipherment.
* New default `certificates_algorithm_curve: secp384r1`.
* New CLI parameters on `deploy` and `auth-bundle`: `--certificate-algorithm {RSA,ECC}`, `--certificate-algorithm-size` (RSA only) and `--certificate-algorithm-curve` (ECC only).
* `ML-DSA` is deliberately **not** offered. Key generation cannot work yet: the `python-cryptography` versions shipped in RHEL do not support it (IDM-1536), and `community.crypto` explicitly blocks ML-DSA key types ([#1044](https://github.com/ansible-collections/community.crypto/pull/1044), [#1056](https://github.com/ansible-collections/community.crypto/issues/1056)). Offering a value that can only fail at deploy time is worse than withholding it, so `ML-DSA` is absent from the `--certificate-algorithm` choices and from `_certificates_supported_algorithms`, and the role asserts the selected algorithm is supported before generating anything — which also covers values set through extra vars or the answers file rather than the CLI. Its key parameters stay in the algorithm table so re-enabling it is a one-line change once the blockers clear. ECC is the stand-in in the meantime; the limitation is documented.

Steps to reproduce:

* Run the unit tests, which generate a real certificate set through the role and validate it with `foreman-certificate-check`:

  ```
  python -m pytest tests/unit/certificates_algorithm_test.py tests/unit/certificates_issue_test.py tests/unit/foreman_ceritificate_check_test.py -vv
  ```

* Deploy with the default algorithm and confirm nothing changed for RSA:

  ```
  foremanctl deploy --foreman-initial-admin-password=changeme --tuning development
  openssl x509 -in /var/lib/foremanctl/certs/certs/$(hostname -f).crt -noout -text | grep -A2 'Public Key Algorithm:\|X509v3 Key Usage'
  ```

  Expected: `rsaEncryption`, 4096 bit, key usage `Key Encipherment`.

* Deploy with ECC on a fresh machine and inspect the generated set:

  ```
  foremanctl deploy --foreman-initial-admin-password=changeme --tuning development --certificate-algorithm=ECC
  for cert in /var/lib/foremanctl/certs/certs/*.crt; do
    echo "== $cert"
    openssl x509 -in "$cert" -noout -text | grep -A2 'Public Key Algorithm:\|X509v3 Key Usage'
  done
  ```

  Expected: `id-ecPublicKey` on the CA, server and client certificates, curve `secp384r1`, key usage `Digital Signature` with no `Key Encipherment`. The `certificate_checks` role passes during the same deploy.

* On an existing RSA deployment, switch algorithm and re-issue:

  ```
  foremanctl deploy --certificate-algorithm=ECC --certificate-ca-renew --certificate-renew
  ```

  Expected: the whole set is replaced with EC certificates. Without the renew flags, existing certificates are kept as-is.

* Check the CLI surface:

  ```
  foremanctl deploy --help | grep certificate-algorithm
  foremanctl auth-bundle --help | grep certificate-algorithm
  ```

  Expected: `--certificate-algorithm {RSA,ECC}`, with no `ML-DSA` in the choices.

* Confirm `ML-DSA` is rejected rather than attempted:

  ```
  foremanctl deploy --certificate-algorithm=ML-DSA
  ```

  Expected: obsah rejects the value at argument parsing. Setting `certificates_algorithm_type=ML-DSA` through extra vars instead fails on the role's assert with `Unsupported certificates_algorithm_type 'ML-DSA'. Supported algorithms: RSA, ECC.`

#### Checklist


src/roles/certificates/vars/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+# Algorithms foremanctl can generate keys for today. Kept separate from the
+# parameter table below so an algorithm can be staged before it is offered.
+_certificates_supported_algorithms:
+  - RSA
+  - ECC
```
pablomh: Could Ed25519 and Ed448 be added as well?

src/roles/certificates/tasks/ca.yml
```diff
@@ -41,8 +41,9 @@
     - name: 'Create CA private key'
       community.crypto.openssl_privatekey:
         path: "{{ certificates_ca_directory_keys }}/ca.key"
-        type: "{{ certificates_algorithm_type }}"
-        size: "{{ certificates_algorithm_size }}"
+        type: "{{ _certificates_key_parameters.type }}"
+        size: "{{ _certificates_key_parameters.size | default(omit) }}"
+        curve: "{{ _certificates_key_parameters.curve | default(omit) }}"
```
pablomh: curve: "{{ _certificates_key_parameters.curve if _certificates_key_parameters.type == 'ECC' else omit }}"

src/roles/certificates/tasks/issue.yml
```diff
@@ -6,8 +6,9 @@
     - name: 'Create server private key'
       community.crypto.openssl_privatekey:
         path: "{{ certificates_output_directory_keys }}/{{ certificates_hostname }}.key"
-        type: "{{ certificates_algorithm_type }}"
-        size: "{{ certificates_algorithm_size }}"
+        type: "{{ _certificates_key_parameters.type }}"
+        size: "{{ _certificates_key_parameters.size | default(omit) }}"
+        curve: "{{ _certificates_key_parameters.curve | default(omit) }}"
```
pablomh: curve: "{{ _certificates_key_parameters.curve if _certificates_key_parameters.type == 'ECC' else omit }}"

src/roles/certificates/tasks/issue.yml
```diff
@@ -38,8 +37,9 @@
 - name: 'Create client private key'
   community.crypto.openssl_privatekey:
     path: "{{ certificates_output_directory_keys }}/{{ certificates_hostname }}-client.key"
-    type: "{{ certificates_algorithm_type }}"
-    size: "{{ certificates_algorithm_size }}"
+    type: "{{ _certificates_key_parameters.type }}"
+    size: "{{ _certificates_key_parameters.size | default(omit) }}"
+    curve: "{{ _certificates_key_parameters.curve | default(omit) }}"
```
pablomh: curve: "{{ _certificates_key_parameters.curve if _certificates_key_parameters.type == 'ECC' else omit }}"

michalgritzbach:
Superseded by #806, which now combines non-RSA certificate generation (#39721) and validation (#39722) into a single commit targeting master.

Issue #220, opened 2025-07-30
Deal with crypto policies


ekohl:
In a non-containerized world you can use crypto-policies (on EL) to configure the TLS protocols & ciphers. Foreman has been moving to that to secure communications in an easy way.

crypto-policies are a user space implementation and containers have their own isolated user space. This means we need a way to manage this at scale.

On a related note, for https://github.com/theforeman/foremanctl/issues/219 it means we need to configure each container based on EL9 to use the `DEFAULT:PQ` policy when the user opts into it.

PR #806, opened 2026-09-01
Fixes #39721,#39722 - Support non-RSA certificate generation and validation

michalgritzbach:
Supersedes #805 by combining non-RSA certificate generation (#39721) and
non-RSA certificate validation (#39722).

### What this does

- The certificate algorithm stays a single choice applied to the whole
  set (CA plus the server and client pair of every hostname), but now
  accepts `ECC` and `ML-DSA-87` alongside `RSA`.
- `--certificate-algorithm-rsa-size` and `--certificate-algorithm-ecc-curve`
  are type-specific; a `forbidden_if` constraint in the playbook metadata
  rejects a size for a non-RSA algorithm and a curve for a non-ECC one.
- A vars file derives the parameters `community.crypto.openssl_privatekey`
  needs from the selected algorithm (a size for RSA, a curve for ECC,
  neither for ML-DSA) and the key usage each algorithm supports.
- CSRs request `keyEncipherment` only for RSA and `digitalSignature` for
  every algorithm, since EC and ML-DSA keys cannot do key encipherment.
  The CA carries `keyCertSign`, `cRLSign` and `digitalSignature` for every
  algorithm.
- `foreman-certificate-check` matches: the key-usage check requires Key
  Encipherment for RSA and Digital Signature otherwise, and the CA
  signing-algorithm check no longer relies on an RSA-centric `pkcs7`
  decode, so SHA-1 signatures are rejected regardless of the signing
  key's algorithm.
- Changing `--certificate-algorithm` on an existing deployment re-issues
  the leaf certificates automatically. If the new algorithm also differs
  from the CA key, the deploy fails and asks for `--certificate-ca-renew`
  rather than silently leaving the CA untouched or wiping it.

### Notes

- ML-DSA needs `community.crypto` 3.3.0 to generate the keys and
  `python-cryptography` 49.0.0 to sign certificates and CSRs. CentOS
  Stream 10 ships a suitable `python-cryptography`; Stream 9 ships 36.0.1,
  on which `ML-DSA-87` fails at key generation. Inspecting and validating
  ML-DSA certificates additionally needs the system OpenSSL 3.5+, so the
  unit tests that exercise ML-DSA are skipped where either piece is
  missing.

Fixes #39721
Fixes #39722


ehelms:
@michalgritzbach Can you update the pull request title? As I think this PR has gone well beyond the current title :grin:

docs/user/certificates.md
```diff
@@ -119,6 +119,31 @@ When certificates are generated by foremanctl, you can control lifetimes in days
...
+foremanctl deploy --certificate-algorithm=ECC
+
+# Generate an ECC certificate set on a different curve
+foremanctl deploy --certificate-algorithm=ECC --certificate-algorithm-curve=secp521r1
+```
+
+Certificates with non-RSA keys only carry the `digitalSignature` key usage, because those keys cannot perform key encipherment. RSA certificates keep both `digitalSignature` and `keyEncipherment`.
+
+> [!NOTE]
+> `ML-DSA` is not offered yet: neither the `python-cryptography` versions shipped in RHEL nor `community.crypto` can generate ML-DSA keys. The certificates role already carries its key parameters, so the option can be added once key generation works. Use `ECC` for non-RSA deployments in the meantime.
```
ehelms: I see three ML-DSA options available: https://docs.ansible.com/projects/ansible/latest/collections/community/crypto/openssl_privatekey_module.html#parameter-type

CentOS Stream 10 has https://mirror.stream.centos.org/10-stream/BaseOS/x86_64/os/Packages/python3-cryptography-49.0.0-1.el10.x86_64.rpm

michalgritzbach:
@ehelms It was originally supposed to be two pull requests and this one was based on the other (#805) (so the title was referring only to the one change that was supposed to be left after 805 was merged), but it turned out that 805 by itself wouldn't do anything and wasn't really mergeable. So the original pull request was closed and this one now contains both non-RSA generation as well as validation.

docs/user/certificates.md
```diff
@@ -119,6 +119,34 @@ When certificates are generated by foremanctl, you can control lifetimes in days
...
 
+### Key algorithm
+
+Certificates generated by foremanctl use RSA keys by default. A single algorithm applies to the whole certificate set: the CA, and the server and client certificates of every hostname.
+
+| CLI flag | Ansible variable | What it controls |
+| -------- | ---------------- | ---------------- |
+| `--certificate-algorithm` | `certificates_algorithm_type` | Key algorithm: `RSA` (default), `ECC`, `ML-DSA-44`, `ML-DSA-65` or `ML-DSA-87` |
+| `--certificate-algorithm-size` | `certificates_algorithm_size` | Key size, RSA only (default 4096) |
+| `--certificate-algorithm-curve` | `certificates_algorithm_curve` | Elliptic curve, ECC only (default `secp384r1`) |
```
ehelms: Given option 2 and 3 are specific to a particular type, I wonder if we should include that in the name?


```suggestion
| `--certificate-algorithm-ecc-curve` | `certificates_algorithm_curve` | Elliptic curve, ECC only (default `secp384r1`) |
```

docs/user/certificates.md
```diff
@@ -119,6 +119,34 @@ When certificates are generated by foremanctl, you can control lifetimes in days
...
+foremanctl deploy --certificate-algorithm=ECC --certificate-algorithm-curve=secp521r1
+
+# Generate an ML-DSA certificate set
+foremanctl deploy --certificate-algorithm=ML-DSA-65
+```
+
+Certificates with non-RSA keys only carry the `digitalSignature` key usage, because those keys cannot perform key encipherment. RSA certificates keep both `digitalSignature` and `keyEncipherment`.
+
+> [!NOTE]
+> The `ML-DSA-44`, `ML-DSA-65` and `ML-DSA-87` algorithms need `community.crypto` 3.3.0 to generate the keys and `python-cryptography` 49.0.0 to sign certificates and CSRs with them. CentOS Stream 10 ships a suitable `python-cryptography`; CentOS Stream 9 ships 36.0.1, on which selecting an ML-DSA algorithm fails at key generation. Candlepin is a further caveat: it is handed this CA and its legacy crypto scheme signs with `SHA256withRSA`, so a deployment that runs Candlepin still needs an RSA CA.
```
ehelms: Candlepin can be configured to use PQC certificates. @ianballou has investigated this and can help shed some light on what's needed.

ehelms: I am thinking we should align on `ML-DSA-87` only.

src/roles/certificates/tasks/main.yml
```diff
@@ -40,6 +40,28 @@
     - auth_bundle is not defined
     - certificates_generate
   block:
+    - name: 'Validate certificate algorithm'
```
ehelms: This should not be necessary, as parameter validation will catch this (https://github.com/theforeman/foremanctl/pull/806/changes#diff-68995961b1f4784a51711c04a402ef23d9b7fd0ba74dcb7654f828cb2fd34e4fR3)

src/roles/certificates/vars/main.yml
```diff
@@ -0,0 +1,34 @@
+---
```
ehelms, line 1: Consider https://github.com/theforeman/foremanctl/pull/806/changes#r3962235213 as that may make this file unnecessary.

tests/fixtures/foreman-certificate-check/certs/ca-ec-sha1-bundle.crt
```diff
@@ -0,0 +1,24 @@
+-----BEGIN CERTIFICATE-----
```
ehelms: This file is named `ca-ec-sha1-bundle`. Is EC different than ECC? If not, I think we should use consistent abbreviations.

tests/fixtures/foreman-certificate-check/certs/ca-mldsa.crt
```diff
@@ -0,0 +1,117 @@
+-----BEGIN CERTIFICATE-----
```
ehelms: Which MLDSA is this file?

ehelms:
Have you considered how to test deployments with different certificate types?

docs/user/certificates.md
```diff
@@ -119,6 +119,34 @@ When certificates are generated by foremanctl, you can control lifetimes in days
...
+# Generate an ML-DSA certificate set
+foremanctl deploy --certificate-algorithm=ML-DSA-87
+```
+
+Certificates with non-RSA keys only carry the `digitalSignature` key usage, because those keys cannot perform key encipherment. RSA certificates keep both `digitalSignature` and `keyEncipherment`.
+
+> [!NOTE]
+> The `ML-DSA-87` algorithm needs `community.crypto` 3.3.0 to generate the keys and `python-cryptography` 49.0.0 to sign certificates and CSRs with them. CentOS Stream 10 ships a suitable `python-cryptography`; CentOS Stream 9 ships 36.0.1, on which selecting `ML-DSA-87` fails at key generation. Candlepin is a further caveat: it is handed this CA and its legacy crypto scheme signs with `SHA256withRSA`, so a deployment that runs Candlepin still needs an RSA CA.
+
+Changing the algorithm does not re-issue existing certificates. Combine it with `--certificate-renew` and `--certificate-ca-renew` to replace a certificate set that already exists.
```
sjha4, line 148: As a user, this would be the flow to migrate deployments with old certificates? 
```
foremanctl deploy --certificate-algorithm=ML-DSA-87
foremanctl deploy --certificate-ca-renew
foremanctl deploy --certificate-renew
```

sjha4:
Would setting the algorithm also affect auth-bundle generation?

docs/user/certificates.md
ianballou: Yep! @aidenfine and I both have set up containerized Katellos in the past with Candlepin using ML-DSA certs. If you're having any troubles, I can share notes.

docs/user/certificates.md
ehelms: I think they would `foremanctl deploy --certificate-algorithm=ML-DSA-87 --certificate-ca-renew` as you should do it all at once and the CA renew ought to trigger certificate renew or else things will be in a weird state.

docs/user/certificates.md
```diff
@@ -119,6 +119,34 @@ When certificates are generated by foremanctl, you can control lifetimes in days
...
 Defaults: both are 7300 days (about 20 years).
 
+### Key algorithm
+
+Certificates generated by foremanctl use RSA keys by default. A single algorithm applies to the whole certificate set: the CA, and the server and client certificates of every hostname.
+
+| CLI flag | Ansible variable | What it controls |
+| -------- | ---------------- | ---------------- |
+| `--certificate-algorithm` | `certificates_algorithm_type` | Key algorithm: `RSA` (default), `ECC` or `ML-DSA-87` |
+| `--certificate-algorithm-rsa-size` | `certificates_algorithm_size` | Key size, RSA only (default 4096) |
```
ehelms, line 129:
```suggestion
| `--certificate-algorithm-rsa-size` | `certificates_algorithm_rsa_size` | Key size, RSA only (default 4096) |
```

docs/user/certificates.md
```diff
@@ -119,6 +119,34 @@ When certificates are generated by foremanctl, you can control lifetimes in days
...
+| `--certificate-algorithm-ecc-curve` | `certificates_algorithm_curve` | Elliptic curve, ECC only (default `secp384r1`) |
```
ehelms, line 130:
```suggestion
| `--certificate-algorithm-ecc-curve` | `certificates_algorithm_ecc_curve` | Elliptic curve, ECC only (default `secp384r1`) |
```

docs/user/certificates.md
ehelms: I am also questioning if this should be the behaviour. Why change the algorithm and not renew? What would be the use case for that?

docs/user/certificates.md
```diff
@@ -119,6 +119,34 @@ When certificates are generated by foremanctl, you can control lifetimes in days
 
 Defaults: both are 7300 days (about 20 years).
 
+### Key algorithm
+
+Certificates generated by foremanctl use RSA keys by default. A single algorithm applies to the whole certificate set: the CA, and the server and client certificates of every hostname.
+
+| CLI flag | Ansible variable | What it controls |
```
ehelms, line 126: Have you considered trying `forbidden_if`? e.g.

```
  forbidden_if:
    - ['certificates_algorithm_type', 'ECC', ['certificates_algorithm_rsa_size']]
    - ['certificates_algorithm_type', 'RSA', ['certificates_algorithm_ecc_curve']]
    - ['certificates_algorithm_type', 'ML-DSA-87', ['certificates_algorithm_ecc_curve']]
```

etc.

src/roles/certificates/vars/main.yml
michalgritzbach: I've moved the validation to forbidden_if as suggested, so this file ended up serving as a home for the two helpers that are used in multiple other locations, which feels like a correct location for these. Any thoughts?

docs/user/certificates.md
lhellebr: I agree this is counterintuitive and already bit me in the butt while testing. Changing certificate algorithm should IMO at least trigger certificate renewal and it's worth considering whether it should renew CA cert as well.

And actually from the first glance at my testing environment, the certs *are* actually renewed without that parameter, but *not* the CA certs - which is a weird inconsistency worth at least being documented.

lhellebr:
The whole setting of `DEFAULT:PQ` and mounting of `/etc/crypto-policies` is questionable.

1) Is it even necessary? It seems like ML-KEM and ML-DSA are part of RHEL 10's `DEFAULT` already. Previously, PQC was opt-in but now it seems to be opt-out using `DEFAULT:NO-PQ`.
Additionally, `DEFAULT:PQ` doesn't exist so that command just fails:
```
stderr: 'Unknown policy `PQ`: file `PQ.pmod` not found in (., policies/modules, /etc/crypto-policies/policies/modules,
        /usr/share/crypto-policies/policies/modules)'
```

2) Is it a good design, mounting **host's** crypto-policy when it can be an entirely different OS? I know there were discussions on whether copy c-p name, c-p content, etc. but I'm not sure we reached some conclusion. Still, this may be entirely unnecessary due to 1)

src/roles/candlepin/tasks/main.yml
```diff
@@ -87,6 +87,8 @@
       - 'candlepin-tomcat-logging-properties,target=/etc/tomcat/logging.properties,{{ candlepin_secret_mount_opts }}'
       - 'candlepin-logback-xml,target=/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/logback.xml,{{ candlepin_secret_mount_opts }}'
       - 'candlepin-db-ca,target={{ candlepin_database_ssl_ca_path }},{{ candlepin_secret_mount_opts }}'
+    volume:
+      - '/etc/crypto-policies:/etc/crypto-policies:ro'
```
ehelms: Is this required as part of this PR?

lhellebr: I think not. Just as I wrote in my PR-level comment.

src/roles/candlepin/templates/server.xml.j2
```diff
@@ -22,13 +22,27 @@
                protocol="org.apache.coyote.http11.Http11NioProtocol"
                SSLEnabled="true"
                maxThreads="150" scheme="https" secure="true">
+{% if _candlepin_ca_key_algorithm == 'ML-DSA-87' %}
+        {# ML-DSA keys only work over TLS 1.3, and Tomcat's FFM OpenSSL bridge
```
ehelms, line 26: Do you have a reference for this?

src/roles/candlepin/vars/main.yml
```diff
@@ -0,0 +1,41 @@
...
+# (SHA256withRSA), so Candlepin refuses to start when handed a non-RSA CA key
+# unless the scheme is redefined. We override the legacy scheme in
+# candlepin.conf.j2 using the values below.
+#
+# The algorithm is derived from the CA key that was actually deployed (inspected
+# in certs.yml) rather than from the requested certificate algorithm, so custom
+# and migrated CAs are handled the same way as generated ones.
+
+# openssl_privatekey_info reports the CA key type as RSA, ECC or ML-DSA-87.
+_candlepin_ca_key_algorithm: "{{ _candlepin_ca_key_info.type }}"
```
ehelms, line 14: Does this actually work? Referencing a registered variable within the vars file? I would have assumed this would be undefined based on the order of operations of Ansible.

src/roles/candlepin/templates/server.xml.j2
lhellebr: https://www.rfc-editor.org/info/rfc9851/

In 2024, NIST released standards for [[ML-KEM](https://www.rfc-editor.org/info/rfc9851/#ML-KEM)], [[ML-DSA](https://www.rfc-editor.org/info/rfc9851/#ML-DSA)], and [[SLH-DSA](https://www.rfc-editor.org/info/rfc9851/#SLH-DSA)]. Many other countries and organizations are publishing their roadmaps, including the multi-national standards organization ETSI [[ETSI](https://www.rfc-editor.org/info/rfc9851/#ETSI)].

While the industry was waiting for NIST to finish standardization, the IETF has had several efforts underway. A working group was formed in early 2023 to work on the use of Post-Quantum Cryptography (PQC) in IETF protocols [[PQUIPWG](https://www.rfc-editor.org/info/rfc9851/#PQUIPWG)]. Several other working groups, including TLS [[TLSWG](https://www.rfc-editor.org/info/rfc9851/#TLSWG)], are working on specifications to support hybrid algorithms and identifiers, for use during a transition from classic to a post-quantum world.

It is important to note that effort within the TLS Working Group is focused exclusively on TLS 1.3 or later. Put bluntly, PQC for TLS 1.2 will not be specified (see [Section 4](https://www.rfc-editor.org/info/rfc9851/#iana)) at any time; anyone wishing to deploy PQC should expect to use TLS 1.3.

src/roles/certificates/tasks/main.yml
```diff
@@ -5,6 +5,25 @@
       - python3-cryptography
     state: present
 
+# ML-DSA certificates can only be negotiated over TLS when the system crypto
```
ehelms: I do not think we should do this in this role. This going above and beyond the role (pun intended) of this. Not to mention, this does not necessarily change what the containers themselves are doing.

The default policy on EL 10 has PQC by default. EL 9 does not. I think it's fine to support this only on EL 10 right now. And whether we should change the crypto policy through `foremanctl` is a bigger discussion, we should have that separately as a design question.

ehelms: I should also note that this only applies to EL systems. The crypto policy concept doesn't exist for Debian.

src/roles/certificates/vars/main.yml
```diff
@@ -0,0 +1,24 @@
...
+# Private key parameters derived from the selected algorithm, shaped for the
+# arguments community.crypto.openssl_privatekey expects: RSA takes a size, ECC
+# takes a curve, and ML-DSA takes neither. The accepted algorithms and the
+# rule that size only applies to RSA and curve only to ECC are enforced by the
+# choices and forbidden_if constraints in _certificate_algorithm's metadata;
+# here we just hand the module the parameter that matches and omit the rest.
+#
+# ML-DSA-87 needs community.crypto 3.3.0 to generate the keys and
```
ehelms, line 15: We are starting to repeat ourselves.

src/roles/certificates/vars/main.yml
```diff
@@ -0,0 +1,24 @@
+---
```
ehelms, line 1: Have you considered a vars file for each type with the variables set based on each type and then loading them based on the users choice?

So you'd have a statement like:

```
---
- name: "Load type variables"
  include_vars: "{{ certificates_algorithm_type }}.yml"
```

ehelms: Here's an example of that pattern: https://github.com/theforeman/forklift/blob/master/roles/libvirt/tasks/main.yml#L1-L3

src/roles/candlepin/templates/candlepin.conf.j2
```diff
@@ -13,6 +13,16 @@ module.config.adapter_module=org.candlepin.katello.KatelloModule
 
 candlepin.ca_key=/etc/candlepin/certs/candlepin-ca.key
 candlepin.ca_cert=/etc/candlepin/certs/candlepin-ca.crt
+{% if _candlepin_ca_key_algorithm != 'RSA' %}
+
+# The CA key is not RSA, so redefine Candlepin's built-in "legacy" crypto scheme
+# to match; without this Candlepin assumes SHA256withRSA and fails to start.
+candlepin.crypto.scheme.legacy.key_algorithm={{ _candlepin_scheme_key_algorithm }}
```
lhellebr, line 20: This is what Ian suggested after his investigation but should it really be necessary? Shouldn't Candlepin just obey crypto-policies in combination with looking at what key/cert it has actually been provided, and not require settings like these?
@ianballou

src/roles/foreman/tasks/main.yaml
```diff
@@ -117,6 +117,7 @@
     hostname: "{{ ansible_facts['hostname'] }}.local"
     volume:
       - 'foreman-data-run:/var/run/foreman:rw,z,U'
+      - '/etc/crypto-policies:/etc/crypto-policies:ro'
```
ehelms: Before we do this, we need to resolve the design. See:

https://github.com/theforeman/foremanctl/pull/489
https://github.com/theforeman/foremanctl/issues/220

src/roles/candlepin/templates/server.xml.j2
ehelms: Sorry, I should have been clearer. I meant the Tomcat FFM support with `PROFILE=SYSTEM` and inheriting the system crypto policy.

lhellebr:
Two separate bugs affecting Candlepin

**1) ECC: Candlepin never starts (`/candlepin` context fails, everything 404s)**

Candlepin's Guice injector throws:
```
ConfigurationException: Unable to read private key at path: /etc/candlepin/certs/candlepin-ca.key
Caused by: KeyException: No private key declaration found in the source stream
```

Cause: `community.crypto.openssl_privatekey` defaults `type: ECC` to legacy **SEC1** format (`-----BEGIN EC PRIVATE KEY-----`) - only ML-DSA (and similar) get PKCS8 by default. Candlepin's `AbstractPrivateKeyReader` only recognizes `RSA PRIVATE KEY` and plain `PRIVATE KEY` (PKCS8) headers; `EC PRIVATE KEY` matches nothing, so it silently falls through and throws that (misleading) error.

**Attempted fix:** explicitly request `format: pkcs8` for the EC key, in **four places**, since it's not enough to just add it to `_certificates_key_parameters` - `ca.yml`/`issue.yml` destructure only `.type`/`.size`/`.curve` from that dict:
- `certificates/vars/main.yml`: add `format: pkcs8` to `_certificates_key_parameters` for ECC.
- `certificates/tasks/ca.yml`: add `format: "{{ _certificates_key_parameters.format | default(omit) }}"` to "Create CA private key".
- `certificates/tasks/issue.yml`: same, on "Create server private key"
- `certificates/tasks/issue.yml`: same, on "Create client private key"

It's questionable whether we should change to pkcs8 for RSA as well, but for stability sake, I would prefer not to (or do it later).
Also worth noting: `openssl_privatekey`'s idempotency check doesn't treat a `format`-only change as reason to regenerate an existing key, so testing this against an already-provisioned ECC deployment needs the key files deleted first, not just `--certificate-ca-renew`.

I think this is in scope of this PR because it's about format of the generated key.

**2) ML-DSA-87: Candlepin starts but handshakes fail**

From the client side, the TLS handshake against Candlepin's Tomcat connector appears to complete (certificate exchange, `Finished` messages), but the connection then dies. On the server side, Tomcat logs `java.io.EOFException` from `SecureNioChannel.handshakeUnwrap`, suggesting it doesn't consider the handshake complete at all - i.e. the two sides disagree about handshake state. Still investigating.

lhellebr:
Also, one of the CI failures seems to be due to mounting `/etc/crypto-policies/` in FIPS - and it seems we kinda agree it's not needed.

src/roles/candlepin/templates/server.xml.j2
```diff
@@ -22,13 +22,29 @@
...
+           does not pick up the system crypto policy's signature algorithms, so
+           the ML-DSA signature algorithm must be advertised explicitly here. #}
+        <SSLHostConfig certificateVerification="optional"
+                       protocols="TLSv1.3"
+                       caCertificateFile="/etc/candlepin/certs/candlepin-ca.crt">
+            <Certificate certificateFile="/etc/candlepin/certs/tomcat.crt"
+                         certificateKeyFile="/etc/candlepin/certs/tomcat.key" />
+            <OpenSSLConf>
+                <OpenSSLConfCmd name="SignatureAlgorithms"
+                                value="mldsa87:ECDSA+SHA256:ECDSA+SHA384:rsa_pss_rsae_sha256:rsa_pss_rsae_sha384:RSA+SHA256:RSA+SHA384" />
```
lhellebr: This seems very much hardcoded, can we use some resource to generate it?

lhellebr:
# Tomcat + OpenSSL-FFM stalls the TLS handshake with ML-DSA-87

Tested on RHEL 10.2 with `tomcat-10.1.49-3.el10_2`, `java-25-openjdk-headless-25.0.4.1.1`, `openssl-3.5.8`.

The only thing that changes between the broken run and the working run is which certificate file Tomcat is pointed at.

---

## 1. Install Tomcat, a Java 25 runtime, and openssl-devel

```bash
dnf -y install tomcat java-25-openjdk-headless openssl-devel
```

## 2. Generate two certificate sets: ML-DSA-87 and ML-DSA-65

Each set is a self-signed CA plus a server certificate signed by it. A chain of two certificates is required, a single self-signed certificate is too small to trigger the bug.

```bash
mkdir -p /etc/pki/tomcat-pq && cd /etc/pki/tomcat-pq

for ALG in ML-DSA-87 ML-DSA-65; do
  # self-signed CA
  openssl req -x509 -newkey $ALG -noenc -keyout ca-$ALG.key -out ca-$ALG.crt \
          -subj "/CN=PQ Test CA $ALG" -days 365

  # server key + CSR, signed by that CA
  openssl req -new -newkey $ALG -noenc -keyout srv-$ALG.key -out srv-$ALG.csr \
          -subj "/CN=localhost"
  openssl x509 -req -in srv-$ALG.csr -CA ca-$ALG.crt -CAkey ca-$ALG.key \
          -out srv-$ALG.crt -days 365
done

chown root:tomcat /etc/pki/tomcat-pq/*
chmod 640 /etc/pki/tomcat-pq/*
```

Check the sizes:

```bash
for A in ML-DSA-87 ML-DSA-65; do
  echo "$A: leaf=$(openssl x509 -in srv-$A.crt -outform DER | wc -c) \
ca=$(openssl x509 -in ca-$A.crt -outform DER | wc -c)"
done
```

```
ML-DSA-87: leaf=7462 ca=7490      -> chain 14952 B
ML-DSA-65: leaf=5504 ca=5532      -> chain 11036 B
```

---

## 3. Point Tomcat at Java 25 and enable FFM

Edit `/etc/tomcat/tomcat.conf`:

```
JAVA_HOME="/usr/lib/jvm/jre-25"
JAVA_OPTS="--enable-native-access=ALL-UNNAMED"
```

---

## 4. Give Tomcat something to serve

```bash
mkdir -p /var/lib/tomcat/webapps/ROOT
echo 'IT WORKS' > /var/lib/tomcat/webapps/ROOT/index.html
```

---

## 5. Write `/etc/tomcat/server.xml`

```xml
<Server port="8005" shutdown="SHUTDOWN">

  <!-- Switches Tomcat's TLS from Java's own implementation over to the
       system OpenSSL, through FFM.
       As per https://github.com/theforeman/foremanctl/pull/472 -->
  <Listener className="org.apache.catalina.core.OpenSSLLifecycleListener" />

  <Service name="Catalina">
    <Connector port="9443" protocol="org.apache.coyote.http11.Http11NioProtocol"
               SSLEnabled="true" scheme="https" secure="true">
      <SSLHostConfig protocols="TLSv1.3">
        <Certificate certificateFile="/etc/pki/tomcat-pq/srv-ML-DSA-87.crt"
                     certificateKeyFile="/etc/pki/tomcat-pq/srv-ML-DSA-87.key"
                     certificateChainFile="/etc/pki/tomcat-pq/ca-ML-DSA-87.crt" />
      </SSLHostConfig>
    </Connector>
    <Engine name="Catalina" defaultHost="localhost">
      <Host name="localhost" appBase="webapps" unpackWARs="true" autoDeploy="false"/>
    </Engine>
  </Service>
</Server>
```

`certificateChainFile` matters: without it Tomcat sends only the leaf, the
handshake stays small, and the bug does not appear.

---

## 6. Start it and confirm OpenSSL-FFM is really active

```bash
systemctl restart tomcat
sleep 10
journalctl -u tomcat --since "-30s" --no-pager | grep -oE 'Starting ProtocolHandler \[[^]]*\]'
```

Must print:

```
Starting ProtocolHandler ["https-opensslffm-nio-9443"]
```

---

## 7. ML-DSA-87

```bash
# curl -k https://localhost:9443/
curl: (56) OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 0
```

Times out

---

## 8. ML-DSA-65

Edit the three paths in `server.xml` from `ML-DSA-87` to `ML-DSA-65`:

```bash
sed -i 's/ML-DSA-87/ML-DSA-65/g' /etc/tomcat/server.xml
systemctl restart tomcat
sleep 10
curl -k https://localhost:9443/
```

**Result:**

```
IT WORKS
```

---

## 9. Why: the handshake outgrew a 17408-byte buffer

tl;dr: ML-DSA 87 cert chain is so big that Tomcat chokes on it.

Measure what the server actually sends in each case:

```bash
timeout 20 openssl s_client -connect localhost:9443 -tls1_3 </dev/null 2>&1 \
  | grep -oE 'handshake has read [0-9]+ bytes'
```


ML-DSA-87, leaf only (drop `certificateChainFile`) => 13477 B => HTTP 200
ML-DSA-65, leaf + CA => 15738 B => HTTP 200 
ML-DSA-87, leaf + CA => 20900 B => hangs

The cutoff is **17408 bytes** (`17 * 1024`), OpenSSL's default size for the memory buffer Tomcat uses to collect encrypted bytes from OpenSSL before writing them to the socket. Tomcat asks for that default by passing `0` in `OpenSSLEngine.java`:

```java
BIO_new_bio_pair(internalBIOPointer, 0, networkBIOPointer, 0);
```

OpenSSL documents that default as "sufficient for a maximum size TLS record": one *record*, not one whole *burst*. When the burst does not fit, OpenSSL fills the buffer and holds the rest back. Tomcat drains the buffer, then decides whether more remains to send using only "is the buffer empty?":

```java
if (sendHandshakeError || BIO_ctrl_pending(state.networkBIO) != 0) {
    return SSLEngineResult.HandshakeStatus.NEED_WRAP;   // more to send
}
...
return SSLEngineResult.HandshakeStatus.NEED_UNWRAP;      // client's turn
```

The buffer *is* empty - Tomcat just emptied it - but that does not mean OpenSSL has finished. Tomcat concludes it is the client's turn and waits; the client is still waiting for the tail of the server's burst. Deadlock, until the client times out and closes. `EOF during handshake` is the server noticing the hangup.

src/roles/candlepin/tasks/main.yml
```diff
@@ -17,6 +17,28 @@
   notify:
     - Restart candlepin
 
+# Tomcat's FFM OpenSSL bridge does not pick up the system crypto policy's
+# signature algorithms on its own, so the ML-DSA connector in server.xml must
+# advertise them explicitly. Rather than hardcode the list, read it back from
+# the crypto policy's OpenSSL backend, which is the authoritative source and
+# already lists mldsa87 in the EL10 default policy. This keeps the connector in
+# sync with the host policy (including FIPS or custom policies) instead of a
+# frozen list. Only the ML-DSA CA needs it; RSA and ECC use the cipher-based
```
lhellebr, line 26: Me thinking out loud: for now, we only support ML-DSA on RHEL 10, so we don't have to think about systems without crypto-policies, because foremanctl will error out as soon as user tells it to generate ML_DSA on those systems. Right?

lhellebr:
My latest finding IMO means that we need to decide to either
1) fix Tomcat (I'll probably file a bugreport anyway), or
2) limit ML-DSA size to 65, at least for now

@ehelms I see you suggested allowing only ML-DSA 87 early in this PR. I suppose that was an ad-hoc decision not backed by anything, right? I didn't decide to require 87 either. Is there any reason not to enforce ML-DSA 65 other than "we want the strongest greatest algorithm"? Does someone actually require 87? Or is it better for compatibility reasons, do we think most people do/will use 87?

ehelms:
My reasoning behind ML-DSA-87 was the default is that seems to be the most widely used and mentioned across things I have read. For example, that is what Red Hat has started signing packages with. As I understand it, certain governments policies (e.g. CNSA 2.0) are expecting ML-DSA-87 whereas the other variants might be the go to for other use cases. In the end, I figured default to the strongest (albeit also the largest) and let users downgrade per their use case.

lhellebr:
Another argument against just limiting to ML-DSA 65 is that it's still possible to create a large enough CA chain from ML-DSA 65 certs that it will still fail. It worked for my chain but someone else may have a longer one. This just seems like a bug in Tomcat that should be fixed.

But even if we fix it, will it be fixed in the version of Tomcat that we use in foreman/satellite? I doubt that.

lhellebr:
Attached
[0001-Fix-TLS-handshake-stall-when-the-server-flight-excee.patch](https://github.com/user-attachments/files/32340953/0001-Fix-TLS-handshake-stall-when-the-server-flight-excee.patch)
 is a "fix" for Tomcat that makes it work but is probably not really correct because https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html says
"""
The SSLEngine produces/consumes **complete** SSL/TLS packets only, and does not store application data internally between calls to wrap()/unwrap(). Thus input and output ByteBuffers must be sized appropriately to hold the maximum record that can be produced. Calls to [SSLSession.getPacketBufferSize()](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLSession.html#getPacketBufferSize--) and [SSLSession.getApplicationBufferSize()](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLSession.html#getApplicationBufferSize--) should be used to determine the appropriate buffer sizes. The size of the outbound application data buffer generally does not matter. If buffer conditions do not allow for the proper consumption/production of data, the application must determine (via [SSLEngineResult](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngineResult.html)) and correct the problem, and then try the call again.
"""

lhellebr:
Bugreport: https://bz.apache.org/bugzilla/show_bug.cgi?id=70236

PR: https://github.com/apache/tomcat/pull/1070

lhellebr:
The Tomcat PR has been merged, cherrypicked and can be expected in **Tomcat 10.1.61**.

docs/user/certificates.md
```diff
@@ -145,6 +145,9 @@ Server and client (leaf) certificates with non-RSA keys only carry the `digitalS
 > [!NOTE]
 > The `ML-DSA-87` algorithm needs `community.crypto` 3.3.0 to generate the keys and `python-cryptography` 49.0.0 to sign certificates and CSRs with them. CentOS Stream 10 ships a suitable `python-cryptography`; CentOS Stream 9 ships 36.0.1, on which selecting `ML-DSA-87` fails at key generation.
 
+> [!WARNING]
+> Candlepin does not yet come up with an `ML-DSA-87` certificate set. Its Tomcat TLS connector uses the FFM OpenSSL bridge, whose BIO buffer (17408 bytes) is smaller than the ML-DSA-87 certificate chain (~20 KB), so the TLS handshake stalls and Candlepin never becomes healthy. This is upstream Apache Tomcat bug 70236 (fixed in Tomcat PR #1070), not yet backported to the Tomcat shipped on RHEL. Until that fix lands, deploy Katello with `RSA` or `ECC`. Non-Candlepin services are unaffected.
```
lhellebr: Do we want to even include this in the PR, given the fix should land sometime next week?

michalgritzbach: Thanks, removed, I should go through the other comments and clean them up, they are really abundant.

Issue #467, opened 2026-04-22
Blocked hosts error in logs


jeremylenz:
```
13:34:20 rails.1   | 2026-04-22T13:34:20 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: host.containers.internal, host.containers.internal
```

we may need to add an entry to allowed hosts.

jeremylenz:
I saw that log after running
```
systemctl start iop-service-vuln-vmaas-sync.service
```
but not sure if it's related.

ekohl:
In https://github.com/theforeman/foremanctl/pull/471 (recently merged) the middleware is correctly configured. However, I don't think `host.containers.internal` is actually configured. Is this still a problem?

jeremylenz:
I think this fixed itself a while back and I haven't had any issues with it. Claude also agrees:

  ▎ host.containers.internal is used in two contexts: (1) IOP database connections (just TCP, not HTTP to Foreman), and (2) the IOP gateway's relay.conf.j2, which does proxy_pass to 
  ▎ host.containers.internal but sets proxy_set_header Host localhost — so Foreman sees localhost as the Host header, which Rails allows by default. So I don't think host.containers.internal
  ▎ needs to be in the :hosts: allowlist and this shouldn't be an issue.

Closing for now; we can always reopen if needed.

jeremylenz:
Reopening — the reasoning that closed this was incorrect. `relay.conf.j2` does set `proxy_set_header Host localhost;`, but Foreman doesn't allow `localhost` by default in production: `config.hosts` is populated from `SETTINGS[:hosts]`, which only ever contains the Foreman FQDN and explicit aliases (`src/roles/foreman/templates/settings.yaml.j2`). So a request proxied through the gateway with `Host: localhost` gets rejected by `ActionDispatch::HostAuthorization` with a 403, before it even reaches the app.

Confirmed this in the wild: it broke VMAAS reposcan's Katello sync (`GET /katello/api/v2/organizations` via the gateway came back 403), which meant CVE/repo data never synced and the vulnerability dashboard silently showed 0 affected hosts.

Fix is up in #807 — sends the real Foreman FQDN as the `Host` header instead, matching the existing `proxy_ssl_name`.

PR #807, opened 2026-09-01
Send the real Foreman FQDN as the Host header in the IOP gateway relay

jeremylenz:
## Summary
- `relay.conf.j2` hardcoded `Host: localhost` when the IOP gateway proxies requests back to Foreman. Foreman's `config.hosts` only trusts its own FQDN (`src/roles/foreman/templates/settings.yaml.j2`), so Rails' `ActionDispatch::HostAuthorization` middleware rejected these with a 403 before the request ever reached the app.
- This broke VMAAS reposcan's Katello repository/CVE sync (`GET /katello/api/v2/organizations` via the gateway), causing `foreman_rh_cloud`'s trusted-smart-proxy access to silently fail with `403` and repository/CVE data to never sync — which in turn left the vulnerability dashboard showing 0 affected hosts even when real errata existed.
- Fix: send the real Foreman FQDN as the `Host` header instead, matching the existing `proxy_ssl_name` on the line above.

Related to #467, which closed this exact "Blocked hosts" error based on the (incorrect) assumption that Rails allows `localhost` by default. It doesn't in production — `config.hosts` is explicitly populated from `SETTINGS[:hosts]`, which never includes `localhost` unless it's the FQDN or an explicit alias.

## Test plan
- [x] `ansible-lint roles/iop_gateway` passes clean
- [x] Confirmed no existing test asserts the literal `Host` header value (`tests/iop/test_gateway.py`, `tests/iop/test_integration.py`)
- [x] Verified live on an affected deployment: after applying the equivalent change (via a `foreman-settings-yaml` secret workaround adding `localhost` to `:hosts:`), `GET /katello/api/v2/organizations` through the gateway went from `403` to `200`, VMAAS successfully synced Katello repos/CVEs, and the vulnerability dashboard began correctly showing affected hosts
- [ ] Full deploy + IOP smoker test run in CI

### DIFF
--- a/src/roles/iop_gateway/templates/relay.conf.j2
+++ b/src/roles/iop_gateway/templates/relay.conf.j2
@@ -3,5 +3,5 @@ proxy_ssl_name "{{ foreman_servername | default(ansible_fqdn) }}";
 
 # URI to forman
 # Example of host.containers.internal is the container network gateway.
-proxy_set_header Host localhost;
+proxy_set_header Host "{{ foreman_servername | default(ansible_fqdn) }}";
 proxy_pass "https://host.containers.internal";
--- a/src/roles/iop_gateway/templates/relay.conf.j2
+++ b/src/roles/iop_gateway/templates/relay.conf.j2
@@ -1,7 +1,7 @@
 # (REQUIRED) CName of the Foreman instance (must match Foreman's TLS certificate)
-proxy_ssl_name "{{ foreman_servername | default(ansible_fqdn) }}";
+proxy_ssl_name "{{ iop_gateway_foreman_name }}";
 
 # URI to forman
 # Example of host.containers.internal is the container network gateway.
-proxy_set_header Host "{{ foreman_servername | default(ansible_fqdn) }}";
+proxy_set_header Host "{{ iop_gateway_foreman_name }}";
 proxy_pass "https://host.containers.internal";
--- a/tests/feature/iop/test_gateway.py
+++ b/tests/feature/iop/test_gateway.py
@@ -25,3 +25,17 @@ def test_gateway_secrets(server):
 
     for secret_name in secrets:
         assert secret_name in result.stdout
+
+
+def test_gateway_relay_reaches_foreman(server, iop_image):
+    # Regression test for https://github.com/theforeman/foremanctl/issues/467:
+    # the relay used to send "Host: localhost" to Foreman, which Rails'
+    # ActionDispatch::HostAuthorization rejected with a 403 before the
+    # request reached the app. The Katello organizations endpoint is a
+    # convenient real-world path that is relayed through the gateway and
+    # only succeeds once the Host header matches Foreman's allowed hosts.
+    result = server.run(
+        f"podman run --network=iop-core-network --rm {iop_image('iop-inventory')} "
+        "curl -s -o /dev/null -w '%{http_code}' http://iop-core-gateway:9090/katello/api/v2/organizations"
+    )
+    assert "200" in result.stdout
--- a/tests/feature/iop/test_gateway.py
+++ b/tests/feature/iop/test_gateway.py
@@ -36,6 +36,7 @@ def test_gateway_relay_reaches_foreman(server, iop_image):
     # only succeeds once the Host header matches Foreman's allowed hosts.
     result = server.run(
         f"podman run --network=iop-core-network --rm {iop_image('iop-inventory')} "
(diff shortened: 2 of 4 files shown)


jeremylenz:
um @coderabbitai it's a 1-line change :smile:

ekohl:
Is there a way we could could cover this with integration tests?

src/roles/iop_gateway/templates/relay.conf.j2
```diff
@@ -3,5 +3,5 @@ proxy_ssl_name "{{ foreman_servername | default(ansible_fqdn) }}";
 
 # URI to forman
 # Example of host.containers.internal is the container network gateway.
-proxy_set_header Host localhost;
+proxy_set_header Host "{{ foreman_servername | default(ansible_fqdn) }}";
```
evgeni: Where is foreman_servername defined?

evgeni: We have https://github.com/theforeman/foremanctl/blob/e969bce165e4e96961450c7f1896da25ac98db21/src/vars/base.yaml#L25 so it could be

```suggestion
proxy_set_header Host "{{ foreman_name | default(ansible_facts['fqdn']) }}";
```

But `foreman_name` is not scoped to the role, so I guess the most correct way would be to
- define `iop_gateway_foreman_name: "{{ ansible_facts['fqdn'] }}"` in `src/roles/iop_gateway/defaults/main.yaml`
- override `iop_gateway_foreman_name: "{{ foreman_name }}"` in `src/vars/base.yaml`

And then
```suggestion
proxy_set_header Host "{{ iop_gateway_foreman_name }}";
```

(and also in `proxy_ssl_name` above)

jeremylenz:
Investigated the smoker failure (\`test_menu_item[Subscriptions]\`) and Evgeni's role-scoping suggestion — both addressed:

**Role scoping**: pushed a follow-up commit replacing the undefined \`foreman_servername\` with a properly role-scoped \`iop_gateway_foreman_name\`, overridden from \`foreman_name\` in \`src/vars/base.yaml\`, matching the existing \`iop_core_foreman_url\` pattern.

**Integration test**: added \`test_gateway_relay_reaches_foreman\` in \`tests/feature/iop/test_gateway.py\`, which proxies a request through the gateway to Foreman's Katello organizations endpoint and asserts success — this would have caught the original "Blocked hosts" regression.

**The smoker failure is unrelated to this PR.** \`test_menu_item[Subscriptions]\` fails because loading that page triggers \`GET /api/v2/users/:id/table_preferences/subscriptions\`, which 404s and gets logged as a browser console error. This is the exact same known, already-partially-handled pattern as \`table_preferences/hosts\` (see the \`EXCLUDE_ERRORS\` allowlist in \`theforeman/smoker\`'s \`tests/test_pages.py\`) — Katello recently migrated the Subscriptions page to the same Foreman core PF4 table-preferences hook, and a 404 there is expected/harmless the first time any table is loaded for a given user, it's just not allowlisted yet for this table name. Filed theforeman/smoker#51 with a suggested fix (generalize the exclusion to match any \`table_preferences/\` path instead of allowlisting table names one at a time). Confirmed this isn't caused by our diff: master's last green run at the same base commit (\`d1e550b\`) passed both \`iop: enabled\` jobs cleanly.

tests/feature/iop/test_gateway.py
```diff
@@ -25,3 +25,17 @@ def test_gateway_secrets(server):
...
+def test_gateway_relay_reaches_foreman(server, iop_image):
+    # Regression test for https://github.com/theforeman/foremanctl/issues/467:
+    # the relay used to send "Host: localhost" to Foreman, which Rails'
+    # ActionDispatch::HostAuthorization rejected with a 403 before the
+    # request reached the app. The Katello organizations endpoint is a
+    # convenient real-world path that is relayed through the gateway and
+    # only succeeds once the Host header matches Foreman's allowed hosts.
+    result = server.run(
+        f"podman run --network=iop-core-network --rm {iop_image('iop-inventory')} "
+        "curl -s -o /dev/null -w '%{http_code}' http://iop-core-gateway:9090/katello/api/v2/organizations"
```
evgeni:
```suggestion
        "curl --silent --output /dev/null --write-out '%{http_code}' http://iop-core-gateway:9090/katello/api/v2/organizations"
```

I am picky about short opts ;)

evgeni (approved):
ACK on the change

jeremylenz:
Updated (that test failure seems like a fluke..)

evgeni (approved):

PR #816, opened 2026-09-07
Review of deployment/development guides

lzap:
Since I am new to foremanctl I took the opportunity to review the docs. In fact, some things were a bit misleading, or incomplete. I cannot rule out that I am too dense :)

FIrst off, I was constantly running `vagrant up` since I was using `forklift` the week before. This halts with an ugly error, so I added a hard-error. Not sure if this does not block some other workflows tho - can drop.

Then I was working on proxy deployment and the current documentation somewhat expects the `foremanctl` to be executed from within the VMs, which currently does not work - I was trying from the rsynced `/vagrant` folder and running into issues. Therefore, I am changing the forkflow to be executed solely on the conrol node.

Finally, I thought that vagrant hostmanager will do the job as in Forklift, but that was not the case. I had misconfigured domain in my default libvirt network, so I added a section which covers that.


docs/developer/deployment.md
```diff
@@ -43,18 +43,18 @@ Before running the proxy deployment, an auth bundle must be generated on the For
 > [!NOTE]
 > The bundle includes the proxy's certificates and OAuth credentials needed for the proxy to communicate with the Foreman server.
 
-2. Copy the bundle to the **proxy VM**:
+2. Copy the bundle to the **control node**:
 
-   ```console
-   # scp /var/lib/foremanctl/certs/bundles/proxy.example.com.tar.gz root@proxy.example.com:/root/proxy.example.com.tar.gz
+   ```bash
+   vagrant ssh quadlet -- sudo cat /var/lib/foremanctl/certs/bundles/proxy.example.com.tar.gz > proxy.example.com.tar.gz
```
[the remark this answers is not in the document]
lzap: I mean, this is a dev docs... Shall we really make it ugly @evgeni ? :)

I can do perhaps this:

```
vagrant ssh quadlet -- sudo cat /var/lib/foremanctl/certs/bundles/proxy.example.com.tar.gz | install -m 600 /dev/stdin proxy.example.com.tar.gz
```

docs/developer/deployment.md
```diff
@@ -43,18 +43,18 @@ Before running the proxy deployment, an auth bundle must be generated on the For
...
    ```
 
-3. On the **proxy VM**, run the deployment:
+3. On the **control node** (where foremanctl is installed), run the deployment remotely via SSH:
```
ehelms, line 52: I think this depends on how a user is doing their development and from where they are set up. There are some of us that have a checkout on our laptop and interact with VMs remotely. There are others who clone it onto the VMS and run it locally within the VM as a user would. I realize that makes the docs a bit harder to account for both. Technically I would agree `control node` accounts for both as the control node could be separate or be localhost itself. 

TLDR; Might help to define control node near the top of this document for clarity.

docs/developer/development-environment.md
```diff
@@ -28,13 +28,16 @@ The development environment provides:
...
 2. **Access the environment:**
-   - SSH into the VM: `vagrant ssh`
+   - To run commands on the VM, SSH into it: `vagrant ssh quadlet`
    - Navigate to Foreman directory: `cd /home/vagrant/foreman`
    - Start Rails server: `bundle exec foreman start`
 
 3. **Access URLs:**
-   - Foreman UI: `http://$(hostname -f):3000` (development server)
-   - Production-style UI: `https://$(hostname -f)` (via Apache proxy)
+   - Foreman UI: `http://localhost:3000` (development server)
```
ehelms: This doesn't work if it's running on a VM.

docs/developer/deployment.md
lzap: I had the same thought. It is already there. I can make it a link if you wish.

```
docs/developer/deployment.md
6:As `foremanctl` is Ansible-based, this means that the ["control node"](https://docs.ansible.com/projects/ansible/latest/network/getting_started/basic_concepts.html#control-node) and the ["managed node"](https://docs.ansible.com/projects/ansible/latest/network/getting_started/basic_concepts.html#managed-nodes) are the same system (`localhost`).
```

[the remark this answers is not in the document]
lzap: Added.

ehelms: Oh fair! I didn't read up far enough.

docs/developer/deployment.md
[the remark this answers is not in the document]
evgeni: Don't listen to the Rabbit

docs/developer/deployment.md
```diff
@@ -30,7 +30,7 @@ Deploys a Foreman server. This is the primary deployment type and the default en
 
 Deploys a Foreman Proxy node that connects to a Foreman server.
 
-Before running the proxy deployment, an auth bundle must be generated on the Foreman server and copied to the proxy VM:
+Before running the proxy deployment, an auth bundle must be generated on the Foreman server and copied to the control node:
 
 1. On the **Foreman server**, generate an auth bundle for the proxy hostname:
```
evgeni, line 35: Technically, you still call `foremanctl` on the control node, but the bundle ends up on the Foreman server.

src/playbooks/deploy-proxy/metadata.obsah.yaml
```diff
@@ -7,6 +7,9 @@ variables:
     help: Base flavor to use in this deployment.
     choices:
       - foreman-proxy-content
+  target_host:
```
evgeni, line 10: please don't do this. our default (prod) deployment is not using remote machines and there are too many corners users will run into when trying to do this in prod. 

See https://github.com/theforeman/foremanctl/issues/630 and https://github.com/theforeman/foremanctl/pull/633 for details

docs/developer/development-environment.md
```diff
@@ -67,6 +70,26 @@ katello-production:
...
+
+```bash
+virsh net-edit default
+```
+
+And ensuring that the domain name tag matches your `VAGRANT_DOMAIN` environment variable (which defaults to `example.com`):
+
+```xml
+<network>
+  <domain name='example.com'/>
```
evgeni, line 87:
```suggestion
  <domain name='example.com' localOnly='yes' register='yes'/>
```

to follow https://community.theforeman.org/t/setting-up-libvirt-with-dns-resolution-for-vagrant/46567

docs/developer/development-environment.md
```diff
@@ -67,6 +70,26 @@ katello-production:
...
+</network>
+```
+
+Once VMs are running, the guest hostnames should resolve automatically on modern Linux distributions (tested on Fedora).
```
evgeni, line 91: the kicker here is `systemd-resolved`, which is what `register=yes` hooks into.

if you don't have resolved, you won't get it automatically.

src/playbooks/deploy-proxy/metadata.obsah.yaml
lzap: Dropped this commit.

lzap:
Rebased.

evgeni (approved):

docs/developer/development-environment.md
```diff
@@ -67,6 +70,26 @@ katello-production:
...
+
+And ensuring that the domain name tag matches your `VAGRANT_DOMAIN` environment variable (which defaults to `example.com`):
+
+```xml
+<network>
+  <domain name='example.com' localOnly='yes' register='yes'/>
+</network>
+```
+
+Once VMs are running, the guest hostnames should resolve automatically on modern Linux distributions with `systemd-resolved`.
```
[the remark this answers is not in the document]
evgeni: Oh, this (the coderabbit comment) is actually correct. You need a `virsh net-destroy default && virsh net-start default`

ehelms (approved):

ehelms:
Tests are unrelated to documentation changes.

PR #819, opened 2026-09-08
Don't require oauth credential from parameters in the restore task

sjha4:
Restore on proxy was requiring oauth key and secret in parameters.yaml. The commit https://github.com/theforeman/foremanctl/pull/698/changes#diff-b005bb841010b9f9006d7ff1e0fa0d8988b8b6a066091afc982a40abd7e3102aR135 removed the need for it and now oauth creds can be read directly from state directory/oauth.

* Stop reading oauth from parameters.yaml since those don't exist anymore since change above. Let the creds come from default oauth directory that restore extracts state into.

Steps to reproduce:

* Set up a proxy
* Run foremanctl backup /tmp/backup --target-host proxy
* Run foremanctl restore /tmp/backup/backup_dir --target-host proxy --force

#### Checklist

### DIFF
--- a/src/playbooks/restore/restore.yaml
+++ b/src/playbooks/restore/restore.yaml
@@ -37,8 +37,6 @@
     - name: Set proxy deploy parameters
       ansible.builtin.set_fact:
         foreman_name: "{{ restore_params.foreman_name }}"
-        foreman_proxy_oauth_consumer_key: "{{ restore_params.foreman_proxy_oauth_consumer_key }}"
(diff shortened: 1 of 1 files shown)


ekohl (commented):
Could this be covered by automated tests?

sjha4:
Reverting tests here since running as part of forge tests slows down all the tests as restore involves running a deploy. I'll play with some other options separately like adding a github action that tests a backup-restore workflow on the lines of upgrade test GH action. We do have coverage for these in robottelo for this which caught this.

ehelms (approved):

PR #822, opened 2026-09-10
Bind mount ansible.cfg into foreman-proxy container

adamruzicka:

Tweaking ansible.cfg on the filesystem was possible in traditional deployments, but not in containers. This change closes that gap.

* Declare a file to be bind-mounted from the host to the container

Steps to reproduce:

* Deploy foreman with ansible feature
* Add some config options to `/var/lib/foreman-proxy/ansible/ansible.cfg`
* Run a ansible job
* Check that the config options were honored 

#### Checklist


adamruzicka:
>  Mount the parent directory, or restart/remount the container when ansible.cfg changes.

As an alternative approach, we could mount an entire directory (like `/var/lib/foreman-proxy/ansible/config`) to a custom location (like `/etc/foreman-proxy/ansible`) and point ansible to its contents using environment variables. This however feels like quite an overkill for a single file. I also don't like the idea of introducing yet another custom location.

ehelms:
> > Mount the parent directory, or restart/remount the container when ansible.cfg changes.
> 
> As an alternative approach, we could mount an entire directory (like `/var/lib/foreman-proxy/ansible/config`) to a custom location (like `/etc/foreman-proxy/ansible`) and point ansible to its contents using environment variables. This however feels like quite an overkill for a single file. I also don't like the idea of introducing yet another custom location.

What about mounting `/var/lib/foreman-proxy/ansible` to `/etc/ansible` ? Does that override things we need to be present by default?

We should probably add some docs about how the behaviour of the ansible feature works in foremanctl, such as these config options, collections and roles.

adamruzicka:
> Does that override things we need to be present by default?

Not necessarily, the things we _need_ there are placed under `/etc/foreman-proxy` anyway, so we should be mostly safe.

.ansible/.lock
ehelms, line 1: What is this file?

adamruzicka: What the..? Dropped

docs/developer/how-to-add-a-feature.md
```diff
@@ -128,6 +128,73 @@ For example, REX needs SSH keys generated and mounted into the container, see [`
 
 Configuration Tasks must notify `Restart Foreman Proxy` and `Refresh Foreman Proxy` handlers when making changes to configs or secrets.
 
+### Ansible feature: volume and secret bindings
```
ehelms: I see partially why you added this here but it feels way more about how the Ansible feature itself works which then feels strange to live in this file. This feels like it combines developer relevant documentation and user. I could see a `developer/feature.md` or an `architecture/feature.md` that has sections for features in them that provide this description. 

We could also argue a good chunk of this can be figured out by reading the code and the important part to document is the "how do I as a user provide roles and collections, and how do I customize Ansible's execution"

adamruzicka: > I see partially why you added this here

Mostly for the lack of a better place, yes

> the important part to document is the "how do I as a user provide roles and collections, and how do I customize Ansible's execution"

We already have that in foreman-documentation (or will have once https://github.com/theforeman/foreman-documentation/pull/5302 goes in). Would that be enough?

ehelms: Ehh...I still think a concise "architecture" view of the feature is handy to have here. As a developer, I hate having to go read foreman-documentation as a repository if I am being honest. And it takes a very concise user centric view. 

I quite like what you have already here, I would just make it slightly more concise and move where it lives.

adamruzicka: very well, how about this https://github.com/theforeman/foremanctl/pull/822/commits/e16b6d4c0cc4b0e6cc472c00b27a2ace12890ee5 ?

ehelms (approved):

PR #823, opened 2026-09-10
fix: preserve client IPs for proxied requests

Alleny244:

When requests reach Foreman through a Capsule or reverse proxy, Foreman needs to trust the proxy and receive the original client IP from `X-Forwarded-For`. Without this, Foreman may see the Capsule IP instead of the real client IP.

foremanctl currently does not provide a supported way to configure Foreman’s `:trusted_proxies`, and the content Capsule Apache configuration does not enable `ProxyAddHeaders` for proxied requests.

* Add `--foreman-trusted-proxies` as a repeatable server deployment parameter.
* Configure Foreman’s `:trusted_proxies` with the default localhost ranges and user-provided proxy addresses.
* Enable `ProxyAddHeaders On` in the content Capsule Apache SSL vhost so `X-Forwarded-For` is forwarded to Foreman.
* Add tests and documentation for the new configuration.

Deploy the Foreman server with a trusted Capsule IP:

```bash
./foremanctl deploy --tuning development \
  --foreman-trusted-proxies 192.168.121.120
```

Verify the Foreman settings:

```bash
podman secret inspect foreman-settings-yaml --showsecret | grep -A8 trusted_proxies
```

Verify the Capsule configuration:

```bash
grep ProxyAddHeaders /etc/httpd/conf.d/pulpcore-ssl.conf
```

Run the tests:

```bash
./forge test --pytest-args="tests/feature/foreman/trusted_proxies_test.py -vv"
```

```bash
./forge test --pytest-args="--server-hostname=proxy tests/flavor/foreman-proxy-content/httpd_test.py -vv"
```

#### Checklist


docs/user/parameters.md
```diff
@@ -58,6 +58,7 @@ There are multiple use cases from the users perspective that dictate what parame
 | `--content-export-path` | Extra file path that Pulp can use for content exports | |
 | `--external-authentication={ipa,ipa_with_api}` | Enable configuration for external authentication via IPA for web UI (or webUI and API for `ipa_with_api`), expects the target machine to [be enrolled into FreeIPA/IDM](https://docs.theforeman.org/3.16/Configuring_User_Authentication/index-katello.html#enrolling-foreman-server-in-freeipa-domain) | `--foreman-ipa-authentication`<br/> `--foreman-ipa-authentication-api` |
 | `--external-authentication-pam-service` | PAM service used for host-based access control in IPA | `--foreman-pam-service` |
+| `--foreman-trusted-proxies` | IP address or CIDR that Foreman trusts for `X-Forwarded-For` (for example a Capsule or load balancer in front of Foreman). May be specified multiple times; values are stored in the installation parameters and merged with `append_unique`. Localhost ranges (`127.0.0.0/8` and `::1`) are always included in Foreman settings. Example: `--foreman-trusted-proxies 10.10.10.20` when a Capsule at that address forwards client requests to Foreman. | `--foreman-trusted-proxies` |
```
evgeni:
```suggestion
| `--foreman-trusted-proxies` | IP address or CIDR that Foreman trusts for `X-Forwarded-For` (for example a Foreman Proxy or load balancer in front of Foreman). May be specified multiple times. Localhost ranges (`127.0.0.0/8` and `::1`) are always included in Foreman settings. Example: `--foreman-trusted-proxies 10.10.10.20` when a Foreman Proxy at that address forwards client requests to Foreman. | `--foreman-trusted-proxies` |
```

- Capsule is a downstream term
- We should not expose details like "append_unique" to users, IMHO

src/playbooks/_foreman/metadata.obsah.yaml
```diff
@@ -14,6 +14,12 @@ variables:
     parameter: --initial-location
   foreman_puma_workers:
     help: Number of workers for Puma.
+  foreman_trusted_proxies:
+    help: >-
+      IP address or CIDR Foreman should trust for X-Forwarded-For (e.g. Capsule).
```
evgeni:
```suggestion
      IP address or CIDR Foreman should trust for X-Forwarded-For (e.g. Foreman Proxy).
```

docs/user/parameters.md
Alleny244: okay got it. have updated the doc

src/playbooks/_foreman/metadata.obsah.yaml
```diff
@@ -14,6 +14,12 @@ variables:
     parameter: --initial-location
   foreman_puma_workers:
     help: Number of workers for Puma.
+  foreman_trusted_proxies:
```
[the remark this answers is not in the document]
evgeni: We could extend https://github.com/theforeman/obsah/blob/master/obsah/data_types.py to have a IP type and use it here.

src/roles/httpd/templates/foreman-ssl-vhost.conf.j2
```diff
@@ -130,6 +130,7 @@
   ProxyPass / {{ httpd_foreman_backend }}/ retry=0 timeout=900 upgrade=websocket
   ProxyPassReverse / {{ httpd_foreman_backend }}/
 {% else %}
+  ProxyAddHeaders On
```
evgeni: The `if` branch also has it, so maybe we can just move it outside the `if`/`else` block?

Alleny244: ahh yess, have updated it

tests/feature/foreman/trusted_proxies_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pytest.fixture(scope="module")
+def foreman_settings_yaml(server):
+    result = server.run(
+        "podman secret inspect foreman-settings-yaml "
+        "--format '{{.SecretData}}' --showsecret"
+    )
+    assert result.succeeded, result.stderr
+    return yaml.safe_load(result.stdout)
```
ekohl: Running a command and raising an exception it built into testinfra (https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.host.Host.check_output)
```suggestion
    result = server.check_output(
        "podman secret inspect foreman-settings-yaml "
        "--format '{{.SecretData}}' --showsecret"
    )
    return yaml.safe_load(result)
```

tests/feature/foreman/trusted_proxies_test.py
```diff
@@ -0,0 +1,31 @@
...
+    assert '::1' in trusted
+
+
+def test_foreman_settings_trusted_proxies_include_configured_entries(
+    foreman_settings_yaml, obsah_params,
+):
+    configured = obsah_params.get('foreman_trusted_proxies') or []
+    trusted = foreman_settings_yaml[TRUSTED_PROXIES_KEY]
+    for entry in configured:
+        assert entry in trusted
```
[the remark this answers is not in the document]
ekohl: It's a good practice to only have a single assert statement in a test. https://mfranc.com/unit-testing/good-unit-test-one-assert/ is a bit longer and has slightly more nuance.
```suggestion
    configured = obsah_params.get('foreman_trusted_proxies') or []
    trusted = foreman_settings_yaml[TRUSTED_PROXIES_KEY]
    assert all(entry in trusted for entry in configured)
```

evgeni:
Tested this on a local setup with a proxy on 192.168.122.118 and a client on 192.168.122.221.
When the client executes `curl -k https://proxy.example.com:8443/unattended/built`, without the change here the logs on the Foreman side read:
```
Sep 11 10:17:27 quadlet.example.com foreman[16605]: 2026-09-11T10:17:27 [I|app|e4e379f8] Started GET "/unattended/built?url=http%3A%2F%2Fproxy.example.com%3A8000" for 192.168.122.118 at 2026-09-11 10:17:27 +0000
Sep 11 10:17:27 quadlet.example.com foreman[16605]: 2026-09-11T10:17:27 [I|app|e4e379f8] Processing by UnattendedController#built as TEXT
Sep 11 10:17:27 quadlet.example.com foreman[16605]: 2026-09-11T10:17:27 [I|app|e4e379f8]   Parameters: {"url"=>"http://proxy.example.com:8000", "unattended"=>{}}
Sep 11 10:17:27 quadlet.example.com foreman[16605]: 2026-09-11T10:17:27 [E|app|e4e379f8] unattended: unable to find a host that matches the request from 192.168.122.118. Search paths: ip: 192.168.122.118
Sep 11 10:17:27 quadlet.example.com foreman[16605]: 2026-09-11T10:17:27 [I|app|e4e379f8]   Rendered text template (Duration: 0.0ms | Allocations: 2)
Sep 11 10:17:27 quadlet.example.com foreman[16605]: 2026-09-11T10:17:27 [I|app|e4e379f8] Completed 404 Not Found in 32ms (Views: 0.7ms | ActiveRecord: 17.1ms | Allocations: 2059)
```

Ignore the 404 -- there is no host in build mode right now. The important part is that it sees the request coming in from `192.168.122.118` (= the proxy!) and also searches a host with that IP.

After applying the change, for the same curl we see:

```
Sep 11 10:25:09 quadlet.example.com foreman[51220]: 2026-09-11T10:25:09 [I|app|bbe15aa5] Started GET "/unattended/built?url=http%3A%2F%2Fproxy.example.com%3A8000" for 192.168.122.221 at 2026-09-11 10:25:09 +0000
Sep 11 10:25:09 quadlet.example.com foreman[51220]: 2026-09-11T10:25:09 [I|app|bbe15aa5] Processing by UnattendedController#built as TEXT
Sep 11 10:25:09 quadlet.example.com foreman[51220]: 2026-09-11T10:25:09 [I|app|bbe15aa5]   Parameters: {"url"=>"http://proxy.example.com:8000", "unattended"=>{}}
Sep 11 10:25:09 quadlet.example.com foreman[51220]: 2026-09-11T10:25:09 [E|app|bbe15aa5] unattended: unable to find a host that matches the request from 192.168.122.221. Search paths: ip: 192.168.122.221
Sep 11 10:25:09 quadlet.example.com foreman[51220]: 2026-09-11T10:25:09 [I|app|bbe15aa5]   Rendered text template (Duration: 0.0ms | Allocations: 4)
Sep 11 10:25:09 quadlet.example.com foreman[51220]: 2026-09-11T10:25:09 [I|app|bbe15aa5] Completed 404 Not Found in 42ms (Views: 8.3ms | ActiveRecord: 16.1ms | Allocations: 2765)
```
Now correctly identifying as `192.168.122.221`

The same applies for `/rhsm`: `curl -k https://proxy.example.com/rhsm/status` results in 
```
Sep 11 10:32:39 quadlet.example.com foreman[51220]: 2026-09-11T10:32:39 [I|app|449b226e] Started GET "/rhsm/status" for 192.168.122.221 at 2026-09-11 10:32:39 +0000
Sep 11 10:32:39 quadlet.example.com foreman[51220]: 2026-09-11T10:32:39 [I|app|449b226e] Processing by Katello::Api::Rhsm::CandlepinProxiesController#server_status as */*
Sep 11 10:32:39 quadlet.example.com foreman[51220]: 2026-09-11T10:32:39 [I|app|449b226e] Completed 200 OK in 119ms (Views: 1.9ms | ActiveRecord: 31.4ms | Allocations: 31122)
```
after the patch has been applied.

src/playbooks/_foreman/metadata.obsah.yaml
```diff
@@ -14,6 +14,12 @@ variables:
...
+    help: >-
+      IP address or CIDR Foreman should trust for X-Forwarded-For (e.g. Foreman Proxy).
+      Can be specified multiple times. Localhost ranges are always included.
+    parameter: --foreman-trusted-proxies
+    action: append_unique
```
evgeni: There is currently no way to remove a trusted proxy once added.
Have a look how `--add-feature` and `--remove-feature` work :)

evgeni (changes requested):
Overall ACK on the architecture
- request changes for the removal of trusted proxies
- the "do more validaton on user input" is a nice to have, but not a must

ekohl (commented):
Apologies for looking too much at the code. I was looking for a related documentation PR. They're very minor comments so not blocking.

src/playbooks/_foreman/metadata.obsah.yaml
Alleny244: Currently I have added `--foreman-trusted-proxies` flag since we are about to support removal of trusted proxies can we have something like ` --add-foreman-trusted-proxies` and  `--remove-foreman-trusted-proxies` or should we maintain the same flag for addition

evgeni: or `--(add|remove)-foreman-trusted-proxy` (singular)?

honestly not too tied to a specific param name here.

maybe @jafiala has an opinion with his docs hat on what's better readable?

ekohl: I'm wondering about `--foreman-trusted-proxy-(add|remove)` so they're stored alphabetically together. Perhaps even consider a third: `--foreman-trusted-proxy-set` to set the value (even if we don't do that immediately).

evgeni: I think `set` is too much choice for the user. Also it'd need to be able to parse things into an array.

src/playbooks/_foreman/metadata.obsah.yaml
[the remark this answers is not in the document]
Alleny244: > We could extend https://github.com/theforeman/obsah/blob/master/obsah/data_types.py to have a IP type and use it here.

Actually it's good to have that, so validation can be more easily done, can I open a pr there and link it here then. (IP types -> accepts x.x.x.x and x.x.x.x/x )?

src/playbooks/_foreman/metadata.obsah.yaml
ekohl: How about reset and then add multiple times? So could this work: `--foreman-trusted-proxy-reset --foreman-trusted-proxy-add $ip1 --foreman-trusted-proxy-add $ip2`

tests/feature/foreman/trusted_proxies_test.py
Alleny244: Sure, got it

src/playbooks/_foreman/metadata.obsah.yaml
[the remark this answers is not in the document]
evgeni: Yeah, plus the same for IPv6.

src/playbooks/_foreman/metadata.obsah.yaml
evgeni: We have reset, so that could already work?

ekohl: Then that would be great to see in our user documentation.

src/playbooks/_foreman/metadata.obsah.yaml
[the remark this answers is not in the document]
Alleny244: > We could extend https://github.com/theforeman/obsah/blob/master/obsah/data_types.py to have a IP type and use it here.

https://github.com/theforeman/obsah/pull/137

src/playbooks/_foreman/metadata.obsah.yaml
Alleny244: okay then , I ll update to` foreman-trusted-proxy-add ` and` foreman-trusted-proxy-remove`

evgeni:
https://github.com/theforeman/obsah/pull/138 is preping the release of obsah with the new validaton

src/playbooks/_foreman/metadata.obsah.yaml
```diff
@@ -14,6 +14,23 @@ variables:
     parameter: --initial-location
   foreman_puma_workers:
     help: Number of workers for Puma.
+  foreman_trusted_proxies:
+    help: >-
+      IP address or CIDR Foreman should trust for X-Forwarded-For (e.g. Foreman Proxy).
+      Can be specified multiple times. Localhost ranges are always included.
+    parameter: --foreman-trusted-proxy-add
+    type: IP
```
evgeni, line 22: this deserves a bump of the minimum obsah version to 1.11.0 now :)

docs/user/parameters.md
```diff
@@ -58,6 +58,9 @@ There are multiple use cases from the users perspective that dictate what parame
...
+| `--foreman-trusted-proxy-remove` | Remove an IPv4/IPv6 address or CIDR from the persisted trusted proxy list (same format as `--foreman-trusted-proxy-add`). May be specified multiple times. Does not remove localhost ranges or addresses added from registered Smart Proxies during deploy. | |
+| `--reset-foreman-trusted-proxy-add` | Clear the entire persisted trusted proxy list for this deployment (Obsah reset flag; hidden from the main option list—see `foremanctl deploy --help` epilog). Use with repeated `--foreman-trusted-proxy-add` to replace the list, for example: `foremanctl deploy --reset-foreman-trusted-proxy-add --foreman-trusted-proxy-add 10.0.0.1 --foreman-trusted-proxy-add 10.0.0.2`. | |
```
evgeni, line 63: `--reset-foreman-trusted-proxy-add` reads very awkwardly to me, but I guess that's the price for automatic `--reset-*` handling and a `add` in the param name

Alleny244: yeahh, do you want to revisit the flag's naming convention

src/playbooks/_foreman/metadata.obsah.yaml
Alleny244: Have bumbed the version in `src/requirements.txt`

docs/user/parameters.md
evgeni: Honestly don't know.
We could do `--foreman-trusted-proxy` for adding (but then the parameter doesn't indicate adding) and `--foreman-trusted-proxy-remove` for removing, then we get `--reset-foreman-trustred-proxy` which sounds better.
OTOH, how often will people use reset? idk.

Probably easiest to keep as is for now and listen to user feedback?

@ekohl @jafiala what do you think?

src/requirements.txt
```diff
@@ -1,3 +1,3 @@
-obsah >= 1.9.1
+obsah >= 1.11.1
```
evgeni:
```suggestion
obsah >= 1.11.0
```

there is no .1 yet

src/playbooks/_foreman/metadata.obsah.yaml
```diff
@@ -14,6 +14,23 @@ variables:
...
+    action: append_unique
+  remove_foreman_trusted_proxies:
+    help: >-
+      Remove an IP address or CIDR from the persisted trusted proxy list.
+      Can be specified multiple times. Localhost ranges are always included;
+      proxies discovered from registered Smart Proxies on deploy are not
+      controlled by this flag.
```
evgeni, line 29: What does that mean?

evgeni:
The `upgrade` failure is unrelated, fix in https://github.com/theforeman/foremanctl/pull/833

docs/user/parameters.md
ekohl: > We could do `--foreman-trusted-proxy` for adding (but then the parameter doesn't indicate adding) and `--foreman-trusted-proxy-remove` for removing, then we get `--reset-foreman-trustred-proxy` which sounds better.

This looks cleanest to me now.

> Probably easiest to keep as is for now and listen to user feedback?

I like the idea, but wonder how we best do that. Perhaps a user feedback post on Discourse? Should we have 1 big thread or a separate one?

src/playbooks/_foreman/metadata.obsah.yaml
Alleny244: `--remove ` flag only removes entries from the list that were added with `-add.` It doesn’t affect the localhost ranges in settings.yaml. So those entries won't be affected and I tried rephrasing the sentence with AI but turned out to be this

evgeni: Isn't that already covered by `Localhost ranges are always included`? I was pondering what "discovered from registered Smart Proxies" means.

Alleny244: What I meant was something like this
 Localhost ranges (127.0.0.0/8 and ::1) are always included in Foreman settings and are not removed by this flag.

Alleny244: > Isn't that already covered by `Localhost ranges are always included`? I was pondering what "discovered from registered Smart Proxies" means.

yeah

Alleny244: Have updated it now

evgeni (approved):

docs/user/parameters.md
jafiala: Can we remove "foreman" to at least shorten the params?

jafiala: @evgeni Shouldn't it be `--foreman-trustred-proxy-reset` to keep with the pattern?

My slight preference would be

```
--add-trusted-proxy
--remove-trusted-proxy
--reset-trusted-proxy
```

or 


```
--trusted-proxy-add
--trusted-proxy-remove
--trusted-proxy-reset
```

jafiala: Oh, is the "reset" automatically prepended to the param it resets? In that case I agree with Evgeni's proposal, but would like to see the `foreman` dropped, if possible.

evgeni: Correct, `--reset-<thing>` is auto-generated, while the others are coming from the metadata explicitly

jafiala: And after Evgeni's presentation, if there's also a pulp proxy, let's keep the `foreman` in there to distinguish it. So I'm for Evgeni's proposal without changes. Sorry for the noise.

evgeni: Those are not the proxies you are looking for.

Proxy (in the context of this parameter) is any system that forwards HTTP requests of a client to our backend. That can be the Foreman Proxy (with the Template feature), it can be Apache httpd on the external proxy (using mod_proxy), it can be Apache httpd on the main server (using mod_proxy, which is why we always add 127.0.0.1 here), it can be some weird loadbalancer the user really wants to put in front of our deployment.

The Smart Proxy that runs inside Pulp is not one of them.

The `foreman` in this param indicates "it's the Foreman service that trusts those proxies, not any other service".

stejskalleos: @Alleny244, I see that `--foreman-trusted-proxy-add` has not been changed to `--foreman-trusted-proxy`; are you planning to update it before the merge?

Alleny244: Yeah , I am updating it , will push the changes

Alleny244: Done

tests/feature/foreman/trusted_proxies_test.py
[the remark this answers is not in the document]
Alleny244: Yeah, updated!

docs/user/parameters.md
Alleny244: Updated!!

PR #825, opened 2026-09-10
Share RH Cloud inventory reports between Foreman and Dynflow with a named volume

nofaralfasi:

`foreman_rh_cloud` writes inventory archives under `/var/lib/foreman/red_hat_inventory` (`generated_reports/`, then `uploads/` / `uploads/done/`). Generation runs in Dynflow; download is Foreman `send_file` in a different container. Without a shared mount, the file is not visible to Foreman.

Binding all of `/var/lib/foreman` overlays the image `db/` and `public/` trees, so Foreman cannot boot. Sharing only `red_hat_inventory` via a named Podman volume keeps those image paths intact. Users download from the UI/API; they are not expected to read the host filesystem.

See #755.

Fixes: https://redhat.atlassian.net/browse/SAT-48305

* When `rh-cloud` is enabled, mount named volume `foreman-rh-cloud-inventory` (override: `foreman_rh_cloud_inventory_volume`) at `/var/lib/foreman/red_hat_inventory` in Foreman and Dynflow (`rw,z,U`)
* Let Podman create that volume when the containers use it (same as `foreman-data-run`)
* Add tests that the volume exists, is mounted in Foreman and Dynflow, and is writable/shared (including `uploads/`)

UI/API generate and download are the supported path. `rh_cloud_inventory:report:generate` is not added to the `foreman-rake` wrapper.

* Deploy, then `./foremanctl deploy --add-feature rh-cloud` (or `--add-feature iop`)
* Confirm the named volume (not a host bind-mount):
  `podman inspect foreman --format '{{json .Mounts}}'`
  `podman inspect dynflow-sidekiq-worker --format '{{json .Mounts}}'`
  Destination should be `/var/lib/foreman/red_hat_inventory`, type `volume`, name `foreman-rh-cloud-inventory`
* Generate from the Inventory Upload UI, or `POST /api/v2/organizations/:id/rh_cloud/report`, wait until the Dynflow task finishes, then download from the UI or `GET` the same path
* From the machine that SSHes into the VM (not on the VM itself):
  `./forge test --pytest-args 'tests/feature/foreman/inventory_storage_test.py -vv'`


src/roles/foreman/tasks/main.yaml
```diff
@@ -100,6 +100,24 @@
...
+    group: root
+    mode: "0755"
+
+# foreman_rh_cloud writes inventory reports under red_hat_inventory/
+# (generated_reports, uploads, outputs). Bind-mount that directory so Foreman
+# and Dynflow share it and `foreman-rake rh_cloud_inventory:report:generate`
+# can drop files on the host.
+- name: Create RH Cloud inventory directory
+  ansible.builtin.file:
+    path: /var/lib/foreman/red_hat_inventory/exports
```
ehelms: Let's just bind mount `/var/lib/foreman`

ehelms:
Is there an issue tracking re-design of the report generation to not rely on an admin SSHing into their Satellite but rather being able to use the API and download the report?

nofaralfasi:
> Is there an issue tracking re-design of the report generation to not rely on an admin SSHing into their Satellite but rather being able to use the API and download the report?

I haven’t found a tracker for that redesign. SAT-48305 is only this PR: make the current rake path work in containers by bind-mounting the report directory.

Generate and download via API already exist (`POST`/`GET` `/api/v2/organizations/:id/rh_cloud/report`). The Inventory Upload “Download report” button already uses that GET. Generation from the UI uses a different endpoint (`POST /foreman_inventory_upload/:org_id/reports`). Both still write or serve the archive from disk, which is why the shared mount is needed for containers.

I can file a foreman_rh_cloud issue if we want to track moving this off the local filesystem.

ehelms:
If the API and UI handle this for the user already, why do we need the rake task?

lfu:
I deployed this branch (4bb9eab) with `foremanctl deploy` on a clean RHEL 9.8 host. **Foreman doesn't boot** — the deploy fails at `RUNNING HANDLER [foreman : Restart foreman]`:

```
cannot load such file -- /usr/share/foreman/db/seeds.d/020-roles_list.rb (LoadError)
    from /usr/share/foreman/lib/seed_helper.rb:1:in `<top (required)>'
```

`/var/lib/foreman` isn't an empty data dir in the image — the app symlinks into it:

```
/usr/share/foreman/public -> /var/lib/foreman/public   # assets/ + webpack/, 168 MB
/usr/share/foreman/db     -> /var/lib/foreman/db       # migrate, seeds.rb, seeds.d
```

Mounting an empty host dir over it hides both. `seed_helper.rb` requires every file in `db/seeds.d/` during the Rails finisher, so boot dies there; `public/` being gone would break the UI's CSS/JS too. Same for the three `dynflow-sidekiq@` instances. Only `foreman-db-migrate` survives, because it has no volume.

I then narrowed both `Volume=` lines on the deployed host to `/var/lib/foreman/red_hat_inventory:/var/lib/foreman/red_hat_inventory:rw,z,U`, `daemon-reload`ed and started the services:

- `foreman` and all three `dynflow-sidekiq@` instances active
- image content intact in the container: 177 files in `public/assets`, 20 in `db/seeds.d`
- `/api/v2/ping` healthy, and `GET /assets/application-<digest>.css` returns 200 / 444 KB
- dynflow writes `red_hat_inventory/generated_reports/report_for_1.tar.xz`, foreman reads it, and it's on the host as `994:994`

So the original `red_hat_inventory` target both fixes the boot failure and delivers the sharing this PR is after. One adjustment needed: the `file` task has to create `/var/lib/foreman/red_hat_inventory`, not just the parent.

Worth noting the mount is added unconditionally, while `foreman_rh_cloud` is gated behind the `rh-cloud` feature (`FOREMAN_ENABLED_PLUGINS`). On my deploy without it, `foreman-rake rh_cloud_inventory:report:generate` returns `Don't know how to build task` — so every deployment takes the boot risk for a feature only some enable. Might be worth conditioning the mount on `rh-cloud`.

Three smaller things:

1. `test_foreman_storage_directory` can't pass live — `:U` chowns whatever the mount source is to the container user, so `root:root` doesn't survive container start; on the deployed host `/var/lib/foreman` came out `drwxr-xr-x 2 994 994`. The `file` task also re-chowns it every run, so it reports `changed` forever. (If you narrow the mount, keep that assertion on the parent — the `red_hat_inventory` subdir becomes 994.)
2. `exports` isn't a directory foreman_rh_cloud uses (it has `generated_reports/`, `uploads/`, `uploads/done/`, `outputs/`). The test `mkdir -p`s it, so it passes vacuously — `generated_reports` is the one that matters.
3. The rake allowlist entry alone doesn't make the report retrievable. `report:generate` uses `base_folder = ENV['target'] || Dir.pwd`; `Dir.pwd` is the root-owned `/usr/share/foreman`, so it falls back to `Dir.mktmpdir`. I ran it on the deployed host:

   ```
   /usr/share/foreman is not writable by the current process
   Using /tmp/d20260911-455-p46nz9 for the output
   ```

   That tmpdir is created in the **foreman** container, but the report is written by `HostInventoryReportJob` in a **dynflow worker** — which doesn't have that path (`/tmp` isn't shared). It ended up empty in the foreman container and absent from both the worker and the host. So the task needs `target=` pointing at the shared mount, e.g. `/var/lib/foreman/red_hat_inventory/generated_reports`. (`report:generate_upload` already passes `generated_reports_folder`, so only `generate` has this.)

The UI/API path does justify the shared mount either way: generation runs in a dynflow worker writing to `generated_reports_folder`, and download is served by the foreman container reading the same path.

nofaralfasi:
> If the API and UI handle this for the user already, why do we need the rake task?

If UI/API is the path we want to support, I don’t think we need the rake allowlist. I added `rh_cloud_inventory:report:generate` because it is still documented (plugin README) and mentioned in SAT-48305. The UI and API do not use that task.

We do still need a shared mount of `/var/lib/foreman/red_hat_inventory` (not all of `/var/lib/foreman` - that overlays the image’s `db/` and `public/`). Dynflow writes the archive there, and Foreman serves it with `send_file` from a separate container. Without that directory shared, download cannot see the file.

Happy to drop the wrapper change and keep that mount.

ehelms:
@evgeni What do you think about this shared volume approach? If we do go that route, this should likely be a named podman volume instead of `/var/lib/foreman/`

evgeni:
We already share `/var/run/foreman` (aka `/usr/share/foreman/tmp`), isn't that sufficient?

Anyway: Who is the consumer of that data? If it's only Foreman/Dynflow, it should be a named volume. If users are supposed to look at that data too, it's probably better a "stable" path on disk instead.

nofaralfasi:
`/var/run/foreman` is already shared between Foreman and Dynflow (`foreman-data-run`). Inventory reports are not written there. UI/API generation uses `/var/lib/foreman/red_hat_inventory/` (`generated_reports` / `uploads` / `outputs`). Sharing tmp does not make Dynflow’s archive visible to Foreman’s `send_file`.

We also cannot bind-mount all of `/var/lib/foreman` (that overlays image `db/` and `public/`).

For UI/API the consumers are Foreman and Dynflow, so a named volume mounted at `/var/lib/foreman/red_hat_inventory` would be enough. The plugin still displays that path (“Report saved to”), and the README documents rake `target=` there. Official docs tell people to download from the UI.

Do you want the bind-mount of that host path, or a named volume at the same container path?

evgeni:
I prefer if the user never has to interact with the filesystem on their own, so named volume and updates to UI/docs.

nofaralfasi:
Named volume it is. I'll change this PR to a named volume at `/var/lib/foreman/red_hat_inventory` (not a host bind-mount, and not all of `/var/lib/foreman`) so Foreman and Dynflow still share the plugin path.

Pointing users at the filesystem is in `foreman_rh_cloud` (Inventory Upload “Report saved to”, and the README `target=`). I’ll follow up there so the UI/docs match “download from the UI/API, don’t touch the disk.”

ehelms:
How long does the report need to exist for? Do they get cleaned up?

src/roles/foreman/templates/foreman-rake.j2
```diff
@@ -24,6 +24,7 @@ validate_actions() {
         "reports:expire"
         "reports:monthly"
         "reports:weekly"
+        "rh_cloud_inventory:report:generate"
```
lfu: Allowlisting this doesn't yet make the report retrievable. I flagged this last round on an empty inventory; here it is again with a real host **and** with `xz` present in the worker, so neither of those can be the cause.

Run exactly as a user would:

```
$ foreman-rake rh_cloud_inventory:report:generate organization_id=1
/usr/share/foreman is not writable by the current process
Using /tmp/d20260915-1057-cxo1ce for the output
Check the Uploading tab for report uploading status.
```

`base_folder = ENV['target'] || Dir.pwd`, and `Dir.pwd` is the root-owned `/usr/share/foreman`, so it falls back to `Dir.mktmpdir` — in the **foreman** container. The report is written by `HostInventoryReportJob` in a **dynflow worker**, which has no such path:

```
[I|app|] slice eee187a8-86e5-4df6-9755-f84130f353a0 was created with 1 hosts
[I|app|] Archiving generated report
[I|app|] tar: tar (child): /tmp/d20260915-1057-cxo1ce/report_for_1.tar.xz: Cannot open./
 | tar: Child returned status 2
[I|app|] Tar command failed: pid 566 exit 0
```

Data collection works; the write fails; no file lands anywhere; and the rake task prints no error at all. With `target=/var/lib/foreman/red_hat_inventory/generated_reports` it works — 6768 bytes, valid XZ, same 1-host slice.

Two things worth deciding on:

1. **The task never tells you where the file went, even on success.** `Setting[:subscription_connection_enabled]` is true by default, so it takes the `Check the Uploading tab` branch and skips the `Generated #{target_file} for organization id …` message entirely.
2. **The named volume makes the output harder to reach.** The only reason to prefer `report:generate` over the UI is to get a file out, but this PR's commit message says users aren't expected to read that tree on disk — and it's now under `/var/lib/containers/storage/volumes/…`.

If this entry is meant to be usable, it probably wants either a documented `target=` or a wrapper that defaults it to the shared path.

Separately, omitting `organization_id` gives a 40-line backtrace rather than the intended message. That's an upstream bug, but this line is what exposes it:

```ruby
organization_ids = [ENV['organization_id']]
...
if organization_ids.empty?
  puts "Must specify organization_id"
```

`[nil]` is never empty, so the guard is dead code and it proceeds to `Organization.find(nil)` → `Couldn't find Organization without an ID (ActiveRecord::RecordNotFound)`. Probably worth filing against `foreman_rh_cloud`.

src/roles/foreman/tasks/main.yaml
```diff
@@ -100,6 +100,20 @@
     - Restart foreman
     - Restart dynflow-sidekiq@
 
+- name: Create Foreman storage directory
```
lfu: This looks like leftover from the bind-mount approach. Now that the mount is a named volume, nothing in this role touches host `/var/lib/foreman` — and `iop_cvemap_downloader` and `iop_vex_downloader` already create that directory with the same `root:root` / `0755`. Can this just go?

tests/feature/foreman/inventory_storage_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+FOREMAN_AND_DYNFLOW_CONTAINERS = [
+    "foreman",
+    "dynflow-sidekiq-orchestrator",
+    "dynflow-sidekiq-worker",
+    "dynflow-sidekiq-worker-hosts-queue",
+]
+
+
+def test_foreman_storage_directory(server):
```
lfu: Companion to the comment on the `Create Foreman storage directory` task — if that task goes, this should go with it. As written it's gated behind `pytest.mark.feature("rh-cloud")` but asserts on a directory that has nothing to do with rh-cloud.

tests/feature/foreman/inventory_storage_test.py
```diff
@@ -0,0 +1,83 @@
...
+        result = server.run(f"podman exec dynflow-sidekiq-worker touch {path}")
+        assert result.succeeded, result.stderr
+        result = server.run(f"podman exec foreman test -f {path}")
+        assert result.succeeded, result.stderr
+    finally:
+        server.run(f"podman exec dynflow-sidekiq-worker rm -f {path}")
+
+
+def test_foreman_inventory_generated_reports_writable_from_foreman_and_dynflow(server):
+    path = f"{GENERATED_REPORTS_PATH}/foremanctl-generated-reports-test"
```
lfu: This targets a directory neither flow populates on its own. A real API-driven run lands in `uploads/` and then `uploads/done/`; `report:generate` writes wherever `target=` points. `generated_reports/` is only the default for `report:generate_upload`. Since the test `mkdir -p`s the directory itself, it passes regardless of what the product does.

Mounting the `red_hat_inventory` parent is the right call precisely because it covers all of these — the test would just be more meaningful pointed at `uploads/`.

lfu:
I retested at `8b7fab0` with a full `foremanctl deploy` on a clean RHEL 9.8 host (`rh-cloud` enabled), with a real RHEL 9.8 content host registered to it. The named volume addresses everything from my last round, and I was able to verify the sharing this PR is after end to end, with real report content. 

## Verified working

Deploy completes. Inside the `foreman` container `public/assets` and `db/seeds.d` are intact, `/api/v2/ping` returns 200 with `foreman_rh_cloud` in the results, and the UI serves normally — the boot failure from `4bb9eab` is gone.

The UI/API generate → download path works over the volume:

```
POST /api/v2/organizations/1/rh_cloud/report   -> {"action_status":"success"}
# dynflow-sidekiq-worker writes into the volume, then moves it on upload:
/var/lib/foreman/red_hat_inventory/uploads/done/report_for_1.tar.xz   6772 bytes, 994:994
GET  /api/v2/organizations/1/rh_cloud/report   -> HTTP 200, 6772 bytes, served by the foreman container
```

The archive is real: `metadata.json` naming one slice with `number_hosts: 1`, and a slice file with the registered host's `fqdn`, its `subscription_manager_id` matching the `katello_subscription_facets` row, and a full `system_profile` (RHEL 9.8, x86_64, 536 `installed_packages`, `yum_repos`, `network_interfaces`, `installed_products`). So generation in the worker and download from the foreman container really are sharing the same storage.

Test runs against that deployment:

- `tests/feature/foreman/inventory_storage_test.py` + `foreman_rake_test.py` — 17/17 pass
- full `tests/feature/foreman` + `tests/feature/httpd` — 88 passed, 4 skipped, 2 failed; both failures are `test_foreman_initial_organization` / `test_foreman_initial_location` expecting CI's "Foreman CI" / "Internet" against a deployment using the `src/vars/foreman.yml` defaults, so unrelated to this branch
- `ansible-lint` and `tests/unit` + `tests/ansible_lint` match the pre-existing baseline on `master`

## Heads-up: the image has no `xz`

Not this PR's problem and not a reason to hold it — but it does mean a stock deployment can't produce a usable report yet, and you need to know about it to reproduce the results above.

`foreman_rh_cloud` archives the report with `tar` piped through `xz`. The `xz` CLI isn't installed in `quay.io/foreman/foreman:nightly` — `liblzma.so.5` is present (as a library dependency), but `command -v xz` finds nothing and `rpm -q xz` reports "package xz is not installed", in the `foreman` container and all three `dynflow-sidekiq-*` containers. The result is silent:

```
[I|app|] Archiving generated report
 | /bin/sh: line 1: xz: command not found
 | tar: Child returned status 127
[I|app|] Tar command failed: pid 109 exit 0
```

A 0-byte `report_for_1.tar.xz` is still written, the dynflow task finishes `stopped`/`success`, the API returns `{"action_status":"success"}`, and the download returns HTTP 200 with `size=0`.

Everything above was therefore measured with the host's `xz` copied into the worker first (`podman cp /usr/bin/xz dynflow-sidekiq-worker:/usr/local/bin/xz`). With that one change and nothing else, the same request produces the valid 6772-byte archive — which is what convinced me the volume, permissions and cross-container sharing in this PR are all correct. 

🤖 Testing and write-up done with [Claude Code](https://claude.com/claude-code)

src/roles/foreman/tasks/main.yaml
```diff
@@ -100,6 +100,20 @@
...
+  ansible.builtin.file:
+    path: /var/lib/foreman
+    state: directory
+    owner: root
+    group: root
+    mode: "0755"
+
+- name: Create RH Cloud inventory volume
```
evgeni: You don't need to explicitly create it. A container using it is sufficient.

nofaralfasi:
> How long does the report need to exist for? Do they get cleaned up?

No timed retention, and nothing deletes them on a schedule.
They stay under `/var/lib/foreman/red_hat_inventory/` until the next report with the same filename overwrites them.

tests/feature/foreman/inventory_storage_test.py
nofaralfasi: Agreed the `generated_reports/` check was vacuous because the test created the directory itself.

UI/API still writes there first, then Dynflow moves the file to `uploads/` and on success to `uploads/done/`. I’ll retarget the nested write/share test to `uploads/`. The named volume on `red_hat_inventory` stays as-is.

src/roles/foreman/tasks/main.yaml
nofaralfasi: Agreed, this is a leftover from the bind-mount. Removed the host `/var/lib/foreman` task and the test that asserted it. The named volume only needs the container path. IoP still creates the host directory for cvemap/vex when that feature is on.

src/roles/foreman/tasks/main.yaml
nofaralfasi: Removed the explicit `podman_volume` task. Foreman and Dynflow already reference the named volume, which is enough for Podman to create it.

jeremylenz:
> > How long does the report need to exist for? Do they get cleaned up?
> 
> No timed retention, and nothing deletes them on a schedule. They stay under `/var/lib/foreman/red_hat_inventory/` until the next report with the same filename overwrites them.

Since they're overwritten by the next report (for the same org/host), I don't think there should really be any cleanup concerns.

jeremylenz:
> Is there an issue tracking re-design of the report generation to not rely on an admin SSHing into their Satellite but rather being able to use the API and download the report?

This will require a documentation update: see https://github.com/theforeman/foreman-documentation/pull/5314#discussion_r4074469847

ehelms:
> We already share `/var/run/foreman` (aka `/usr/share/foreman/tmp`), isn't that sufficient?

Based on that information, this is a viable strategy and would keep us from mounting yet another volume which is a good architectural choice. This would require a code change in `foreman_rh_cloud`? Which sounds like a minor change for better architecture.

lfu:
- `inventory_storage_test.py` + `foreman_rake_test.py` — 15/15 pass
- Full `tests/feature/foreman` + `tests/feature/httpd` — 86 passed, 4 skipped, 2 failed
  - Both failures are pre-existing: `test_foreman_initial_organization` / `test_foreman_initial_location` (CI vs default org/location names)

**End-to-end volume sharing verified**
```
POST /api/v2/organizations/1/rh_cloud/report  →  {"action_status":"success"}
dynflow-sidekiq-worker writes → /var/lib/foreman/red_hat_inventory/uploads/done/report_for_1.tar.xz  (6760 bytes)
GET  /api/v2/organizations/1/rh_cloud/report  →  HTTP 200, served by the foreman container
```

Mount confirms named volume on both containers:
```
{"Type":"volume","Name":"foreman-rh-cloud-inventory","Destination":"/var/lib/foreman/red_hat_inventory"}
```
Volume spec correctly uses the `foreman_rh_cloud_inventory_volume` variable — no hardcoded name.

**Note on xz**
As flagged in my previous review, the `xz` CLI is still missing from `quay.io/foreman/foreman:nightly` — report generation silently produces 0-byte archives. The 6760-byte report above was produced after manually copying `xz` into the worker (`podman cp /usr/bin/xz dynflow-sidekiq-worker:/usr/local/bin/xz`). This is not caused by this PR and doesn't block it, but the feature it plumbs can't deliver a usable report on a stock image until `xz` is added.

**One suggestion for merge**
Commits 1+2 (`3c27ff7` + `cb42a81`) should be squashed — the first mounts all of `/var/lib/foreman` (which breaks boot by shadowing the image's `public/` and `db/`), and the second immediately fixes it. A git bisect would land on a broken state otherwise.

lfu:
Where should the `xz` issue be reported?
The `xz` binary needs to be in the `quay.io/foreman/foreman` container image itself.
The right fix is a one-line addition of `xz` to the image's Dockerfile (or whatever builds the foreman image).

jeremylenz:
> > We already share `/var/run/foreman` (aka `/usr/share/foreman/tmp`), isn't that sufficient?
> 
> Based on that information, this is a viable strategy and would keep us from mounting yet another volume which is a good architectural choice. This would require a code change in `foreman_rh_cloud`?

Keep in mind there that I think we use `Rails.root` in foreman_rh_cloud code, which conveniently places the file in `/var/lib/foreman` in production and in `/home/vagrant/foreman` in development. If we go back to hard-coding, we'll lose that, which may or may not be a concern.

ehelms:
> Where should the `xz` issue be reported?
> The `xz` binary needs to be in the `quay.io/foreman/foreman` container image itself.
> The right fix is a one-line addition of `xz` to the image's Dockerfile (or whatever builds the foreman image).

Open a PR to https://github.com/theforeman/foreman-oci-images/blob/master/images/foreman/Containerfile

> Keep in mind there that I think we use `Rails.root` in foreman_rh_cloud code, which conveniently places the file in `/var/lib/foreman` in production and in `/home/vagrant/foreman` in development. If we go back to hard-coding, we'll lose that, which may or may not be a concern.

You do not have to go back to hard-coding: `Rails.root.join('tmp')` is what you would use.

jeremylenz:
Seems my memory was incorrect about hard-coding: https://github.com/theforeman/foreman_rh_cloud/blob/develop/lib/foreman_inventory_upload.rb#L7

I am trying to remember why we didn't go with `Rails.root` at that time. Probably because it would've put the report in `/home/vagrant/foreman` in development and then you'd have to move it before git committing anything.

jeremylenz:
@nofaralfasi Per Eric's [comment above](https://github.com/theforeman/foremanctl/pull/825#issuecomment-5781405095), let's try changing foreman_rh_cloud to use `/var/run/foreman` per [Evgeni's suggestion](https://github.com/theforeman/foremanctl/pull/825#issuecomment-5679383852). Then we may not need this PR at all.

nofaralfasi:
Per @jeremylenz's [suggestion](https://github.com/theforeman/foremanctl/pull/825#issuecomment-5794949677), I opened a [PR](https://github.com/theforeman/foreman_rh_cloud/pull/1262) in foreman_rh_cloud to change the plugin's `base_folder` from `/var/lib/foreman/red_hat_inventory/` to `/var/run/foreman/red_hat_inventory/`.

Since `/var/run/foreman` is already shared between Foreman and Dynflow via the `foreman-data-run` volume, this eliminates the need for the `foreman-rh-cloud-inventory` named volume introduced here. No foremanctl changes are needed.

Closing this PR in favor of that approach.

PR #830, opened 2026-09-11
Mark content/ostree feature as internal

Gauravtalreja1:
`foremanctl features` lists `content/ostree` feature as available, which should be internal feature like other content features

* Marking content/ostree feature as internal, so its not visible in available features list

1. $ foremanctl features
2. Verify `content/ostree` isn't listed as available feature

Steps to reproduce:

*

#### Checklist

### DIFF
--- a/src/features.d/content.yaml
+++ b/src/features.d/content.yaml
@@ -26,5 +26,6 @@ content/python:
     - pulp
 content/ostree:
   description: OSTree content type for Pulp
+  internal: true
   dependencies:
     - pulp


ekohl (commented):
I don't think so because we have explicit code to enable this feature in our docs (https://docs.theforeman.org/nightly/Managing_Content/index-containerized-katello.html#managing-ostree-content). If this is made internal that wouldn't work anymore.

ehelms:
Worth noting -- https://community.theforeman.org/t/pulp-ostree-plugin-archival-planned/47532

ekohl:
But then it would be removed, not marked internal.

ehelms:
@ianballou @sjha4 Based on Pulp's news around ostree, what should we do here? Should we address this? Should we just drop it?

sjha4:
I'd vote to remove it now since we are in the process of upgrading pulp and plugins for Katello 5.1 but I am not sure it has been communicated properly upstream for 5.1..Although users with existing ostree content (I don't know of any) won't be able to carry it into 5.2, the upgrade-able version even if we wait a version to drop it.

ianballou:
I was actually anticipating to ship the ostree RPM for another release since it should continue working at least until the next Pulpcore breaking changes release. That gives users a bit more time to consider options.

PR #835, opened 2026-09-16
install foreman-fapolicyd for the container rule

evgeni:
the ansible one remains, as that is specific to our Vagrant setup

*

Steps to reproduce:

*

#### Checklist


lzap (approved):
Makes sense, I haven't tested.

evgeni:
I'm quite confident CI will tell us :)

.github/workflows/test.yml
```diff
@@ -142,12 +142,9 @@ jobs:
...
         run: |
           ./forge security --mode ${{ matrix.security }}
-      - name: Apply fapolicyd workarounds
-        # https://access.redhat.com/solutions/7072618 / https://issues.redhat.com/browse/RHEL-37912
-        # https://github.com/theforeman/foreman-fapolicyd/blob/develop/15-foreman-container.rules
+      - name: Apply fapolicyd workaround
         if: matrix.security == 'fapolicyd'
         run: |
-          vagrant ssh --command "echo 'allow perm=any pattern=ld_so exe=/usr/bin/crun : path=/usr/lib64/libsystemd.so.0' | sudo tee -a /etc/fapolicyd/rules.d/15-foremanctl.rules" quadlet
           vagrant ssh --command "echo 'allow perm=any all : dir=/home/vagrant/.ansible/tmp/' | sudo tee -a /etc/fapolicyd/rules.d/15-foremanctl.rules" quadlet
```
Gauravtalreja1, line 148: Shouldn't this workaround as well be part of foreman-fapolicyd?

.github/workflows/test.yml
```diff
@@ -142,12 +142,9 @@ jobs:
...
           vagrant ssh quadlet -- sudo systemctl restart fapolicyd
```
Gauravtalreja1, line 149: And, this workaround be part of securing development playbook after we install foreman-fapolicyd pkg?

.github/workflows/test.yml
evgeni: There is no Vagrant in the prod deployment case, so no.

.github/workflows/test.yml
evgeni: The restart is only needed because we update the rules.

`foreman-fapolicyd` does that for its rules too, but we need to allow ansible earlier.

.github/workflows/test.yml
evgeni: Actually, it seems we can drop the Ansible part totally: https://github.com/theforeman/foremanctl/pull/837

ehelms (approved):

evgeni:
Going to merge as-is and drop the rest in #837

.github/workflows/test.yml
Gauravtalreja1: Nice, Thanks for checking!

PR #838, opened 2026-09-16
install collections via git/tarball not Galaxy -- galaxy is slow

evgeni:

*

Steps to reproduce:

*

#### Checklist


ehelms:
@evgeni Did you notice a difference with this change?

evgeni:
While Galaxy was having a hard time: yes.
But it's not something I'd want to do forever, so let's keep it as is now that galaxy is mostly back

PR #839, opened 2026-09-16
Add iop to upgrade features

ehelms:
#### Why are you introducing these changes? (Problem description, related links)

To ensure iop is incorporated in upgrade concerns.

#### What are the changes introduced in this pull request?

* Adds iop feature to upgrade testing

#### How to test this pull request

Steps to reproduce:

*

#### Checklist
* [ ] Tests added/updated (if applicable)
* [ ] Documentation updated (if applicable)

PR #840, opened 2026-09-16
Remove leftover iop-core-network before creating foreman-core-network

Alleny244:

After [PR #728](https://github.com/theforeman/foremanctl/pull/728), deploy always creates `foreman-core-network` on `10.130.0.0/24`. This is the same subnet previously used by `iop-core-network`.

Existing installs can still have `iop-core-network` and attached IOP containers, causing the network creation task to fail:

```text
Error: subnet 10.130.0.0/24 is already used on the host or by another config
```

`iop-core-network` was a Podman network (`podman_network` with `state: present`), not a Quadlet `.network` unit, so `NetworkDeleteOnStop` does not apply. `podman network rm --force` can also leave the netavark bridge (`podman1`, `podman2`, etc.) holding `10.130.0.1`, which still blocks creation of the new network.

- Stop systemd units and containers still attached to the leftover `iop-core-network` before removing it.
- Force-remove `iop-core-network`.
- If a leftover `podman[0-9]+` bridge still has `10.130.0.1`, delete that interface before creating `foreman-core-network`.
- Document the upgrade path in `docs/architecture/network.md`.

Steps to reproduce:

- On a host with an existing IOP deployment from before #728 (`iop-core-network` present, with gateway/kafka/etc. running on it), run `foremanctl deploy`.
- Confirm that deploy no longer fails at `Create foreman-core-network`.
- Confirm `podman network ls` shows `foreman-core-network` and no longer shows `iop-core-network`.
- Confirm `ip -4 addr show to 10.130.0.1` shows the address on the new network's bridge and not on an orphaned `podmanN` bridge.
- On a fresh install with no leftover `iop-core-network`, confirm the cleanup is a no-op and network creation succeeds.

#### Checklist

- [ ] Tests added/updated (if applicable)
- [ ] Documentation updated (if applicable)


ehelms:
You may want to cherry-pick https://github.com/theforeman/foremanctl/pull/839 and include it in this PR so we can see that upgrades pass.

Alleny244:
> You may want to cherry-pick #839 and include it in this PR so we can see that upgrades pass.

oh yes, have updated it

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
```diff
@@ -0,0 +1,81 @@
...
+  loop: "{{ _foreman_core_network_legacy_container_names }}"
+  failed_when: false
+
+- name: Remove leftover IOP network from previous deploys
+  containers.podman.podman_network:
+    name: "{{ foreman_core_network_legacy_iop_network_name }}"
+    state: absent
+    force: true
+
+- name: Check if foreman-core-network already exists
```
ehelms: This should never exist at this point in time unless we are trying to having continuity with "nightly" as we have not released the `foreman-core-network` change anywhere.

Alleny244: The check was originally added because during the pipeline redeploy, the cleanup could mistake the `foreman-core-network` bridge for a leftover IOP bridge and delete it, which caused PostgreSQL to become unreachable. I also saw that sometimes even after removing the Podman network, the bridge could still remain on the host. So now I check whether that bridge is still associated with any Podman network and only remove it if it is truly orphaned.

ehelms: Which pipeline redeploy?

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
```diff
@@ -0,0 +1,92 @@
+---
+- name: List containers still using leftover IOP network
```
evgeni, line 2: Instead of using commands, can't we use `containers.podman.podman_container_info` and look for containers which have `iop-core-network` in their `NetworkSettings.Networks` hash? Something like:

```
    - name: get containers
      containers.podman.podman_container_info:
      register: _c
    - name: dump foreman network users
      ansible.builtin.debug:
        msg: "{{ _c['containers'] | selectattr('NetworkSettings.Networks', 'contains', 'foreman-core-network') | map(attribute='Name') }}"
```

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
```diff
@@ -0,0 +1,92 @@
...
+      | list }}"
+
+- name: Stop leftover IOP network systemd units
+  ansible.builtin.systemd:
+    name: "{{ item }}"
+    state: stopped
+  loop: "{{ _foreman_core_network_legacy_container_names }}"
+  failed_when: false
+
+- name: Stop leftover IOP network containers
```
evgeni: if you stop the units, the containers should be stopped already?
or is there a case where this is not happening?

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
Alleny244: https://github.com/theforeman/foremanctl/actions/runs/35155070290/job/104993910448
In our test workflow we run foremanctl deploy twice , `Run deloyment `creates `foreman-core-network`. The second step `Deploy features` runs the same playbook again , on second deploy , the earlier fix was deleting the live bridge that  held `10.130.0.1. ` which was in use by foreman-core-network

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
Alleny244: yeah, this can do the job , it's much cleaner,  I ll update it. Thanks!!

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
Alleny244: yeah there shouldn't be any , just added it as a safeguard in case, there will be an orphaned container with its corresponding unit deleted , I ll remove the check

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
```diff
@@ -0,0 +1,81 @@
+---
+- name: List containers still using leftover IOP network
+  containers.podman.podman_container_info:
+  register: _foreman_core_network_legacy_info
+
+- name: Collect leftover IOP network container names
+  ansible.builtin.set_fact:
```
ehelms: Instead of `set_fact` can this be set as a `vars` on the next task? I think that should work and is a technique @evgeni taught me.

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
```diff
@@ -0,0 +1,81 @@
...
+  loop: "{{ _foreman_core_network_legacy_container_names }}"
+  failed_when: false
+
+- name: Remove leftover IOP network from previous deploys
+  containers.podman.podman_network:
+    name: "{{ foreman_core_network_legacy_iop_network_name }}"
+    state: absent
+    force: true
+
+- name: Check for leftover IOP subnet on the host
```
ehelms: If it was removed in the previous task, why would it still remain? Wouldn't the task abov ehave failed?

src/roles/foreman_core_network/tasks/main.yaml
```diff
@@ -1,4 +1,7 @@
 ---
+- name: Clean up leftover IOP network from previous deploys
+  ansible.builtin.include_tasks: cleanup_legacy_iop_network.yaml
```
ehelms, line 7: Right now all of the actions for this set of tasks will run every time. That can be unnecessary for an operation that will in reality only need to run once. Could we gate this inclusion, or gate within the task file itself to only run if `iop-core-network` exists?

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
Alleny244: yeah , sure I ll update it

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
Alleny244: Actually I had run into an instance, where the podman network was removed and that succeeded but somehow the underlying bridge for that network was remaining and causing this subnet already in use error, but it might be because of teardown issues in between which might happen rarely, if that's not an issue we can remove that safeguard rails where we check for orphaned bridges?

src/roles/foreman_core_network/tasks/main.yaml
Alleny244: Yeah, I ll modify it to run only when `iop-core-network` exists

ehelms (approved):

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
```diff
@@ -0,0 +1,26 @@
...
+  register: _foreman_core_network_legacy_info
+
+- name: Stop leftover IOP network systemd units
+  vars:
+    _foreman_core_network_legacy_container_names: "{{
+      _foreman_core_network_legacy_info.containers
+      | selectattr('NetworkSettings.Networks', 'defined')
+      | selectattr('NetworkSettings.Networks', 'contains', foreman_core_network_legacy_iop_network_name)
+      | map(attribute='Name')
+      | map('regex_replace', '^/', '')
```
evgeni: Where does this leading slash come from?

evgeni (approved):

evgeni:
I pushed a change to accommodate for the new test we introduced in #843 

The IOP enablement made the memory usage rise, so I pushed #850 to `2.y-stable` so machines get 12G now for upgrades too.

The rest of the failures is because cloud.centos.org seems to fail :/

src/roles/foreman_core_network/tasks/cleanup_legacy_iop_network.yaml
Alleny244: I had it in mind that the leading `/` would be returned by `podman_container_info` because when querying the Podman API with curl through the Unix socket, the `Names` field is returned with a leading `/`. I assumed `podman_container_info` would return it the same way, but yeah when I checked it now, it is right there is no leading slash, I ll update it

evgeni:
https://github.com/theforeman/foremanctl/pull/851 should fix the centos.org problems

evgeni:
I have *no* idea why backup tests are failing :/

evgeni:
Ooooh, it's running them on EL9, and forgets to upgrade the PostgreSQL client tools.

How the F did that work before?!

evgeni:
Yay, it passed

PR #845, opened 2026-09-18
Do not enable rex by default in capsule flavor

adamruzicka:

Deploying a new capsule could break established REX workflows due to the possibility of jobs getting routed to the new instance before its key is deployed to the hosts.

* Making remote execution opt-in in the capsule flavor

Steps to reproduce:

* Run deploy with the capsule flavor
* Check script feature is missing on that newly deployed proxy instance
* Run deploy again with `--add-feature remote-execution`
* Check script feature is present on that newly deployed proxy instance

#### Checklist


ekohl (approved):
This is what we discussed internally: adding REX in the flavor means the user can't disable it. By default the Smart Proxy selection algorithm could break things in subtle ways and we expect users to explicitly opt in.

Gauravtalreja1 (approved):

